PR #29, opened 2020-01-17
Fix explore.md

shiatsumat:
Fixed typos and minor errors in `explore.md`.

### DIFF
--- a/docs/explore.md
+++ b/docs/explore.md
@@ -11,7 +11,7 @@ There are two distinct aspects to system programming:
   - Can treat memory directly as bits and bytes
   - Little or no abstraction on the hardware
 
-To implement many low-level systems (for instance, a memory manager), you need raw access.
+To implement a low-level system of various kinds (for example, a memory manager), you need raw access.
 In some sense, the memory manager is producing an abstraction on the machine that high-level services can consume.
 Guaranteeing safety through a type system for programmers with raw access has not been achieved.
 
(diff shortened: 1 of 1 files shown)


docs/explore.md
```diff
@@ -26,16 +26,16 @@ This is an explicit non-goal of the project.
...
 This dynamic consensus that an object is no longer in use can cause 
 either latency issues or throughput.
 There is amazing research in doing this better
 but in Verona we want to take an alternative approach: ownership.
 
-Ownership has been used a in a few languages to support scalable memory management.
+Ownership has been used in a few languages to support scalable memory management.
 Rather than allowing multiple threads to access a single object, we ensure that at most one thread can access an object at a time.
-This is fundamental in languages like [Pony](https://www.ponylang.io) and [Rust](https://www.rust-lang.org).
+This is fundamental in languages like [Pony](www.ponylang.io) and [Rust](https://www.rust-lang.org).
```
mjp41: Looks like you edit is reverting #28.

docs/explore.md
```diff
@@ -11,7 +11,7 @@ There are two distinct aspect to system programming:
   - Can treat memory directly as bits and bytes
   - Little or no abstraction on the hardware
 
-To implement many low-level systems, for instance a memory manager, you need raw access.
+To implement many low-level systems (for instance, a memory manager), you need raw access.
```
davidchisnall: If you're going to correct the punctuation here, it's probably also a good idea to correct the vocabulary.  'For instance' is informal and should be 'for example' in formal writing.

shiatsumat: OK, I'll change 'for instance' to 'for example'.
Rather, I found 'many low-level systems' vague here; 'a low-level system of various kinds' may be clearer?

docs/explore.md
shiatsumat: Oops, sorry and thank you for notifying this.

PR #32, opened 2020-01-17
Instructions to build on macOS.

NickNaso:
Added instructions to build the project on macOS.

### DIFF
--- a/docs/building.md
+++ b/docs/building.md
@@ -70,8 +70,9 @@ Furthermore, the `cmake` command is different:
 cmake .. -G "Visual Studio 15 2017 Win64"
 ```
 
+# Building on UNIX-like platform
 
-# Building on Linux
+## Prerequisites to build on Linux
 
 These steps were tested on Ubuntu 18.04.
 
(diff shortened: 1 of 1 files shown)


mjp41:
@plietar or @davidchisnall could you confirm these instructions for macOS.  Thanks

@NickNaso thanks for submitting this.

davidchisnall:
These are correct, but I don't think that we want to have a complete copy of the cmake instructions for every *NIX platform.  Please can you separate out the sections on installing dependencies for macOS and for Linux and then have a common section for building on UNIX-like platforms?

docs/building.md
```diff
@@ -84,6 +85,27 @@ sudo apt install cmake ninja-build python3 \
                  llvm
 ```
 
+## Prerequisites to build on macOS
+
+These steps were tested on macOS Catalina (10.15).
+
+First, you will need to install dependencies:
+
+- Install Python 3.x
```
plietar: FWIW Catalina seems to include Python 3 already.
Since previous versions don't, I'm fine with keeping this line.

docs/building.md
```diff
@@ -84,6 +85,27 @@ sudo apt install cmake ninja-build python3 \
...
+   installed, you can find them under the menu 
+   `Xcode -> Open Developer Tool -> More Developer Tools...`. This step will 
+   install `clang`, `clang++`, and `make`.
+- Install [Homebrew](https://brew.sh/)
+- Install `ninja` and `cmake` running the following command:
+```
+brew install ninja cmake
+```
+
+# Build instructions:
```
mjp41: Shouldn't this have an extra "#".  This is part of the UNIX-like section?

NickNaso: Yes, you right. I just fixed it.

PR #34, opened 2020-01-17
Add primitive types

plietar:
Primitive types are those which have a special meaning to the compiler / runtime. This replaces special type like `U64` and `cown` which we had before, removing a lot of bespoke code since they are regular Entity types.

Additionally these can now have both static and dynamic methods (see the binary operators on `U64`, or `cown.create`)

The PR has been carefully split into reasonably sized and standalone commits. It may be easier to review by going through those rather than the full PR. I'm happy to make distinct PRs if you prefer, but the earlier commits make little sense without the rest (eg. adding primitives but not using them).

As usual, GitHub pointlessly renders commits in chronological order instead of topological. Here's the more useful order:
- Correctly follow the calling convention on program entry.
- Add primitive entities
- Add a special keyword to identify builtins.
- Replace U64 by a primitive
- Allow dynamic method calls on integers
- Change binary operator desugaring.
- Make freeze a builtin function rather than a keyword.
- Turn cown into a primitive type.

### DIFF
--- a/src/compiler/ast.h
+++ b/src/compiler/ast.h
@@ -445,11 +445,6 @@ namespace verona::compiler
     const Entity* definition = nullptr;
   };
 
-  struct NewCownExpr : public Expression
-  {
-    ASTPtr<Expression> contents;
-  };
-
   struct Argument : public ASTContainer
   {
     ASTPtr<Expression> inner;
@@ -597,11 +592,6 @@ namespace verona::compiler
 
     // Added during resolution
     Symbol symbol = ErrorSymbol();
-  };
-
-  struct CownTypeExpr final : public TypeExpression
-  {
-    ASTPtr<TypeExpression> contents;
   };
 
   struct UnionTypeExpr final : public TypeExpression
--- a/src/compiler/ast.h
+++ b/src/compiler/ast.h
@@ -288,9 +288,11 @@ namespace verona::compiler
     {
       Class,
       Interface,
+      Primitive,
     };
     static constexpr Kind Class = Kind::Class;
     static constexpr Kind Interface = Kind::Interface;
+    static constexpr Kind Primitive = Kind::Primitive;
 
     ASTPtr<ASTConstant<Kind>> kind;
     ASTChild<Name> name;
--- a/src/compiler/ast.h
+++ b/src/compiler/ast.h
@@ -329,6 +329,7 @@ namespace verona::compiler
     {
       return name;
     }
+
     Kind kind() const
     {
       if (kind_)
--- a/src/compiler/ast.h
+++ b/src/compiler/ast.h
@@ -555,11 +555,6 @@ namespace verona::compiler
     ASTPtr<Expression> expr;
   };
 
-  struct FreezeExpr : public Expression
-  {
-    ASTPtr<Expression> expr;
-  };
-
   enum class BinaryOperator
   {
     Add,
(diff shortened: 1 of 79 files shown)


jemc:
This is great. Thanks for making the commits very easy to follow.

plietar:
For people with a Pony background, Pony has two uses for primitives, compiler-specific types (eg. U64) and singleton types (eg. None).

This only implements the former, as I've always disliked the confusion between the two anyway. I'm open to naming suggestions instead of `primitive`, especially if we want to add the singleton type meaning.

Potentially `builtin class U64 { ... }`.

src/compiler/ast.h
```diff
@@ -308,27 +307,56 @@ namespace verona::compiler
...
 
-    std::string instantiated_path(const Instantiation& instantiation) const;
+    std::string
+    instantiated_path(const Instantiation& instantiation) const final;
+
+    const Name& get_name() const final
+    {
+      return name;
+    }
+    Kind kind() const
```
mjp41: New line?

src/compiler/ast.h
```diff
@@ -269,14 +269,13 @@ namespace verona::compiler
...
   {
-    ASTChild<Name> name;
-
     /**
      * Pointer back to the Entity which contains this member.
      * Added during resolution
      */
     const Entity* parent = nullptr;
 
+    virtual const Name& get_name() const = 0;
```
mjp41: Why do you want this to be virtual? And why did you drop the common field from Member.  Is this due to Pegmatite restrictions on parsing in order of fields?

src/compiler/parser.cc
```diff
@@ -210,7 +210,10 @@ namespace verona::compiler
     Rule fn_signature = generics >> function_params >> -(":" >> type) >>
       -ExprPtr(where_clauses);
     Rule fn_body = block;
-    Rule method = def_ident >> fn_signature >> (fn_body | ";");
+
+    Rule method_builtin = "builtin"_E;
```
mjp41: Should a builtin have a body? I guess can give better error message if we parse it, and then say why we are unhappy?

src/compiler/parser.cc
```diff
@@ -83,7 +83,7 @@ namespace verona::compiler
 
     Rule keyword = term(
       "while"_E | "if" | "class" | "interface" | "primitive" | "var" | "unit" |
-      "match" | "String" | "iso" | "mut" | "imm" | "mut-view" | "in" | "cown" |
+      "match" | "String" | "iso" | "mut" | "imm" | "mut-view" | "in" |
```
mjp41: Can user program now define a cown type? I guess core library will define, an if we get two, then we error?

plietar: Yes and yes. If and when we get modules, it will be possible to define a type cown in a different namespace.

src/compiler/ast.h
plietar: Pegmatite requires fields to be laid out in the order they appear in the syntax, where fields of the base type come before fields of the derived class.

Previously both Field and Method started with their name, so it was put in Entity. But because Method now has a kind before name, we need to move the name into the derived class.

We want this virtual method so we can get the name of a Entity*.

plietar:
I don’t think all primitives will be imm in the future.
For example arrays could be implemented as a primitive, but we’ll want iso and mut arrays.

Similarly “linear promises” would be iso to make sure they can’t be shared.

mjp41:
I wonder if we might want a concept of an immutable primitive? A lot of the types are going to be in that category and having to write `& imm` is annoying, but it seems that would be easy to add near the front-end.

src/compiler/parser.cc
plietar: Yes. I've added a test for the error message.

Issue #35, opened 2020-01-17
Reachability analysis doesn't always visit receiver


lukecheeseman:
Compiling the following:

```
class Cell[T] {
  foo(self: mut) {}
}

class B {
  bar(): Cell[U64Obj] & iso {
    new Cell
  }

  baz() {
    var a = B.bar();
    (mut-view a).foo();
  }
}

class Main {
  main() {
    B.baz();
  }
}
```
gives the error:
```
terminate called after throwing an instance of 'std::out_of_range'
  what():  map::at
```

It looks like the type Cell hasn't been visited in the Reachabilty pass as the following doesn't cause the same issue:

```
baz() {
  baz() {
    var c = new Cell;
    B.refine(mut-view c);
    var a = B.bar();
    (mut-view a).foo();
  }
}
```

plietar:
This issue here is that the reachability pass has an invariant that the receiver type is always  visited before its methods. The rationale is that you can only call a method if you have a value, and you can only have a value if you've visited its type. It sort of worked so far, but apparently there's edge cases where it doesn't :(.

There's probably a way to fix that edge case, bu there's really no point keeping that assumption. @lukecheeseman has a simple fix coming soon.

plietar:
Even smaller test case:
```
class A {
  foo(self: iso) {}
}

class B {
  bar(): A & iso {
    new A
  }
}

class Main {
  main() {
    (B.bar()).foo();
  }
}
```

I'm amazed we hadn't hit this bug before given how trivial it is to trigger it.

PR #38, opened 2020-01-17
Reachability

lukecheeseman:
Fix bug https://github.com/microsoft/verona/issues/35


plietar:
Now we actually use an issue tracker, I think we should reference the issue number in test cases, either in the file name or as a comment. Helps understand what the test is doing. This is what [Rust does for example](https://github.com/rust-lang/rust/blob/master/src/test/codegen/issue-45466.rs).

PR #46, opened 2020-01-19
Support relative importing without requiring "./" in the input file path

0paIescent:
Currently the following code:
```rust
// foo.verona
use "bar.verona"
```
```java
// bar.verona
class Main {
  main() {}
}
```
and running `veronac foo.verona` will produce the following error:
```
Cannot open file ""/bar.verona""
```
You can only run a file with dependencies, from the current directory, if you prepend it with a `./`.
For example this, `veronac ./foo.verona`, works perfectly fine.

This PR makes it so that is no longer required.

### DIFF
--- a/src/compiler/main.cc
+++ b/src/compiler/main.cc
@@ -120,9 +120,11 @@ namespace verona::compiler
       // Add nested includes to the work list.
       for (auto& include : file->modules)
       {
-        auto directory = input_file.remove_filename();
-        directory += "/" + (std::string)*include;
-        work_list.push(directory);
+        auto new_input_file = input_file.remove_filename();
+        if (new_input_file.empty())
+          new_input_file = ".";
+        new_input_file += "/" + (std::string)*include;
+        work_list.push(new_input_file);
       }
 
       program->files.push_back(std::move(file));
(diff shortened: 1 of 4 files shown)


src/compiler/main.cc
```diff
@@ -120,9 +120,11 @@ namespace verona::compiler
       // Add nested includes to the work list.
       for (auto& include : file->modules)
       {
-        auto directory = input_file.remove_filename();
-        directory += "/" + (std::string)*include;
-        work_list.push(directory);
+        auto new_input_file = input_file.remove_filename();
+        if (new_input_file.empty())
+          new_input_file = ".";
+        new_input_file += "/" + (std::string)*include;
```
mjp41: I would sooner keep the directory name, but change this line to 
```
auto new_input_file = directory + "/" + (std::string)*include;
```

My original was not clear, but I think your change is unclear in a different way.

mjp41:
Also could you add a test case for this. Something like the following files:
```
testsuite\features\compiler-pass\use.verona
testsuite\features\compiler-pass\library\child.verona
testsuite\features\compiler-pass\library\sibling.verona
```
where `child.verona`
```Rust
use "sibling.verona"
```
and `use.verona` is
```Rust
use "library\child.verona"
```

0paIescent:
Of course, I appreciate the points in the right direction.

PR #47, opened 2020-01-19
Turkish description of the processes

04603aslan:
I added the Turkish description of the Swicth cycle


mjp41:
Sorry, we are not currently prepared to take comments in languages other than English.  We do not have the resources to check or maintain documentation in other languages.

PR #50, opened 2020-01-19
Fixes #11

### LINKED ISSUE #11: Errors in included file are not reported correctly.
```pony
// foo.verona
use "bar.verona"
```

```pony
// bar.verona
class A {
  m(x: T) { } // Use a non-existent type T
}
```

Compiling `foo.verona` will cause an error, as expected, but the location is incorrect:
```
$ ./dist/veronac foo.bar
foo.verona:0:6: error: Cannot find value for symbol 'T'
// foo.verona
     ^
1 error generated
```

The error should be at `bar.verona:3:8`, but `foo.verona:0:6` is reported instead.

0paIescent:
`extract_bits` is implemented correctly for everything except when it is used in `is_small_source_location`.

The current implementation of `is_small_source_location` is inconsistent in that the values it considers small are not contiguous; as it currently gets the value of the 30th bit.

For example:
```c++
uint32_t min = 0;
uint32_t max = -1; // 4,294,967,295

// some intermediate values
uint32_t quarter = 1 << 30;               // 1,073,741,824 -- 30th bit is set, while 31st is not
uint32_t half = 1 << 31;                  // 2,147,483,648 -- 31st bit is set, while 30th is not
uint32_t three_quarters = half + quarter; // 3,221,225,472 -- 30th and 31st bits are set

// verify that min is small and max is large
assert(is_small_source_location(min) == 0);
assert(is_small_source_location(max) == 1);

/**
 * The following ranges are inclusive on the beginning, and non inclusive on the end
 *
 * [min .. quarter]  and [half .. three_quarters]    are all considered small
 * [quarter .. half] and [three_quarters .. max + 1] are all considered large
 * 
 * Here are the full ranges for those interested
 * 
 * small -- [0             .. 1,073,741,824] and [2,147,483,648 .. 3,221,225,472]
 * large -- [1,073,741,824 .. 2,147,483,648] and [3,221,225,472 .. 4,294,967,296]
 */

// these are considered small
assert(is_small_source_location(quarter - 1) == 0);
assert(is_small_source_location(half) == 0);
assert(is_small_source_location(three_quarters - 1) == 0);

// and these are considered large
assert(is_small_source_location(quarter) == 0);
assert(is_small_source_location(half - 1) == 0);
assert(is_small_source_location(three_quarters) == 0);
```

To elaborate, after `1,073,741,824` source locations are created, everything from that point forward will be considered large. Until then, they will be written to `large_location_index` and `large_locations`, but the index returned from `make_source_location` will be interpreted as a small source location.

The solution as I see it is to extract bits from the 31st bit instead of the 30th, however this leads into another problem:
```c++
template<typename T, unsigned HighBit, unsigned Width>
static T extract_bits(T bitfield)
{
    ...
    static_assert((sizeof(T) * CHAR_BIT) > HighBit);
    ...
}
```

This assert requires the HighBit to be less than `32` in this case, therefore we can't pass in `discriminator_offset + 1`.

If we subtract one from `discriminator_offset` in `is_small_source_location`, as @mjp41 suggested, we get even further from the bit we actually want. Now we're getting the 29th bit:
```c++
return extract_bits<SourceLocation, discriminator_offset - 1, 1>(l) == 0;
```

The solutions that I've come up with are:
 * change `extract_bits` to `static_assert((sizeof(T) * CHAR_BIT) >= HighBit)` and `is_small_source_location` to `discriminator_offset + 1`
 * remove the offending assert entirely and change `is_small_source_location` to `discriminator_offset + 1`
 * reimplement `extract_bits` in `is_small_source_location` without the asserts
 * change `next_source_location` to
    ```
    (static_cast<SourceLocation>(1) << discriminator_offset) + (static_cast<SourceLocation>(1) << (discriminator_offset - 1))
    ```
    so its starting value will be the lowest value in the highest contiguous range (`3,221,225,472`), this however will lead to half the  expanded storage space

I'd rather not change the innards of `extract_bits` at all, because I don't know why that assert is there (if anyone can clarify that for me, I'd be grateful).

My recommendation is the third solution explained above:
```c++
const SourceLocation mask = (static_cast<SourceLocation>(1) << 1) - 1;
return (l >> discriminator_offset) & mask;

// OR since we already know that the mask is 1

return (l >> discriminator_offset) & static_cast<SourceLocation>(1);
```
\
On a somewhat unrelated note, I'm not very well versed in bit operations or C++, so any criticism is very welcome. This is also my first *real world* PR, so if you have any critiques about length, detail, etc., or if I've committed any PR sins, criticism is welcome there as well.


0paIescent:
Fixes #11

davidchisnall:
The static assert *should* be checking that we're not trying to extract bits that don't exist.  The indexes are from 0, so we don't want to extract a high bit that is past the end of the word (which would trigger UB from the shifts).  It is entirely possible that some of the arithmetic here is wrong!

davidchisnall:
The static assert *should* be checking that we're not trying to extract bits that don't exist.  The indexes are from 0, so we don't want to extract a high bit that is past the end of the word (which would trigger UB from the shifts).  It is entirely possible that some of the arithmetic here is wrong!

Looking at the code, I think there is an off-by-one error in that assert, because the sizes are one-indexed (maybe Verona should have distinct types for cardinal and ordinal numbers).  

I think the correct thing to do is fix the assert and add some more static asserts to check that we are correctly able to extract the top bit from a value.

davidchisnall:
I think this looks correct, but please can you add some more `static_assert`s to ensure that it is doing the right thing?

0paIescent:
Thanks for taking a look, but I haven't been able to think of any more `static_asserts` that would ensure something that hasn't already been ensured, I could be easily overlooking something though.

Since `(sizeof(T) * CHAR_BIT)` gives the size of storage, we guarantee with the current assertions that `(HighBit - Width)` will never be more than `(sizeof(T) * CHAR_BIT) - 1`.

davidchisnall:
Hi, I meant `static_assert`s for the specific corner cases that this should fix.  For example `static_assert(extract_bits<uint32_t, 31, 1>(1UL<<31) == 1)`.

davidchisnall:
Looks good to me, thank you for fixing this.  @mjp41, do you want to press the merge button?

0paIescent:
Of course, thanks for the pointers too, I appreciate it.

PR #53, opened 2020-01-20
Information on debugging the runtime/interpreter/language.

mjp41:
Some rough notes on how to debug the various components on the system.

Needs more details, but should be a good starting point.

### DIFF
--- a/docs/debugging.md
+++ b/docs/debugging.md
@@ -1,0 +1,130 @@
+# Debugging Project Verona
+
+We have built several features into the codebase to aid debugging.
+This document explains those features.
+If anything is unclear, please PR fixes or raise issues.
+
+We will split the rest of this document based on the various stages of running a Verona program
+
+* Compiler
+* Interpreter
+* Runtime
+
+## Debugging the compiler
+
+The simplist way to debug the compiler is to get it to dump all the phases of compiling every method:
+```
+veronac.exe --dump-path dmp region101.verona
+```
+This dumps all the stages of compiling each method into a sub-directory `dmp`.
+For each Class and Method, you should find:
+```
+  [Class].[Method].ast.txt
+  [Class].[Method].resolved-ast.txt
+  [Class].[Method].elaborated-ast.txt
+  [Class].[Method].wf-types.txt
+  [Class].[Method].ir.txt
+  [Class].[Method].liveness.txt
+  [Class].[Method].constraints.txt
+  [Class].[Method].infer.txt
+  [Class].[Method].solver.txt
+  [Class].[Method].substitution.txt
+  [Class].[Method].types.txt
+  [Class].[Method].typed-ir.txt
+  [Class].[Method].make-region-graph.txt
+  [Class].[Method].region-graph.txt
+```
+The order is the order of the phases in the compiler, and can be used to track where something has gone wrong.
+
+Individual, phases and results for particular methods can be output using other options to the compiler.
+
+### Debugging the parser
+
+The Verona parser currently uses [Pegmatite](https://github.com/CompilerTeaching/Pegmatite), a PEG parser design the teaching and rapid prototyping.
+This has two compile options that can aid debugging:
+
+* DEBUG_PARSING  - this outputs tracing during parsing the string
+* DEBUG_AST_CONSTRUCTION  - this outputs the construction of the AST nodes
+
+
+## Debugging the interpreter
+
+As the majority of code running in the interpreter is actually the runtime, the subsequent section could also prove useful.
+The interpreter provides a trace of all the instructions that it executes.
+Every option that you can pass to the interpreter, can also be passed to the compiler prepended with `run-`, for example:
+```
+interpreter foo.bc --verbose
+```
+or 
+```
+veronac foo.verona --run --run-verbose
+```
+The `--verbose` option basically provides a trace of the interpreters operations:
+for example:
+```
+veronac.exe --run --run-verbose region101.verona
+```
+produces
+```
+[ 27b]: Entering function Main.main, argc=1 retc=1 locals=9
+[ 28d]: LOAD_DESCRIPTOR r1, 0x1
+[ 293]: COPY r8, r1
+[ 296]: CALL 7, 0x1
+[ 296]:  Calling function Main.test1, base=r8 argc=1 retc=1 locals=37
+[ 12c]:  LOAD_DESCRIPTOR r1, 0x3
+...
+```
+
+Both the interpreter, and the compiler's run option, can also provide systematic testing for concurrency:
+```
+  veronac-sys.exe --run testsuite/demo/run-pass/dining_phil.verona --run-seed 100 --run-seed_upper 200
+```
+runs 100 seeds sequentially, but testing various interleavings of the runtime. The seeds are replayable.
+
+## Debugging the runtime
+
+The runtime provides two key features to aid debugging:
+
+* Systematic testing
+* Crash logging (A flight recorder that provides the last few seconds of logging)
+
+Both of these can produce high detail logging of the runtime operations.
+
+It is considerably easier to debug a crash in systematic testing, than in standard concurrent execution.
+The majority of tests of the runtime cover a large number of pseudo-random traces, so that the failures can easily be replayed.  The code in
+
+* https://github.com/microsoft/verona/blob/master/src/rt/test/harness.h#L38-L70
+* https://github.com/microsoft/verona/blob/master/src/rt/test/harness.h#L73-L114
+
+highlights all of the features for running with the various logging and random seed approaches.
+
+To compile the run-time with systematic testing build with
+```C++
+USE_SYSTEMATIC_TESTING
+```
+defined, and then call
+```C++
+Systematic::enable_logging();
+```
+to print the logging, and
+```C++
+Scheduler::get().set_seed(seed);
+```
+specifies the seed for the pseudo-random interleaving.
+The seed to interleaving should be consistent across platforms.
+
+To enable crash logging build with
+```C++
+USE_FLIGHT_RECORDER
+```
+defined, and call
+```C++
+Systematic::enable_crash_logging();
+```
+to enable its output.
+
+The systematic testing log has the earliest entry first, while the crash logging is in reverse, and contains timestamps.
+This is a quirk of how the crash log is reconstructed after the runtime crashes.
+
+
+Enabling ASAN on Clang builds by passing CMake `-DUSE_ASAN=On` will aid in detecting memory corruption issues.
--- a/docs/debugging.md
+++ b/docs/debugging.md
@@ -160,7 +160,7 @@ The first line comes from the following line:
 It prints the address of the `cown` and the `get_epoch_mark` on that object.
 
 If neither systematic testing or the crash log is enabled, then all the logging
-should be erased by the compiler.
+should be erased when compiling the runtime.
 
 ### ASAN
 
--- a/docs/debugging.md
+++ b/docs/debugging.md
@@ -140,7 +140,7 @@ Both logs use the same printing in the runtime:
 ```C++
 Systematic::cout() << "My messages" << ... << std::endl;
 ```
-For, crash log the pretty print functions are not run until the crash, so they
+For the crash log, the pretty print functions are not run until the crash, so they
 should not depend on any state other than the `uintptr_t` sized argument they
 are provided.  Here is an example log from systematic testing:
 ```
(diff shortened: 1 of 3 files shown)


mjp41:
@jemc could you let me know if any of this is useful to you.

I also welcome suggestions for more things to add/places you got confused.

docs/debugging.md
```diff
@@ -0,0 +1,130 @@
...
+  [Class].[Method].solver.txt
+  [Class].[Method].substitution.txt
+  [Class].[Method].types.txt
+  [Class].[Method].typed-ir.txt
+  [Class].[Method].make-region-graph.txt
+  [Class].[Method].region-graph.txt
+```
+The order is the order of the phases in the compiler, and can be used to track where something has gone wrong.
+
+Individual, phases and results for particular methods can be output using other options to the compiler.
```
plietar: I think the comma after "Individual" is misplaced. A more detailed note of how to use this would be neat. eg:

> Individual results can be printed to the standard output instead by passing the `--print=` flag to the compiler.
For example, the following command will print the intermediate representation for the `Main.main` method.

```
$ ./dist/veronac ../testsuite/demo/run-pass/bank1.verona --print=Main.main.ir
IR for Main.main:
  Basic block BB0:
    0 <- static Main
    1 <- integer 100
    2 <- call 0.example()
[...]

File: docs/debugging.md [hunk 2]
```diff
@@ -0,0 +1,130 @@
...
+
+### Debugging the parser
+
+The Verona parser currently uses [Pegmatite](https://github.com/CompilerTeaching/Pegmatite), a PEG parser design the teaching and rapid prototyping.
+This has two compile options that can aid debugging:
+
+* DEBUG_PARSING  - this outputs tracing during parsing the string
+* DEBUG_AST_CONSTRUCTION  - this outputs the construction of the AST nodes
+
```
plietar: Explain where this needs to be set. I guess this is a cmake flag, so `cmake -DDEBUG_PARSING=ON`?

docs/debugging.md
```diff
@@ -0,0 +1,130 @@
...
+interpreter foo.bc --verbose
+```
+or 
+```
+veronac foo.verona --run --run-verbose
+```
+The `--verbose` option basically provides a trace of the interpreters operations:
+for example:
+```
+veronac.exe --run --run-verbose region101.verona
```
plietar: Quite a detail, but you use `.exe` in some places but not others. Better to be consistent. I'd say no `.exe`, especially since I think windows doesn't require it.

docs/debugging.md
```diff
@@ -0,0 +1,130 @@
...
+* Systematic testing
+* Crash logging (A flight recorder that provides the last few seconds of logging)
+
+Both of these can produce high detail logging of the runtime operations.
+
+It is considerably easier to debug a crash in systematic testing, than in standard concurrent execution.
+The majority of tests of the runtime cover a large number of pseudo-random traces, so that the failures can easily be replayed.  The code in
+
+* https://github.com/microsoft/verona/blob/master/src/rt/test/harness.h#L38-L70
+* https://github.com/microsoft/verona/blob/master/src/rt/test/harness.h#L73-L114
```
plietar: Can you use URLs with commit hashes, otherwise the line numbers will be out-of-date very quickly.

docs/debugging.md
```diff
@@ -0,0 +1,130 @@
...
+
+highlights all of the features for running with the various logging and random seed approaches.
+
+To compile the run-time with systematic testing build with
+```C++
+USE_SYSTEMATIC_TESTING
```
plietar: Is this a CMake flag?

docs/debugging.md
mjp41: One has a CMake flag, and one does not in the pegamatite CMake.  I'll PR a fix, and then update the documentation.

docs/debugging.md
mjp41: Not yet.

docs/debugging.md
mjp41: Added

docs/debugging.md
mjp41: Fixed

docs/debugging.md
mjp41: Added CMake flags for these, and also for the Pegmatite options.

docs/debugging.md
```diff
@@ -0,0 +1,138 @@
...
+
+We will split the rest of this document based on the various stages of running a Verona program
+
+* Compiler
+* Interpreter
+* Runtime
+
+## Debugging the compiler
+
+The simplist way to debug the compiler is to get it to dump all the phases of compiling every method:
```
mhyee: Typo: simplist -> simplest

docs/debugging.md
```diff
@@ -0,0 +1,138 @@
...
+  [Class].[Method].solver.txt
+  [Class].[Method].substitution.txt
+  [Class].[Method].types.txt
+  [Class].[Method].typed-ir.txt
+  [Class].[Method].make-region-graph.txt
+  [Class].[Method].region-graph.txt
+```
+The order is the order of the phases in the compiler, and can be used to track where something has gone wrong.
+
+Individual results can be printed to the standard output instead by passing the --print= flag to the compiler.
```
mhyee: Backticks for `--print=`?

docs/debugging.md
```diff
@@ -0,0 +1,138 @@
...
+The seed to interleaving should be consistent across platforms.
+
+To enable crash logging you can pass CMake `-DUSE_CRASH_LOGGING=On`
+and in your program call
+```C++
+Systematic::enable_crash_logging();
+```
+to enable its output.
+
+Some build artifacts automatically have these settings enabled by default.
```
mhyee: Maybe list what these build artifacts are?

docs/debugging.md
```diff
@@ -0,0 +1,138 @@
...
+To enable crash logging you can pass CMake `-DUSE_CRASH_LOGGING=On`
+and in your program call
+```C++
+Systematic::enable_crash_logging();
+```
+to enable its output.
+
+Some build artifacts automatically have these settings enabled by default. 
+
+The systematic testing log has the earliest entry first, while the crash logging is in reverse, and contains timestamps.
```
mhyee: I think this should be emphasized somehow, because it's really easy to get wrong and then get really confused.

Actually, as a separate change, maybe the flight recorder dump could start with a note that the log is in reverse?

docs/debugging.md
```diff
@@ -0,0 +1,138 @@
...
+* [src/rt/test/harness.h/SystematicTestingHarness::run](https://github.com/microsoft/verona/blob/82a00c9d7e0e4f12b08e2b829a599a9ef94c1402/src/rt/test/harness.h#L73-L114)
+
+highlights all of the features for running with the various logging and random seed approaches.
+
+To compile the run-time with systematic testing you can pass CMake `-DUSE_SYSTEMATIC_TESTING=On`, 
+and then in your program call
+```C++
+Systematic::enable_logging();
+```
+to print the logging, and
```
mhyee: Mention that you can do print debugging with `Systematic::cout()`?

docs/debugging.md
```diff
@@ -0,0 +1,138 @@
...
+```
+to enable its output.
+
+Some build artifacts automatically have these settings enabled by default. 
+
+The systematic testing log has the earliest entry first, while the crash logging is in reverse, and contains timestamps.
+This is a quirk of how the crash log is reconstructed after the runtime crashes.
+
+
+Enabling ASAN on Clang builds by passing CMake `-DUSE_ASAN=On` will aid in detecting memory corruption issues.
```
mhyee: This is a little buried here, maybe making it a subsection (or subsubsection) can help.

docs/debugging.md
```diff
@@ -0,0 +1,138 @@
...
+runs 100 seeds sequentially, but testing various interleavings of the runtime. The seeds are replayable.
+
+## Debugging the runtime
+
+The runtime provides two key features to aid debugging:
+
+* Systematic testing
+* Crash logging (A flight recorder that provides the last few seconds of logging)
+
+Both of these can produce high detail logging of the runtime operations.
```
mhyee: This would be a lot of work, but walking through an example of debugging a crash in systematic testing would be super helpful. Or at the very least, showing a few lines of systematic output and explaining the convention (the "prefix" numbers represent threads, each object prints its address, etc.).

docs/debugging.md
```diff
@@ -0,0 +1,138 @@
...
+  Basic block BB0:
+    0 <- static Main
+    1 <- integer 100
+    2 <- call 0.example()
+[...]
+```
+
+### Debugging the parser
+
+The Verona parser currently uses [Pegmatite](https://github.com/CompilerTeaching/Pegmatite), a PEG parser design the teaching and rapid prototyping.
```
plietar: design the teaching -> designed for teaching?

docs/debugging.md
```diff
@@ -0,0 +1,138 @@
...
+The interpreter provides a trace of all the instructions that it executes.
+Every option that you can pass to the interpreter, can also be passed to the compiler prepended with `run-`, for example:
+```
+interpreter foo.bc --verbose
+```
+or 
+```
+veronac foo.verona --run --run-verbose
+```
+The `--verbose` option basically provides a trace of the interpreters operations:
```
plietar: interpreters -> interpreter’s

docs/debugging.md
```diff
@@ -0,0 +1,138 @@
...
+
+* [src/rt/test/harness.h/SystematicTestingHarness constructor](https://github.com/microsoft/verona/blob/82a00c9d7e0e4f12b08e2b829a599a9ef94c1402/src/rt/test/harness.h#L38-L70)
+* [src/rt/test/harness.h/SystematicTestingHarness::run](https://github.com/microsoft/verona/blob/82a00c9d7e0e4f12b08e2b829a599a9ef94c1402/src/rt/test/harness.h#L73-L114)
+
+highlights all of the features for running with the various logging and random seed approaches.
+
+To compile the run-time with systematic testing you can pass CMake `-DUSE_SYSTEMATIC_TESTING=On`, 
+and then in your program call
+```C++
+Systematic::enable_logging();
```
jemc: Just a small question for clarification: is any such logging possible without the runtime being compiled for systematic order of execution?

docs/debugging.md
```diff
@@ -0,0 +1,138 @@
...
+```
+to print the logging, and
+```C++
+Scheduler::get().set_seed(seed);
+```
+specifies the seed for the pseudo-random interleaving.
+The seed to interleaving should be consistent across platforms.
+
+To enable crash logging you can pass CMake `-DUSE_CRASH_LOGGING=On`
```
jemc: Does this option also require `-DUSE_SYSTEMATIC_TESTING=On`,or is it independent of that option?

docs/debugging.md
mjp41: It is all erased, if neither option is set.

docs/debugging.md
mjp41: Independent

docs/debugging.md
mjp41: I'll add a snippet, not sure I know how to explain that until I get some questions from people.

docs/debugging.md
mjp41: Good idea, I'll add a message at the top of the log.

docs/debugging.md
```diff
@@ -0,0 +1,169 @@
...
+Some build artifacts automatically have these settings enabled by default including some of the runtime unit tests, veronac-sys and interpreter-sys.
+
+The systematic testing log has the earliest entry first, while the crash logging is in reverse, and contains timestamps.
+This is a quirk of how the crash log is reconstructed after the runtime crashes.
+
+Both logs use the same printing in the runtime:
+```C++
+Systematic::cout() << "My messages" << ... << std::endl;
+```
+For, crash log the pretty print functions are not run until the crash, so they
```
jemc:
```suggestion
For the crash log, the pretty print functions are not run until the crash, so they
```

docs/debugging.md
```diff
@@ -0,0 +1,169 @@
...
+The rest of the line is the message that is being logged.
+The first line comes from the following line:
+```C++
+        Systematic::cout() << "Scheduled Cown: " << cown << " ("
+                           << cown->get_epoch_mark() << ")" << std::endl;
+```
+It prints the address of the `cown` and the `get_epoch_mark` on that object.
+
+If neither systematic testing or the crash log is enabled, then all the logging
+should be erased by the compiler.
```
jemc:
```suggestion
should be erased by the runtime.
```

I'm guessing you meant "runtime" and not "compiler" here, but feel free to ignore met if I'm wrong

PR #55, opened 2020-01-21
Adding a Channel example

lukecheeseman:
Building a channel from cowns, promises and queues.

### DIFF
--- a/testsuite/demo/run-pass/channel.verona
+++ b/testsuite/demo/run-pass/channel.verona
@@ -172,7 +172,9 @@ class Main {
     };
 
     Channel.write(channel, U64Obj.create(52));
-    // CHECK-L: Contents: 7
-    // CHECK-L: Contents: 52
+
+    // Check-dag as the output could appear in either order
+    // CHECK-DAG: Contents: 7
+    // CHECK-DAG: Contents: 52
   }
 }
--- a/testsuite/demo/run-pass/channel.verona
+++ b/testsuite/demo/run-pass/channel.verona
@@ -1,72 +1,15 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
 /*****************************************************************************
- * This example demonstrates how one could build a channel.
+ * This example demonstrates how one could build a channel in verona using
+ * cowns.
  *
- * It illustrates creating and using promises and cowns and communicating
- * between behaviours.
+ * As well as cowns, this implementation using promises to communicate between
+ * behaviours.
  *****************************************************************************/
 
 use "library/queue.verona"
-
-/*
- * Create a channel from two queues:
- *   - A queue of reads
- *   - A queue of writes
- *
- * Writing a value to the channel services the first waiting read
- * if one doesn't exist then the write is queued.
- *
- * Reading a value from the channel returns a promise of a value.
- * If a write has been queued then the read is immediately serviced,
- * otherwise the read is queued.
- */
-class Channel[T] {
-  _ws: Queue[Writer[T]] & iso;
-  _rs: Queue[Reader[T]] & iso;
-
-  create(): cown[Channel[T]] & imm {
-    var result = new Channel;
-    result._ws = Queue.create(0);
-    result._rs = Queue.create(1);
-    cown.create(result)
-  }
-
-  /*
-   * Write is a synchronous method.
-   * The method schedules a behaviour to log a write
-   * to the channel.
-   */
-  write(channel: cown[Channel[T]] & imm, value: T) {
-    var cell = Cell.create(value);
-    when(channel) {
-      match((mut-view (channel._rs)).remove()) {
-        var _: None => (mut-view (channel._ws)).add(Writer.create(cell)),
-        var r: Reader[T] => r.write(cell),
-      }
-    }
-  }
-
-  /*
-   * Read is an asynchronous method.
-   * The method returns the wait part of a promise
-   * that will be fulfilled when a value is available.
-   */
-  read(channel: cown[Channel[T]] & imm): cown[Cell[T]] & imm {
-    var r = Reader.create();
-    var handle = (mut-view r).await();
-    when(channel) {
-      match((mut-view (channel._ws)).remove()) {
-        var _: None => (mut-view (channel._rs)).add(r),
-        var w: Writer[T] => {
-          match (w.read()) {
-            var _: None => {} /* shouldn't be possible */
-            var c: Cell[T] => r.write(c),
-          }
-        }
-      }
-    };
-    handle
-  }
-}
 
 /*
  * A wrapper class around data, this allows data to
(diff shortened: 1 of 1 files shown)


testsuite/demo/run-pass/channel.verona
```diff
@@ -0,0 +1,156 @@
+/*****************************************************************************
+ * This example demonstrates how one could build a channel.
+ *
```
sophiaIC: This example demonstrates how one could build channels in Verona using cowns

testsuite/demo/run-pass/channel.verona
```diff
@@ -0,0 +1,156 @@
...
+ * It illustrates creating and using promises and cowns and communicating
+ * between behaviours.
```
sophiaIC: Does it?

mjp41:
CI requires you put at least one
```C++
   // CHECK: ...
```
or
```C++
   //CHECK-L: ... 
```
Preferably, add something that actually checks it does what you expect, if possible.

testsuite/demo/run-pass/channel.verona
sophiaIC: As well as cowns, the implementation uses promises to communicate between behaviours.

testsuite/demo/run-pass/channel.verona
```diff
@@ -0,0 +1,156 @@
...
+ *****************************************************************************/
+
+use "library/queue.verona"
+
+/*
+ * Create a channel from two queues:
+ *   - A queue of reads
```
mjp41: Are both queues required at the same time? Or could you play the same trick as in `Scheduler`?  It is fine to be different, just want to understand.

testsuite/demo/run-pass/channel.verona
```diff
@@ -0,0 +1,156 @@
...
+  create(data: Cell[T] & iso): Writer[T] & iso {
+    var result = new Writer;
+    result._data = data;
+    result
+  }
+
+  read(self: iso): (Cell[T] & iso) | (None & imm) {
+    self._data = Builtin.freeze(new None)
+  }
+}
```
mjp41: I would like these classes to appear above Channel.  So I can read it from top to bottom and understand the types.

testsuite/demo/run-pass/channel.verona
```diff
@@ -0,0 +1,156 @@
...
+  /*
+   * Write is a synchronous method.
+   * The method schedules a behaviour to log a write
+   * to the channel.
+   */
+  write(channel: cown[Channel[T]] & imm, value: T) {
+    var cell = Cell.create(value);
+    when(channel) {
+      match((mut-view (channel._rs)).remove()) {
+        var _: None => (mut-view (channel._ws)).add(Writer.create(cell)),
```
mjp41: Add comments to explain the two cases. 
```
// No readers are waiting, so add this `value` to the end of the queue
```
and
```
// Fulfill the longest waiting reader.
```

testsuite/demo/run-pass/channel.verona
```diff
@@ -0,0 +1,156 @@
...
+   */
+  read(channel: cown[Channel[T]] & imm): cown[Cell[T]] & imm {
+    var r = Reader.create();
+    var handle = (mut-view r).await();
+    when(channel) {
+      match((mut-view (channel._ws)).remove()) {
+        var _: None => (mut-view (channel._rs)).add(r),
+        var w: Writer[T] => {
+          match (w.read()) {
+            var _: None => {} /* shouldn't be possible */
```
mjp41: Perhaps elaborate on why?

testsuite/demo/run-pass/channel.verona
```diff
@@ -0,0 +1,156 @@
...
+   * Read is an asynchronous method.
+   * The method returns the wait part of a promise
+   * that will be fulfilled when a value is available.
+   */
+  read(channel: cown[Channel[T]] & imm): cown[Cell[T]] & imm {
+    var r = Reader.create();
+    var handle = (mut-view r).await();
+    when(channel) {
+      match((mut-view (channel._ws)).remove()) {
+        var _: None => (mut-view (channel._rs)).add(r),
```
mjp41: Explain two cases here.

testsuite/demo/run-pass/channel.verona
```diff
@@ -0,0 +1,156 @@
...
+  }
+
+  /*
+   * Read is an asynchronous method.
+   * The method returns the wait part of a promise
+   * that will be fulfilled when a value is available.
+   */
+  read(channel: cown[Channel[T]] & imm): cown[Cell[T]] & imm {
+    var r = Reader.create();
+    var handle = (mut-view r).await();
```
mjp41: This is nice as the code can return before this has even entered the promise in the queue, explain this.

testsuite/demo/run-pass/channel.verona
```diff
@@ -0,0 +1,156 @@
...
+
+    when(var received = Channel.read(channel)) {
+      match (received.extract()) {
+        var _: None => Builtin.print("Empty\n"),
+        var o: U64Obj => Builtin.print1("Contents: {:#}\n", o),
+      }
+    };
+
+    Channel.write(channel, U64Obj.create(52));
+  }
```
mjp41: Add 
// CHECK-L: Contents: 7
// CHECK-L: Contents 52

testsuite/demo/run-pass/channel.verona
lukecheeseman: You can play the same trick, I experimented with the single queue. This is, in some ways, just to be different but also the implementation with a single queue makes the read/write methods larger. The extra logic for changing the queue type distracts a little from the meaning of the read/write methods.

testsuite/demo/run-pass/channel.verona
```diff
@@ -0,0 +1,180 @@
...
+    var channel = Channel.create();
+
+    /* write a message into the channel */
+    Channel.write(channel, U64Obj.create(7));
+
+    /* wait for a message to be received on the channel */
+    when(var received = Channel.read(channel)) {
+      match (received.extract()) {
+        var _: None => Builtin.print("Empty\n"),
+        var o: U64Obj => Builtin.print1("Contents: {:#}\n", o.v),
```
mjp41: The DAG aspect is interesting.  Could you change this to 
"Contents 1:"
and the subsequent one to "Contents 2:"

I believe you would then have 
```
// CHECK-DAG: Contents 1: 7
// CHECK-DAG: Contents 2: 52
```
But wanted to clarify.  Also, refining the test will help with understanding.

lukecheeseman: Yes, that would be the expected output. Thanks, those changes are a good idea.

Although, looking at the CI again, it uses OutputCheck and not FileCheck so I can't use CHECK-DAG.

mjp41: You could just check one of them?

PR #66, opened 2020-01-27
Fix the "extract in finaliser" pattern.

### LINKED ISSUE #62: Finaliser runs too early
The follow program:
```
use "list.verona"

class A {
  _id: U64 & imm;

  create(id: U64 & imm): A & iso {
    var result = new A;
    result._id = id;
    result
  }

  final(self: mut) {
    Builtin.print1("Finalising A: {:#}\n", self)
  }
}

class Cell {
  _pool: cown[Pool] & imm;
  _f: A & iso;

  create(pool: cown[Pool] & imm, f: A & iso): Cell & iso {
    var result = new Cell;
    result._pool = pool;
    result._f = f;
    result
  }

  final(self: mut) {
    Builtin.print1("Finalising Cell: {:#}\n", self);
    Pool.return(self._pool, self._f = A.create(1));
  }
}

class Pool {
  _f: A & iso;

  create(): cown[Pool] & imm {
    var result = new Pool;
    cown.create(result)
  }

  return(pool: cown[Pool] & imm, f: A & iso) {
    Builtin.print1("Returning A: {:#}\n", mut-view f);
    when (var p = pool) {
      // FIXME: at this point A stops existing
      Builtin.print1("A: {:#}\n", mut-view f);
      p._f = f;
    }
  }
}

class Main {
  main() {
    var pool = Pool.create();
    var c = Cell.create(pool, A.create(0));
  }
}
```
When run gives the following output:
```
Finalising Cell: Cell { cown[Pool](0x7fde1b010020), A { 0 } }
Returning A: A { 0 }
Finalising A: A { 0 }
fish: '../verona/build/dist/veronac ./…' terminated by signal SIGSEGV (Address boundary error)
```
The value passed to `return` is finalised before it's passed into the behaviour.

plietar:
This adds a desctructor to every object. The destructor runs just before an object is deallocated. The VM uses this destructor to deallocate the array of fields, allowing us to move the "find iso fields" phase to after the finaliser has run.

Fixes #62.

### DIFF
--- a/src/interpreter/object.cc
+++ b/src/interpreter/object.cc
@@ -22,14 +22,27 @@ namespace verona::interpreter
   {
     rt::Descriptor::size = sizeof(VMObject);
     rt::Descriptor::trace = VMObject::trace_fn;
-    rt::Descriptor::trace_possibly_iso = VMObject::trace_fn;
-    rt::Descriptor::finaliser = VMObject::finaliser_fn;
+
+    // Try to be on the trivial ring as much as possible. This requires the
+    // following three methods to be null.
+    //
+    // If there are no fields, then VMObject's destructor is a no-op and we can
+    // skip it. In that case, there are also obviously no iso fields.
+    rt::Descriptor::destructor =
+      field_count > 0 ? VMObject::destructor_fn : nullptr;
+    rt::Descriptor::trace_possibly_iso =
+      field_count > 0 ? VMObject::trace_fn : nullptr;
+    rt::Descriptor::finaliser =
+      finaliser_ip > 0 ? VMObject::finaliser_fn : nullptr;
   }
 
-  VMObject::VMObject(VMObject* region)
-  : fields(std::make_unique<FieldValue[]>(descriptor()->field_count)),
-    parent_(region)
-  {}
+  VMObject::VMObject(VMObject* region) : parent_(region)
+  {
+    if (descriptor()->field_count > 0)
+      fields = std::make_unique<FieldValue[]>(descriptor()->field_count);
+    else
+      fields = nullptr;
+  }
 
   VMObject* VMObject::region()
   {
@@ -58,9 +71,12 @@ namespace verona::interpreter
   void VMObject::finaliser_fn(rt::Object* base_object)
   {
     VMObject* object = static_cast<VMObject*>(base_object);
-
     VM::execute_finaliser(object);
+  }
 
+  void VMObject::destructor_fn(rt::Object* base_object)
+  {
+    VMObject* object = static_cast<VMObject*>(base_object);
     (object)->~VMObject();
   }
 }
--- a/src/interpreter/vm.cc
+++ b/src/interpreter/vm.cc
@@ -67,30 +67,28 @@ namespace verona::interpreter
   void VM::execute_finaliser(VMObject* object)
   {
     const VMDescriptor* descriptor = object->descriptor();
-    uint32_t finaliser_ip = descriptor->finaliser_ip;
-    if (finaliser_ip != 0)
-    {
-      auto vm = VM::local_vm;
-      vm->trace("Finaliser for: {}", descriptor->name);
+    assert(descriptor->finaliser_ip > 0);
 
-      auto old_halt = vm->halt_;
-      vm->halt_ = false;
+    auto vm = VM::local_vm;
+    vm->trace("Finaliser for: {}", descriptor->name);
 
-      // Set up a new frame for finaliser
-      vm->write(Register(vm->frame_.locals - 1), Value::mut(object));
-      vm->call(finaliser_ip, (uint8_t)1);
+    auto old_halt = vm->halt_;
+    vm->halt_ = false;
 
-      // Save call stack, so we can jump back into normal execution.
-      auto backup = std::move(vm->cfstack_);
+    // Set up a new frame for finaliser
+    vm->write(Register(vm->frame_.locals - 1), Value::mut(object));
+    vm->call(descriptor->finaliser_ip, (uint8_t)1);
 
-      // Run finaliser to completion
-      vm->dispatch_loop();
+    // Save call stack, so we can jump back into normal execution.
+    auto backup = std::move(vm->cfstack_);
 
-      // Put back normal execution
-      vm->cfstack_ = std::move(backup);
-      vm->opcode_return();
-      vm->halt_ = old_halt;
-    }
+    // Run finaliser to completion
+    vm->dispatch_loop();
+
+    // Put back normal execution
+    vm->cfstack_ = std::move(backup);
+    vm->opcode_return();
+    vm->halt_ = old_halt;
   }
 
   void VM::grow_stack(size_t size)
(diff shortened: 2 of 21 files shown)


mjp41:
@mhyee could you take a look at this.  This is very close to the code from your internship, so this is probably very fresh for you.

mjp41:
I don't see where this code calls destructors on arena objects that aren't large or have custom finalisers.

I think the choice is right for objects with destructors to end up in the plain ring for trace regions, but for arena regions it seems we actually want them in the finaliser end of the region, so objects without destructors or finalisers can be collected in bulk.  Otherwise, we will need to look at the whole region for dispose making it much more costly.

mjp41:
I guess there is a design choice here.  If an object is complex in terms of its allocation, like the VM Objects with a second allocation for there fields, or when we add resizeable arrays, where again it will probably be two allocations.  Should the second allocation be in the arena?

Is it the case that "destructors" are only allowed to deallocate objects and nothing else?  Can they close file handles, or should that be the "finaliser"?

jemc:
Could some documentation be included (for example, some code comments in `object.h` or elsewhere) that explains concretely the difference between a "finaliser" and a "destructor"; for example, to explain what each one is responsible for, when they run, and when to use one vs the other.

I know that there is some context on the need for this in issue #62, but it would be good if the repo itself had the documentation needed to understand the need for this, so that runtime API consumers can readily understand the fields they are responsible for populating in the descriptor without tracking it down in tickets outside of the code/docs.

src/rt/region/region_trace.h
```diff
@@ -538,6 +541,14 @@ namespace verona::rt
       current_memory_used += size;
     }
 
+    static bool needs_finaliser_ring(Object* o)
+    {
+      // Objects with just a destructor don't need much special treatment, so
```
mhyee: Can you also update the comment at the top of this file?

src/rt/region/region_arena.h
```diff
@@ -578,6 +580,21 @@ namespace verona::rt
       dealloc(alloc);
     }
 
+    /*
+     * Unlike traced regions, objects with a destructor but no finaliser are
+     * placed with the objects that do have a finaliser. This avoids having to
+     * walk all the objects on deallocation to find the ones with destructors.
+     */
+    static bool needs_finaliser_area(const Descriptor* desc)
```
mhyee: I don't think we ever agreed on what to call the finaliser and non-finaliser... areas... of an arena. My only nitpick with `area` is that it looks too much like `arena`. (`non-finaliser` is also a bit of a mouthful to say, and I guess it's no longer true, now that objects without finalisers but with destructors go into the other end.)

src/rt/region/region_arena.h
```diff
@@ -578,6 +580,21 @@ namespace verona::rt
       dealloc(alloc);
     }
 
+    /*
+     * Unlike traced regions, objects with a destructor but no finaliser are
```
mhyee: Can you update the comment/diagram for the Arena class?

src/rt/region/region_arena.h
mjp41: I think we had discusses the C++ non-trivially destructable.  But that is even more of a mouthful.

Rust uses the "Drop" term, so we could perhaps TrivialDrop and NonTrivialDrop.

src/interpreter/object.cc
```diff
@@ -22,8 +22,16 @@ namespace verona::interpreter
   {
     rt::Descriptor::size = sizeof(VMObject);
     rt::Descriptor::trace = VMObject::trace_fn;
-    rt::Descriptor::trace_possibly_iso = VMObject::trace_fn;
-    rt::Descriptor::finaliser = VMObject::finaliser_fn;
+    rt::Descriptor::destructor = VMObject::destructor_fn;
```
mjp41: Could you use `is_trivially_destructible` trait here on VMObject to decide if we should have a destructor or not.  This will make the code more future proof.

src/interpreter/vm.cc
```diff
@@ -67,30 +67,28 @@ namespace verona::interpreter
   void VM::execute_finaliser(VMObject* object)
   {
     const VMDescriptor* descriptor = object->descriptor();
-    uint32_t finaliser_ip = descriptor->finaliser_ip;
-    if (finaliser_ip != 0)
-    {
-      auto vm = VM::local_vm;
-      vm->trace("Finaliser for: {}", descriptor->name);
+    assert(descriptor->finaliser_ip > 0);
```
mjp41: Nice change to make this code non-conditional by making the VM do the conditional work.

plietar:
> Does src/rt/cpp/vobject.h need updating?

Currently vobject runs the C++ destructor in the verona finaliser. We could switch it to run as the verona destructor, and instead allow implementations to define a `finaliser` method.

mjp41:
> > Does src/rt/cpp/vobject.h need updating?
> 
> Currently vobject runs the C++ destructor in the verona finaliser. We could switch it to run as the verona destructor, and instead allow implementations to define a `finaliser` method.

I think that makes sense from the runtime implementation.  As long as the C++ destructor only does C++ stuff, and doesn't do Verona stuff.  Previously, @davidchisnall and @theodus built a C++ wrapper classes to handle various forms of references, so then they performed Verona decrefs and region reclaims in destructors.  I think this semantics is not something we should support.

plietar:
I’ve run into more difficulty than I expected, both in terms of edge cases and in trying to reduce the confusion in how rings/areas are named, so this might take me a bit longer to finish.

Meanwhile, I’ve found at least one existing C++ test that could benefit from this. The [Philosopher class](https://github.com/microsoft/verona/blob/master/src/rt/test/func/diningphilosophers/diningphil.cc#L58) has a vector of cowns, which currently gets deallocated in the finaliser.

There’s some tricks in place to work around this, but it should become simpler once we have proper destructors. It’s technically UB to call methods on the object after the C++ destructor has run, so a Sufficiently Smart Compiler could optimize the `!final` condition in the trace method (but neither clang nor gcc are that smart).

plietar:
> As long as the C++ destructor only does C++ stuff, and doesn't do Verona stuff.

I think the finaliser must always leave the object in a “valid” state, as other finalisers running later could traverse the already finalised object and read its fields.

This means (if the RS isn’t used), decrementing reference counts must be done in the destructor, not the finaliser.

Obviously if you know the whole region topology you can play fast and loose (which the C++ tests do a lot), but in the Verona language we’ll have to stay safe.

plietar:
I've finally gotten around to finishing this:
- The traced and arena regions use the same criteria to segregate objects.
- Objects are now categorized as trivial / non-trivial. An object is trivial if it has no finaliser, no destructor and no iso field.
- Traced regions need three iterations of the non-trivial ring. This is because all finalisers must run before we can start looking for sub-regions, and we must find all the sub-regions before we can start deallocating anything. The latter is because we look at the header bits of the object pointed by the field to decide whether it is a child region. If we had enough type information from the compiler, we could maybe merge the last two iterations.
- The cown leak detector now looks at every object, not just the ones on the non-trivial ring. This allows fewer objects to appear on that ring, optimizing the common non-LD code paths.

PR #68, opened 2020-01-29
Fix issue #67

### LINKED ISSUE #67: Reachability of variable renaming type
The following crashes the compiler:
```
class B {
  foo(self: mut) {}
}

class Main {
  main() {
    var b = new B;
    var mutb = mut-view b;
    while (1) {
      mutb.foo();
    };
  }
}
```
with error:
```
Unhandled case N6verona8compiler20VariableRenamingTypeE in visitor N6verona8compiler19ReachabilityVisitor16CallReachabilityE.
```

lukecheeseman:
Loops introduce variable renamings and their types. The
reachability pass doesn't need to do anything here but
that visitor needs a variable renaming type case.

Fixes #67

### DIFF
--- a/src/compiler/codegen/reachability.cc
+++ b/src/compiler/codegen/reachability.cc
@@ -399,6 +399,9 @@ namespace verona::compiler
         visit_types(ty->elements);
       }
 
+      void visit_variable_renaming_type(const VariableRenamingTypePtr& ty) final
+      {}
+
       void visit_capability(const CapabilityTypePtr& ty) final {}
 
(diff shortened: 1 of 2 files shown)


plietar:
I’m surprised this is needed. VariableRenaming types are only needed if there’s an inference variable, otherwise they are simplified straight away.

By the time inference is done all inference variables should have been erased already.

What does the typed IR look for this function?

plietar:
I'll investigate later what's going on here, as it looks a bit suspicious, but the code change itself looks fine.

PR #69, opened 2020-01-29
Remove uninitialized field from rt test harness

Theodus:
The `seed` field was uninitialized on the construction of `SystematicTestHarness`. This has resulted in the use of uninitialized values for the xoroshiro PRNG in multiple runtime tests.

### DIFF
--- a/src/rt/test/func/cowngc1/cowngc1.cc
+++ b/src/rt/test/func/cowngc1/cowngc1.cc
@@ -64,6 +64,15 @@ struct PRNG
     return rand();
 #endif
   }
+
+  void seed(size_t seed)
+  {
+#ifdef USE_SYSTEMATIC_TESTING
+    return rand.set_state(seed);
+#else
+    return rand.seed(seed);
+#endif
+  }
 };
 
 static constexpr uint64_t others_count = 3;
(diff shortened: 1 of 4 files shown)


mjp41:
@Theodus  thanks for finding this.  So I really want systematic testing to have the following property
```
my_test --seed 100 --seed_upper 150
```
If this fails on 122 say, then I want 
```
my_test --seed 122
```
to be exactly the same run.  I think the use of the PRNG here is persisting across tests, and that will not satisfy this property.

I had some tests that took the harness as the parameter, something like this:
```C++
int main(int argc, char** argv)
{
  SystematicTestHarness harness(argc, argv);
 
  size_t ring = harness.opt.is<size_t>("--ring", 10);
  uint64_t forward = harness.opt.is<uint64_t>("--forward", 10);

  harness.run(test_cown_gc, forward, ring, &harness);
  harness.run(test_cown_gc_before_sched);

  return 0;
}
```
and for us to add 
```C++
void test_cown_gc(uint64_t forward_count, size_t ring_size, SystematicTestingHarness* harness)
{
  PRNG rand(harness->seed);
  rcown_first = nullptr;
  auto a = new RCown(ring_size, forward_count);
  Cown::schedule<Ping>(a, a, rand);
} 
```
as a field of the `harness`.  Perhaps, make `seed` private and have an accessor that faults if you access it outside a `run`.

I think one test does it this way now:
https://github.com/microsoft/verona/blob/5ad39dbb9f53bbd08c1e6f7c87bbad9f0a439d71/src/rt/test/func/diningphilosophers/diningphil.cc#L148-L163

mjp41:
LGTM

PR #70, opened 2020-01-30
Made Verona CI build output stack traces on crashs

mjp41:
This makes the veronac and interpreter output exceptions
and not produce a dialog box on Windows for debugging
when inside a CI build.


mjp41:
There seems to be an intermittent fault in the
```
testsuite\feature\run-pass\when.verona
```
example.  I am hoping this will give a clue, and help with future issues.

PR #71, opened 2020-01-30
Cmake now uses a flag rather than environment variables

Yuhanun:
Applied the proper cmake flags to set the C and C++ compiler.

```
CC=...
CXX=...
```

These environment variables can have side effects for the rest of the system, it's better to have cmake use it for 1 specific build.


mjp41:
Thanks, @Yuhanun

PR #74, opened 2020-01-31
Pools

lukecheeseman:
An example of how one may build a Pool of resources. This allows requests of N resources of the same type. The requests are fulfilled when there are enough resources available and the resources are automatically returned to the Pool when they are no longer being used.


testsuite/demo/run-pass/library/pool.verona
```diff
@@ -0,0 +1,235 @@
...
+    };
+  }
+
+  /*
+   * Check the first request in the queue, if it can be satisified then
+   * fulfill the requests and schedule a behaviour to check the next
+   * waiting request.
+   */
+  _check_requests(pool: cown[Pool[T]] & imm) {
+    when (var p = pool) {
```
mjp41: So all calls to `_check_requests` are when the pool is already owned by the caller.  Is it important to repeatedly acquire the `pool` here.  Does this improve fairness?  It seems like it is creating more `when`s without improving anything algorithmically.

testsuite/demo/run-pass/pool_example.verona
```diff
@@ -0,0 +1,70 @@
...
+    when (var group = Pool.request(p, 3)) {
+      Builtin.print("-----------------------------------\n");
+      Builtin.print1("Got group of size: {:#}\n", (mut-view group).size());
+      Main.print_resources((mut-view group).extract());
+    };
+
+    // For lack of a better check
+    // CHECK-L: Got group of size: 3
+
+    // FIXME: Requesting too many resources will currently cause
```
mjp41: Can you elaborate?  Do we need to fix the runtime, the compiler, the interpreter, or your example?

rengolin:
@lukecheeseman @mjp41, this is an old PR that only changes test files. It seems that the idea is good and we can make it work with an updated tree. Is it worth the work or should be just close this PR?

rengolin:
This PR is too old, with an unnecessary revert in it and adds tests that don't pass. They're still available in Luke's `group` branch, so it should be easy to resubmit once it's in better shape.

Issue #72, opened 2020-01-30
Interpreter crash when interpreting sourcefile


Yuhanun:
Title says a lot:

```cpp
 ~/g/verona_test  verona-interpreter main.verona 
terminate called after throwing an instance of 'std::logic_error'
  what():  Instruction overflow 6 8307
fish: “verona-interpreter main.verona” terminated by signal SIGABRT (Abort)
```

```verona
class Main {
    main() {
       Builtin.print2("Hello World!"); 
    }
}
```

Yuhanun:
I'm aware that you ould use 
```
veronac main.verona --output a.out
```

And then use the interpreter to interpret the bytecode, but still :)

mjp41:
The interpreter runs on bytecode produced by the compiler:

```
veronac main.verona --output main.bc
interpreter main.bc
```

You are trying to interpet the source as bytecode.  We haven't put checks in for that yet.

Your program should be 
```
Builtin.print("Hello World!"); 
```
as there are no arguments other than the format.

Yuhanun:
Ah okay, thank you :) 

General question, I'd like to contribute to the project, is this possible or is this meant to be more of a "microsoft employee"-only thing?

mjp41:
We are definitely keen to build a community and it is definitely not a Microsoft employee project: @plietar has the largest changes since we open sourced and is a student at Imperial College London.

But at this stage it we don't have many issues that are easy to give to new people.  As the project grows, I hope that will change, and we will flag issues as good for new people.

Thanks for your interest.

Yuhanun:
Alright, Is there any way you could tell me if there's an issue that someone new to the project could pick up? 

I have experience in C++, not in language development, however I'd love to help out.

plietar:
We should add some magic number at the start of the bytecode, and give a nice error message if it isn't found.

Yuhanun:
> We should add some magic number at the start of the bytecode, and give a nice error message if it isn't found.

I'm fairly confident that's how python identifies versions in their bytecode (pyc) files, it starts with some magic number that indicates the version.

mjp41:
@yuhanun If you want to get involved adding that to the bytecode format would be a useful feature.

orcmid:
You might want to look into some representation of a prefix and elements that also allow for verification of the integrity of a pyc file, with the prospect of a digital signature.   You want to know it is likely a pyc and also the version, I suspect.

[...]

PR #76, opened 2020-02-03
Adds magic numbers and versions into bytecode

Yuhanun:
Fixes issue #72 

Format:
4 bytes: 4 random numbers I generated, in this case: 0xB0, 0x45, 0x46, 0xBE
4 bytes: Major version (0 right now)
4 bytes: Minor version (0 right now)
4 bytes: Build version (0 right now)

This would allow us to run separate versions of functions when running the VM for specific versions of bytecode.

Any changes, when requested, will be applied.

Everything has been tested and documented, however no unittests were written.

### DIFF
--- a/src/compiler/codegen/codegen.cc
+++ b/src/compiler/codegen/codegen.cc
@@ -18,6 +18,15 @@ namespace verona::compiler
   using bytecode::Opcode;
   using bytecode::SelectorIdx;
 
+  /**
+   * Randomly generated verona bytes :)
+   */
+  constexpr static const char* verona_bytes = "\xB0\x45\x46\xBE";
+
+  constexpr static uint32_t MAJOR_VERSION = 0;
+  constexpr static uint32_t MINOR_VERSION = 0;
+  constexpr static uint32_t BUILD_VERSION = 0;
+
   bool is_valid_main_signature(Context& context, const FnSignature& signature)
   {
     return signature.generics->types.empty() && signature.receiver == nullptr &&
@@ -92,6 +101,32 @@ namespace verona::compiler
     return std::make_pair(class_item, method_item);
   }
 
+  void write_magic_bytes(std::vector<uint8_t>& vector)
+  {
+    for (uint8_t i = 0; i < 4; i++)
+    {
+      vector.push_back(verona_bytes[i]);
+    }
+
+    auto* major_version = reinterpret_cast<uint8_t*>(MAJOR_VERSION);
+    auto* minor_version = reinterpret_cast<uint8_t*>(MINOR_VERSION);
+    auto* build_version = reinterpret_cast<uint8_t*>(BUILD_VERSION);
+
+    for (size_t i = 0; i < 4; i++)
+    {
+        vector.push_back(major_version[i]);
+    }
+
+    for (size_t i = 0; i < 4; i++)
+    {
+        vector.push_back(minor_version[i]);
+    }
+    for (size_t i = 0; i < 4; i++)
+    {
+        vector.push_back(build_version[i]);
+    }
+  }
+
   std::vector<uint8_t> codegen(
     Context& context, const Program& program, const AnalysisResults& analysis)
   {
@@ -100,6 +135,11 @@ namespace verona::compiler
       return {};
 
     std::vector<uint8_t> code;
+
+    // TODO: write bytes
+
+    write_magic_bytes(code);
+
     Generator gen(code);
 
     Reachability reachability = compute_reachability(
--- a/src/compiler/codegen/codegen.cc
+++ b/src/compiler/codegen/codegen.cc
@@ -13,6 +13,8 @@
 #include "ds/helpers.h"
 #include "interpreter/bytecode.h"
 
+#include <array>
+
 namespace verona::compiler
 {
   using bytecode::Opcode;
--- a/src/compiler/codegen/codegen.cc
+++ b/src/compiler/codegen/codegen.cc
@@ -21,10 +21,21 @@ namespace verona::compiler
   /**
    * Randomly generated verona bytes :)
    */
-  constexpr static const char* verona_bytes = "\xB0\x45\x46\xBE";
+  constexpr static const std::array<uint8_t, 4> VERONA_BYTES = {0xB0, 0x45, 0x46, 0xBE};
 
+  /**
+   * Major verona version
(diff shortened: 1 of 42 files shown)


Yuhanun:
What version is macOS and windows on and why do they not conform with the standard?

Yuhanun:
Nevermind, forgot to include a header

plietar:
Thanks for tackling this! A few points:
- Could you please update the comment at the top of `bytecode.h`, in the program header section.
- There should be only one definition of the magic bytes. `bytecode.h` is a good place to put that, as that’s where all the shared stuff already is.
- It’s simpler to represent the magic bytes as a 32 bit integer. You can use the functions in Generator and Code to write / read those, rather than manual loops. It’ll be little-endian, but that’s fine as long as the comment around the constant makes that clear.
- I don’t think it’s worth tackling the problem of version numbers yet. We’re very far from a stable release, and we’re not making any kind of guarantees yet. In particular there’s no use in storing those in the Special Descriptors table. You could just drop it, or reduce it to a single `BYTECODE_VERSION` number, defined in one place and enforced at the same time as the magic number when loading the bytecode.

Yuhanun:
Managed to resolve another issue, load<uint32_t> would load a uint16 instead.

src/compiler/codegen/codegen.cc
```diff
@@ -13,6 +13,8 @@
 #include "ds/helpers.h"
 #include "interpreter/bytecode.h"
 
+#include <array>
```
plietar: Not needed anymore.

src/compiler/codegen/codegen.cc
```diff
@@ -92,6 +94,24 @@ namespace verona::compiler
...
+   * Writes the verona bytes, literally writes VERONA_NUMBEr
+   * @param code Generator to gen from
+   */
+  void write_verona_bytes(Generator& code)
+  {
+    code.u32(bytecode::VERONA_NUMBER);
+  }
+
+  /**
+   * Writes the verona integers to the bytecode
```
plietar: "Verona integers" is vague. "magic number" would be better.
Also `write_magic_bytes` and `write_verona_bytes` can be one function (eg. `write_magic_number`).

src/interpreter/bytecode.h
```diff
@@ -131,6 +132,8 @@ namespace verona::bytecode
     uint32_t size;
   };
 
+  constexpr static uint32_t VERONA_NUMBER = 4085854915;
```
plietar: Hex notation is better for this, so it's clear all 4 bytes are used, and it gives something to look for when hexdump'ing a bytecode file.

Should be called `MAGIC_NUMBER` instead, as verona number is too vague.

src/interpreter/code.h
```diff
@@ -5,6 +5,7 @@
 #include "interpreter/bytecode.h"
 #include "interpreter/object.h"
 
+#include <array>
```
plietar: Not needed anymore

src/interpreter/code.h
```diff
@@ -133,10 +136,19 @@ namespace verona::interpreter
       return header;
     }
 
+    void parse_verona_nums(size_t& ip) {
```
plietar: This doesn't really parse anything.
How about `check_magic_number` instead? Also this should be a private method.

src/interpreter/bytecode.h
```diff
@@ -13,6 +13,7 @@
  * Bytecode has the following layout:
  *
  * Program header:
+ * - 32-bit verona magic number, 4085854915
```
plietar: The number should be in hex, or doesn't even have to appear. Just add a "equal to the MAGIC_NUMBER constant" instead.

src/compiler/codegen/codegen.cc
Yuhanun: I kept write_verona_bytes because we might want to add more later, but okay sure

orcmid:
> What version is macOS and windows on and why do they not conform with the standard?

@Yuhanun I think thisissue has been handled.  I just don't understand this question, so I am asking.  What standard?

Yuhanun:
@orcmid The C++ standard, but it was my mistake, I forgot to include and it gave some weird errors on the CI whereas it was fine locally

plietar:
The failure on `features/run-pass/when` is a flaky test, I'll disable and then this LGTM.

mjp41:
@yuhanun I have fixed some issues in the CI process that wasn't enforcing clangformat. Please could you rebase on top of the current master.  The CI is only letting me run without updating the merge point :(

I think you have some clangformat issues.  There is a build target
```
make clangformat
```
or
```
ninja clangformat
```

Should build you a diff to commit to get through the CI.

Yuhanun:
> @Yuhanun I have fixed some issues in the CI process that wasn't enforcing clangformat. Please could you rebase on top of the current master. The CI is only letting me run without updating the merge point :(
> 
> I think you have some clangformat issues. There is a build target
> 
> ```
> make clangformat
> ```
> 
> or
> 
> ```
> ninja clangformat
> ```
> 
> Should build you a diff to commit to get through the CI.

How do I rebase onto it?

Yuhanun:
> @Yuhanun I have fixed some issues in the CI process that wasn't enforcing clangformat. Please could you rebase on top of the current master. The CI is only letting me run without updating the merge point :(
> 
> I think you have some clangformat issues. There is a build target
> 
> ```
> make clangformat
> ```
> 
> or
> 
> ```
> ninja clangformat
> ```
> 
> Should build you a diff to commit to get through the CI.

How do I rebase onto it?

Yuhanun:
Should be good, let me apply linting

mjp41:
Thanks.  For doing this.  I think you have used clang-format 9, but our CI is using an older version.  So they disagree on the style.  I will migrate the CI to a more recent version of clang-format, and then I'll fix any merges for the PR.

Yuhanun:
> 
> 
> Thanks. For doing this. I think you have used clang-format 9, but our CI is using an older version. So they disagree on the style. I will migrate the CI to a more recent version of clang-format, and then I'll fix any merges for the PR.

I'm using clang10, but yeah, cool :) no problem, lmk if anything else can be done by me

mjp41:
I have fixed the CI and created a new PR#81 that incorporates your changes.

PR #81, opened 2020-02-10
Yuhanun feature/bytecode magicnum

mjp41:
This is #76 updated to the with fixes to the CI and formatting.


Yuhanun:
Any idea on when verona will be native? Very excited for that.

Is the plan to transpile to C++ and compile from there? Or generate the bytecode ourselves?

mjp41:
@yuhanun current thoughts are to parse into MLIR, perform type checking and region optimisations at the MLIR level, and then lower to LLVM, and then to native.  There is a lot of work to get this done though.  At the moment, we are focussing on making sure we have the right type system and language design before adding more inertia to change.

PR #85, opened 2020-02-21
Fix a few build issues.

plietar:


src/rt/test/systematic.h
```diff
@@ -443,8 +443,8 @@ namespace Systematic
   inline static void signal_handler(int sig, siginfo_t*, void*)
   {
     auto str = strsignal(sig);
-    write(1, str, strlen(str));
-    write(1, "\n", 1);
+    (void)write(1, str, strlen(str));
```
mjp41: Can we use the UNUSED macro for clarity?

plietar: So far we don't use UNUSED for complex expressions, only for function parameters.

I'd be happy to switch it, as it expands to the same thing, but I don't really like putting things with side effects into macro calls.

plietar: Should I just add a comment above instead.

plietar:
> Does the building.md need updating, or will that install clang-8?

Good point. I think on 18.04 it will install clang 6 actually. On the other hand GCC 9 isn't available by default, so it won't have the compatibility issue we have on the CI images.

We can say clang 8 is what we "officially" support (it's available on 18.04, just not the default), and I'll change the instructions to use that.

Issue #84, opened 2020-02-18
Leaking a weak cown reference causes the cown to be collected twice


plietar:
```c++
#include <test/harness.h>

struct MyCown : VCown<MyCown> {};
struct Msg : VAction<Msg>
{
  MyCown* m;
  Msg(MyCown* m) : m(m) {}

  void f()
  {
    Systematic::cout() << "Msg on " << m << std::endl;
  }
};

void run_test()
{
  MyCown* t = new MyCown;
  // HERE: the weak RC is never released.
  t->weak_acquire();

  Cown::schedule<Msg, YesTransfer>(t, t);
}

int main(int argc, char** argv)
{
  SystematicTestHarness h(argc, argv);
  h.run(run_test);
}
```

I would expect a `debug_is_empty: found non-empty allocator` error. Instead the cown is collected twice, once when the strong RC hits zero and once on "teardown".

You can see in the log below the two `Collecting: 0x10c010220` lines:
```
 --cores 4
Seed: 5489
0         Set fair: 1
0         Object state :0x10c0100a0: EPOCH_NONE -> SCANNED
0         Object state :0x10c010100: EPOCH_NONE -> SCANNED
0         Object state :0x10c010160: EPOCH_NONE -> SCANNED
0         Object state :0x10c0101c0: EPOCH_NONE -> SCANNED
0         Epoch change on: 0x10c010220 : 7 -> EPOCH_A
0         Object state :0x10c010220: EPOCH_NONE -> EPOCH_A
0         0x10c0010a0 state change EPOCH_NONE -> EPOCH_NONE
0         MultiMessage 0x10c0010a0 payload 0x0 (EPOCH_NONE)
0         Schedule behaviour of type: 3Msg
0         MultiMessageBody 0x10c020020
0         0x10c0010d0 state change EPOCH_NONE -> EPOCH_A
0         MultiMessage 0x10c0010d0 payload 0x10c020020 (EPOCH_A)
0         MultiMessage 0x10c0010d0 index 0 fast requesting 0x10c010220
0         MultiMessage fast acquired 0x10c010220
0         MultiMessage fast send completed, rescheduling cown 0x10c010220
0         LIFO Scheduled Cown: 0x10c010220
0         Unpausing other threads.
0         Starting all threads
   3      Pausing
    4     Popped cown:0x10c010220
    4     Scheduled Cown: 0x10c010220 (EPOCH_A)
    4     Running Cown: 0x10c010220
[...]

PR #87, opened 2020-02-24
Weak leak bug

mjp41:
This fixes the bug reported in #84.


mjp41:
We need a more comprehensive fix around how we treat weak references. I will submit a new PR next week.

PR #93, opened 2020-04-06
Dynamically sized objects

plietar:
This PR adds support for dynamically sized arrays, through a Pointer primitive and a wrapper Array class. The codegen and VM have special knowledge of the Array class, as they need to provide a custom trace and destructor function.

The chunk of memory referenced by the Array is uniquely owned by it. Ideally we would be able to have many slices all pointing into the same block, without needing any indirection, but there's a few issues around tracing and GC that prevents that. I'll write it up in some issue later to see if anyone has a solution.

### DIFF
--- a/src/interpreter/vm.cc
+++ b/src/interpreter/vm.cc
@@ -289,6 +289,11 @@ namespace verona::interpreter
     return std::move(src);
   }
 
+  void VM::opcode_error(std::string_view reason)
+  {
+    fatal("{}", reason);
+  }
+
   Value VM::opcode_int64(uint64_t imm)
   {
     return Value::u64(imm);
@@ -572,6 +577,7 @@ namespace verona::interpreter
       OP(Clear, opcode_clear);
       OP(Copy, opcode_copy);
       OP(FulfillSleepingCown, opcode_fulfill_sleeping_cown);
+      OP(Error, opcode_error);
       OP(Freeze, opcode_freeze);
       OP(Int64, opcode_int64);
       OP(Jump, opcode_jump);
--- a/src/interpreter/vm.cc
+++ b/src/interpreter/vm.cc
@@ -92,6 +92,25 @@ namespace verona::interpreter
     vm->halt_ = old_halt;
   }
 
+  void VM::dealloc_array_data(const VMObject* object)
+  {
+    ArrayFields array = load_array_fields(object);
+    for (size_t i = 0; i < array.size; i++)
+    {
+      array.data[i].~FieldValue();
+    }
+    alloc_->dealloc(array.data, array.capacity * sizeof(FieldValue));
+  }
+
+  void VM::trace_array_data(const VMObject* object, rt::ObjectStack* stack)
+  {
+    ArrayFields array = load_array_fields(object);
+    for (size_t i = 0; i < array.size; i++)
+    {
+      array.data[i].trace(stack);
+    }
+  }
+
   void VM::grow_stack(size_t size)
   {
     size = snmalloc::bits::next_pow2(size);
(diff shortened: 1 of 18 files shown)


plietar:
After some more thought, the approach I’ve taken here isn’t going to work, so please ignore this PR for now.

src/stdlib/builtin.verona
```diff
@@ -175,3 +175,106 @@ primitive Pointer[T] {
...
+ */
+class Array[T] {
+  _data: Pointer[T] & imm;
+  _size: U64 & imm;
+  _capacity: U64 & imm;
+
+  create(): Array[T] & iso
+  {
+    var this = new Array;
+    this._data = Pointer.allocate(0);
```
mjp41, line 193: This is going to perform an allocation will you are saying has zero capacity, so it will just get deallocated. Why not pick a larger size, so the work is atleast useful.

I think really we want `snmalloc` to say the size it allocates for the size you request.  Then capacity can be based on `snmalloc`s size classes.  I don't propose you solve that on this PR, as that requires changes that might affect `snmalloc`, but perhaps leave a TODO, or raise some issues.

src/stdlib/builtin.verona
```diff
@@ -175,3 +175,106 @@ primitive Pointer[T] {
...
 }
 
+/**
+ * Dynamically resizable array.
+ *
+ * Codegen and the VM have special support for this class. The trace function
+ * will traverse the first _size elements of the _data field. The class'
+ * destructor clears the data and frees the underlying memory allocation.
+ */
+class Array[T] {
```
mjp41, line 185: Perhaps, add an `_invariant()`, which checks 
```
  if (_size > _capacity) errror()
```

src/stdlib/builtin.verona
mjp41: Do we need an allocation if `_capacity` is zero? As we only access the Pointer is `_capacity` and `_size` are greater than 0.  Makes correctness harder to justify?

testsuite/features/run-pass/array.verona
```diff
@@ -0,0 +1,20 @@
+class Main
+{
+  main()
+  {
+    var x = mut-view (Array.create());
+    var i = 1;
+    while i <= 10 {
+      x.push_back(i);
+      i = i + 1;
```
mjp41, line 9: Not testings all the methods of array here: `get`, `apply`, and `update`.

src/interpreter/vm.cc
```diff
@@ -563,6 +564,80 @@ namespace verona::interpreter
...
+    const Value& parent, const Value& ptr, uint64_t index, Value value)
+  {
+    check_type(ptr, Value::POINTER);
+    check_type(parent, {Value::ISO, Value::MUT});
+    VMObject* region = parent->object->region();
+    Value old = ptr->pointer[index].exchange(alloc_, region, std::move(value));
+    return std::move(old);
+  }
+
+  void VM::opcode_pointer_move(
```
mjp41, line 670: Do we need to expose this, versus having a loop in Verona code using `swap` and `set`

src/stdlib/builtin.verona
```diff
@@ -112,3 +112,66 @@ primitive U64 {
...
+ * should instead be traced through some other mean. For example, the trace
+ * function of Array is special cased to cover its backing array.
+ */
+primitive Pointer[T] {
+  /**
+   * Allocate a block of memory.
+   *
+   * The size is specified in number of elements, not in bytes.
+   */
+  builtin allocate(size: U64 & imm): Pointer[T] & imm;
```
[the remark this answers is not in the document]
plietar: An annoying part of making something iso/mut is that it implies it belongs to a region, therefore you can use it to do allocation. This isn't true of imm, which is why I've used it here.

I do eventually want a solution for linear values that aren't in a region, for eg. unique promises, but for now making Pointer immutable is the simplest option.

src/interpreter/vm.cc
plietar: We would need an additional method on Pointer, which extracts a value and leaves the value unitialized. On other hand this method can also be used to implement swap, so it's probably worth doing it.

rengolin:
@plietar @mjp41, is it worth continue working on the old compiler/interpreter at this stage? Should we close this PR instead?

plietar:
I’m sorry, this kind of fell off my plate and I never got back to it.

I’d argue there is still a future to the interpreter (which is what most of this PR changes) in that we could, fairly easily, lower the MLIR Verona dialect into bytecode. Moreover, unlike the prototype compiler, the interpreter is really simple and robust.

In the short term, this can happen sooner than a native backend (no need to lower to standard/LLVM dialect, no need to call LLVM or a linker, no complex integration with the runtime, a lot of the infrastructure can be reused).

In the long term, I do like the idea of using this interpreter as a constant evaluation engine, as a REPL, and as a faster compile time / slower run time backend than LLVM.

If we decide we want to keep that part of the prototype, then I can finish the PR in the coming weeks.

rengolin:
> In the short term, this can happen sooner than a native backend (no need to lower to standard/LLVM dialect, no need to call LLVM or a linker, no complex integration with the runtime, a lot of the infrastructure can be reused).

We'll already have to lower to LLVM IR, I'd rather work on one lowering target than two. Also, you don't need to create executables for simple cases, LLVM can JIT-compile and run IR files directly (if self-contained).

> In the long term, I do like the idea of using this interpreter as a constant evaluation engine, as a REPL, and as a faster compile time / slower run time backend than LLVM.

Lessons learnt from editor code navigation tools: they're all moving to Clang. Using anything but a single and complete compiler will generate slightly different results and at best misdirect the user, but likely break things in subtle ways.

The design of the Verona front-end and MLIR infrastructure is taking partial compilation and evaluation at its core from the beginning, which will completely replace the need for any interpreter as soon as it has the equivalent features.

rengolin:
This PR is outdated and we're not even sure we'll want to keep the interpreter long term. If this becomes important again, we can always refresh the branch and re-submit a new PR, referencing this one.

Issue #77, opened 2020-02-03
features/run-pass/when test randomly times out


plietar:
The solver [randomly goes into an infinite loop](https://dev.azure.com/ProjectVeronaCI/Project%20Verona/_build/results?buildId=271&view=logs&j=e632f1d6-0a34-55c8-b58a-b7d0f9490cee&t=6eec3b4f-8232-50fe-9366-b417b07238a8) for this test.

I'm disabling it in #78 in order to not block the rest of the work.

plietar:
`features/run-pass/loop` also randomly goes into an infinite loop (with a lower probability).
Disabling it in #97.

Solver log for both tests:
[loop.txt](https://github.com/microsoft/verona/files/4486476/loop.txt)
[when.txt](https://github.com/microsoft/verona/files/4486477/when.txt)

PR #94, opened 2020-04-14
Fixed test generation of CI

mjp41:
The test generation for CI is problematic.  When the CHUNK size was
made larger for the testing it meant the systematic test names no longer
match the regexp for making a smaller set of running tests in CI.

This test reduces the batch size in CI, so the original regexp still
works.


mjp41:
Re-run this several times, the Mac OS X version only broke once with a timeout.  I think this may be related to the non-termination in the prover based on hashconsing.  @plietar

plietar:
The good news is I'm able to reproduce it. Happens about 1 in 200 runs on my machine.
It's probably going to take me some time to debug, so in the meantime the best we can do is disable it. Hopefully this is the same issue as #77.

PR #95, opened 2020-04-14
Update snmalloc submodule

### LINKED ISSUE #88: snmalloc is out-of-date
Submodule at path `src/rt/external/snmalloc` is at revision`3775a62`, whereas upstream is at revision `8723ae2`.

You can update the submodule by running:

    git submodule update --remote src/rt/external/snmalloc

plietar:
Fixes #88

### DIFF
--- a/src/rt/ds/barrier.h
+++ b/src/rt/ds/barrier.h
@@ -1,7 +1,7 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
 #pragma once
-
+#include <cassert>
 #ifdef _WIN32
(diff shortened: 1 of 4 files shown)


plietar:
This doesn't work because of the "no two ABA operations" rule that was added in snmalloc.
In particular the [SPMCQ code uses snmalloc](https://github.com/microsoft/verona/blob/f193b1ce246333d6826e4538bebb29e117996ce6/src/rt/sched/spmcq.h#L71-L81) (via the Epoch api) inside a CAS loop.

I don't know that code at all, @mjp41 any ideas?

PR #96, opened 2020-04-15
Simplify the CF stack and finalisers implementation.

plietar:
- The current frame is stored at the top of the CF stack, rather than
  as a separate field.
- The current ip is stored in the top most frame, avoiding the need to
  save / restore it on call / return.
- Frames have a "halt on return" flag, used to make the return opcode
  in finalisers return control to the C++ code.
- Remove ident_. It can be inferred from the size of the cf stack.

This was prompted by a bug in the way finalisers used to be dispatched.
The receiver of the finaliser would be written to the last register of
the current frame. This could overwrite useful values, or underflow
and try to write to register 255, causing a crash. We were getting a
flaky CI because of it.

### DIFF
--- a/src/interpreter/vm.cc
+++ b/src/interpreter/vm.cc
@@ -90,11 +90,15 @@ namespace verona::interpreter
     vm->trace("Running the finaliser for: {}", descriptor->name);
 
     // Set up a new frame for the finaliser. The frame starts past the current,
-    // with no overlap.
+    // with no overlap, or at zero if there are no executing frames.
     //
     // The frame is marked as "halt on return" so we gain back control when the
     // finaliser is done.
-    size_t base = vm->frame().base + vm->frame().locals;
+    size_t base;
+    if (vm->cfstack_.empty())
+      base = 0;
+    else
+      base = vm->frame().base + vm->frame().locals;
     vm->push_frame(descriptor->finaliser_ip, base, OnReturn::Halt);
     if (vm->frame().argc != 1)
     {
--- a/src/interpreter/vm.cc
+++ b/src/interpreter/vm.cc
@@ -12,30 +12,12 @@ namespace verona::interpreter
   void VM::run(std::vector<Value> args, size_t cown_count, size_t start)
   {
     halt_ = false;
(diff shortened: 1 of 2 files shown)


src/interpreter/vm.cc
```diff
@@ -12,30 +12,12 @@ namespace verona::interpreter
...
-
-    trace(
-      "Entering function {}, argc={:d} retc={:d} locals={:d}",
-      name,
-      frame_.argc,
-      frame_.retc,
-      frame_.locals);
+    assert(static_cast<size_t>(frame().argc) == args.size());
 
     // Load registers with cowns values.
```
mjp41, line 19: Is this comment in the right place?

src/interpreter/vm.h
```diff
@@ -186,7 +197,7 @@ namespace verona::interpreter
        *
        * This is unused in the lowest frame, as exiting that frame halts the VM.
        */
-      size_t return_address;
+      size_t ip;
```
mjp41, line 200: Does this comment need updating for `Halt` frames?

src/interpreter/vm.h
```diff
@@ -146,20 +161,16 @@ namespace verona::interpreter
...
-
-    /**
-     * Instruction Pointer
-     */
-    size_t ip_;
+    rt::Alloc* const alloc_;
+    const bool verbose_;
 
     /**
      * Address of the currently executing instruction.
```
mjp41, line 168: `start_ip_` is "Address of the currently executing function"?

plietar: No, the comment is correct. `start_ip` gets updated in each iteration of `dispatch_loop`.
On the other hand there, it wasn't being saved correctly by `execute_finaliser`, so I've fixed that.

src/interpreter/vm.cc
plietar: I've removed it, since it was pretty out-of-date I think, and expanded the one below a bit.

src/interpreter/vm.h
plietar: Done.

PR #100, opened 2020-05-05
Set CMAKE_EXPORT_COMPILE_COMMANDS=1

rengolin:
Super helpful for YouCompleteMe users.


rengolin:
@mjp41

davidchisnall:
I've never seen that set in a CMake file, normally it's set by users.  That said, I can't see any reason for a user wanting to turn it off...

rengolin:
LLVM does it. :)

PR #109, opened 2020-05-08
Fix error when compiling RT tests

Theodus:
This PR fixes a compilation error by moving the declaration of `CCown` from cowngc1 and cowngc4 tests before its use as a class derrived from `Object`.

### DIFF
--- a/src/rt/test/func/cowngc1/cowngc1.cc
+++ b/src/rt/test/func/cowngc1/cowngc1.cc
@@ -81,6 +81,18 @@ struct CCown;
 struct RCown;
 static RCown* rcown_first;
 
+struct CCown : public VCown<CCown>
+{
+  CCown* child;
+  CCown(CCown* child_) : child(child_) {}
+
+  void trace(ObjectStack* fields) const
+  {
+    if (child != nullptr)
+      fields->push(child);
+  }
+};
+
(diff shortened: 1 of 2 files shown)


rengolin:
Hi @Theodus. I just tried to build and had the same problem. Though, given the nature of the test `CCown cown = nullptr`, I wonder why it was built that way instead of `Cown cown = nullptr`.

The problem there, it seems, is that `CCown` is derived from `VCown <- V<Cown>`, while `Cown` is derived from `Object`, which is the type `<T>` in `ObjectStack`. A blind C-type cast "works", but I'd rather change it without blind casts, if possible.

Unfortunately, `git blame` doesn't help, so I'm going to blame @mjp41. :D

Theodus:
Hi @rengolin, `V<T>` is derived from `Object` as well. I believe the source of the error is that `CCown*` is being used as an `Object*` before `CCown` is defined as a complete type.

rengolin:
Ah, that makes more sense. Looks like a better fix.

I'll wait for @mjp41 to approve, but LGTM, thanks!

Theodus:
I am using clang 10.0.0 on Arch Linux

rengolin:
It failed for me too, on Arch Linux with gcc 9.3.0.

PR #113, opened 2020-05-18
Fix compilation error with clang 10

rengolin:
Missing include <limits> generated the error:
```
../src/./interpreter/bytecode.h:124:10:
  error: no member named 'numeric_limits' in namespace 'std'
    std::numeric_limits<DescriptorIdx>::max();
```


rengolin:
@mjp41

PR #115, opened 2020-05-18
Add `ninja check` target to run ctest with deps

### LINKED ISSUE #114: Better error message on ctest failure
I tried to build Verona today and _all_ tests failed on master, which doesn't make sense.

When running `ctest -R <test> -V` I noticed the error was:
```
9: CMake Error at .../verona/testsuite/common.cmake:11 (message):
9:    To run tests you must build the install target.
```

It's not a widespread practice to require `ninja install` before testing, but it's not uncommon either. 

The best "fix" would be to create a `ninja check` which depends on `install` and just runs `ctest`. 

Another option would be to not require installing and take the binaries and libraries from the build directory.

rengolin:
Just running `ctest` can silently crash if the user hasn't also ran
`ninja install`, which is confusing. This commit adds a new `check`
target that not only makes the dependency clear, but also defaults the
test to run on the maximum number of cores available.

Fixes #114

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -41,7 +41,7 @@ add_custom_target(install_for_test
 
 set(CTEST "ctest")
 add_custom_target("check"
-                  COMMAND ${CTEST} ${PARALLEL_ARGS} -C$<CONFIG>
+                  COMMAND ${CTEST} ${PARALLEL_ARGS} -C $<CONFIG>
                   COMMENT "Run ctest tests"
                   USES_TERMINAL COMMAND_EXPAND_LISTS)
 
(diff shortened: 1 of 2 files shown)


rengolin:
Interesting, Windows builds don't have the "install" target:
```
CMake Error at CMakeLists.txt:37 (add_dependencies):
  The dependency target "install" of target "check" does not exist.
```

mjp41:
According to the cmake docs this shouldn't work:

> `add_dependencies(<target> [<target-dependency>]...)`
   Makes a top-level `<target>` depend on other top-level targets to ensure that they build before 
   `<target>` does. A top-level target is one created by one of the `add_executable()`, `add_library()`, or `add_custom_target()` commands (but not targets generated by CMake like `install`).

  https://cmake.org/cmake/help/latest/command/add_dependencies.html

rengolin:
@mjp41 thanks for the changes!

PR #116, opened 2020-05-19
Initial commit for a cpp-peglib based parser and untyped AST gen

sylvanc:
This uses cpp-peglib to generate a packrat PEG for a grammar defined
in a text file, and then does some very basic AST manipulation. It
sets up scopes and symbol tables, and does a bit of fixity analysis.
The intention is to use this for a very slim parser front-end that
goes to MLIR before any serious heavy lifting is done. No tests are
included in the initial commit, as the grammar.peg is preliminary.Z

### DIFF
--- a/src/ast/CMakeLists.txt
+++ b/src/ast/CMakeLists.txt
@@ -1,3 +1,5 @@
+find_package(Threads REQUIRED)
+
 add_library(verona-ast-lib
   ast.cc
   err.cc
@@ -8,6 +10,7 @@ add_library(verona-ast-lib
 
 target_link_libraries(verona-ast-lib CLI11::CLI11)
 target_link_libraries(verona-ast-lib cpp-peglib)
+target_link_libraries(verona-ast-lib Threads::Threads)
 
 add_executable(verona-ast main.cc cli.cc path.cc)
 target_link_libraries(verona-ast verona-ast-lib)
--- a/src/ast/CMakeLists.txt
+++ b/src/ast/CMakeLists.txt
@@ -1,0 +1,16 @@
+add_library(verona-ast-lib
+  ast.cc
+  err.cc
+  files.cc
+  parser.cc
+  sym.cc
+  )
+
+target_link_libraries(verona-ast-lib CLI11::CLI11)
+target_link_libraries(verona-ast-lib cpp-peglib)
+
+add_executable(verona-ast main.cc cli.cc path.cc)
+target_link_libraries(verona-ast verona-ast-lib)
+
+install(TARGETS verona-ast RUNTIME DESTINATION .)
+install(FILES ${CMAKE_CURRENT_SOURCE_DIR}/grammar.peg DESTINATION .)
--- a/src/ast/main.cc
+++ b/src/ast/main.cc
@@ -1,0 +1,40 @@
+#include "cli.h"
+#include "files.h"
+#include "parser.h"
+#include "sym.h"
+
+// #include "llvm/ExecutionEngine/ExecutionEngine.h"
+// #include "llvm/ExecutionEngine/GenericValue.h"
+// #include "llvm/ExecutionEngine/MCJIT.h"
+// #include <llvm/IR/IRBuilder.h>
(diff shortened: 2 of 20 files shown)


mjp41:
Can you use
```
https://github.com/yhirose/cpp-peglib.git
```
as the submodule reference, rather than an ssh based one.

rengolin:
Quick overall comment, there are a number of assumptions (make_shared onto nullptr, fields were set before use) that could segfault without errors. 

I think adding user errors for those is wrong, because it's not a user error, so assert would be the right level, I think. CMake has release builds with assert, which are good for testing that kind of thing.

src/ast/CMakeLists.txt
```diff
@@ -0,0 +1,16 @@
+add_library(verona-ast-lib
+  ast.cc
+  err.cc
+  files.cc
+  parser.cc
+  sym.cc
+  )
+
+target_link_libraries(verona-ast-lib CLI11::CLI11)
+target_link_libraries(verona-ast-lib cpp-peglib)
```
Theodus, line 10: peglib crashes with a `std::system_error` from inside `std::call_once` on Linux systems if it's linked without pthread.

rengolin:
The lack of tests should be fine, given the state of the grammar and the prototype. I think this can be the next task after merge, to go through the list of tests we already have and make sure they have the syntax we want and change the parser/tests to match it. 

This will be a long process, with a lot of discussion, and can happen in parallel with other work (ex. the MLIR bridge, runtime libraries).

rengolin:
Initial list of files from the testsuite (marked as `compile-pass`) that this front-end parses into AST:

```
./resolution/compile-pass/circular.verona
./resolution/compile-pass/interleaved.verona
./reachability/compile-pass/visit-variable-renaming-type.verona
./reachability/compile-pass/visit-receiver.verona
./features/compile-pass/library/sibling.verona
./features/compile-pass/cell.verona
./typechecking/compile-pass/field-write.verona
```

None of the `run-pass` parses.

It seems most (if not all) errors are on `class` declaration.

This is just a comment, for future reference, doesn't preclude merging the code as is.

src/ast/main.cc
```diff
@@ -0,0 +1,40 @@
+#include "cli.h"
+#include "files.h"
+#include "parser.h"
+#include "sym.h"
+
+// #include "llvm/ExecutionEngine/ExecutionEngine.h"
+// #include "llvm/ExecutionEngine/GenericValue.h"
+// #include "llvm/ExecutionEngine/MCJIT.h"
+// #include <llvm/IR/IRBuilder.h>
```
rengolin, line 9: I think those commented out lines can go. I wouldn't use this same main file as the main driver, and would only have it as an AST testing framework. I'm doing the same thing with the MLIR side, `verona-opt`, which will take in MLIR and spit out MLIR or LLVM. Same for the bridge between AST->MLIR.

I think the main compiler should use this AST framework as a library, as well as the conversion bridge between AST and MLIR and the MLIR transformation pipeline, which will also be tested for whole-program support.

If we keep things separate, it's a lot easier to keep the API generic and well tested.

src/ast/path.h
```diff
@@ -0,0 +1,9 @@
+#pragma once
+
+#include <string>
+
+namespace path
```
rengolin, line 5: We're using c++17, why not use [std::filesystem](https://en.cppreference.com/w/cpp/filesystem/path)?

mjp41: filesystem support on Mac is not very good.  There are a bunch of things I implemented that can expand the platform support here.
```
get_builtin_library()
```
Has a more comprehensive `executable()` implementation, but not factored out.  And `remove_filename()` in `fs.h` covers some other bits.  

I think this can be tidied after this PR.

mjp41:
Actually can we remove it from building in CI.  I don't really want failed CI ever on master.

src/ast/path.h
sylvanc: Thanks for get_builtin_library() pointer, lifted some implementation from there.

PR #119, opened 2020-05-20
Parse more common string representation

Theodus:
Change string syntax from `` `(...)`` to `"..."`, including common escape
sequences.


sylvanc:
Let's not merge this until we have an in-depth discussion on string literal representation.

Theodus:
In that case, we should also discuss pattern matching

rengolin:
We could enhance the discussion with explicit tests, as proposed in #120

Theodus:
Closing for now

PR #121, opened 2020-05-20
Add ast parser tests

### LINKED ISSUE #120: Create an initial set of tests for the parser
It'd be nice to have an initial set of tests for the parser, from the existing examples in `testsuite` that we kinda agree yields an "ok" AST, and add some checks for it.

Even if they're not what we ultimately want, any changes in the grammar or parser would come alongside changes in the test checks, which will make explicit the effect of the change.

This can avoid having to write long commit messages or referring to unlinked documentation.

As the AST dump is a simple tree, we can easily use `OutputCheck` instead of `FileCheck` for them, so no big changes needed.

Theodus:
Closes #120


src/ast/test/tests.cmake
```diff
@@ -0,0 +1,44 @@
+if(NOT EXISTS ${PARSER})
+  message(FATAL_ERROR " To run tests you must build the install target." ${PARSER})
+endif()
+
+find_package(Python3 COMPONENTS Interpreter)
+
+if (NOT Python3_FOUND)
+  message(WARNING " No python interpreter - Test Suite not compiled")
```
rengolin, line 8: nit: "no python 3" . Users might `python` and see that it's there, but it's python2.

src/ast/test/tests.cmake
```diff
@@ -0,0 +1,44 @@
...
+
+find_program(FILECHECK NAMES OutputCheck)
+
+if(NOT FILECHECK)
+  message(WARNING " Could not find OutputCheck - Test Suite not compiled")
+  message(WARNING "   Run: pip install OutputCheck")
+  return()
+endif()
+
+file(GLOB INPUT_FILES ${TEST_DIR}/*.verona)
```
rengolin, line 20: we may want to have multiple sub-dirs for different kinds of tests, can CMake GLOB recurse?

src/ast/test/tests.cmake
```diff
@@ -0,0 +1,44 @@
...
+foreach(INPUT_FILE ${INPUT_FILES})
+  get_filename_component(TEST_NAME ${INPUT_FILE} NAME_WE)
+  message(STATUS parse-${TEST_NAME})
+  set(AST_DUMP_FILE ${CMAKE_CURRENT_BINARY_DIR}/${TEST_NAME}.output)
+  execute_process(
+    COMMAND ${PARSER} -g ${GRAMMAR} -a ${INPUT_FILE}
+    OUTPUT_FILE ${AST_DUMP_FILE})
+
+  if(NOT ${TEST_EXIT} EQUAL 0)
+    message(FATAL_ERROR "Parser Exit " ${TEST_EXIT})
```
rengolin, line 30: Nice, it's good to have separate errors for "cannot parse" and "checks don't match".

src/ast/grammar.peg
```diff
@@ -85,10 +85,9 @@ sym <- <[!-'*+\x2D./:<-@\\^|~]+> # Reserved: , _ () [] {} ` ;
 float <- <digits '.' digits ([eE][-+]? digits)?>
 int <- <digits>
 digits <- [0-9][_0-9]*
-hex <- <'0x' [_a-fA-F]*>
+hex <- <'0x' [_0-9a-fA-F]*>
```
rengolin, line 88: oops :)

src/ast/grammar.peg
```diff
@@ -85,10 +85,9 @@ sym <- <[!-'*+\x2D./:<-@\\^|~]+> # Reserved: , _ () [] {} ` ;
...
 binary <- <'0b' [_01]*>
-string <- '`' ('(' <((!'(' !')' .) / substring)*> ')' / <%word>)
-substring <- '(' ('\\)' / (!'(' !')' .) / substring)* ')'
+string <- '`(' <('\\)' / string / (!'(' !')' .))*> ')'
```
rengolin, line 90: My peg-foo isn't good enough yet to know the difference. @sylvanc?

src/ast/test/tests.cmake
Theodus: It can if you use `GLOB_RECURSE`

src/ast/grammar.peg
Theodus: The previous implementation didn't handle nested strings or the escaped `\)`

Theodus: I'd like to hear @sylvanc's thoughts on this in case I was missing something

plietar:
Being the compiler’s internal representation, a simple refactor can cause changes to dozens of test outputs. I’m worried using OutputCheck/FileCheck is going to lead to lots of manual updating even for minor changes.

In some existing compile-pass and run-pass tests, the expected output is a separate file, which allows a single command, `ninja update-dump`, to update everything.

Separate files has the added benefit that checking is a lot more strict, eg. written as is the test allow extra lines to be emitted by the compiler without failing them.

src/ast/grammar.peg
mjp41: Just checking my understanding
```
escaped_closing_paren <- '\\)'
not_paren <- (!'(' !')' .)
 '`(' <(escaped_closing_paren / string / not_paren)*> ')'
```
In the [paper](https://bford.info/pub/lang/peg.pdf), for the nested comments, they don't restrict the opening token, just the closing one.  
```
Comment <- ’(*’ (Comment / !’*)’ .)* ’*)’
```
I guess for your implementation this means that
```
`( foo `( )
```
Is not a valid string, and cannot be escaped either?

rengolin:
> In some existing compile-pass and run-pass tests, the expected output is a separate file, which allows a single command, `ninja update-dump`, to update everything.

My worry with this approach is that we may ignore some changes that weren't meant to be. I think there are two types of tests: strict ones that have a precise syntax, and functionality ones, that have more loose checks on syntax and worry about semantics.

Large dump updates work well on the latter, not at all on the former.

The new tests @Theodus added here seem to be the former, while the original proposal in #120 was more towards the latter.

The core idea of #120 was, however, achieved with these tests as well: to make sure we reflect grammar changes on test changes that are intentional and clear. This can be done with either type of tests.

For now, given the grammar isn't fixed, the difference is muddled a bit, so I'd be happy either way.

src/ast/grammar.peg
Theodus: Fixed, you're right. I added the `\(` escape from the original grammar

sylvanc: I think the original implementation is correct. It allows balanced parens, ignoring the backtick inside a string. It currently allows escaping unbalanced right parens - perhaps we should allow an unbalanced left paren as well:

```
string <- '`' ('(' <('\\(' / '\\)' / (!'(' !')' .) / substring)*> ')' / <%word>)
substring <- '(' ('\\(' / '\\)' / (!'(' !')' .) / substring)* ')'
```

rengolin:
Do we really need the string change in this commit? This was supposed to just add tests, not change the grammar (other than the clear typo in hex).

I don't mind it going in, now that we already spent the time reviewing, but in the future it would be good to separate functional from non-functional pull requests, to avoid dragging simple ones unnecessarily.

src/ast/grammar.peg
Theodus: I think we can leave the unbalanced `` `(`` for a future PR

Theodus:
@rengolin Sorry, I had inadvertently modified the grammar from what we had agreed upon in an attempt to fix an escape sequence. I should have tested my fix for the string grammar more thoroughly before opening the PR.

rengolin:
No worries, it's hard to notice when we get side tracked. :)

I think the lesson learned here is to not even bring functional changes on an NFC patch in the first place, so there's no need to test it in the first place (for this patch).

What I usually do when finding functional problems when working on an NFC patch is to create issues for them, and dealing with them later. Helps me keep on track and not lose sight of them in the long term.

rengolin:
Regarding the tests, @sylvanc is heavily changing the representation, and that will probably continue for some time until it stabilises. That means @plietar's idea is probably the best approach in the near future.

Recap:
 * Revert the string change, keep it for a separate patch, don't test strings here.
 * The `hex` change is clearly a typo, so it's ok to stay here, given you have tests for that.
 * Generate the files with `ninja` as recommended by Paul, add them to the commit.

This should allow you to work on other (more interesting) issues around the grammar and parser. :)

utils/update_dump.py
```diff
@@ -24,7 +24,7 @@ def error(self, *args):
     self.has_error = True
 
   def generate_dump(self, source, dump_dir):
-    cmd = [self.compiler, "--dump-path=%s" % dump_dir, source]
+    cmd = self.compiler.split(" ") + ["--dump-path=%s" % dump_dir, source]
```
plietar, line 27: Not that it makes any real difference, but I’ve got a slight preference for `shlex.split` here.

testsuite/parse/ast-parse/region.verona
```diff
@@ -0,0 +1,9 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
```
plietar, line 1: Not sure about the file name. Nothing in here talks about regions.

Theodus: I intended for it to have something about allocation and regions, but there doesn't seem to be anything analogous to `new Foo in r`.

src/ast/cli.cc
```diff
@@ -14,6 +14,8 @@ namespace cli
 
     Opt opt;
     app.add_flag("-a,--ast", opt.ast, "Emit an abstract syntax tree.");
+    app.add_option(
+      "-d,--dump-path", opt.dump_path, "Dump abstract syntax tree to file.");
```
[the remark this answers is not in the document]
Theodus: This was chosen for compatibility with the Verona compiler's `--dump-path` option used by the `update_dump.py` script. Would you like me to make this change to the current Verona compiler as well, or pass the flag name to `update_dump.py`?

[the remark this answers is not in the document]
rengolin: No, if the change spans across tools, ignore. If we do that change, we should do it in another commit. By "other compilers" I meant non-Verona, but that's not important right now.

src/ast/main.cc
```diff
@@ -30,5 +30,8 @@ int main(int argc, char** argv)
   if (opt.ast)
     std::cout << peg::ast_to_s(ast) << std::endl;
 
+  if (opt.dump_path != "")
```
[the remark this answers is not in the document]
rengolin: Perhaps we should leave this to a different commit, too. If feels like it's too out of scope for this one.

src/ast/cli.h
```diff
@@ -13,6 +13,7 @@ namespace cli
     bool exec = false;
     std::string grammar;
     std::string filename;
+    std::string dump_path = "";
```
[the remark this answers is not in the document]
rengolin: let's ignore this one for now, too.

src/ast/CMakeLists.txt
```diff
@@ -5,8 +5,7 @@ add_library(verona-ast-lib
   err.cc
   files.cc
   parser.cc
-  sym.cc
-  )
+  sym.cc)
```
rengolin, line 8: small nit: avoid unnecessary changes on files that have no functional change. We always want to touch the least number of files on every commit.

Theodus: Thanks, this was left over from the initial commit where more changes were made to that file.

src/ast/cli.h
sylvanc, line 16: This shouldn't have an initialiser.

src/ast/files.cc
```diff
@@ -35,4 +35,11 @@ namespace files
 
     return data;
   }
+
+  void write(const std::string& file, std::string content)
```
sylvanc, line 39: Should be `const std::string& content`. This needs error handling in case the file can't be written. Sadly, output filenames can have path components, which means needing to parse them and create intermediate directories if necessary. That seems like too much for this feature - we can either error out there (annoying to the user) or perhaps dump to stdout instead of a file and rely on redirects?

src/ast/files.h
```diff
@@ -10,4 +10,5 @@
 namespace files
 {
   std::vector<char> slurp(const std::string& file, bool optional = false);
+  void write(const std::string& file, std::string content);
```
sylvanc, line 13: `const std::string& content`

src/ast/main.cc
sylvanc, line 33: `if (!opt.dump_path.empty())`

src/ast/main.cc
```diff
@@ -30,5 +30,8 @@ int main(int argc, char** argv)
...
+    files::write(opt.dump_path + "/ast.txt", peg::ast_to_s(ast));
+
```
sylvanc, line 35: This seems surprising, why isn't the output filename just the output filename, without appending something to it? Also, see other comment above on error handling and directories in paths.

Theodus: I intend to modify the test scripts to handle output redirect and remove the need for this option. But the goal of this PR is to integrate these tests with minimal changes to other tools.

Theodus: For now, the behavior of `--dump-path` is required by `update-dump.py` which we use to generate the test files we check against

testsuite/parse/ast-parse/literal/ast.txt
```diff
@@ -0,0 +1,47 @@
++ module
+  + function
+    + funcname
+      - id (f)
```
mjp41, line 4: What is the difference between `+` and `-` in the ast format?

Theodus: peglib shows a `-` if the node is a token, and `+` otherwise

rengolin:
@sylvanc, have your concerns being addressed? The output handling has been changed, but you're still marked as "requested changes".

Theodus:
I have removed the `--dump-path` option from `verona-ast` in favor of regenerating the test files from stdout using cmake

utils/update_dump.py
```diff
@@ -25,7 +24,7 @@ def error(self, *args):
     self.has_error = True
 
   def generate_dump(self, source, dump_dir):
-    cmd = shlex.split(self.compiler) + ["--dump-path=%s" % dump_dir, source]
+    cmd = [self.compiler, "--dump-path=%s" % dump_dir, source]
```
rengolin: `--dump=path` was removed?

Theodus: Yes, and `update_dump.py` is no longer being used to regenerate the test files for the parser tests. It should no longer be touched by the PR

Theodus: this just reverts the source to where it was prior to my changes

rengolin: of course! I only saw the commit, not the final state. :)

rengolin:
Right, the dump and const string were the two issues @sylvanc requested to change. I believe this PR is now good to merge. Thanks!

PR #123, opened 2020-05-24
Sched fixup

Theodus:
This PR fixes the documentation and naming of the spmcq implementation to be more consistent with the mpscq. Statically known alloc and dealloc sizes are also passed to snmalloc for `MultiMessage` and epoch `DecNode`.


rengolin:
This makes sense to me. NFC change with clearer meaning.

PR #124, opened 2020-05-26
Improving systematic testing and bug fixes to noticeboards

mjp41:
The current choice over expensive testing locally, and cheap in CI seems
incorrect.
* The variable used originally is incorrect, so no testing was performed
  in CI.
* Second imposing the high-bar on all local testing is too much for most
  changes and machines, and is only necessary for complex changes. The
  PR process should ultimately have an option to add an additionally
  pipeline with the expensive testing.

This PR also fixes a perf regression in systematic testing.  When the `coin` was introduced into the heuristics it caused the expensive case of `epoch` advancement to occur very frequently, and this cost a lot of performance.  This PR as a probability to the coin, so that we still cover these cases, but the performance is better.


mjp41:
FYI: The change in settings in the heuristics has revealed a bug exposed by the noticeboard test.

mjp41:
Will tidy this once #135 and #136 go in.

PR #126, opened 2020-05-28
Fix string escaping

Theodus:
Allow escape sequences for `` ` ``, `\`, and `)`.

### DIFF
--- a/src/ast/grammar.peg
+++ b/src/ast/grammar.peg
@@ -87,8 +87,9 @@ int <- <digits>
 digits <- [0-9][_0-9]*
 hex <- <'0x' [_0-9a-fA-F]*>
 binary <- <'0b' [_01]*>
-string <- '`' ('(' <((!'(' !')' .) / substring)*> ')' / <%word>)
-substring <- '(' ('\\)' / (!'(' !')' .) / substring)* ')'
+string <- '`' (<%word> / '(' <string_inner> ')')
+string_inner <- <(string_escape / (!'`(' !')' .) / string)*>
+string_escape <- '\\' ('`' / ')')
 
 list(x) <- x (',' x)*
 
--- a/src/ast/parser.h
+++ b/src/ast/parser.h
@@ -23,7 +23,6 @@ namespace parser
     if (!parser.load_grammar(grammar.data(), grammar.size()))
       exit(-1);
 
-    parser.enable_packrat_parsing();
     parser.enable_ast<ast::AstImpl>();
     return parser;
   }
--- a/src/ast/parser.h
+++ b/src/ast/parser.h
@@ -23,6 +23,7 @@ namespace parser
     if (!parser.load_grammar(grammar.data(), grammar.size()))
       exit(-1);
 
+    parser.enable_packrat_parsing();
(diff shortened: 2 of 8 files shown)


src/ast/parser.h
```diff
@@ -23,7 +23,6 @@ namespace parser
     if (!parser.load_grammar(grammar.data(), grammar.size()))
       exit(-1);
 
-    parser.enable_packrat_parsing();
```
sylvanc, line 26: Why aren't we enabling packrat parsing?

src/ast/grammar.peg
```diff
@@ -87,8 +87,9 @@ int <- <digits>
 digits <- [0-9][_0-9]*
 hex <- <'0x' [_0-9a-fA-F]*>
 binary <- <'0b' [_01]*>
-string <- '`' ('(' <((!'(' !')' .) / substring)*> ')' / <%word>)
-substring <- '(' ('\\)' / (!'(' !')' .) / substring)* ')'
+string <- '`' (<%word> / '(' <string_inner> ')')
+string_inner <- <(string_escape / (!'`(' !')' .) / string)*>
+string_escape <- '\\' ('`' / ')')
```
sylvanc, line 92: Nice.

src/ast/sym.cc
```diff
@@ -149,15 +149,7 @@ namespace sym
...
-        // A missing function name is sugar for "apply"
-        if (ast->nodes[0]->nodes.size() == 0)
-        {
-          ast->nodes[0]->nodes.push_back(
-            ast::from(ast->nodes[0], "id", "apply"));
-        }
-
-        auto node = ast->nodes[0]->nodes[0];
-        add_symbol(node->token, ast, err);
+        add_symbol(ast->nodes[0]->nodes[0]->token, ast, err);
```
sylvanc, line 152: Why is apply sugar being removed?

src/ast/parser.h
Theodus: I don't think our grammar currently requires it in any case, but I'm fine with putting it back if you think it may be needed in the future.

sylvanc: There are a few examples already that require arbitrary look-ahead, like lambda syntax (which looks like a tuple followed by a block). I think we should leave it on.

src/ast/sym.cc
Theodus: I think this node is only for function declarations. I think apply sugar only applies to function application, where it is parsed as:
```
+ term
    - prefix (foo)
    + tuple
```

src/ast/parser.h
Theodus: Alright, I've put it back

src/ast/sym.cc
sylvanc: This changes an empty function name at declaration to instead be `apply`. This allows `apply` to be used to reference the function. There will be `apply` sugar at invocation time as well, but it's not in yet.

Theodus: Ok, I missed that the grammar allowed an empty function name. A test has been added for this sugar.

PR #127, opened 2020-05-31
Build name-resolution aware expression tree

sylvanc:
The parser builds an AST stream for expressions. This precedence
pass rewrites that stream to an expression tree. By doing this
after name-resolution, the expression tree is aware of which
identifiers are parameters, locals, types, etc.


testsuite/parse/ast-parse/alloc/ast.txt
```diff
@@ -1,71 +1,4 @@
...
-                        + type_ref
-                          - id (U64)
-                  + inittype
-            + block
-              + seq
-                + expr/1
-                  - ref (a)
-                  - infix (+)
-                  - int (42)
+/home/sylvanc/dev/verona/testsuite/parse/ast-parse/alloc.verona:7:20: Expected an infix operator here.
```
rengolin, line 1: I would avoid transforming passing tests into negative tests. Right now, with the syntax still evolving, I'd change the verona file to convey the same idea (and still pass) and add another file, with `fail` in its name, and check for error messages.

sylvanc: Yup, these are actual test errors, fixing.

SeanTAllen:
Nice

sylvanc:
This will eventually need to handle inheriting default implementations from nominally declared interfaces, and integrate with the module system.

src/ast/sym.cc
```diff
@@ -27,53 +34,94 @@ namespace
...
+          assert(0);
+        }
+      }
+    }
+  }
+
+  void add_typeargs(ast::Ast& ast, ast::Ast& typeargs)
+  {
+    // Adds a typeargs node to ast and moves to the next node in the expr.
+    // If it isn't a typeargs node, adds an empty typeargs onde and does not
```
mjp41, line 84: "onde"

src/ast/sym.cc
```diff
@@ -247,27 +296,389 @@ namespace sym
...
+  {
+    if (ast && (ast->tag == "typeref"_))
+      prec_staticcall(ast, err);
+  }
+
+  void prec_member(ast::Ast& ast, err::Errors& err)
+  {
+    prec_obj(ast, err);
+    auto next = ast::get_next_in_expr(ast);
+
```
mjp41, line 423: A brief comment here might be helpful.

src/ast/sym.cc
```diff
@@ -247,27 +296,389 @@ namespace sym
...
+        return;
+      }
+
+      if (op.empty())
+      {
+        op = next->token;
+      }
+      else if (op != next->token)
+      {
+        err << next << "Use parentheses to establish precedence between " << op
```
mjp41, line 545: Nice

src/ast/sym.cc
```diff
@@ -247,27 +296,389 @@ namespace sym
...
+  {
+    // (call function typeargs lhs (args rhs))
+    // Infix operators are left associative.
+    prec_prefix(ast, err);
+    auto next = ast::get_next_in_expr(ast);
+    std::string op;
+
+    while (next)
+    {
+      if (next->tag == "assign"_)
```
mjp41, line 530: This feels a little awkward.  I wonder if the peg could be redesigned a little to give you a list of assigns, then you wouldn't have to know about the outer level here?  We are treating `=` specially.

sylvanc: That may be a good approach, but I don't know how I would do it right now. I'd rather get the PR in and look at that later.

mjp41: Sure

src/ast/ast.cc
```diff
@@ -42,7 +65,29 @@ namespace ast
...
+    if (ast->is_token)
+    {
+      next = token(ast, name, ast->token);
     }
+    else
+    {
+      next = node(ast, name);
+
+      for (auto& node : ast->nodes)
+        push_back(next, node);
```
Theodus, line 85: This should probably be `std::move(node)` to avoid unnecessary atomic incref/decref.

src/ast/sym.cc
```diff
@@ -247,27 +296,392 @@ namespace sym
...
+      if (next)
+      {
+        ast::remove(next);
+        ast::push_back(args, next);
+        next = ast::get_next_in_expr(ast);
+      }
+    }
+  }
+
+  void prec_assign(ast::Ast& ast, err::Errors& err)
```
mjp41, line 582: I wonder if this might be better named
```C++
prec_for_assign
```
I was thinking it is parsing assignments to start with.  And as it doesn't always require an assign, this was a little confusing. Adding `for` makes it less about there being an assign in my mind.

src/ast/sym.cc
```diff
@@ -247,27 +296,392 @@ namespace sym
...
+      auto call = ast::node(ast, "call");
+      ast::replace(ast, call);
+      ast::remove(next);
+      ast::rename(next, "function");
+      ast::push_back(ast, next);
+      next = ast::get_next_in_expr(ast);
+      add_typeargs(ast, next);
+      ast::push_back(ast, lhs);
+      auto args = ast::node(ast, "args");
+      ast::push_back(ast, args);
```
Theodus, line 562: Shouldn't the lhs be pushed into args? Also, I think there should be a test somewhere for a cases like `1 + 2 + 3` and `1 + (2 + 3)` to prevent regressions.

sylvanc: No, the lhs is the receiver, not an argument. Feel free to add new tests.

src/ast/ast.cc
sylvanc: It works now. Optimisations can be made in a future PR.

src/ast/sym.cc
sylvanc: I'm sorry, that doesn't make sense to me. It's the assign rule in the precedence pass. Not sure I want to rename every rewrite rule as `for_$rule` ? There may be no infix in `prec_infix` and no prefix in `prec_prefix`.

PR #128, opened 2020-06-01
Add tests for operator precedence

Theodus:


testsuite/ast-parse.cmake
```diff
@@ -4,10 +4,16 @@ set(ACTUAL_DUMP ${CMAKE_CURRENT_BINARY_DIR}/${TEST_NAME})
 
 PrepareTest(VERONAC_FLAGS EXPECTED_DUMP ACTUAL_DUMP)
 
-CheckStatus(
+execute_process(
```
rengolin, line 7: This change is not really necessary, and I'd only change this to check for error messages in negative tests (the ones expecting to fail and checking for error messages).

Theodus: I find it convenient when debugging to have the error message printed when a test fails instead of having to find the file where it was logged. But I understand if this is not worth modifying the CMake.

rengolin: Yeah, it's just to separate the concerns. This can easily be another quick PR.

Issue #125, opened 2020-05-27
Verona Build Failing: attempting to reference a deleted function


Nate-Watts:
Freshly cloned repo following build steps in docs

**Versions**
Os: Windows 10
cmake version 3.17.20032601-MSVC_2
Visual Studio 2019 v16.6.0

```Build FAILED.

       "C:\Users\Nate\source\repos\verona\build\INSTALL.vcxproj" (default target) (1) ->
       "C:\Users\Nate\source\repos\verona\build\ALL_BUILD.vcxproj" (default target) (3) ->
       "C:\Users\Nate\source\repos\verona\build\src\interpreter\interpreter-sys.vcxproj" (default target) (9) ->
       (ClCompile target) ->
         C:\Users\Nate\source\repos\verona\src\rt\region\externalreference.h(55): error C2280: 'verona::rt::Object::Obj
       ect(void)': attempting to reference a deleted function [C:\Users\Nate\source\repos\verona\build\src\interpreter\
       interpreter-sys.vcxproj]
         C:\Users\Nate\source\repos\verona\src\rt\region\region_arena.h(237): error C2280: 'verona::rt::RegionBase::Reg
       ionBase(void)': attempting to reference a deleted function [C:\Users\Nate\source\repos\verona\build\src\interpre
       ter\interpreter-sys.vcxproj]
         C:\Users\Nate\source\repos\verona\src\rt\region\region_trace.h(72): error C2280: 'verona::rt::RegionBase::Regi
       onBase(void)': attempting to reference a deleted function [C:\Users\Nate\source\repos\verona\build\src\interpret
       er\interpreter-sys.vcxproj]
         C:\Users\Nate\source\repos\verona\src\interpreter/object.h(66): error C2280: 'verona::rt::V<verona::interprete
       r::VMCown,verona::rt::RegionType::Cown,verona::rt::Cown>::V(void)': attempting to reference a deleted function [
       C:\Users\Nate\source\repos\verona\build\src\interpreter\interpreter-sys.vcxproj]
         C:\Users\Nate\source\repos\verona\src\interpreter/object.h(75): error C2280: 'verona::rt::V<verona::interprete
       r::VMCown,verona::rt::RegionType::Cown,verona::rt::Cown>::V(void)': attempting to reference a deleted function [
       C:\Users\Nate\source\repos\verona\build\src\interpreter\interpreter-sys.vcxproj]
[...]

PR #131, opened 2020-06-02
VS 2019 fixes

mjp41:
The anonymous union types in Verona and snmalloc were not compiling on VS2019 due to a lack of default constructor. I could not find a way to add a default constructor to an anonymous union. This addresses those issues by making the unions types have default constructors.

### DIFF
--- a/src/interpreter/object.cc
+++ b/src/interpreter/object.cc
@@ -36,7 +36,8 @@ namespace verona::interpreter
       finaliser_ip > 0 ? VMObject::finaliser_fn : nullptr;
   }
 
-  VMObject::VMObject(VMObject* region) : parent_(region)
+  VMObject::VMObject(VMObject* region, const VMDescriptor* desc)
+  : Object(desc), parent_(region)
   {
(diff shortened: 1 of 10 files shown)


mjp41:
Fixes #125

mjp41:
Defaults construction of anonymous enums seems a little missing in the standard.  VS2019 was complaining that this field (of anon enum type) is not default constructable, and thus the class itself is not default constructable.  And somewhere else is calling the default constructor and it complains it has been deleted.

I seem to have failed the CI comprehensively, so need to do some local testing.

rengolin:
Ah, I see, so nothing wrong with they being anonymous, you just named them to give a proper constructor. Makes sense.

mjp41:
@Nate-Watts this should fix #125 for you.

rengolin:
Is this what @davidchisnall had proposed?

mjp41:
Sorry I can't find a proposal from @davidchisnall? I guess I missed something in Teams.

PR #132, opened 2020-06-02
String literals

sylvanc:
This PR changes from `(...) syntax to allow two forms of string
literal:

1. C-style strings, which are "..." delimited, not nested, not
   interpolated, and allow C-style escapes. Unlike C strings, they
   are multiline and allow arbitrary UTF-8 content.
2. Verona strings, which are "{...}" delimited, nested, have no
   escapes at all, and interpolate $id and ${expr} in the outer-
   most nesting. They are also multiline (and do indentation
   mangling), and allow arbitrary UTF-8 content.

### DIFF
--- a/src/ast/CMakeLists.txt
+++ b/src/ast/CMakeLists.txt
@@ -4,7 +4,9 @@ add_library(verona-ast-lib
   ast.cc
   err.cc
   files.cc
+  lit.cc
   parser.cc
+  prec.cc
   sym.cc
   )
 
--- a/src/ast/ast.cc
+++ b/src/ast/ast.cc
@@ -90,6 +90,13 @@ namespace ast
     replace(ast, next);
   }
 
+  void elide(Ast& ast)
+  {
+    assert(ast->nodes.size() == 1);
+    auto child = ast->nodes.front();
+    ast::replace(ast, child);
+  }
+
   Ast get_closest(Ast ast, Tag tag)
   {
     while (ast && (ast->tag != tag))
--- a/src/ast/ast.h
+++ b/src/ast/ast.h
@@ -32,6 +32,7 @@ namespace ast
   void replace(Ast& prev, Ast next);
   void remove(Ast ast);
   void rename(Ast& ast, const char* name);
+  void elide(Ast& ast);
 
   Ast get_closest(Ast ast, Tag tag);
   Ast get_scope(Ast ast);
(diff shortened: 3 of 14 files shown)


rengolin:
I see this as a very premature decision with all that comes with it, mainly, the freedom to be wrong and likely having to change it a few times before we get it right.

However, this proposal has two things that I like:
1. It has a "standard string", which behaves just like every other string, and we may have to support horrible things in there, (just like everyone else). This makes it easier to integrate with external code and interfaces.
2. The idea of multi-line string isn't new, nor controversial. I mean, even bash has it and doesn't make a mess of itself (like everywhere else). If we stick to the basics, we could hope to simplify a lot of the horrible cases of ype (1) without compromising compatibility.

But, as I said above, this is well before hind-sight, so I am definitely missing a ton of stuff that will go wrong. But as general directions go, it's sensible.

If we're open to be wrong in the future, and fix things as we go, it can definitely work.

mjp41:
So here are some comments I received today about this.

* Going to be a lot of characters to escape a dollar: should we add `$$`
* Nested string is cute, but optimising for quoting the Verona language in Verona is not really an end goal.  What are the use cases for this?  Are they aligned with our end goal, or are we over PLing.  The concept of forbidding escaping is very useful. But what language should we optimise this for. The current design would be pretty bad for quoting bash scripts.
* Do we really need the C style string?  Would one be enough?  Two before we even have a complete design seems overkill.
* Leave room for future expansion.  We don't want a third type of string in the future.  Don't parse everything with a $ that isn't {...} or a id as a string, or you can never add anything extra there.  Also, could lead to unexpected behaviour
    `"{Too many $'s is great}"`    
   is very different to
     `"{Too many $s is great}"`

The indent sensitive I like, but can I write a literal that has two indents on each element, e.g.
```
"{
    should be indented two
    should be indented two
}"
```
I think you are currently looking for the shortest indent, so this kind of thing wouldn't work?

sylvanc:
This is almost certainly not the final string literal implementation, but I'll merge this.

* David's point about stdlib integration is important. It has to do with how interpolated expressions become strings. Roughly, it will be conformance to a structural interface that has some sort of `string(self): String` method, but this needs a bunch of discussion.
* The feedback Matt gathered is also important. For now we aren't addressing these, but we will continue to discuss them.

PR #134, opened 2020-06-03
Add tests for object literal syntax

Theodus:


rengolin:
This makes sense to me, but I'm only now getting used to the AST, so...

testsuite/parse/ast-parse/literal.verona
```diff
@@ -3,15 +3,17 @@
...
+  let float1 = 27.18e-1;
+  let string1 = `foo;
+  let string2 = `(wat ℵ \n);
+  let string3 = `(foo\nbar\) baz);
+  let string4 = `(foo \`\( \ ( \\ baz);
+  let string5 = `(foo `(bar) `bar baz);
+  let precedence1 = 1 + 2 + 3;
+  let precedence2 = 1 + (2 + 3);
+  let object1 = new { x : U64; };
+  let object2 = new Foo { x : U64; };
```
mjp41, line 18: Why are there no values for x here?  I would have thought they would be part of the syntax?

Also, what is `new Foo { x : U64 ; };`, a `Foo` with an additional field?

Theodus: > Why are there no values for x here? I would have thought they would be part of the syntax?

Added.

> Also, what is `new Foo { x : U64 ; };`, a `Foo` with an additional field?

The new object literal would be inheriting from some interface `Foo`.

PR #135, opened 2020-06-03
Only build and test runtime if changed

mjp41:
The runtime systematic testing can be slow, and there are a lot of
language changes that do not need to retest the runtime. This changed
detects if anything has changed in the runtime, and if it has not
changed, then it does not rebuild or run the tests.

### DIFF
--- a/.vsts-ci.yml
+++ b/.vsts-ci.yml
@@ -2,8 +2,23 @@ resources:
 - repo: self
 
 jobs:
+- job: HasRuntimeChanged
+  pool:
+    vmImage: 'ubuntu-16.04'
+  steps:
+  - script: |
+      echo "Determine if runtime changed."
+      git diff --quiet origin/master -- src/rt || echo "##vso[task.setvariable variable=testRuntime;isOutput=true]On" #set variable to testRuntime to On
+      git diff --quiet origin/master -- src/rt && echo "##vso[task.setvariable variable=testRuntime;isOutput=true]Off" #set variable to testRuntime to Off
+      git diff --quiet origin/master -- src/rt && echo "Runtime unchanged!"
+      git diff --quiet origin/master -- src/rt || echo "Runtime changed!"
+    displayName: Check for runtime changes.
+    name: setVarStep
 - job: 
   displayName: Linux
+  dependsOn: HasRuntimeChanged
+  variables: 
+    RTTests: $[ dependencies.HasRuntimeChanged.outputs['setVarStep.testRuntime'] ]
   pool:
     vmImage: 'ubuntu-16.04'
   strategy:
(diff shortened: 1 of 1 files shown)


mjp41:
@plietar

mjp41:
This roughly halves CI time, and will be a bigger improvement once I fix the systematic testing, which will probably take CI time up to 30mins.  With this it will be 10mins if you are working on the front end, and 30mins if you change the runtime.

.vsts-ci.yml
```diff
@@ -2,8 +2,23 @@ resources:
 - repo: self
 
 jobs:
+- job: HasRuntimeChanged
+  pool:
+    vmImage: 'ubuntu-16.04'
+  steps:
+  - script: |
+      echo "Determine if runtime changed."
+      git diff --quiet origin/master -- src/rt || echo "##vso[task.setvariable variable=testRuntime;isOutput=true]On" #set variable to testRuntime to On
```
rengolin, line 11: This could be simplified with:
```
    if git diff...; then
      ... turn RT tests on
    else
      .. turn RT tests off
    fi
```
as Paul suggested.

.vsts-ci.yml
```diff
@@ -2,8 +2,23 @@ resources:
...
+      git diff --quiet origin/master -- src/rt && echo "##vso[task.setvariable variable=testRuntime;isOutput=true]Off" #set variable to testRuntime to Off
+      git diff --quiet origin/master -- src/rt && echo "Runtime unchanged!"
+      git diff --quiet origin/master -- src/rt || echo "Runtime changed!"
+    displayName: Check for runtime changes.
+    name: setVarStep
 - job: 
   displayName: Linux
+  dependsOn: HasRuntimeChanged
+  variables: 
+    RTTests: $[ dependencies.HasRuntimeChanged.outputs['setVarStep.testRuntime'] ]
```
rengolin, line 21: Can't this be just "ON" and "OFF", to be used on the CMake flags and let CMake do the rest?

For example, if CMake would create a dependency from RT_TEST to "check" target, so that "ninja check" would do all that's required automatically.

mjp41: This is how you pass variables from one job to the next in Azure Dev Ops.  This is just setting up a variable in this job that is either "ON" or "OFF", and it gets passed into cmake using 
```
  -DRT_TESTS=$(RTTests)
```
This then sets up all the build targets and ctest tests to either include or not the runtime tests.

mjp41:
@rengolin I factored this as a separate job, so that all the different platforms had a consistent choice, and so I didn't have to implement the logic at the start of each platforms job.  

It seems to take about 6 seconds to calculate the variable, and then it has to schedule machines, I think it probably adds about 15seconds to the overall time to do this in a separate job.  But if we expand the selective compilation more, then I think the simplification in logic will be worth it.

.vsts-ci.yml
```diff
@@ -42,13 +60,14 @@ jobs:
       sudo apt-get install -y ninja-build
 
       sudo pip install wheel OutputCheck
+      echo RTTESTS=$(RTTests)
```
Theodus, line 63: Can this echo be removed?

rengolin:
It seems that "Format" and "HasRuntimeChanged" could run in parallel, but they don't ATM. Is this a simple thing to do?

If s, it would release all other jobs to run while format checks are in progress.

mjp41:
They do run in parallel just we have three CI requests, and the other two PRs have all the build machines.

PR #136, opened 2020-06-04
Fix potential double collection

mjp41:
Make the ref count the trigger for final collection on a cown.  Leak
detection can trigger the body of the cown to be collected, but
the stub will remain until the weak_count hits 0.  This fixes a bug
where the delayed dec ref lists are ignored by the leak detector. Now
when the leak detector frees an unreachable component, it performs the
decrefs, but does not signal that the stub can be collected until the
ref count reachs zero.

Previously, the thread field was set to zero to indicate the body had
been collected, now we use the bottom bit, so we don't confuse the
unadopted by a scheduler thread state, and the collected state.

### DIFF
--- a/src/rt/sched/cown.h
+++ b/src/rt/sched/cown.h
@@ -65,7 +65,10 @@ namespace verona::rt
     verona::rt::MPSCQ<MultiMessage> queue;
 
     // Used for garbage collection of cyclic cowns only.
-    std::atomic<SchedulerThread<Cown>*> thread;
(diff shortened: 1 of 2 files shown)


src/rt/sched/cown.h
```diff
@@ -273,7 +305,7 @@ namespace verona::rt
       Systematic::cout() << "Weak release " << this << std::endl;
```
Theodus, line 305: The comment above this function should be updated.

PR #140, opened 2020-06-05
Initial documentation on how Verona / C++ interop should work.

davidchisnall:
Also add a couple of experiments exploring parts of how this can work:

 - A process-based sandbox using snmalloc for memory management.
 - Some Clang code to explore interfacing with C++ via clang.

### DIFF
--- a/docs/foreign_code.md
+++ b/docs/foreign_code.md
@@ -1,0 +1,156 @@
+Verona Foreign Code Interop Design
+==================================
+
+The Verona language does not exist in isolation.
+Rewriting all of the useful C/C++ code that could be written in Verona would cost trillions of dollars and so is not a feasible option.
+Languages such as TypeScript and C# provide mature application development ecosystems that we wish to enable on top of Verona services.
+Verona will have to exist in a world with other languages but that should not compromise the security guarantees that we aim to provide.
+
+Core principles of interoperability
+-----------------------------------
+
+The Verona interoperability model will be built on the following principles:
+
+ - Foreign code should not imply unsafety.
+ - Interoperability should not compromise any security guarantees of the non-Verona language or library API.
+ - Most libraries are stateful, the correct granularity for interoperability is a library, not a function.
+
+The first of these relates to one of the core principles of Verona: there is no `unsafe` keyword.
+Verona targets security-critical workloads and it's very hard to make strong security claims about a language if any call to another language is able to compromise any and all of our safety guarantees.
+The Java JNI specification, for example, contains this disclaimer in the Design Overview chapter:
+
+> The programmer must not pass illegal pointers or arguments of the wrong type to JNI functions. Doing so could result in arbitrary consequences, including a corrupted system state or VM crash.
+
+The language makes an explicit declaration that any call to native code can undermine any of the invariants that the Java security model depends on. 
+This is a pragmatic decision to allow quick adoption but it means that no Java program can be considered safe until all native components have been removed.
+
+Modern C++ codebases work hard to provide memory safety guarantees at the API level.
+These guarantees are often lost in interop layers.
+For example, LLVM makes extensive use of typed smart pointers in the C++ APIs, but the Python bindings are written on top of the C API, which make type safety the caller's responsibility.
+Even if we can sandbox the resulting code, using a C++ library from Verona should not be more likely to introduce security vulnerabilities than using the same library from C++.
+The 2017 paper [*Finding and Preventing Bugs in JavaScript Bindings*](https://ieeexplore.ieee.org/document/7958598) showed that memory safety bugs in the C++ / JavaScript interop layer in Chrome are surprisingly common.
+Any time that a user has to hand-write code at a bindings layer to get memory safety right, we can assume that they will get it wrong and so we aim to automate as much as possible.
+
+The desire for sandboxing also leads to the conclusion that a 'foreign function interface' is exposing the wrong level of abstraction.
+What sandbox should a function run in?
+Our prior [CHERI-JNI](https://www.cl.cam.ac.uk/research/security/ctsrd/pdfs/201704-asplos-cherijni.pdf) work showed that treating libraries (in the loose sense, including any dependent libraries) as the unit of foreign code provided an abstraction that fitted well with sandboxing and allowed the user to determine the degree of state sharing between foreign code invocations.
+This abstraction also has software engineering benefits.
+LLVM, for example, does not have stable ABI (or API) guarantees across programs and so can cause linking errors if, for example, an application depends on one library that uses libclang from LLVM 10 to parse syntax highlighting and another that uses the JIT from LLVM 11 for GPU shaders.
+In a Verona world, these would be in different sandboxes with different 'global' namespaces and so would not conflict.
+
+### Sandboxes are surfaced as regions
+
+Sandboxing is a security *policy* that can be implemented with a variety of *mechanisms*.
+Sandboxing runs a component with a reduced set of privileges and limits its interactions with the outside world.
+It can be enforced with a separate process that runs with reduced OS privilege (e.g. the OpenSSH privesp model or Chromium's renderer sandboxes), with SFI techniques such as those used in WebAssembly compilers or NaCl (potentially with hardware acceleration from features such as Intel MPK), or with in-address-space hardware isolation such as CHERI.
+
+Sandboxes are difficult to program with in C-like languages because there is no useful construct in the abstract machine that can represent them.
+Web browsers are the most aggressively compartmentalised applications and yet run with a handful of different sandbox types.
+Because there is no abstraction in C for distinct sets of objects with mediated entry points, sandboxing code written in C leaks mechanism into the security code (for example, the idea that sandboxes are process).
+Any call between sandboxes must go via some RPC layer.
+Allocating memory inside a sandbox from outside and directly manipulating data structures is difficult to get right because the language doesn't impose any restrictions on where any given pointer can be stored.
+
+A Verona region is a set of related objects, with arbitrary topology, that have a single entry point.
+Sandboxes map cleanly to this abstraction.
+A sandbox's heap can contain pointers to any other data structures owned by the sandbox.
+It cannot contain pointers outside of the sandbox because following them would violate the sandbox policy.
+The Verona type system enforces this automatically by disallowing cross-region pointers.
+
+We hope that the region abstraction will make programming with sandboxes natural within the language.
+
+### Foreign libraries are classes
+
+If the unit of interoperability is a library, how do we surface that within Verona?
+We aim to treat foreign libraries just like any other module: as classes that can contain nested types.
+For example, we may import LLVM in something like this (straw-man syntax):
+
+```verona
+class LLVM
+{
+	[#native, #interface_language="c++"]
+	imports library("libLLVM.so", "llvm.h");
+}
+```
+
+This includes two attributes that tell the compiler the format of the library (native code) and how to parse the header (which, in this example, would have to be something provided by the user that included the desired headers).
+Once C++ module support is more widespread, module imports can take the place of header inclusions.
+
+This instructs the compiler to take the C++ namespace structure and inject each namespace as a class within the LLVM class and all C++ classes as nested classes.
+
+Instantiating this class would require the user to specify a sandboxing policy, for example (again, straw-man syntax):
+
+```verona
+var llvm = LLVM.Create(PROCESS);
+```
+
+This would create a new process that loaded the LLVM library and had RPC invocations for any of the LLVM functions that this program called.
+The `llvm` object in this example is a new isolated region.
+The programmer can then construct LLVM types within that region.
+
+Implementation
+--------------
+
+The proposed implementation strategy for C/C++ interoperability is to deeply embed clang.
+We are already depending on LLVM (and MLIR) for our compiler infrastructure and so embedding clang is not a significant addition.
+In the above example, the we would pass the `llvm.h` header to clang and generate an AST.
+Any lookup of symbols within the `LLVM` class's namespace would inspect this AST to find equivalent types.
+
+We want to avoid having to deal with the C/C++ ABI details in the Verona compiler.
+A suggestion from Chris Lattner provides a path to avoiding that by having clang synthesise wrapper functions.
+
+For example, consider some Verona code that wants to call `llvm::IRBuilder<>::CreateRetVoid`.
+This is a method in a templated class, so the template will need to be instantiated before it can be called and the caller then needs to be aware of C++ ABI details.
+This, like every other problem in computer science, can be addressed by a level of indirection, using clang to synthesise a function roughly equivalent to this :
+
+```c++
+__attribute__((used,always_inline))
+extern "C" llvm::ReturnInst *irbuilder_createretvoid(llvm::IRBuilder<> *b)
+{
+        return b->CreateRetVoid();
+}
+```
+
+This takes and returns all values as pointers and so the LLVM lowering is predictable as a function that takes and returns only pointers.
+For complex return values, the caller will need to allocate space and pass a pointer to the return.
+Clang generates IR somewhat like the following for this function.
+
+```llvm
+define dso_local %"class.llvm::ReturnInst"* @irbuilder_createretvoid(%"class.llvm::IRBuilder"*) #0 {
+  %2 = alloca %"class.llvm::IRBuilder"*, align 8
+  store %"class.llvm::IRBuilder"* %0, %"class.llvm::IRBuilder"** %2, align 8
+  %3 = load %"class.llvm::IRBuilder"*, %"class.llvm::IRBuilder"** %2, align 8
+  %4 = bitcast %"class.llvm::IRBuilder"* %3 to %"class.llvm::IRBuilderBase"*
+  %5 = call %"class.llvm::ReturnInst"* @_ZN4llvm13IRBuilderBase13CreateRetVoidEv(%"class.llvm::IRBuilderBase"* %4)
+  ret %"class.llvm::ReturnInst"* %5
+}
+```
+
+Creating this forced clang to instantiate a lot of templates.
+Fortunately, [Hal Finkel's C++ JIT prototype](https://github.com/hfinkel/llvm-project-cxxjit) demonstrated that this is possible without changes to the clang AST.
+
+This generated function is then trivial for the Verona compiler to call (all details of the C/C++ ABI are abstracted) and can then be inlined during the optimisation phase if desirable.
+More interestingly, LLVM will inline the `IRBuilder<>::CreateRetVoid` call into this function.
+The Verona compiler can then examine the result and determine if it is safe to inline at the caller without breaking the sandboxing invariants.
+For example, a function that ends up flattened can be made safe simply by enforcing the sandbox bounds checks on every LVM IR load and store instruction.
+Functions that do a lot of work, in contrast, may require a full domain switch.
+
+Exactly the same technique can be applied to field accessors.
+Getting and setting a field in a C or C++ POD `struct` is easy in the common cases, but may involve some complex ABI interactions for features such as bitfields in structures where some fields have custom alignment.
+To access these fields, the Verona compiler would simply make clang synthesize a 'get' and 'set' function for each field.
+This can then be inlined (with correct bounds checks applied).
+
+A similar technique applies to memory allocation.
+Without optimisation, any 'on stack' value types that refer to a sandbox's heap should be allocated in that heap and have their lifetimes tracked with the owning stack frame.
+Ideally, we'd promote these to the stack.
+This is possible if every function that is called on them can be safely inlined.
+For example, consider `std::shared_ptr` or `std::lock_guard`.
+These are almost stateless in their common implementation and refer to state stored on the heap.
+These would be safe to allocate on the stack (or even in a register: the storage is a single pointer).
+Methods may invoke calls into the sandbox, but those calls would not take the address of the on-stack object.  
+
+There are lots of open questions in terms of exposing C++ syntax (for example, operator overloading).
+
+Exporting Verona code to other languages
+----------------------------------------
+
+**TODO**
--- a/docs/foreign_code.md
+++ b/docs/foreign_code.md
@@ -11,7 +11,7 @@ Core principles of interoperability
 
 The Verona interoperability model will be built on the following principles:
 
- - Foreign code should not imply unsafety.
+ - Using foreign code should not require giving up on the safety guarantees of Verona.
  - Interoperability should not compromise any security guarantees of the non-Verona language or library API.
  - Most libraries are stateful, the correct granularity for interoperability is a library, not a function.
 
@@ -75,7 +75,9 @@ class LLVM
 This includes two attributes that tell the compiler the format of the library (native code) and how to parse the header (which, in this example, would have to be something provided by the user that included the desired headers).
 Once C++ module support is more widespread, module imports can take the place of header inclusions.
 
-This instructs the compiler to take the C++ namespace structure and inject each namespace as a class within the LLVM class and all C++ classes as nested classes.
+This instructs the compiler to expose the C++ namespace and class structure the `LLVM` Verona class.
+C++ namespaces and classes will appear as nested classes within the Verona `LLVM` class.
+For example, `LLVM.llvm.SmallVector[LLVM.int, 12]` corresponds to `llvm::SmallVector<int, 12>`.
 
 Instantiating this class would require the user to specify a sandboxing policy, for example (again, straw-man syntax):
 
@@ -90,7 +92,19 @@ The programmer can then construct LLVM types within that region.
 Implementation
 --------------
 
-The proposed implementation strategy for C/C++ interoperability is to deeply embed clang.
+C/C++ are among the most difficult cases for interoperability, for two reasons:
+
+ - They lack of a clear module system or interface format. 
+   C/C++ 'interfaces' are header files.  
+   These are just text concatenation and can contain arbitrary C/C++ code.
+   They have to be extracted by compiling the code and cannot be recreated with 100% fidelity by anything that is not a complete C/C++ preprocess, parser, and semantic analysis toolchain.
+ - ABI complexity. 
+   The standard C ABI is not too hard to handle for any given platform (except bitfields) but there are a load of non-standard GCC and MSVC extensions that give fine-grained control over structure layout.
+   Beyond that, C++ is even harder.
+   Things like diamond inheritance (handled differently in Itanium / MSVC ABIs), RTTI (handled completely differently in MSVC vs Itanium, subtly differently in Itanium vs Aarch32 vs Fuchsia), exception ABIs (totally different between Win32, Win64, and Itanium, with some significant differences between ARM and x86 Windows) add up to a very large amount of ABI logic that differs between platforms and architectures.
+   Getting these right is really, really hard in anything that isn't a complete C/C++ compiler.
+
+To address this, the proposed implementation strategy for C/C++ interoperability is to deeply embed clang.
 We are already depending on LLVM (and MLIR) for our compiler infrastructure and so embedding clang is not a significant addition.
 In the above example, the we would pass the `llvm.h` header to clang and generate an AST.
 Any lookup of symbols within the `LLVM` class's namespace would inspect this AST to find equivalent types.
--- a/experiments/clang_interface/Makefile
+++ b/experiments/clang_interface/Makefile
@@ -1,0 +1,70 @@
+
+all: clang_interface clang_interface-release
+
+clang_interface: clang_interface.cc Makefile
+	clang++80 clang_interface.cc -o clang_interface \
+		-fcolor-diagnostics \
+		-std=c++17 \
+		-fsanitize=address \
+		-g \
+		-I ~/llvm-project/clang/include/ \
+		-I ~/llvm-project/llvm/include/ \
+		-I ~/build/llvm-trunk-debug/include/ \
+		-I ~/build/llvm-trunk-debug/tools/clang/include/ \
+		-I /usr/local/llvm80/include \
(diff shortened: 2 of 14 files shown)


docs/foreign_code.md
```diff
@@ -0,0 +1,156 @@
...
+
+This would create a new process that loaded the LLVM library and had RPC invocations for any of the LLVM functions that this program called.
+The `llvm` object in this example is a new isolated region.
+The programmer can then construct LLVM types within that region.
+
+Implementation
+--------------
+
+The proposed implementation strategy for C/C++ interoperability is to deeply embed clang.
+We are already depending on LLVM (and MLIR) for our compiler infrastructure and so embedding clang is not a significant addition.
```
rengolin, line 94: Does this mean interoperation with other languages will also need to embed their front-ends?

Rust and any other LLVM-based language should be ok (module different LLVM versions they require), but we also want to embed C# and other non-LLVM based languages, no?

It would be bad to have two-tiered support for external languages.

docs/foreign_code.md
```diff
@@ -0,0 +1,156 @@
...
+In the above example, the we would pass the `llvm.h` header to clang and generate an AST.
+Any lookup of symbols within the `LLVM` class's namespace would inspect this AST to find equivalent types.
+
+We want to avoid having to deal with the C/C++ ABI details in the Verona compiler.
```
rengolin, line 98: `s/in the Verona compiler/in general/` :D

docs/foreign_code.md
```diff
@@ -0,0 +1,156 @@
...
+```c++
+__attribute__((used,always_inline))
+extern "C" llvm::ReturnInst *irbuilder_createretvoid(llvm::IRBuilder<> *b)
+{
+        return b->CreateRetVoid();
+}
+```
+
+This takes and returns all values as pointers and so the LLVM lowering is predictable as a function that takes and returns only pointers.
+For complex return values, the caller will need to allocate space and pass a pointer to the return.
```
rengolin, line 114: These will need to be special Verona containers with C++ objects inside, because we want to control their life-time inside Verona, where they will be used, and because the C++ side cannot see them.

If we return a pointer allocated by the C++ code, how do we make sure it isn't also freed by the C++ code before its lifetime expires in Verona?

docs/foreign_code.md
```diff
@@ -0,0 +1,156 @@
...
+Fortunately, [Hal Finkel's C++ JIT prototype](https://github.com/hfinkel/llvm-project-cxxjit) demonstrated that this is possible without changes to the clang AST.
+
+This generated function is then trivial for the Verona compiler to call (all details of the C/C++ ABI are abstracted) and can then be inlined during the optimisation phase if desirable.
+More interestingly, LLVM will inline the `IRBuilder<>::CreateRetVoid` call into this function.
+The Verona compiler can then examine the result and determine if it is safe to inline at the caller without breaking the sandboxing invariants.
+For example, a function that ends up flattened can be made safe simply by enforcing the sandbox bounds checks on every LVM IR load and store instruction.
+Functions that do a lot of work, in contrast, may require a full domain switch.
+
+Exactly the same technique can be applied to field accessors.
+Getting and setting a field in a C or C++ POD `struct` is easy in the common cases, but may involve some complex ABI interactions for features such as bitfields in structures where some fields have custom alignment.
```
rengolin, line 138: C++ inheritance is a particular PITA, and different across architectures.

docs/foreign_code.md
```diff
@@ -0,0 +1,156 @@
...
+To access these fields, the Verona compiler would simply make clang synthesize a 'get' and 'set' function for each field.
+This can then be inlined (with correct bounds checks applied).
```
rengolin, line 140: Would this be allowed to violate encapsulation? If not, would the compiler be required to understand the target language's properties?

experiments/clang_interface/Makefile
```diff
@@ -0,0 +1,70 @@
+
+all: clang_interface clang_interface-release
+
+clang_interface: clang_interface.cc Makefile
+	clang++80 clang_interface.cc -o clang_interface \
```
rengolin, line 5: Better to use "clang++", we don't want that to break when we upgrade. Let the environment control which version we use.

experiments/clang_interface/clang_interface.cc
```diff
@@ -0,0 +1,771 @@
...
+
+      fprintf(stderr, "Computing precompiled preamble\n");
+      {
+        auto t = TimeReport("Building PCH");
+        pchBuffer = generatePCH(headerFile, pchArgs);
+      }
+
+      fprintf(stderr, "Parsing fake CU\n");
+      const char* args[] = {
+        "clang", "-x", langName, "-I", "/usr/local/llvm80/include/", cu_name};
```
rengolin, line 283: `s/clang80/clang/`

experiments/process_sandbox/README.md
```diff
@@ -0,0 +1,36 @@
...
+One object is large and will be used for the sandbox's heap.
+The large object is mapped at the same address in both the parent and a child process, allowing intra-sandbox pointers to have the same values in both.
+This is mapped read-write in both processes.
+
+The smaller object is a single page, used for the snmalloc page map that covers the large region.
+This is mapped read-write in the parent over the region of the pagemap that contains metadata for the large region.
+It is mapped read-only in the child.
+
+The parent has an allocator associated with the sandbox that allows allocation in the sandbox's heap.
+The parent allocator's message queue is in the shared region, but the rest of the state is not.
```
rengolin, line 19: The shared region is writable by the child, wouldn't that be a potential problem?

experiments/process_sandbox/README.md
```diff
@@ -0,0 +1,36 @@
...
+On systems with Capsicum support, it then enters capability mode, at which point it has no access to any global namespaces other than those explicitly passed by the parent.
+The child then closes the file descriptors for the shared mapping and opens the requested library.
+The library initialisation code is the first untrusted code that runs in the child and it does so only after the child has dropped all OS rights.
+
+Finally, the child process enters a run loop waiting for messages from the parent.
+
+Note that, for this to be efficient, the OS must implement lazy commit so that allocating a large (e.g. 1GiB) shared memory region does not consume 1GiB of physical memory or swap unless it is actually used.
+
+Currently, the in-memory RPC mechanism used to invoke methods in the child is very high latency.
+This may not be a problem for Verona, where foreign calls are likely to be wrapped in `when` clauses.
```
rengolin, line 35: Better explain this with a bit more context. Not all readers will be familiar with Verona.

experiments/process_sandbox/README.md
```diff
@@ -0,0 +1,36 @@
...
+The parent allocator's message queue is in the shared region, but the rest of the state is not.
+When the allocator in the parent needs to update the pagemap, it does so directly.
+When an allocator in the child needs to update the parent, it communicates the update via a pipe to the parent.
+The parent then validates the requested update and performs the write.
+
+When the child starts, it first maps the shared region and initialises a version of snmalloc as its malloc implementation.
+The child process' snmalloc uses a custom PAL that is backed by the shared memory region and a custom PageMap to write updates.
+On systems with Capsicum support, it then enters capability mode, at which point it has no access to any global namespaces other than those explicitly passed by the parent.
+The child then closes the file descriptors for the shared mapping and opens the requested library.
+The library initialisation code is the first untrusted code that runs in the child and it does so only after the child has dropped all OS rights.
```
rengolin, line 28: So, `mmap` -> `fread` -> `fclose` -> `dlsym`?

docs/foreign_code.md
```diff
@@ -0,0 +1,156 @@
...
+Rewriting all of the useful C/C++ code that could be written in Verona would cost trillions of dollars and so is not a feasible option.
+Languages such as TypeScript and C# provide mature application development ecosystems that we wish to enable on top of Verona services.
+Verona will have to exist in a world with other languages but that should not compromise the security guarantees that we aim to provide.
+
+Core principles of interoperability
+-----------------------------------
+
+The Verona interoperability model will be built on the following principles:
+
+ - Foreign code should not imply unsafety.
```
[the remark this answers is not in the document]
mjp41: Maybe
> Integrating foreign code should not compromise the safety of the system.

docs/foreign_code.md
```diff
@@ -0,0 +1,156 @@
...
+Even if we can sandbox the resulting code, using a C++ library from Verona should not be more likely to introduce security vulnerabilities than using the same library from C++.
+The 2017 paper [*Finding and Preventing Bugs in JavaScript Bindings*](https://ieeexplore.ieee.org/document/7958598) showed that memory safety bugs in the C++ / JavaScript interop layer in Chrome are surprisingly common.
+Any time that a user has to hand-write code at a bindings layer to get memory safety right, we can assume that they will get it wrong and so we aim to automate as much as possible.
+
+The desire for sandboxing also leads to the conclusion that a 'foreign function interface' is exposing the wrong level of abstraction.
+What sandbox should a function run in?
+Our prior [CHERI-JNI](https://www.cl.cam.ac.uk/research/security/ctsrd/pdfs/201704-asplos-cherijni.pdf) work showed that treating libraries (in the loose sense, including any dependent libraries) as the unit of foreign code provided an abstraction that fitted well with sandboxing and allowed the user to determine the degree of state sharing between foreign code invocations.
```
mjp41, line 36: Perhaps remove "Our".

experiments/clang_interface/README.md
```diff
@@ -0,0 +1,12 @@
...
+
+This experiment is a quick-and-dirty hack to explore whether we can use clang to:
+
+ - Parse a C/C++ header.
+ - Cache the parse result so that the Verona compiler can do incremental builds.
+ - Resolve types and symbols.
+ - Instantiate templates.
+ - Synthesize functions that provide simple-ABI wrappers.
+
+It is incomplete and ugly code but serves to demonstrate that the clang APIs are sufficient for our purposes.
```
mjp41, line 12: Could you add a test or example of how to use this?  The value for others will be more significant if they can see it used briefly.

experiments/clang_interface/Makefile
```diff
@@ -0,0 +1,70 @@
+
```
mjp41, line 1: Why Make, rather than CMake?

experiments/process_sandbox/README.md
mjp41: Also, sounds like we are okay with "high-latency" in Verona.

experiments/clang_interface/Makefile
davidchisnall: Because this is *very* incomplete.  This is just about enough to test it on my machine, but it's not sufficiently complete for anything other than copying and pasting into other bits of the system.  I doubt actually running it is useful to anyone.

docs/foreign_code.md
```diff
@@ -0,0 +1,156 @@
...
+This abstraction also has software engineering benefits.
+LLVM, for example, does not have stable ABI (or API) guarantees across programs and so can cause linking errors if, for example, an application depends on one library that uses libclang from LLVM 10 to parse syntax highlighting and another that uses the JIT from LLVM 11 for GPU shaders.
```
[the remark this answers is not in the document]
davidchisnall: I'm not sure I understand this comment.

docs/foreign_code.md
```diff
@@ -0,0 +1,156 @@
...
+Web browsers are the most aggressively compartmentalised applications and yet run with a handful of different sandbox types.
+Because there is no abstraction in C for distinct sets of objects with mediated entry points, sandboxing code written in C leaks mechanism into the security code (for example, the idea that sandboxes are process).
+Any call between sandboxes must go via some RPC layer.
+Allocating memory inside a sandbox from outside and directly manipulating data structures is difficult to get right because the language doesn't impose any restrictions on where any given pointer can be stored.
+
+A Verona region is a set of related objects, with arbitrary topology, that have a single entry point.
+Sandboxes map cleanly to this abstraction.
+A sandbox's heap can contain pointers to any other data structures owned by the sandbox.
+It cannot contain pointers outside of the sandbox because following them would violate the sandbox policy.
+The Verona type system enforces this automatically by disallowing cross-region pointers.
```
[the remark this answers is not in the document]
davidchisnall: This is very implementation-specific.  With CHERI, you need to carve out a range of virtual address space.  With the process-based prototype, that's currently the case but it would be relatively easy to extend so that the heap is contiguous in the sandbox side and non-contiguous on the parent side, because the Verona compiler would already need some special logic for handling pointers within the sandbox (validating that they were in the range, at least), so could also handle a small set of non-contiguous mappings.

It's not clear to me that that this is actually useful though, because in practice it's typically important to impose a maximum memory limit on sandboxes to avoid DoS attacks.  In WebAssembly, this is intrinsic to the abstract machine (wasm memory regions are fixed size and the C/C++ lowering to WebAssembly places the entire heap in a single wasm region)

docs/foreign_code.md
```diff
@@ -0,0 +1,156 @@
...
+
+We hope that the region abstraction will make programming with sandboxes natural within the language.
```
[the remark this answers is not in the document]
davidchisnall: Creating a region and loading a shared library.  The region may or may not be wrapped in a single-region cown.

docs/foreign_code.md
davidchisnall: Not necessarily.  C/C++ are among the most difficult cases for two reasons:

 - Lack of a clear module system or interface format.  C/C++ 'interfaces' have to be extracted by compiling the code and cannot be recreated with 100% fidelity by anything that is not a complete C/C++ preprocess, parser, and semantic analysis toolchain.
 - ABI complexity.  The standard C ABI is not too hard to handle for any given platform (except bitfields) but there are a load of non-standard GCC and MSVC extensions that give fine-grained control over structure layout.  Beyond that, C++ is even harder. Things like diamond inheritance (handled differently in Itanium / MSVC ABIs), RTTI (handled completely differently in MSVC vs Itanium, subtly differently in Itanium vs Aarch32 vs Fuchsia), exception ABIs (totally different between Win32, Win64, and Itanium, with some significant differences between ARM and x86 Windows). Getting these right is really, really hard in anything that isn't a complete C/C++ compiler.

Most other languages that we'd want to interop with have some combination of simpler ABIs, cleaner module / interface definitions, and so on.  For .NET, interfaces are well defined and the binary format for assemblies is specified, as are various bits of the interop layer, so it should be fairly easy to add a .NET interface without needing to embed a .NET compiler.

docs/foreign_code.md
davidchisnall: This should be handled by value types already.  Verona value types can live only on the stack and have well-defined lifetimes (module @sylvanc actually extracting that bit of the spec from his brain and writing it on paper).

docs/foreign_code.md
davidchisnall: Good question.  I would expect it not to be allowed to violate encapsulation, so the clang code would need to perform the relevant access checks.  This would involve reusing / copying and pasting some sema logic.

experiments/clang_interface/Makefile
davidchisnall: I'm tempted to just remove this Makefile.  It works on my system (clang++ won't, I think, because it's too old to support C++17), but it hard codes things like the locations of my LLVM builds.  It's mostly there as an example: I don't expect anyone to actually use this code as-is.

experiments/process_sandbox/README.md
davidchisnall: In theory, yes.  That's why we have (had? @mjp41, did we publish it?) code in snmalloc to validate those pointers.  The worst that should be able to happen is that malicious code can trigger a UAF, but since any data in the sandbox is intrinsically untrusted this doesn't affect the threat model.

experiments/process_sandbox/README.md
davidchisnall: `mmap` -> `close` -> `cap_enter` -> `dlopen`.

The library is opened in the parent and a file descriptor to it and directory descriptors to the system library locations are inherited by the child.  This is quite complex logic, but there are comments.

experiments/process_sandbox/README.md
davidchisnall: I assume anyone who digs into the `experiments` subdirectory in the Verona repository will be familiar with Verona.  

How about:

```suggestion
This may not be a problem for Verona, where foreign calls are likely to be wrapped in `when` clauses, which can batch multiple operations within the library.
The asynchronous operation of `when` clauses hides latency, avoiding the blocking operations in the C++ proof-of-concept.
```

docs/foreign_code.md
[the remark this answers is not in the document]
davidchisnall: How about:

```suggestion
 - Using foreign code should not require giving up on the safety guarantees of Verona.
```

docs/foreign_code.md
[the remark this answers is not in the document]
davidchisnall: Note: It is fine for Verona *code* to operate on data in a foreign region, as long as the compiler inserts the correct dynamic checks.  It is not fine to depend on type safety for Verona *objects* in a foreign region.  We will have to impose some constraints on the kinds of object that can be allocated in a foreign region.

docs/foreign_code.md
```diff
@@ -0,0 +1,156 @@
...
+{
+	[#native, #interface_language="c++"]
+	imports library("libLLVM.so", "llvm.h");
+}
+```
+
+This includes two attributes that tell the compiler the format of the library (native code) and how to parse the header (which, in this example, would have to be something provided by the user that included the desired headers).
+Once C++ module support is more widespread, module imports can take the place of header inclusions.
+
+This instructs the compiler to take the C++ namespace structure and inject each namespace as a class within the LLVM class and all C++ classes as nested classes.
```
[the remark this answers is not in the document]
davidchisnall: How about:
```suggestion
This instructs the compiler to expose the C++ namespace and class structure the `LLVM` Verona class.
C++ namespaces and classes will appear as nested classes within the Verona `LLVM` class.
For example, `LLVM.llvm.SmallVector[LLVM.int, 12]` corresponds to `llvm::SmallVector<int, 12>`.
```

docs/foreign_code.md
```diff
@@ -0,0 +1,156 @@
...
+
+Instantiating this class would require the user to specify a sandboxing policy, for example (again, straw-man syntax):
+
+```verona
+var llvm = LLVM.Create(PROCESS);
+```
+
+This would create a new process that loaded the LLVM library and had RPC invocations for any of the LLVM functions that this program called.
+The `llvm` object in this example is a new isolated region.
+The programmer can then construct LLVM types within that region.
```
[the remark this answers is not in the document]
davidchisnall: The former.  How about:

```suggestion
The programmer can then construct types exposed via the `LLVM` class within that region.
```

docs/foreign_code.md
davidchisnall:
```suggestion
For complex return values, the caller will need to allocate space and pass a pointer to the return.
These will be exposed in Verona as value types, which cannot be stored on the heap in Verona.
```

experiments/process_sandbox/README.md
mjp41: I think at the moment it is "had thought about". I reviewed a lot of places it would be required, but did not implement it.

docs/foreign_code.md
davidchisnall: Agreed, not sure it needs calling out explicitly.

docs/foreign_code.md
davidchisnall: I want to call out that we're building on prior experience with designing safe interop systems.  Can you suggest a way of doing that?

experiments/clang_interface/README.md
davidchisnall: This is not intended to be used as-is, it's to demonstrate how various bits of clang the APIs can be used.  As the README says, it is very incomplete and just contains example code for the things in the list.  If you run it, it will spit out some LLVM IR for a function in a compilation unit that includes some C++ things.

docs/foreign_code.md
```diff
@@ -0,0 +1,156 @@
...
+
+Implementation
+--------------
+
+The proposed implementation strategy for C/C++ interoperability is to deeply embed clang.
```
davidchisnall, line 93:
```suggestion
C/C++ are among the most difficult cases for interoperability, for two reasons:

 - They lack of a clear module system or interface format. 
   C/C++ 'interfaces' are header files.  
   These are just text concatenation and can contain arbitrary C/C++ code.
   They have to be extracted by compiling the code and cannot be recreated with 100% fidelity by anything that is not a complete C/C++ preprocess, parser, and semantic analysis toolchain.
 - ABI complexity. 
   The standard C ABI is not too hard to handle for any given platform (except bitfields) but there are a load of non-standard GCC and MSVC extensions that give fine-grained control over structure layout.
   Beyond that, C++ is even harder.
   Things like diamond inheritance (handled differently in Itanium / MSVC ABIs), RTTI (handled completely differently in MSVC vs Itanium, subtly differently in Itanium vs Aarch32 vs Fuchsia), exception ABIs (totally different between Win32, Win64, and Itanium, with some significant differences between ARM and x86 Windows) add up to a very large amount of ABI logic that differs between platforms and architectures.
   Getting these right is really, really hard in anything that isn't a complete C/C++ compiler.

To address this, the proposed implementation strategy for C/C++ interoperability is to deeply embed clang.
```

docs/foreign_code.md
davidchisnall: I've added a suggestion to add most of this comment to the text above.

docs/foreign_code.md
davidchisnall: Now discussed above in the proposed change.

docs/foreign_code.md
mjp41: Does it mater that it is our experience.   I guess I have written too many double blind reviewed papers now.  Feel free to ignore.

experiments/clang_interface/README.md
mjp41: If you add a 
>  If you run it, it will spit out some LLVM IR for a function in a compilation unit that includes some C++ things.

to the readme, possibly with a command line, then I would know what to expect.

docs/foreign_code.md
```diff
@@ -0,0 +1,156 @@
...
+
+The desire for sandboxing also leads to the conclusion that a 'foreign function interface' is exposing the wrong level of abstraction.
+What sandbox should a function run in?
+Our prior [CHERI-JNI](https://www.cl.cam.ac.uk/research/security/ctsrd/pdfs/201704-asplos-cherijni.pdf) work showed that treating libraries (in the loose sense, including any dependent libraries) as the unit of foreign code provided an abstraction that fitted well with sandboxing and allowed the user to determine the degree of state sharing between foreign code invocations.
+This abstraction also has software engineering benefits.
+LLVM, for example, does not have stable ABI (or API) guarantees across programs and so can cause linking errors if, for example, an application depends on one library that uses libclang from LLVM 10 to parse syntax highlighting and another that uses the JIT from LLVM 11 for GPU shaders.
+In a Verona world, these would be in different sandboxes with different 'global' namespaces and so would not conflict.
```
SeanTAllen, line 39: I feel like this is touching the surface of something and then glides along.  I'm not sure what you re getting at, but it feels very important.

docs/foreign_code.md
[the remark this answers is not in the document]
SeanTAllen: "We will have to impose some constraints on the kinds of object that can be allocated in a foreign region."

Can you elaborate on that more?

experiments/clang_interface/README.md
davidchisnall: I honestly don't think there is any benefit to anyone in running this code.  The benefit is in reading it and incorporating bits of it elsewhere.  The only reason that I run it is to check that it works (for what it does).

docs/foreign_code.md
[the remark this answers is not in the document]
rengolin: If the errors are not a link time but, say, use of the wrong enum values and is only observable at run time. Things will crash in unexpected ways and we want to make it easier for the developer to be able to dig through the Verona stack trace until they get to the actual error.

Sandboxing model fits with its expected usage, but we should also aim to simplify error handling to isolate the sandbox error from the rest of Verona (this is vague, I know). Just thinking out loud, don't mind this comment.

docs/foreign_code.md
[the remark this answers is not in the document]
rengolin: This is true for web browsers that keep the memory while the page is open. But what about system software that allocates a bunch of stuff, use it and then throw it away. We need an elastic approach to memory allocation by giving back to the OS the memory that is no longer in use, even if just temporary.

docs/foreign_code.md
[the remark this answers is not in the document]
rengolin: I thought all Verona objects would be allocated by the stub caller (for arguments and return values) and not the external code. I believe that's what JNI used to do. I don't know how we could guarantee that the external code hasn't messed up a Verona object internally if it's in its own memory.

experiments/process_sandbox/README.md
rengolin: Right, close enough. :)

docs/foreign_code.md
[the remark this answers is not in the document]
davidchisnall: Ah, I understand.  Yes, we can't guarantee ABI's aren't subtly broken in foreign code (though we can provide tooling to make it easier for anything that uses semantic versioning).  

Sandbox failure is an interesting topic that we don't yet have a complete story on.  We want to kill the process on contract violation.  We shouldn't kill the process when untrusted code misbehaves because doing so gives malicious code an easy DoS vector.  We also don't want to turn every sandbox call into something that returns an option type that needs checking between calls.  Unless we make every argument to sandboxed functions implicitly take an option type, so we can chain together calls, but need to check them when we return to Verona.  That might actually be a nice intermediate.  For example, the following C++ method:

```c++
Value *CreateRet(Value *);
```

Would be exposed in Verona as:

```verona
Value* | SandboxError CreateRet(Value* | SandboxError);
```

You could then pass around all of the `Value* | SandboxError` values from similar calls but when you wanted to store them anywhere you'd need to check if the sandbox had done a bad thing. If you pass the failure value into any of the wrapper functions, the compiler would insert code to skip the call.  

This would let you write a `when` clause that the compiler would turn into an asynchronous invocation that ran a bunch of commands in the sandbox asynchronously and returned a promise that would evaluate to `SandboxError` if the sandbox died.  If we catch the signal and preserve the error state within the sandbox, then we could get a stack trace (or even detach the child and hand if over to an interactive debugger) when the promise is accessed.

docs/foreign_code.md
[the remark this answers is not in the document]
davidchisnall: See the process isolation demo.  This allows allocating objects in the child from the parent *without* an RPC.  We don't aim to guarantee that the object hasn't been messed up, we aim to guarantee that messing up the object only lets you corrupt other objects within the sandbox.

docs/foreign_code.md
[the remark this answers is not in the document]
rengolin: I like the chain of optionals to only be checked at usage and the sandbox error state to be kept (not necessarily inside a sandbox object) until it's out of scope (whether or not it's consumed). 

This will be tricky to implement but it would be a really clean interface for sandbox ewrror handling.

Another way of DoS would be to have multiple failures with large error contexts that never go away. Say this case (in clearly no Verona code):

```
  Sandbox s;
  while (true) {
    when (resource L) {
       s(L).dispatch(); // or something
       // crashes, (s)'s error is kept, but not out of context
       s.tearDown(); // Doesn't remove the error
    }
    // all previous invocations of the external code accumulate their errors
    // they never go out of context, we never make use of them
    // How many of these can we keep in memory?
  }
```

So, perhaps we should also make sure that the same sandbox can only have one error context (the last one) and to be cleared once it's invoked again. What's not clear to me is how to differentiate them if they're running on different resources `L` (which could be a template or inherited parameter).

docs/foreign_code.md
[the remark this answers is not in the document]
davidchisnall: @SeanTAllen, any object that is allocated in a sandbox region becomes untrusted immediately after any untrusted code is allowed to execute in the untrusted region (which, with process sandboxing, is immediately, unless we have OS support for suspending all threads in a child process - this is much easier with SFI or CHERI sandboxes, where we control their scheduling).  This means that we need to check any pointers that we follow and we can't trust that they haven't been tampered with.  For example, we can't trust a vtable pointer for a verona object.  We may be able to load the vtable pointer, validate it, and then invoke a method, but we'll need the compiler to generate methods that check access.

We especially need to be careful of tearing and TOCTOU vulnerabilities, because malicious code in the sandbox can potentially race you in accessing different fields in an object, so any pair of fields that need accessing atomically will need to be copied out and the copies used.  

We probably need to start with some *very* tight constraints on what kinds of objects can be allocated in untrusted regions.  Something similar to C++'s POD types: no dynamic dispatch, all fields are primitive or point to other objects with the same restrictions.

docs/foreign_code.md
davidchisnall: Basically, there is no global namespace.  If you import a shared library, each instantiation of that library will see its own copy of all globals.  If you import two versions of the same library in different sandboxes, their namespaces are isolated as well as their heaps.  This prevents a bunch of annoying things that come from the C/UNIX dynamic linkage model.  Glibc and Android have some support in their linkers for loading libraries in different namespaces to address this - in prototyping, we should definitely use that, so even without sandboxing we don't touch the global namespace by accident.

docs/foreign_code.md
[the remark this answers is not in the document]
davidchisnall: Once a sandbox has hit an error, it *must* be torn down (or, at least, never invoked again).  We did that in CHERI-JNI because calling into something in an undefined state is... ill-advised.  

Your example wouldn't type check because `s` is a region and so the `when` clause would claim ownership of it and `s` effectively drops out of scope after the first loop iteration, so the second loop iteration is invalid code.  I'm not sure if I can transform it into something that typechecks and shows the problem.

PR #141, opened 2020-06-08
Make CI ignore changes to just doc or experiments.

mjp41:
If a PR only has changes in the docs/experiments directory or to markdown files, then we can ignore CI on this PR.


rengolin:
Weird, this change was marked as "Doc changes only". I don't think it should.

mjp41:
@rengolin I had forgotten to compare to origin/master.  So it decided there were no changes.

PR #142, opened 2020-06-08
Fix pattern matching syntax

Theodus:


mjp41:
So a semi-independent thing that might be helpful.  I want a syntax for providing explicit type annotations at an arbitrary point in the code. This is required to help when type inference has failed. 
```
  x : T;
```
Was what I was thinking about, but at the moment, `:` is not a reserved symbol.  If we made : a reserved symbol would that help you here, as you would know it is not part of atom?

Theodus:
> If we made : a reserved symbol would that help you here, as you would know it is not part of atom?

I agree that that would be a convenient feature. But I don't think it would help parsing types as patterns since it's probably an issue of using `|` as a type operator and also as a pattern separator. Enforcing capitalization of type names may also help.

mjp41:
If you lose the colon, then the `|` would become ambiguous, but it doesn't actually matter.  
```
match (e)
  U64 | I32 {}
  ...
```
and
```
match (e)
  : U64 | I32 {}
  ...
```
have the same meaning but different syntaxs.

Also, in principle one could provide a `|` operator on a type, but you would not be able to use that in a pattern expression.  This would break a substitution principle:
```
let x = e 
match (y)
  x {...}
  ...
```
may have a different meaning to
```
match (y)
  e {...}
  ...
```
As the `|` will be parsed differently in `e` if it contains one.

rengolin:
This seems like an unresolved RFC, but without a description, I can't make out what the problem is just from the title and the comments. Is this still relevant? Should we continue working on this as an RFC PR (moving it to draft) or has this issue been deprecated by more recent developments (in which case we should close this)?

rengolin:
Added issue #279 to track the discussion about the syntax.

PR #144, opened 2020-06-09
DevOps LLVM Build

rengolin:
Builds LLVM, creates a tarball and uploads that as an artefact. This can
be used by the Verona build instead of rebuilding LLVM every time.

For now, the DevOps pipeline needs to set the $(GOOD_HASH) manually, but
the idea is to automate this later in the process for finding the new
LLVM commit that we want to rebase on top of.

### DIFF
--- a/.llvm-ci.yml
+++ b/.llvm-ci.yml
@@ -1,0 +1,64 @@
+resources:
+  repositories:
+  - repository: llvm
+    type: github
+    endpoint: LLVMServiceConnection
+    name: llvm/llvm-project
+
+trigger: none
+
+pr:
+- llvm-build
+
+jobs:
+- job: 
+  displayName: LLVM Linux Build
+  pool: VeronaLinuxLLVMBuild
+  strategy:
+    matrix:
+      Clang-8 Debug:
+        CC: clang
+        CXX: clang++
+        CXXFLAGS: -stdlib=libstdc++
+        BuildType: Debug
+        Asan: Off
+  steps:
+  - checkout: llvm
+
+  - script: |
+      set -eo pipefail
+      git checkout $(GOOD_HASH)
+    displayName: 'Move tree to known good LLVM commit for Verona'
+
+  - script: |
+      set -eo pipefail
+      sudo apt-get update
+      sudo apt-get install -y clang ninja-build cmake lld
+    displayName: 'Install Build Dependencies'
+
+  - task: CMake@1
+    displayName: 'CMake'
+    inputs:
+      cmakeArgs: |
+        ../llvm -GNinja -DCMAKE_BUILD_TYPE=$(BuildType) -DCMAKE_C_COMPILER=$(CC) -DCMAKE_CXX_COMPILER=$(CXX) -DCMAKE_CXX_FLAGS=$(CXXFLAGS) -DLLVM_ENABLE_PROJECTS=mlir -DLLVM_BUILD_EXAMPLES=ON -DLLVM_TARGETS_TO_BUILD="X86" -DCMAKE_BUILD_TYPE=Debug -DLLVM_ENABLE_ASSERTIONS=ON -DLLVM_ENABLE_LLD=ON
+
+  - script: |
+      set -eo pipefail
+      ninja check-mlir
+    workingDirectory: build
+    failOnStderr: true
+    displayName: 'Compile and check'
+
+  - script: |
+      set -eo pipefail
+      # FIXME: Be more precise on what we keep
+      rm -f /tmp/VeronaLLVM-$(GOOD_HASH).tar.xz
+      CPUS=$(nproc --all)
+      XZ_DEFAULTS="--threads $CPUS" tar Jcf /tmp/VeronaLLVM-$(GOOD_HASH).tar.xz .
+    failOnStderr: true
+    displayName: 'Create artifact'
+
+  - task: PublishPipelineArtifact@1
+    inputs:
+      targetPath: /tmp/VeronaLLVM-$(GOOD_HASH).tar.xz
+      artifactName: VeronaLLVM-$(GOOD_HASH)


rengolin:
This is just Linux for now until we get the infra ready, then we'll need Windows and Mac builders. With Mac builders slower than others, we may need to do the finding process in multiple stages (first on Linux/Windows and then confirm on Mac).

rengolin:
Well, the LLVM Build job is running on pre-commit, this isn't good. I need to disable it.

rengolin:
Ok, now the LLVM Build will not run on pre-commit.

.llvm-ci.yml
```diff
@@ -0,0 +1,59 @@
...
+  displayName: LLVM Linux Build
+  pool: VeronaLinuxLLVMBuild
+  strategy:
+    matrix:
+      Clang-8 Debug:
+        CC: clang
+        CXX: clang++
+        CXXFLAGS: -stdlib=libstdc++
+        BuildType: Debug
+        Asan: Off
```
davidchisnall, line 19: Is it worth turning Asan on?  I normally do release and debug + ASan builds for testing (release is fast, if something goes wrong then debug + ASan is quicker at finding failures than just debug).

.llvm-ci.yml
```diff
@@ -0,0 +1,59 @@
...
+      rm -f /tmp/VeronaLLVM-$(GOOD_HASH).tar.xz
+      CPUS=$(nproc --all)
+      XZ_DEFAULTS="--threads $CPUS" tar Jcf /tmp/VeronaLLVM-$(GOOD_HASH).tar.xz .
+    failOnStderr: true
+    displayName: 'Create artifact'
+
+  - task: PublishPipelineArtifact@1
+    inputs:
+      targetPath: /tmp/VeronaLLVM-$(GOOD_HASH).tar.xz
+      artifactName: VeronaLLVM-$(GOOD_HASH)
```
davidchisnall, line 59: The artefact name should probably contain the build configuration and ASan state as well, so that we can build different LLVM configurations using your matrix and pull the correct one in.

davidchisnall:
Shiny!  Do you think we can extend the CMake for the build so anyone can just pull down the pre-built LLVM instead of building their own?

.llvm-ci.yml
rengolin: Not really. We only want to build LLVM that has been tested upstream, by selecting the commit hashes that had been proven to pass the kinds of buildbots we care about. This is the job for another script, but essentially, if we find a commit that has already passed ASAN on Linux and Windows, we don't need to test it again here. Same for self-hosting, Arm builds, test-suite, etc.

.llvm-ci.yml
rengolin: I'm only planning to build one per arch/OS. I still need to add that, though.

rengolin:
> Shiny! Do you think we can extend the CMake for the build so anyone can just pull down the pre-built LLVM instead of building their own?

What do you mean by that? Extend LLVM's or Verona's CMake? Who's anyone.

The way we did here, DevOps artefacts will be used to "pull down" instead of building. If we want something more universal, we can use LLVM's buildbot build cache, but we'd take a lot more than we need and won't be able to control the build configuration.

.llvm-ci.yml
davidchisnall: That isn't my goal.  I find ASan LLVM builds *incredibly* useful for catching places where I've made memory management errors by not correctly understanding implicit ownership conventions from LLVM APIs *used in my code*.  I'm not expecting to find new LLVM bugs, I'm expecting to find memory-management bugs in our compiler by running the tests with Debug + ASan.

.llvm-ci.yml
davidchisnall: We can't link a Release build of the Verona compiler against a Debug build of LLVM and we should be testing both in CI, so we will need Release builds of LLVM as well.

.llvm-ci.yml
rengolin: I see. Still, why not just build debug+asan and use that for all Verona build types?

davidchisnall:
For reference, with other out-of-tree front ends, I build:

 - A shared-library Release build of LLVM.
 - A static-library Debug + ASan build of LLVM.
 - A Release build of my front end, linked against a Release build of LLVM.
 - A Debug + ASan build of my front end, linked against a Release build of LLVM.

I test in both.  For my normal compile-test cycle, I do my release build and run tests and kick off a debug build in the background (shared linking LLVM takes < 1 second, static linking takes much longer).  If a test fails, I run it again in the Debug + ASan version, which usually gives me an assert failure or ASan violation quickly and lets me test it.  Once everything is passing and the feature I'm working on is complete, I run the full test suite in the Debug + ASan version, which catches a few cases where I've used the LLVM APIs in a subtly incorrect way.

.llvm-ci.yml
rengolin: Ok

davidchisnall:
> > Shiny! Do you think we can extend the CMake for the build so anyone can just pull down the pre-built LLVM instead of building their own?
> 
> What do you mean by that? Extend LLVM's or Verona's CMake? Who's anyone.

Verona's CMake, so that it can pull down a pre-build LLVM matching the build config and LLVM hash if we have one, rather than building locally.
 
> The way we did here, DevOps artefacts will be used to "pull down" instead of building. If we want something more universal, we can use LLVM's buildbot build cache, but we'd take a lot more than we need and won't be able to control the build configuration.

Are DevOps artefacts only accessible from DevOps?  I thought they were public.  If they're public, then adding a CMake step that tries to grab a pre-built version if one is available would be convenient.  I expect the LLVM build to be 95% of the total Verona build time, so this reduces barriers for external contributors.

rengolin:
> I test in both. For my normal compile-test cycle, I do my release build and run tests and kick off a debug build in the background (shared linking LLVM takes < 1 second, static linking takes much longer). If a test fails, I run it again in the Debug + ASan version, which usually gives me an assert failure or ASan violation quickly and lets me test it. Once everything is passing and the feature I'm working on is complete, I run the full test suite in the Debug + ASan version, which catches a few cases where I've used the LLVM APIs in a subtly incorrect way.

So, all you need from this job is { Release, Debug+ASAN } from each Arch/OS. Means that, for now, we only need a Linux matrix of two jobs. Once we introduce Windows/Mac, they'd also have two jobs.

rengolin:
> Verona's CMake, so that it can pull down a pre-build LLVM matching the build config and LLVM hash if we have one, rather than building locally.

Ah, yes, that's the whole idea behind separate jobs, yes.

> Are DevOps artefacts only accessible from DevOps? I thought they were public. If they're public, then adding a CMake step that tries to grab a pre-built version if one is available would be convenient. I expect the LLVM build to be 95% of the total Verona build time, so this reduces barriers for external contributors.

I was more concerned with CI, because although a local LLVM build takes some time, you only really built it once. It would only have an impact on Verona developers when we update the hash.

The biggest problem is, though, if we're bisecting across LLVM hash changes, or if developers have two branches with different hashes, but this happens with local builds or remote dumps and is not specific on how we defer the builds.

rengolin:
Note: This patch should not change anything other than the yaml file. Other pull requests should do that, later, once we're clear on *how* we'll build LLVM asynchronously.

PR #148, opened 2020-06-12
Recusive deallocation tests and fixes

mjp41:
Recursively deallocating structures if implemented incorrectly can cause a stack overflow.  This PR adds twos test
* Deeply nested region test - very long chain of traceable regions
* Deeply nested cown test - very long chain of cowns going via a region at each level

The PR also fixes the deeply nested cowns code to queue recursive calls to avoid stack overflow.

### DIFF
--- a/src/rt/sched/cown.h
+++ b/src/rt/sched/cown.h
@@ -1164,7 +1164,7 @@ namespace verona::rt
       yield();
       weak_release(alloc);
 
-      // Collect recursively reachable cowns 
+      // Collect recursively reachable cowns
       while (!current.empty())
       {
         auto a = (Cown*)current.pop();
--- a/src/rt/sched/cown.h
+++ b/src/rt/sched/cown.h
@@ -1146,18 +1146,18 @@ namespace verona::rt
      **/
     void queue_collect(Alloc* alloc)
     {
-      thread_local ObjectStack* queue = nullptr;
+      thread_local ObjectStack* work_list = nullptr;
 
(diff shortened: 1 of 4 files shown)


rengolin:
Just a comment on choosing "very large" ranges: different machines/systems have different thresholds. What stresses one doesn't stress the other, what works in one, breaks the other. As our validation matrix expands, it will be really hard to find the righ values for every test.

The LLVM sanitizers have a lot of heuristics to know what stresses x86, Arm and each variation. It was a pain to get there and I'm not sure they're still relevant, with the improvement of Arm machines.

I think we need this kind of test, and perhaps for now constant values are ok. But we need to keep in mind that this does not scale and may become irrelevant if we don't make sure the values are actually stressing each system we test on.

Creating dynamic tests (to find the best values) also don't scale. They can crash often or worse, run for very long times. Another approach would be to create a container with precise limits that are as big as our smallest test machine, and then test the boundaries of the algorithm inside the container.

But obviously, this is not a request to implement this now, just heads up to make sure we have a plan to do that regarding stress tests.

sylvanc:
In this particular case, the stress test is over the size of the stack, so less machine sensitive than it may initially seem.

My "eventually" comment is that eventually I'd like to get rid of the `thread_local` (for performance reasons, so this needs validation). But definitely not now. This looks good.

rengolin:
This seems like a generally good change, but I can't see what failed in the build as the results have been flushed away. @mjp41 can you push again to trigger a new build?

PR #149, opened 2020-06-15
Mark all submodules as 'shallow'

rengolin:
This speeds up CI builds as it only checks out the last commit status.

We shouldn't have to submit any changes to the external repositories via
Verona anyway, so it should be good long term.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -2,16 +2,21 @@
 	path = src/rt/external/snmalloc
 	url = https://github.com/Microsoft/snmalloc
 	branch = master
+	shallow = true
 [submodule "external/pegmatite"]
 	path = external/pegmatite
 	url = https://github.com/CompilerTeaching/Pegmatite
+	shallow = true
 [submodule "external/CLI11"]
 	path = external/CLI11
 	url = https://github.com/CLIUtils/CLI11
+	shallow = true
 [submodule "external/fmt"]
 	path = external/fmt
 	url = https://github.com/fmtlib/fmt.git
+	shallow = true
 [submodule "external/cpp-peglib"]
 	path = external/cpp-peglib
 	url = https://github.com/yhirose/cpp-peglib.git
 	branch = master
+	shallow = true


plietar:
I actually do develop and PR snmalloc from within my Verona clone (well not so much these days, but I used to). I’m sure I can work around this, there’s probably a way to unshallow a clone (or just adjust my workflow), so it is only a minor inconvenience.

Can’t this be done by adding a flag to the git clone / git submodule command instead?

rengolin:
You can do `git fetch --unshallow`. ([Ref](https://stackoverflow.com/questions/6802145/how-to-convert-a-git-shallow-clone-to-a-full-clone))

Better to do that once on the few cases we need than make it the default behaviour.

rengolin:
Also, seems this doesn't work as expected. I'm still seeing the whole repos checked out (LLVM was the big teller in my branch). Turns out, this is a [git issue](https://stackoverflow.com/questions/2144406/how-to-make-shallow-git-submodules/47374702#47374702). We should leave it as is in case git changes, but we're all going to get deep branches for now.

SeanTAllen:
I'd check DevOps docs to see if it allows you to control the checkout command for CI.

Previously with CirrusCI and CircleCI, I had it doing the submodule and depth as part of a single CI checkout.

rengolin:
The problem is that you can't directly `git clone` from a specific commit, just HEAD. If you want a particular commit you need to have it in your history (if not HEAD) or clone from a branch where the HEAD is the commit you want (I tihnk).

DevOps provides two options: `fetchDepth` as a number and `submodules` with `true` (don't recurse) and `recursive`. `.gitmodules` provides us with a `branch` option, which would allow us to shallow clone a repo from a special branch that has that specific commit as HEAD.

But that's a number of cross dependencies which will make the CI more brittle (too many things to change during test and production), as well as having to manage an LLVM fork (we can't create a `verona` branch on LLVM's main repo).

plietar:
Can we set the DevOps `submodules` option to false, and then issue our own `git submodule init/update` command?

rengolin:
We could, but it still wouldn't work with shallow clones because we specify a commit hash on each submodule. 

There seems to be a way to clone only the refs, then choose the one you want and then pull the changes of that specific ref, but AFAICS it doesn't work with submodules.

plietar:
Are you sure? The following sequence of commands works for me:
```
git clone https://github.com/microsoft/verona
cd verona/
git submodule init
git submodule update --depth 1 --recursive
```

It is faster than if I do it without the `--depth 1` option, and the console output suggests it did perform a shallow clone (very small number of objects downloaded).

rengolin:
Ok, I'll try that on my branch. The checkout with LLVM is taking more than 7 minutes on the CI.

rengolin:
It worked! From 7min to 40s again. Thanks!

PR #150, opened 2020-06-16
Initial module system

sylvanc:
This adds an initial module system, where a directory is a module, and a module is treated as an enclosing class (and can be generic).

### DIFF
--- a/src/ast/grammar.peg
+++ b/src/ast/grammar.peg
@@ -16,9 +16,12 @@ typeparam <- id
 constraints <- constraint*
 constraint <- 'where' id oftype inittype
 
-field <- id oftype initexpr ';'
+field <- qualifier id oftype initexpr ';'
+qualifier <- (static / private)*
+static <- 'static'
+private <- 'private'
 
-function <- funcname sig (block / ';')
+function <- qualifier funcname sig (block / ';')
 funcname <- (id / sym)?
 sig <- typeparams params oftype constraints
 params <- '(' list(param)? ')'
@@ -61,7 +64,7 @@ guard <- ('if' '(' seq ')')?
 
 atom <-
   break / continue / return / yield / let / new / lambda /
-  hex / binary / float / int / interp_string / string /
+  hex / binary / float / int / interp_string / string / true / false /
   id / sym / tuple / typeargs
 
 break <- 'break'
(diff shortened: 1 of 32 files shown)


rengolin:
The AST is changing radically quite frequently, as expected, due to the evolving design of the language. But these changes will impact the MLIR generation for a while...

To help with that problem, on the MLIR branch, I have a set of helpers to translate ast structure into syntax (getFunctionBody, getFunctionArgs, etc). This won't solve all teething issues but at least will make it clear what is AST logic, what is MLIR and what is in between.

Not for this patch, but once we merge the MLIR changes, it would be good for some of us to go through the examples and enhance them to traverse the tree in a menaingful way. They should be an API for an MLIR generator to query the AST, ie. looks like MLIR calls but is implemented as AST tree search.

The main benefit is that changing the MLIR side should not break the AST and vice versa. When you have changes, even big ones like this, it should be simple to update the API to get the same information, without needing to know how the MLIR generation works internally.

rengolin:
Sorry, I renamed the build and GitHub didn't recognised. This build is green.

plietar:
On the topic of generic modules, can a module “import” itself at another instantiation, like classes do. Of course we can’t support non regular recursion (eg. `M[X]` uses `M[C[X]]`), but we can have `M[X]` use `M[A]`. Can a module name itself, to be used to fulfill an interface?

sylvanc:
That's a great question. Currently no, because it would form a "cyclic dependency". Of course, there is no such actual dependency, and we should be able to handle this as another instance of mutually recursive types.

On the other hand, doing this with a syntax like `type ThisModule = ".";` seems awkward. Perhaps we allow an optional identifier in the `module` syntax? Would its only use be self-reference?

src/ast/grammar.peg
```diff
@@ -74,7 +77,7 @@ lambda <- sig block
 tuple <- '(' list(expr)? ')'
 
 keyword <-
-  'type' / 'class' / 'where' /
+  'type' / 'class' / 'where' / 'static' / 'private' /
```
Theodus, line 80: This rule should include `module`

src/ast/module.cc
```diff
@@ -0,0 +1,249 @@
...
+    const std::string& path,
+    const std::string& ext,
+    err::Errors& err)
+  {
+    std::map<std::string, ModulePtr> modules;
+    std::vector<ModulePtr> stack;
+    bool ok = true;
+
+    auto canonical_path = path::canonical(path);
+    auto m = make_module(path);
```
Theodus, line 104: There should be a check here for `dep->ast == nullptr`

sylvanc: There's no `dep` on line 104. Did you mean line 137? If so, there's no need for `dep->ast` to be valid - `err` will already have the parse error(s), and the module DAG code will work either way.

Theodus: Sorry wrong location. But `sym::build` segfaults due to a null AST when an invalid module path is given.

sylvanc:
Not _yet_.

PR #151, opened 2020-06-16
Refactor hashmap

Theodus:
This PR makes several fixes and simplifications to the user interface of the Robin Hood hash map:
- Use `uintptr_t` instead of `size_t` when messing with `Object*` key bits
- Replace entry wrapper classes and `key_of` with a single set of defaulted static callbacks
- Enforce strict entry types
- Hide manipulation of `Object*` key bits from the user
- Add overload for insert without the need to create a dummy index

### DIFF
--- a/src/rt/ds/hashmap.h
+++ b/src/rt/ds/hashmap.h
@@ -20,6 +20,28 @@ namespace verona::rt
    *
    * The key must be Object*, because the low bits are used to encode marking
    * status (MARK) and distance-to-initial-bucket (DIB).
+   *
+   * Static callbacks may be supplied as the second template parameter as
+   * follows:
+   *   ```cpp
+   *   struct MyMapCallbacks
+   *   {
+   *     static void on_insert(Entry& e)
+   *     {
+   *       // called every time an entry is inserted into the map
+   *       // i.e. when `insert` returns true.
+   *     }
+   *
+   *     static void on_erase(Entry& e)
+   *     {
+   *       // called every time an entry is deleted from the map
+   *     }
+   *   };
+   *
+   *   using MyMap = PtrKeyHashMap<std::pair<Object*, uint8_t>, MyMapCallbacks>;
+   *   ```
+   *   Note that all callbacks will have the low bits of the Entry key unmarked
+   *   so that it may be used as expected.
    */
   template<typename Entry, typename CB = DefaultMapCallbacks<Entry>>
   class PtrKeyHashMap
--- a/src/rt/ds/hashmap.h
+++ b/src/rt/ds/hashmap.h
@@ -160,7 +160,7 @@ namespace verona::rt
       return dib;
     }
 
-    inline void set_entry(size_t index, Entry& entry, size_t dib)
+    inline void set_entry(size_t index, Entry entry, size_t dib)
     {
       // When entry is passed in, it may or may not have a mark bit, but it
       // must have no dib. If the DIB is greater than the maximum DIB, encode
@@ -313,7 +313,7 @@ namespace verona::rt
     }
 
     // Returns true if newly added, false if previously present.
-    bool insert(Alloc* alloc, Entry& entry, size_t& location)
+    bool insert(Alloc* alloc, Entry entry, size_t& location)
     {
       const auto& orig_key = key_of(entry);
       assert(orig_key == unmark_key(orig_key));
@@ -337,8 +337,8 @@ namespace verona::rt
             location = index;
 
           // This index is empty, insert here.
-          set_entry(index, entry, dib_entry);
           CB::on_insert(entry);
+          set_entry(index, std::move(entry), dib_entry);
 
           count++;
           grow(alloc);
@@ -369,7 +369,7 @@ namespace verona::rt
           // The DIB of the entry to insert is greater than the DIB of the
           // entry at this index. Insert o here, and continue looking for
           // somewhere to insert other.
-          set_entry(index, entry, dib_entry);
+          set_entry(index, std::move(entry), dib_entry);
 
           key_of(tmp) = (uintptr_t)get_pointer(key_of(tmp));
           entry = std::move(tmp);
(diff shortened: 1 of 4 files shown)


src/rt/ds/hashmap.h
```diff
@@ -39,7 +49,32 @@ namespace verona::rt
     {
       // The Object* returned will retain its RememberedSet mark bit, but not
       // its dib.
-      return (Object*)(p & ~(size_t)DIB_MAX);
+      return (Object*)(p & ~DIB_MAX);
+    }
+
+    template<typename>
```
mjp41, line 55: I wonder if this would be clearer as something like
```C++
    template<typename>
    struct decompose_entry_type : std::false_type
    {
    };

    template<>
    struct decompose_entry_type<Object*> : std::true_type
    {
      using value_type = Object*;
      static constexpr bool is_set = true;
    };

    template<typename V>
    struct decompose_entry_type<std::pair<Object*, V>> : std::true_type
    {
      using value_type = V;
      static constexpr bool is_set = false;
    };
```
At the moment, `is_valid_pair` is performing two jobs, but that isn't clear on first reading.

```C++
    static_assert(
      decompose_entry_type<Entry>::value,
      "Map Entry must be Object* or std::pair<Object*, V>");

    using ValueType = typename decompose_entry_type<Entry>::value_type;
    static constexpr bool is_set = decompose_entry_type<Entry>::is_set;
```
```

File: src/rt/ds/hashmap.h [hunk 2]
```diff
@@ -351,10 +411,15 @@ namespace verona::rt
...
+    bool insert(Alloc* alloc, Entry entry)
     {
       size_t dummy;
-      auto unique = insert(alloc, entry, dummy);
+      return insert(alloc, std::move(entry), dummy);
+    }
+
+    void insert_unique(Alloc* alloc, Entry entry)
+    {
+      auto unique = insert(alloc, std::move(entry));
```
mjp41, line 422: This feels a bit suspicious.   The Entry has been copied into this call, and then moved?  Given you are using callbacks rather than destructors now, is this even needed now?

mjp41: Also, this is going to call `on_insert` internally, which feels like it is going to have called that repeatedly?

src/rt/region/externalreference.h
```diff
@@ -199,11 +150,11 @@ namespace verona::rt
 
     void merge(Alloc* alloc, ExternalReferenceTable* that)
     {
-      for (auto& e : *that->external_map)
+      for (auto e : *that->external_map)
```
mjp41, line 153: The iterator is now returning copies of the entries, so the `std::move` does nothing useful?  I think this code is going to perform an incref and decref for everything it merges, rather than no inc/dec refs.

Theodus: the entry type here is `std::pair<Object*, ExternalRef*>` so there won't be any implicit increfs/decrefs with the raw pointers. I'll add a commit soon to forward moved arguments properly

src/rt/ds/hashmap.h
Theodus: `on_insert` is only called when the entry is inserted, not when updated. I'll add a note about that in the docs

src/rt/region/externalreference.h
mjp41: If we merge two tables that are completely disjoint then this is going to cause increfs, as the `insert_unique` will call `on_insert` as it wasn't there before, and the `std::move` is not going to invalidate the original version in table being merged from, so the destructor will call `on_delete`?

Could you log the increfs and decrefs before and after your changes?

Theodus: before:
```
theodus|~/src/microsoft/verona/src/rt/build(master !)$ ninja func-con-rememberedset && ./func-con-rememberedset
[2/2] Linking CXX executable func-con-rememberedset
incref 0x603000000130
decref 0x603000000130
incref 0x6030000000a0
decref 0x6030000000a0
decref 0x6030000000a0
decref 0x603000000130
incref 0x6030000002e0
decref 0x6030000002e0
decref 0x603000000250
decref 0x6030000002e0
incref 0x6030000005b0
incref 0x6030000005b0
decref 0x6030000005b0
decref 0x6030000005b0
decref 0x603000000520
decref 0x6030000005b0
decref 0x603000000490
incref 0x603000000820
incref 0x603000000820
decref 0x603000000820
decref 0x603000000820
decref 0x603000000700
decref 0x603000000790
decref 0x603000000820
```

after:
```
theodus|~/src/microsoft/verona/src/rt/build(theo-hashmap !)$ ninja func-con-rememberedset && ./func-con-rememberedset
[2/2] Linking CXX executable func-con-rememberedset
incref 0x603000000130
decref 0x603000000130
incref 0x6030000000a0
decref 0x6030000000a0
decref 0x6030000000a0
decref 0x603000000130
incref 0x6030000002e0
decref 0x6030000002e0
decref 0x603000000250
decref 0x6030000002e0
incref 0x6030000005b0
incref 0x6030000005b0
decref 0x6030000005b0
decref 0x6030000005b0
decref 0x603000000520
decref 0x6030000005b0
decref 0x603000000490
incref 0x603000000820
incref 0x603000000820
decref 0x603000000820
decref 0x603000000820
decref 0x603000000700
decref 0x603000000790
decref 0x603000000820
```

mjp41: Thanks.  Could you run that on the external reference tests?  It is that merge code that worries me.

Theodus: Ah ok, now I see what you mean. This does add infcrefs/decrefs when merging. I'll try to figure out how to fix that

Theodus: I've brought the increfs/decrefs back down to what they were previously. Sorry, I didn't look closely enough at the shenanigans in the wrapper class move constructor.

mjp41: Thanks.  No worries. It was only on second reading that I realised the issue. C++ tricks to make clients harder to get wrong, have implementation hard to get right ;-)

Theodus:
@mjp41 Are there still concerns or is this good to merge?

PR #152, opened 2020-06-17
Add Windows/Mac support for LLVM Build

rengolin:
Updating LLVM's yaml file and Azure's scale set script.

### DIFF
--- a/.llvm-ci.yml
+++ b/.llvm-ci.yml
@@ -6,6 +6,7 @@ resources:
     name: llvm/llvm-project
 
 jobs:
+############################################## Linux Builds
 - job:
   displayName: LLVM Linux Build
   pool: verona-llvm-build-linux
@@ -26,6 +27,9 @@ jobs:
         BuildType: Debug
         BuildName: debug
         Sanitizer: Address
+  variables:
+  - name: PKG_NAME
+    value: verona-llvm-x86_64-linux-$(BuildName)-$(GOOD_HASH)
   steps:
   - checkout: llvm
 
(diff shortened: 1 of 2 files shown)


.llvm-ci.yml
```diff
@@ -82,3 +82,73 @@ jobs:
...
+  timeoutInMinutes: 120
+  strategy:
+    matrix:
+      Debug:
+        BuildType: Debug
+        BuildName: debug
+      Release:
+        BuildType: Release
+        BuildName: release
+  # TODO: Figure how to set version on Windows
```
letmaik, line 98: What's the issue? Can you not do the same as in Linux?

rengolin: I can install `grep` with choco, but `coreutils` (for `cut`) doesn't seem to be available. I just need to figure out but that's not important because none of those builds will be off a release branch anyway, sothe LLVM version doens't matter much.

mjp41: Can you  use `wsl` for somethings?

rengolin: I probably could, but not before I get the MLIR generator upstream. For now, anything reasonable that works is better than trying inprovements.

This build is a step towards getting my changes upstream, so they can be tested and we can continue the development in multiple layers (ast, mlir, rt).

rengolin:
Tested mac support, works too, so bundling both on this PR.

Still no simple way to get the version, but leaving at 11 for now. We'll have time to look at that in the near future, before it flips to 12.

rengolin:
Right, good to go, all tests pass. I'll merge this so I can finish the first MLIR PR. We can revise the verison after that.

PR #153, opened 2020-06-17
initial test for modules

Theodus:


src/ast/module.cc
```diff
@@ -105,7 +105,7 @@ namespace
     bool ok = true;
 
     auto canonical_path = path::canonical(path);
-    auto m = make_module(path);
+    auto m = make_module(canonical_path);
```
sylvanc, line 108: Why is the non-portable `canonical_path` being used here? The module's name is the `path`, i.e. how it was referred to in the source code, appended to the same `path` for the module that "originally" (DAG-ordered) referenced this module.

Theodus: Sorry, I've moved the canonicalization of the module path where it is needed, when parsing files associated with the module path.

sylvanc: That shouldn't be necessary either (it may _be_ necessary, but it shouldn't be). What's the error without that? That change will cause canonical paths to propagate in strange places.

Theodus: The error is that on Windows, if a verona file path is given including `\` characters much of the path will be lopped of. So for example: `C:\Users\tbutler\src\microsoft\verona\testsuite\parse\ast-parse\import.verona` results in a path of `module_foo/` instead of the full path

sylvanc: Ok, I see. Instead of the change you have here, can you please try doing:

```
auto m = make_module(path::from_platform(path));
```

Theodus: Thanks, that works. Updated

PR #158, opened 2020-06-23
Azure script

rengolin:
Small change to allow different machines to be created


rengolin:
Sorry, rogue commit in the middle.

PR #162, opened 2020-06-26
Change CI/Nightly builds

rengolin:
In preparation for the Verona+LLVM builds, this separates the PR tests from the nightly tests.

The CI tests will run on every PR update and contain one build on each platform that is likely to catch most bugs:
 * GCC Linux Release
 * Clang Linux Release with asserts, debug info and ASAN
 * VS2019 Windows Release
 * Clang MacOS Release

These are currently taking ~15 min avg, 30 min max. The new reduces set seems to be 8min avg, 22min max.

Cloning LLVM, downloading the cache and working with LLVM will add another 5~10 minutes, which will keep the old times.

The CI VM pools tend to clog if we use too many, so having less builds per loop means more loops in parallel. If we are worried about some change, nothing stops us from running the nightly build on a PR specifically, only when necessary.

The nightly build will have the following:
 * GCC Linux (Release, Debug)
 * Clang Linux (Release, Debug), with and without ASAN
 * VS2019 Windows (Release, RelDbgInfo)
 * Clang MacOS (Release, Debug)

These are, for now, the same speed as before (15/30) but working with LLVM may increase considerably, given that some builds (ex. Windows Debug) will have to compile LLVM inside the build on a different machine, with less availability.

Once approved and merged to master, I'll change the GitHub policy to require the new CI job instead.


plietar:
What happens on a nightly failure? Can we email the people that have committed that day, or post into the Infrastructure channel?

rengolin:
> What happens on a nightly failure? Can we email the people that have committed that day, or post into the Infrastructure channel?

I believe DevOps has Teams integration, we could dump the results every night. I'm not sure it has the "email all committers", though.

With any segmented CI, we need to decide what happens when an asynchronous build fails. In LLVM, if the fix isn't quick enough (for some definition of speed), we revert the commit until a new one can pass the failed test. In this case, the new commit would come as a PR and we'd fire the nightly job against it, and that would show on Github directly.

If we don't revert, someone that has access to the DevOps will have to liaise with the author and help test until green. For now I think all of us have access, so we can all push to a branch and test directly until green. 

But once we get to the point where we have so many external contributors that we can't deal with the aftermath of failures, we can think of a better approach. I'd be very happy when we get to that point to feel sad to have to re-work the CI again. :)

rengolin:
Ok, so I have created a dashboard and linked that to the Infrastructure channel on the group. In addition, the internal team will receive an email every time the nightly build fails.

The options to pick the committers' email didn't seem obvious to me, but for now, it should sufice. Anyone knowing DevOps more than me, feel free to add that option.

rengolin:
Tried building some recent builds to compare. They're not identical, so the errors are slightly different, but with added GCC to the CI mix, we're getting all error/warning issues straight away.

On Theo's dev branch (theo-backpressure). GCC catches errors that Clang didn't, and so does VS, which is a positive change. Clang tests still fail, so that didn't change.

On Matt's PR (#148), the build doesn't exist anymore, but there were compilation and test failures. The CI picks up a compilation error, while the nightly test picks up the test errors. If fixing the compilation errors would also fix the test errors, we don't need to do anything, else we could move at least one debug build from nightly to CI.

I guess this could be pretty fluid exchange, so we shouldn't need to decide right now. If a lot of stuff ends up leaking to the nightly build to catch, we rebalance towards CI. If the nightly never catches anything more than CI, we rebalance towards nightly.

rengolin:
Ok, everything is done. the nightly fails because master is actually broken on GCC, the old build doesn't start because it doesn't exist in this branch. As soon as I get an approval, I'll change the Github requirements (from "microsoft.verona" to "Verona CI") and then this will be mergeable and all future PR updates will go through the new build).

rengolin:
@mjp41 I have addressed your comments while Github was offline.

PR #163, opened 2020-06-29
Add ENABLE_ASSERTS to cmake

rengolin:
This option enables asserts without the need of a debug build. The
implementation is stolen from LLVM because it's not trivial.

With the new flag, we enable assertions on all CI and nightly builds.
LLVM builds already had assertions enabled before.


rengolin:
Checked that the output in the CMake step is correct (for Linux, Mac and Windows) and ran the nightly just to be sure. Everything looks fine.

PR #164, opened 2020-06-29
Fix gcc warnings

Theodus:
Workaround for gcc bug: https://gcc.gnu.org/bugzilla/show_bug.cgi?id=95975

### DIFF
--- a/devops/ci.yml
+++ b/devops/ci.yml
@@ -60,9 +60,9 @@ jobs:
       set -eo pipefail
       sudo add-apt-repository ppa:ubuntu-toolchain-r/test
       sudo apt-get update
-      sudo apt-get install -y ninja-build gcc-10 g++-10
-      sudo update-alternatives --install /usr/bin/gcc gcc /usr/bin/gcc-10 100
-      sudo update-alternatives --install /usr/bin/g++ g++ /usr/bin/g++-10 100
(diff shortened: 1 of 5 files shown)


rengolin:
There's another warning on GCC: `-Werror=unused-but-set-parameter` on `static void gc_notified(Object* o)` and `static void gc_final(Object* o, Object* region, ObjectStack& sub_regions)`.

I know they're not related, but we could probably fix all warnings in one go and get a green build.

rengolin:
best to update both ci and nightly to gcc-10. If we want more coverage, we can create more jobs in the nightly one with different versions, but I don't think that'd be necessary right now.

Theodus:
Everything compiles now, but it seems that there are many runtime errors in the gcc builds.

Theodus:
I think fixing those should be out of the scope for this PR.

rengolin:
It does, but we cannot merge without the main test passing. :(

Theodus:
Welp. I guess I'll be looking into it tomorrow

rengolin:
It was working as of last week, so it's not something that has never worked before... I think a quick bisect could show you the culprit.

But yeah, tomorrow is fine. :)

Theodus:
Closing. See #165

PR #175, opened 2020-07-02
Use LLVM's cmdline interface

### LINKED ISSUE #167: Refactor cmdline args in verona-mlir to use LLVM's
Initially, the arguments were passed to the AST library as is, so we needed to use the same kind of command line parsing.

Now they interface is opaque, so we can use LLVM's, which is more suited to a large number of arguments and will integrate well with the existing infrastructure for their arguments.

rengolin:
Fixes #167


src/mlir/verona-mlir.cc
```diff
@@ -14,107 +13,111 @@
...
 #include "mlir/InitAllPasses.h"
 
+#include "llvm/Support/CommandLine.h"
 #include "llvm/Support/InitLLVM.h"
 
 namespace
 {
+  namespace cl = llvm::cl;
+  // Input file name
+  static cl::opt<std::string> inputFile(
```
davidchisnall, line 23: `static` is redundant (and confusing) for things in the anonymous namespace.

src/mlir/verona-mlir.cc
```diff
@@ -14,107 +13,111 @@
...
   };
-
-  // Parse cmd-line and set defaults
-  Opt parse(int argc, char** argv)
+  static cl::opt<enum OutputKind> outputKind(
+    "emit",
+    cl::init(OutputKind::None),
+    cl::desc("Output type"),
+    cl::values(clEnumValN(OutputKind::MLIR, "mlir", "MLIR")),
+    cl::values(clEnumValN(OutputKind::LLVM, "llvm", "LLVM IR")));
```
davidchisnall, line 55: Is this giving us a flag like `-emit mlir`, rather than `-emit-mlir`?  If so, that's different from other LLVM front ends in a slightly confusing way.  Does `cl::opt` have a simple way of providing a pair of options like `-emit-llvm` / `-emit-mlir` and using whichever was the last one specified?

src/mlir/verona-mlir.cc
```diff
@@ -14,107 +13,111 @@
...
+
+  // Optimisations enabled
+  static cl::opt<bool> enableOpt("opt", cl::desc("Enable optimizations"));
```
davidchisnall, line 58: Why `opt` not `O`?

src/mlir/verona-mlir.cc
```diff
@@ -14,107 +13,111 @@
...
+
+  // Grammar file
+  static cl::opt<std::string>
+    grammarFile("g", cl::init(""), cl::desc("Grammar file"));
```
davidchisnall, line 62: `-g` is normally used for debug.  Can we use something else here?

src/mlir/verona-mlir.cc
```diff
@@ -14,107 +13,111 @@
...
+      }
       else
-        newName += ".mlir";
-    }
-    else
-    {
-      newName += ".ll";
+      {
+        std::string newName =
+          filename.substr(0, filename.find_last_of('.')).str();
```
davidchisnall, line 103: I don't really like the fact that we're using strings for paths here.  Can we not use `llvm::sys::path` and its `replace_extension` method?

src/mlir/verona-mlir.cc
```diff
@@ -14,107 +13,111 @@
...
+          newName += ".ll";
+        }
+        outputFile = newName;
+      }
     }
-    return newName;
+
+    // Default grammar
+    if (grammarFile.empty())
+      grammarFile = path::directory(path::executable()).append("/grammar.peg");
```
davidchisnall, line 120: What is the use case for allowing the user to override this?  I understand that, for prototyping, we're going to want to be able to quickly iterate on it, but that can be done in the source tree.

Looking for the .peg file in the same location as the executable is also a problem, since it precludes an FHS-compliant installer.

src/mlir/verona-mlir.cc
rengolin: sorry, copy and paste.

src/mlir/verona-mlir.cc
rengolin: AFAIK there is no simple way to do that on a single flag (even in other cli libs). I'd have to create two different options like I had before.

Given that we can't yet emit LLVM anyway, I'd be fine in deferring that for later and just assume we want to always emit MLIR.

src/mlir/verona-mlir.cc
rengolin: Copy&paste from the MLIR example. `-O` usually requires a number while this will for now be a boolean flag. I could implement this with a number and turn on the optimisations when `O > 0`.

src/mlir/verona-mlir.cc
rengolin: Copy&paste from `verona-ast`, trying to use the same flag, but you're right, it is odd. Honestly, I don't think we should be changing the grammar in `verona-mlir` in the same way we do in `verona-ast`. I'll remove this option.

src/mlir/verona-mlir.cc
rengolin: Good point, will do.

src/mlir/verona-mlir.cc
rengolin: `verona-ast` has a case for changing the gramma, `verona-mlir` does not.

However, `verona-mlir` is also a development tool and should never be deployed by an installer.

The actual compiuler will have a driver to resolve all path issues, so this won't be a problem in the actually installable product.

rengolin:
Addressed the reviews, one commit per comment.

rengolin:
@davidchisnall looks good?

PR #177, opened 2020-07-06
Refactor error handling in MLIR generator

### LINKED ISSUE #168: Move verona-mlir error handling to llvm::ErrorOr + Stacking
The current prototype throws exceptions on errors, but that requires LLVM to be built with EH+RTTI, which makes the libraries larger and slower.

We should start using LLVM's `ErrorOr` in a stackable way (a list of errors) so we can emulate stack unwind and MLIR's `LogicalResult` for local true/false methods.

But first, we'll need to disentangle `pegmatite`'s build from LLVM, as it relies on EH itself.

rengolin:
Moving from exception handling to `llvm::Error` and `llvm::Expected` classes. This is a proof-of-concept idea on how to handle errors in the project and the implementation will grow to encompass everything MLIR/LLVM related.

Created two new error types, `RuntimeError`, for generic errors and `ParsingError` for Verona-to-MLIR conversion errors, with file location information.

Two new tests created in `mlir-fail` directory to match the two main errors we can easily reproduce.

The dependency on EH/RTTI continues in LLVM because peglib still uses EH and we need to link them together in `verona-mlir`. Unlinking both projects to move LLVM back to `no-rtti` and `no-exceptions` is left for future work.

Fixes #168


src/mlir/error.h
```diff
@@ -0,0 +1,67 @@
...
+
+#include <string>
+
+// Error Handling
+//
+// This file implements the various types of errors that the generator methods
+// can return when trying to parse Verona files to MLIR, and then lowering to
+// LLVM IR.
+//
+// The all must have a description. Parsing related errors must also have a
```
Theodus, line 19: __They__ all?

src/mlir/verona-mlir.cc
```diff
@@ -142,47 +143,19 @@ int main(int argc, char** argv)
...
-      {
-        std::cerr << "ERROR: cannot convert Verona file " << inputFile
-                  << " into MLIR" << std::endl
-                  << e.what() << std::endl;
-        return 1;
-      }
-      break;
+      check(gen.readAST(m->ast));
     }
+    break;
```
Theodus, line 148: Small nit, feel free to ignore. I prefer the `break`s in a switch statement be at a consistent level of indentation.

rengolin: I'm using clang-format and that's what it chose. If we change that, the clang-format test will fail in CI.

src/mlir/error.h
rengolin: oops

src/mlir/generator.cc
```diff
@@ -223,22 +217,44 @@ namespace mlir::verona
...
     }
-    builder.create<mlir::ReturnOp>(getLocation(ast), last);
+    if (auto err = last.takeError())
+      return std::move(err);
+    builder.create<mlir::ReturnOp>(getLocation(ast), *last);
 
     return func;
   }
 
+  mlir::Type Generator::parseType(const ::ast::Ast& ast)
```
mjp41, line 241: Why did you move the location of this function?

src/mlir/error.h
```diff
@@ -0,0 +1,67 @@
...
+  llvm::Error runtimeError(llvm::StringRef desc);
+
+  // Parsing Error: Errors when parsing Verona into MLIR
+  class ParsingError : public llvm::ErrorInfo<ParsingError>
+  {
+    std::string desc;
+    mlir::Location loc;
+
+  public:
+    static char ID;
```
davidchisnall, line 51:
```suggestion
    /// Unique identifier.  This value is unused but its address is used to uniquely identify this class.
    static char ID;
```

src/mlir/error.h
```diff
@@ -0,0 +1,67 @@
...
+    ParsingError(llvm::StringRef desc, mlir::Location loc)
```
davidchisnall, line 52: Doc comment describing what `desc` is?

src/mlir/error.h
```diff
@@ -0,0 +1,67 @@
...
+    : desc(desc), loc(loc)
+    {}
+    void log(llvm::raw_ostream& OS) const override
```
davidchisnall, line 55: Doc comment?  Is `log` part of some other infrastructure (is there a `dump` method somewhere that uses `log`?)

src/mlir/error.h
```diff
@@ -0,0 +1,67 @@
...
+    {
+      OS << desc << " at ";
+      loc.print(OS);
+    }
+    std::error_code convertToErrorCode() const override
```
davidchisnall, line 60: Same comment as above.

src/mlir/generator.cc
```diff
@@ -31,29 +31,31 @@ namespace mlir::verona
...
-    parseModule(ast);
+    if (auto err = parseModule(ast))
+      return err;
 
     // On error, dump module for debug purposes
     if (mlir::failed(mlir::verify(*module)))
     {
       module->dump();
-      throw std::runtime_error("MLIR verification failed from Verona file");
+      return runtimeError("MLIR verification failed from Verona file");
```
davidchisnall, line 44: First caller and now I see how this API is meant to be used.  

If this string is presented to the user, we should provide an enum value rather than a string and map that to a string later, so that it can be localised without needing to modify huge numbers of unrelated source files.  The parser in the interpreter already does this.

src/mlir/generator.cc
```diff
@@ -31,29 +31,31 @@ namespace mlir::verona
...
-      throw std::runtime_error("No input filename provided");
+      return runtimeError("No input filename provided");
 
     // Read an MLIR file
     auto srcOrErr = llvm::MemoryBuffer::getFileOrSTDIN(filename);
 
     if (auto err = srcOrErr.getError())
-      throw std::runtime_error(
+      return runtimeError(
         "Cannot open file " + filename + ": " + err.message());
```
davidchisnall, line 59: We should not be doing string concatenation like this if this is a human-facing error message.  The existing error reporting code uses the `fmt` library to collect format strings and their arguments so that they can be localised and rendered for humans.

src/mlir/generator.cc
rengolin: to match the order in the header file.

src/mlir/generator.cc
rengolin: We should do that as part of a localisation effort. I'd rather think about it and reach a consensus than rush a half-baked implementation. Internationalisation tends to spiral out of control if done badly initially.

src/mlir/error.h
```diff
@@ -0,0 +1,67 @@
...
+    }
+  };
+  // Create a parsing error and return
+  llvm::Error runtimeError(llvm::StringRef desc);
+
+  // Parsing Error: Errors when parsing Verona into MLIR
+  class ParsingError : public llvm::ErrorInfo<ParsingError>
+  {
+    std::string desc;
+    mlir::Location loc;
```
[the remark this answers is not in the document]
rengolin: These objects can be chained, each having their own `desc` and `loc`. Though, for now, there aren't any needs for chaining errors, as the steps are completed before the next one starts: Verona -> MLIR, MLIR opt, MLIR -> LLVM.

We could have a `RuntimeError` chained with a `ParsingError`, emitting a stack similar to:
```
  Runtime error: failed to lower to MLIR, because:
  Parsing error: operations not implemented yet at file.verona:10:4
```

PR #178, opened 2020-07-07
Add comments to error.h

rengolin:


src/mlir/error.h
```diff
@@ -19,49 +19,78 @@
...
+// string value, which doesn't work when the error could be emitted in
+// multiple languages. We need a mechanism where strings could come from
+// any language with a formatted output for a list of arguments, filled in
+// by the caller.
 namespace mlir::verona
 {
   // RuntimeError: Errors that have no relation to the source file
   class RuntimeError : public llvm::ErrorInfo<RuntimeError>
   {
+    // Description of the error
```
davidchisnall, line 33:
```suggestion
    // Human-readable description of the error
```

src/mlir/error.h
```diff
@@ -19,49 +19,78 @@
...
     std::string desc;
 
   public:
+    // Internal ID to control matching
```
davidchisnall, line 37:
```suggestion
    // Internal ID to control matching.  The value of this field is never used, its address is used as a unique identifier.
```

src/mlir/error.h
```diff
@@ -19,49 +19,78 @@
...
     RuntimeError(llvm::StringRef desc) : desc(desc) {}
+
+    // Public ErrorInfo interface, requires raw_ostream to output
+    // the string in the right way. This is used by error handlers.
     void log(llvm::raw_ostream& OS) const override
     {
       OS << desc;
     }
+    // Public ErrorInfo interface, converts to error_code for
+    // interoperation with other error handling mechanisms.
```
davidchisnall, line 48:
```suggestion
    // interoperation with other LLVM error handling mechanisms.
```

src/mlir/error.h
```diff
@@ -19,49 +19,78 @@
...
+    // For now, we don't care about that, but once we start working
+    // with other error handling we should implement this for real.
```
davidchisnall, line 50:
```suggestion
    // FIXME: For now, we don't care about that, but once we start working
    // with other error handling we should implement this for real.
```

src/mlir/error.h
```diff
@@ -19,49 +19,78 @@
...
     std::error_code convertToErrorCode() const override
     {
       return llvm::inconvertibleErrorCode();
     }
   };
-  // Create a parsing error and return
+
+  /// Create and return a new runtime error.
+  /// The returned error will own a copy of the string passed as `desc`.
```
davidchisnall, line 58: This is the only case where the doc comments are actually doc comments.  I'd prefer if all of them were.  LLVM typically uses `///`, but support for that in editors is a bit unreliable.  I generally perfer:
```c
/**
 * Some comment
 */
```
It uses more vertical space, but also gives a clear visual indicator.

src/mlir/error.h
```diff
@@ -19,49 +19,78 @@
...
   // Parsing Error: Errors when parsing Verona into MLIR
   class ParsingError : public llvm::ErrorInfo<ParsingError>
   {
+    // Description of the error
     std::string desc;
+    // File location where the error occurred
     mlir::Location loc;
 
   public:
+    // Internal ID to control matching
```
davidchisnall, line 70:
```suggestion
    // Internal ID to control matching.  The value of this field is never used, its address is used as a unique identifier.
```

src/mlir/error.h
```diff
@@ -19,49 +19,78 @@
...
+    // the string in the right way. This is used by error handlers.
+    // We must make sure the line info is printed.
     void log(llvm::raw_ostream& OS) const override
     {
       OS << desc << " at ";
       loc.print(OS);
     }
+    // Public ErrorInfo interface, converts to error_code for
+    // interoperation with other error handling mechanisms.
+    // For now, we don't care about that, but once we start working
```
davidchisnall, line 85:
```suggestion
    // FIXME: For now, we don't care about that, but once we start working
    // with other error handling we should implement this for real.
```

src/mlir/error.h
```diff
@@ -19,49 +19,78 @@
...
 
   public:
+    // Internal ID to control matching
     static char ID;
     ParsingError(llvm::StringRef desc, mlir::Location loc)
     : desc(desc), loc(loc)
     {}
+    // Public ErrorInfo interface, requires raw_ostream to output
+    // the string in the right way. This is used by error handlers.
+    // We must make sure the line info is printed.
```
davidchisnall, line 77:
```suggestion
    // Public interface defined by `ErrorInfo`, used for error handlers to report the human-readable error message.
    // This version displays the message description and the source location to which it corresponds.
```

src/mlir/error.h
rengolin: Not necessarily. `std::error_code` is standard and can be used by other mechanisms, standard or not.

rengolin:
I'm using multi-line syntax for multi-line comments, triple slashes for methods and public interfaces and double slashes for inline comments.

src/mlir/error.h
davidchisnall: Ah, I see.  Perhaps then:
```suggestion
interoperation with error-handling mechanisms that use the C++11 error types.
```

rengolin:
Last issue and rebase

rengolin:
@davidchisnall looks good?

PR #183, opened 2020-07-09
Initial Implementation for Runtime Backpressure

### LINKED ISSUE #33: Runtime Backpressure
Verona currently does not have a system to prevent a cown's message queue from growing out of memory. I nearly have a working example of backpressure for many cowns sending to a single cown on my fork, which is based on the backpressure system in the Pony runtime. Before I continue to extend this to multimessages in general, I would like to know if this is a feature that the team working on Verona would like to see implemented.

Theodus:
This is an initial version of the runtime backpressure system.

The primary goal of backpressure is to prevent runaway growth of a cown's message queue that may eventually result in a Verona application running out of memory. This system is loosely based on Pony's generalized runtime backpressure. Two tests (perf-backpressure 1 & 2) are included that, without backpressure, would either run out of memory or complete with a much longer total runtime.

The perf-backpressure3 test shows an example where a single consumer constantly requires a random muted producer to participate in a mutimessage in order to make progress. This third example does not deadlock, however the current implementation is able to throttle the producers enough after the random unmuting to prevent runaway growth of the consumer's queue. More discussion is needed on how to address this problem.

Closes #33


src/rt/sched/backpressure.h
```diff
@@ -0,0 +1,259 @@
...
+ * The backpressure system uses "load" as an approximation for message queue
+ * depth. Load is measured using a token message that occasionally falls out of
+ * its message queue. The "current load" on a cown is equal to the number of
+ * messages processed by the cown since the last token message fell out. Once
+ * the next token falls out of its queue, the current load is stored into a ring
+ * buffer and the current load is reset. The total load for the cown is
+ * calculated as the sum of the current load and previous loads stored in the
+ * ring buffer. Simply measuring the batch size of messages processed by the
+ * cown does not account for additional participants of a message cutting their
+ * batch short once they receive the mutli-message.
```
SeanTAllen, line 31:
```suggestion
 * batch short once they receive the multi-message.
```

src/rt/sched/backpressure.h
```diff
@@ -0,0 +1,259 @@
...
+
+/**
+ * ## Introduction
+ *
+ * The primary goal of backpressure is to prevent runaway growth of a cown's
+ * message queue that might eventually result in a Verona application running
+ * out of memory. This is achieved by tracking the load on each cown to detect
+ * when it is unable to process messages as quickly as messages are added to the
+ * queue. Once a cown is considered "overloaded", message sends to that cown
+ * will result in the senders being muted so that the overloaded cown may
```
SeanTAllen, line 13: This is the first time using the term "muted". I think you should be defining what it means here or nearby.

src/rt/sched/backpressure.h
```diff
@@ -0,0 +1,259 @@
...
+ * temporarily process more messages than it receives. Similarly, cowns sending
+ * to muted receivers will also be muted so that cowns don't experience runaway
+ * queue growth while they are muted. Muted cowns are generally not rescheduled
```
SeanTAllen, line 16: "generally" here raises a lot of questions that should answered. It isn't right here so a note about where that information can be found would be valuable.

src/rt/sched/backpressure.h
```diff
@@ -0,0 +1,259 @@
+// Copyright Microsoft and Project Verona Contributors.
+// SPDX-License-Identifier: MIT
+#pragma once
+
+/**
+ * ## Introduction
+ *
+ * The primary goal of backpressure is to prevent runaway growth of a cown's
```
SeanTAllen, line 8: It might be unclear to people how runaway growth can happen. I think you need to introduce that here at the beginning.

src/rt/sched/backpressure.h
```diff
@@ -0,0 +1,259 @@
...
+ * when it is unable to process messages as quickly as messages are added to the
+ * queue. Once a cown is considered "overloaded", message sends to that cown
+ * will result in the senders being muted so that the overloaded cown may
+ * temporarily process more messages than it receives. Similarly, cowns sending
+ * to muted receivers will also be muted so that cowns don't experience runaway
+ * queue growth while they are muted. Muted cowns are generally not rescheduled
+ * until the receiver that resulted in their muting (the "mutor") is no longer
+ * muted or overloaded.
+ *
+ * ## Cown Load
```
SeanTAllen, line 20: I find this section confusing. I know in general what is happening but found myself reading it several times. current load feels like its a bit overloaded where for eample you say "current load is stored into a ring buffer and the current load is reset" followed by "total load... is calculated as the sum of the current load and previous loads stored in the ring buffer."

src/rt/sched/backpressure.h
```diff
@@ -0,0 +1,259 @@
...
+ *
+ * The backpressure system uses "load" as an approximation for message queue
+ * depth. Load is measured using a token message that occasionally falls out of
+ * its message queue. The "current load" on a cown is equal to the number of
+ * messages processed by the cown since the last token message fell out. Once
+ * the next token falls out of its queue, the current load is stored into a ring
+ * buffer and the current load is reset. The total load for the cown is
+ * calculated as the sum of the current load and previous loads stored in the
+ * ring buffer. Simply measuring the batch size of messages processed by the
```
SeanTAllen, line 29: Why doesn't "simply measuring the batch size of message processed" account for additional participants? I think there's a lot of important details hiding behind that statement.

src/rt/sched/backpressure.h
```diff
@@ -0,0 +1,259 @@
...
+ * cown does not account for additional participants of a message cutting their
+ * batch short once they receive the mutli-message.
+ *
+ * ## Mute Map Scan
+ *
+ * Once a scheduler thread completes a message action that would result in the
```
SeanTAllen, line 35: what is "the muted multimessage cowns"?

rengolin:
How do we measure runtime on benchmarks? If the test would pass either way, but would take "a lot longer" without it, any regressions can cause partial degradation and we still wouldn't notice. Our CI times can be anything between 8min and 1h, depending on the time it takes to acquire the builders from the shared pool, so timeouts won't help here.

Not necessarily for this particular PR, but it would be nice to have a table with the execution time for the tests, perhaps some CMake magic, that would then be released as an artifact, so we can plot results on DevOps. Extra points if we define a safety margin and "fail" the test it if takes more than some threshold.

src/rt/sched/backpressure.h
```diff
@@ -0,0 +1,259 @@
...
+ * muted multimessage cowns, the cowns are acquired by the scheduler thread and
+ * placed in a "mute map" on the scheduler thread. The mute map is a mapping
+ * from mutor => mute set, where "mute set" referrs to the set of cowns muted by
+ * a mutor. Each scheduler thread scans the mutors of its mute map on each
+ * iteration of the run loop. If any mutor is not muted and is not overloaded,
```
SeanTAllen, line 40: "If any mutor is not muted..."

I'm not sure how to change this but it misses the key point that they were once muted/overloaded but something has changed and they no longer are.

src/rt/sched/backpressure.h
```diff
@@ -0,0 +1,259 @@
...
+ * the corresponding mute set is unmuted and the entry is removed from the mute
+ * map. A mutor may exist as the key in mutiple mute maps, but a muted cown may
```
SeanTAllen, line 42: "A mutator may exist as..."

I think this deserve an explanation. How can a mutor be in multiple mute maps. 

"a muted cown may only be tracked"

I'm having a hard time putting that together. Why the "may" and how does it related to mutor maybe existing as the key in multiple maps.

Is may used here as "it is possible" or more like "is allowed".

src/rt/sched/backpressure.h
```diff
@@ -0,0 +1,259 @@
...
+ * only be tracked in a single mute set.
+ *
+ * ## Message Scan
+ *
+ * All messages from one set of sending cowns to a set of receiving cowns is
```
SeanTAllen, line 47:
```suggestion
 * All messages from one set of sending cowns to a set of receiving cowns are
```

src/rt/sched/backpressure.h
```diff
@@ -0,0 +1,259 @@
...
+ * scanned to determine if the senders should be muted. Such a message will
```
SeanTAllen, line 48:
```suggestion
 * scanned to determine if the senders should be muted. Sender will be
```

src/rt/sched/backpressure.h
```diff
@@ -0,0 +1,259 @@
...
+ * result in muted senders if all of the following are true:
```
SeanTAllen, line 49:
```suggestion
 * be muted if all of the following are true:
```

src/rt/sched/backpressure.h
```diff
@@ -0,0 +1,259 @@
...
+ *   1. The receiver set does not contain any of the senders
+ *   2. No sender is overloaded
+ *   3. One of the receivers is either overloaded or muted.
+ *
+ * A message is given "priority" if either condition 1 or 2 are false. A
```
SeanTAllen, line 54: wait, what? 

this sounds wrong. can you explain this?

src/rt/sched/backpressure.h
```diff
@@ -0,0 +1,259 @@
...
+ * priority message will result in any muted receivers being unmuted so that the
+ * senders may make progress sooner.
+ *
+ * The backpressure bits of any cown may only be modified by the scheduler
```
SeanTAllen, line 58: "backpressure bits"... are you referring to something about the implementation, is information stored in bits? I'm not sure what backpressure bits are.

src/rt/sched/backpressure.h
```diff
@@ -0,0 +1,259 @@
...
+ * thread running the cown or holding the muted cown in its mute map. If a
```
SeanTAllen, line 59: these are mutually exclusive yes? i don't think that mutual exclusion will be something someone new to this code would be aware of, and therefore should be explicitly called out.

src/rt/sched/backpressure.h
```diff
@@ -0,0 +1,259 @@
...
+ * priority message requires a recipient cown to be unmuted, then the scheduler
+ * thread running the message will search for the cown in its mute map. If the
+ * cown does not exist, then it will be sent to other scheduler threads until it
```
SeanTAllen, line 62: what is being sent to other scheduler threads?

src/rt/sched/backpressure.h
```diff
@@ -0,0 +1,259 @@
...
+ * is found and subsequently unmuted.
+ *
+ * ## Cown Backpressure States
+ *
+ * These states are not explicitly encoded, but they are useful to consider.
```
SeanTAllen, line 67: I don't understand this section.

src/rt/sched/backpressure.h
```diff
@@ -0,0 +1,259 @@
+// Copyright Microsoft and Project Verona Contributors.
```
SeanTAllen, line 1: How is the decision made that something is overloaded made. There's some somewhat fine grained details in the comments (like that it involves a ring-buffer) but nothing beyond a bit about "summing" to explain the process.

I feel like there's a high-level "here is backpressure and how it fits into the runtime" that is missing from description at the beginning of this.

src/rt/sched/backpressure.h
```diff
@@ -0,0 +1,259 @@
...
+
+    /**
+     * Backpressure bits contain the following fields:
+     *   [31:31] muted
+     *   [30:30] needs_token
+     *   [29:24] (unused)
+     *   [23:08] load_history
+     *   [07:00] current_load
+     *
+     * The `muted` bit indicates if a cown is currently muted. A muted cown must
```
SeanTAllen, line 106: why mustn't it run or be collected?

src/rt/sched/backpressure.h
```diff
@@ -0,0 +1,259 @@
...
+     * not run nor be collected until it has been unmuted.
+     *
+     * The `needs_token` bit indicates if the token message is not in a cown's
```
SeanTAllen, line 109: how does this work? this is stuff that i don't know about and this makes me curious.

I think you need a firm understanding of the scheduler to understand this note. That might a reasonable thing to assume the reader has.

src/rt/sched/backpressure.h
```diff
@@ -0,0 +1,259 @@
...
+     * Return true if the token message is not in this cown's queue and a new
+     * token message should be added if it runs another message.
+     */
+    inline bool needs_token() const
+    {
+      return (bits & needs_token_mask) != 0;
+    }
+
+    /**
+     * Return the amount of messages processed since the last token message fell
```
SeanTAllen, line 143: is amount here the number? 

"Return the number of messages"?

Is this message very current implementation specific? This could change later right? It's getting a representation of the load that happens to be X right now?

If that is the case, I think being less specific in this comment and more along the line s of "get current load" would be better.

src/rt/sched/backpressure.h
```diff
@@ -0,0 +1,259 @@
...
+    }
+
+    /**
+     * Calculate the total load accumulated for this cown. This value will be
+     * in the range [0, 1215].
+     */
+    inline uint32_t total_load() const
+    {
+      // Add the current load to the 4 upper nibbles stored as load history. The
+      // load history may be more efficiently compressed, but the clang SLP
```
SeanTAllen, line 158: is "may be" here meaning "could be"?

src/rt/sched/backpressure.h
```diff
@@ -0,0 +1,259 @@
...
+    /**
+     * Return true if this cown is overloaded.
+     */
+    inline bool overloaded() const
+    {
+      return total_load() > overload_threshold;
+    }
+
+    /**
+     * Return true if this cown is muted or overloaded.
```
SeanTAllen, line 176: this feels very implementation specific. this is really...

"check if sending a message to this cown should cause the sender to be muted" 

yes?

src/rt/sched/backpressure.h
```diff
@@ -0,0 +1,259 @@
...
+    inline bool triggers_muting() const
+    {
+      return muted() || overloaded();
+    }
+
+    /**
+     * Return true if this cown is not muted and if the load is substantially
+     * lower than the threshold for muting.
+     */
+    inline bool triggers_unmuting() const
```
SeanTAllen, line 187: reading through this file and getting this far, i'm confused by the role that triggers_unmuting plays or the logic therein (and the comment as well).

src/rt/sched/backpressure.h
```diff
@@ -0,0 +1,259 @@
...
+     * Mark this cown as muted.
+     */
+    inline void mute()
+    {
+      assert(!muted());
+      bits |= muted_mask;
+    }
+
+    /**
+     * Unmark this cown as muted.
```
SeanTAllen, line 202: For me this reads very oddly, I was scanning and mostly focused on the "muted".

I can see the symmetry with "Mark" above but I spent more time on this then I think I should have.

src/rt/sched/backpressure.h
```diff
@@ -0,0 +1,259 @@
...
+    {
+      assert(muted());
+      bits &= ~muted_mask;
+    }
+
+    /**
+     * Mark that this cown has a token message in its queue and no longer
+     * requires a new token message to be added.
+     */
+    inline void add_token()
```
SeanTAllen, line 214: the comment and the method name seems very confusing.

i would expect "add_token" to add a token.

src/rt/sched/backpressure.h
```diff
@@ -0,0 +1,259 @@
...
+    {
+      bits &= ~needs_token_mask;
+    }
+
+    /**
+     * Mark that this cown has removed the token message from its queue.
+     */
+    inline void remove_token()
```
SeanTAllen, line 222: see above.

src/rt/sched/cown.h
```diff
@@ -771,11 +771,99 @@ namespace verona::rt
...
       // Try to acquire as many cowns as possible without rescheduling,
       // starting from the beginning.
       fast_send(body, epoch);
     }
 
+    /**
+     * Return true if any senders are overloaded or if any sender is in the
+     * receiver set.
+     */
+    static inline bool backpressure_priority(
```
SeanTAllen, line 786: I find this method name opaque and confusing.

src/rt/sched/cown.h
```diff
@@ -842,7 +936,7 @@ namespace verona::rt
           //   return true;
           //
           // TODO: Investigate systematic testing coverage here.
```
SeanTAllen, line 938: Is there an issue for this TODO?

src/rt/sched/cown.h
```diff
@@ -858,32 +952,64 @@ namespace verona::rt
           }
 
           Systematic::cout()
-            << "Cown has no work this time:" << this << std::endl;
+            << "Cown has no work this time: " << this << std::endl;
```
SeanTAllen, line 955: 👍

src/rt/sched/cown.h
```diff
@@ -858,32 +952,64 @@ namespace verona::rt
...
           // Deschedule the cown.
           Cown::release(alloc, this);
           return false;
         }
 
         assert(!queue.is_sleeping());
 
+        // This load may be relaxed because modifications to the backpressure
```
SeanTAllen, line 963: what does it mean for load to be relaxed?

src/rt/sched/cown.h
```diff
@@ -908,14 +1034,16 @@ namespace verona::rt
       if (in_epoch(epoch))
         return false;
 
-      // Check if the Cown is already collected
-      if (!is_collected())
+      // Check if the Cown is already collected or is muted.
```
SeanTAllen, line 1037: why is the muted check needed here?

src/rt/sched/cown.h
```diff
@@ -977,10 +1105,13 @@ namespace verona::rt
         }
       }
 
+      assert(!backpressure.load(std::memory_order_acquire).muted());
+
       // Now we may run our destructor.
       destructor();
 
       MultiMessage* last = queue.destroy();
+      assert(last->next.load(std::memory_order_relaxed) == nullptr);
```
SeanTAllen, line 1114: I feel like I sort of know why this is happening based on reading earlier code, but an explanatory comment here would be helpful for me. It feels like a bit of an odd assert when you run across it.

src/rt/sched/schedulerthread.h
```diff
@@ -168,6 +187,129 @@ namespace verona::rt
...
+        cown->backpressure.store(bp, std::memory_order_release);
+        T::acquire(cown);
+        mute_set.insert(alloc, cown);
+      }
+
+      alloc->dealloc(cowns, count * sizeof(T*));
+    }
+
+    /**
+     * Scan the mute map for cowns that may be unmuted. If `force` is true, then
```
SeanTAllen, line 232: Is this cowns that are unmuted but are still in the map, or cowns that are eligible to be unmuted?

src/rt/sched/schedulerthread.h
```diff
@@ -168,6 +187,129 @@ namespace verona::rt
...
+        auto it = mute_set.find(cown);
+        if (it != mute_set.end())
+        {
+          unmute_local_cown(cown);
+          mute_set.erase(it);
+          return;
+        }
+      }
+      // We are not responsible for this cown, so it must be sent to the next
+      // schdeuler thread to be unmuted.
```
SeanTAllen, line 293:
```suggestion
      // scheduler thread to be unmuted.
```

src/rt/sched/schedulerthread.h
```diff
@@ -168,6 +187,129 @@ namespace verona::rt
...
+        auto& mute_set = *entry.value();
+        auto it = mute_set.find(cown);
+        if (it != mute_set.end())
+        {
+          unmute_local_cown(cown);
+          mute_set.erase(it);
+          return;
+        }
+      }
+      // We are not responsible for this cown, so it must be sent to the next
```
SeanTAllen, line 292: will the next thread unmute it or will it try to unmute? my understanding is that it might be able to unmute and we only know that this cown can't so we pass it along to another.

however, the comment has me questioning my understanding.

src/rt/sched/schedulerthread.h
```diff
@@ -168,6 +187,129 @@ namespace verona::rt
...
+          alloc->dealloc<sizeof(ObjectMap<T*>)>(&mute_set);
+        }
+      }
+
+      if (mute_map.size() == 0)
+        mute_map.clear(alloc);
+    }
+
+    /**
+     * Unmute an individual cown.
```
SeanTAllen, line 277: This is really "attempt to unmute an individual cown" rather than unmute yes?

src/rt/sched/schedulerthread.h
```diff
@@ -168,6 +187,129 @@ namespace verona::rt
...
+      // schdeuler thread to be unmuted.
+      cown->weak_acquire();
+      next->unmute_q.enqueue(new (alloc->alloc<sizeof(UnmuteMessage<T>)>())
+                               UnmuteMessage<T>(cown));
+    }
+
+    /**
+     * Unmute a cown that exists in this mute map.
+     */
+    inline void unmute_local_cown(T* cown)
```
SeanTAllen, line 302: unlike `unmute_cown`, this is guaranteed to work, yes?

src/rt/sched/schedulerthread.h
```diff
@@ -449,8 +594,13 @@ namespace verona::rt
           UNUSED(tsc);
         }
 #endif
-          // Enter sleep only when the queue doesn't contain any real cowns.
-          if (state == ThreadState::NotInLD && q.is_empty())
+          if ((mute_map.size() != 0) || (unmute_q.peek() != nullptr))
```
SeanTAllen, line 597: Formatting looks off here, but I couldn't figure it out well enough to do a suggested change.

src/rt/sched/backpressure.h
Theodus: Yeah, this is a holdover from a previous design where these states were more explicit. I think I'll remove them since they may just lead to confusion.

src/rt/sched/cown.h
Theodus: This is to check that the message queue is empty when it is destroyed. I found this assert useful to debug premature collection of cowns without needing to hit the queue invariant assert whenever a collected cown is run.

src/rt/sched/schedulerthread.h
Theodus: Only muted cowns are in the mute sets.

src/rt/sched/schedulerthread.h
Theodus: I'm not sure what you mean. The cown will be unmuted, either during this function or by the time a request for it to be unmuted reaches the scheduler tracking the cown.

src/rt/sched/backpressure.h
SeanTAllen: so a sender isn't overloaded. the receiver is muted.

so the receiver will be unmuted?

src/rt/sched/cown.h
SeanTAllen: Yes, but still sounds confusing.

src/rt/sched/schedulerthread.h
SeanTAllen: the comment reads like there might be unmuted cowns in the mute map.

src/rt/sched/schedulerthread.h
SeanTAllen: Right. So, this function might not unmute the cown. I don't think the comment reflects what is happening.

src/rt/sched/backpressure.h
Theodus: No, in this case conditions 2 and 3 are true. So no muting occurs.

src/rt/sched/cown.h
rengolin: I think the whole logic here is confusing. There is no relation between which iteration returned `true` here and which receiver will be unmuted later.

From earlier conversations, it seems there can only be two cases: one sender, multiple receivers or multiple senders and one receiver, but the code seems to multiple of both.

Perhaps factoring some of the code out to smaller functions with more accurate names and combining them in the logical way could clear things up.

src/rt/sched/cown.h
rengolin: I wouldn't require every todo to have its own issue yet. Is there a reason why you want this TODO in particular to have an issue?

src/rt/sched/schedulerthread.h
```diff
@@ -168,6 +187,131 @@ namespace verona::rt
...
+
+      alloc->dealloc(cowns, count * sizeof(T*));
+    }
+
+    /**
+     * Unmute all mute sets where the mutor is in a state that triggers
+     * unmuting. If `force` is true, then all mute sets in the map will be
+     * unmuted.
+     */
+    template<bool force = false>
```
rengolin, line 236: Why does `force` need to be a template parameter and not a function argument?

src/rt/sched/schedulerthread.h
```diff
@@ -168,6 +187,131 @@ namespace verona::rt
...
+    void mute_map_scan()
+    {
+      // Handle unmute queue first.
+      while (true)
```
rengolin, line 240: Instead of `while(true)`, I'd rather have something like:
```
while ((auto *msg = unmute_q...) != nullptr)
{
  ...
}
```

It's clear what the exit condition is from the first line. I wouldn't have to dig the code to find out...

src/rt/sched/cown.h
Theodus: There is no relationship between where this function returns true and what receiver would be unmuted. Requesting that a receiver be unmuted before determining if it should be seems incorrect to me, and I would prefer an additional loop over the short array of receivers than an allocation to track which ones might possibly require unmuting.

The name of this function should line up with the description of "priority" from the summary comment in backpressure.h. If the description or the name "priority" are unclear, then that needs to be addressed.

There can be any amount of senders and receivers for a multi-message.

src/rt/test/perf/backpressure1/backpressure1.cc
```diff
@@ -0,0 +1,193 @@
...
+#else
+  UNUSED(seed);
+#endif
+  sched.set_fair(true);
+  sched.init(cores);
+
+  auto* alloc = ThreadAlloc::get();
+
+  for (size_t r = 0; r < receivers; r++)
+    receiver_set.push_back(new (alloc) Receiver);
```
rengolin, line 155: Isn't this what breaks GCC 10?

src/rt/test/perf/backpressure1/backpressure1.cc
```diff
@@ -0,0 +1,193 @@
...
+{
+  size_t index;
+
+  Proxy(size_t index_) : index(index_) {}
+
+  void trace(ObjectStack& st) const
+  {
+    if (this != proxy_chain.back())
+    {
+      st.push(proxy_chain[index + 1]);
```
rengolin, line 60: This looks dodgy. What if index is the last one?

src/rt/test/perf/backpressure1/backpressure1.cc
```diff
@@ -0,0 +1,193 @@
...
+      {
+        Cown::acquire(proxy_chain[0]);
+      }
+      else
+      {
+        for (auto* r : receiver_set)
+          Cown::acquire(r);
+      }
+
+      auto* s = new (alloc) Sender(std::chrono::milliseconds(duration));
```
rengolin, line 174: This could be +- a random percentage to make the test more robust.

src/rt/test/perf/backpressure1/backpressure1.cc
```diff
@@ -0,0 +1,193 @@
+// Copyright Microsoft and Project Verona Contributors.
+// SPDX-License-Identifier: MIT
+
+/**
+ * This tests a simple scenario for backpressure where many individual `Sender`
+ * cowns send messages to a single set of `Receiver` cowns. The `Recevier` cowns
+ * may be placed behind a chain of `Proxy` cowns to test backpressure
+ * propagation.
+ **/
```
rengolin, line 9: It would be nice to have an additional paragraph explaining why this would fail and why does backpressure fixes it.

src/rt/test/perf/backpressure2/backpressure2.cc
```diff
@@ -0,0 +1,158 @@
...
+};
+
+int main(int argc, char** argv)
+{
+  opt::Opt opt(argc, argv);
+  const auto seed = opt.is<size_t>("--seed", 5489);
+  const auto cores = opt.is<size_t>("--cores", 4);
+  const auto senders = opt.is<size_t>("--senders", 100);
+  const auto receivers = opt.is<size_t>("--receivers", 10);
+  const auto duration = opt.is<size_t>("--duration", 10'000);
```
rengolin, line 123: Not necessarily for this commit, but these files have a lot of common setup process that could be joined into one file. One way would be to have different namespaces for the tests, so the different class declarations with same name wouldn't clash.

A more future-proof solution would be to use something like google-test.

src/rt/sched/cown.h
rengolin: > Requesting that a receiver be unmuted before determining if it should be seems incorrect to me

That's not what I meant.

If there could be multiple senders and multiple receivers, and if any sender could return true but only receivers from the same sender, then unmuting *any* sender that is overloaded seems to not match my interpretation of safety.

Then again, it might be that I'm misinterpreting what the semantics are, or what the code does, which means the code is slightly confusing, at least to me, personally.

My proposal was only to clarify what the code is doing. I find it helpful if I break the functionality in small pieces when the code is not one clear step after the other, even if it's only called from one site. Either that, or some expressive comment before each block, but that's less preferable than a nice function name. :)

src/rt/test/perf/backpressure1/backpressure1.cc
Theodus: Not really, the issue is with using placement new with regions.

src/rt/sched/schedulerthread.h
Theodus: I find using an assignment as an expression very confusing when it returns the newly assigned value. But either way, I can't get the compiler to accept either of the following:
```cpp
error: expected ')'
      while ((auto* msg = unmute_q.dequeue(alloc, notify)) != nullptr)
                    ^
```
```cpp
error: expected ')'
      while (auto* msg = unmute_q.dequeue(alloc, notify); msg != nullptr)
                                                        ^
```

src/rt/sched/schedulerthread.h
```diff
@@ -168,6 +187,131 @@ namespace verona::rt
       }
     }
 
+    /**
+     * Mute a set of cowns. This will add the cowns to the mute set of the
+     * mutor.
+     */
+    void mute(T** cowns, size_t count)
```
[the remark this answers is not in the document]
Theodus: I agree, but I think the implementation of such a thing (like vector, but with snmalloc) and using it throughout the runtime should be a separate issue.

src/rt/sched/schedulerthread.h
Theodus: The thought was to avoid a branch in the case where force is true without requiring the function to be inlined. But perhaps that is not worth the syntax of a template parameter?

src/rt/sched/schedulerthread.h
```diff
@@ -168,6 +187,131 @@ namespace verona::rt
...
+    {
+      auto it = mute_map.find(mutor);
```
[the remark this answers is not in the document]
Theodus: Assert added.

src/rt/test/perf/backpressure1/backpressure1.cc
Theodus: If that were the case, then `this == proxy_chain.back()` would be true

src/rt/test/perf/backpressure1/backpressure1.cc
Theodus: Systematic testing would be a better option if we wanted to cover possible interleavings at the end. Existing `func-` prefixed tests cover that with the `coin()`s inserted to randomly trigger backpressure conditions.

src/rt/test/perf/backpressure2/backpressure2.cc
Theodus: Yes, something similar to the test harness for `func-` tests may be nice for the `perf-` tests. But I think some consideration for automatic checking for performance regressions would be necessary to make sure these tests are more useful in the future.

src/rt/test/perf/backpressure1/backpressure1.cc
Theodus: Iv'e added additional explanation to all three tests

src/rt/sched/cown.h
Theodus: Sorry, I think I'm confused about your confusion :)

> if any sender could return true but only receivers from the same sender

What do you mean by "but only receivers from the same sender"?

> then unmuting any sender that is overloaded seems to not match my interpretation of safety

Senders would never be unmuted, because they cannot be muted. A muted cown cannot be running an action and therefore cannot be sending messages.

src/rt/sched/backpressure.h
SeanTAllen: Ok, I think the way it is phrased is very confusing i finally figured out how it is supposed to read. That was really really confusing to me. Too much negation I think.

SeanTAllen: I think saying "a sender is overloaded" would be a lot more clear (and then not having the negation).

src/rt/sched/cown.h
SeanTAllen: In my experience, TODOs without issues never get addressed.

SeanTAllen:
@Theodus I think "priority" might be a confusing term here. I do not have a better name at the moment, but I think it is confusing.

src/rt/sched/schedulerthread.h
rengolin: Sorry, that'd be without parenthesis, but that's not good either and is super confusing. Ignore this for now.

src/rt/sched/cown.h
rengolin: > Sorry, I think I'm confused about your confusion :)

:D

> What do you mean by "but only receivers from the same sender"?

I meant:
```
    if (receiver == sender)
        return true;
```

> > then unmuting any sender that is overloaded seems to not match my interpretation of safety
> 
> Senders would never be unmuted, because they cannot be muted. A muted cown cannot be running an action and therefore cannot be sending messages.

Sorry again, "unmuting any receivers", which is what the code is doing when this function returns true.

src/rt/sched/cown.h
rengolin: And too many issues often get ignored. :D

At Arm, we had a script that would automatically create an issue for every build breakage in the CI. Builds were fixed, but after a few months, we had so many issues that we had no idea had been fixed or not that we just left them there.

src/rt/sched/schedulerthread.h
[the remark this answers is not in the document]
rengolin: Fair enough.

src/rt/sched/schedulerthread.h
rengolin: I'd leave it without template until we figure that loop is super hot and we need to optimise for a case that we know is the vast majority. Right now, it's not clear. I'm not expecting the `mute_map` to be so big as to take considerable computation resources comparable to the threads actual load.

src/rt/test/perf/backpressure1/backpressure1.cc
rengolin: d'oh

src/rt/test/perf/backpressure1/backpressure1.cc
rengolin: makes sense

src/rt/test/perf/backpressure2/backpressure2.cc
rengolin: agree. nice for future work.

src/rt/sched/backpressure.h
```diff
@@ -0,0 +1,249 @@
...
+#pragma once
+
+/**
+ * ## Introduction
+ *
+ * The primary goal of backpressure is to prevent runaway growth of a cown's
+ * message queue that might eventually result in a Verona application running
+ * out of memory. Cowns that receive messages at a higher rate than they are
+ * able to process are considered "overloaded" and may lead to this unbounded
+ * queue growth. Once an overloaded cown is identified, cowns sending messages
```
SeanTAllen, line 12: This crosses lines so I'm not going to do a suggested edit.

How about...

"If a cown receives messages more quickly than it is able to process them, the queue will start to grow in an unbounded fashion. Cowns that aren't able to keep up with their arriving messages are marked as "overloaded". Once a cown is marked as overloaded, cowns sending messages to it will be temporarily descheduled ("muted"). Descheduling cowns that are sending to an overloaded cown helps alleviate pressure on the it by:

- Turning off a source of incoming messges
- Allows for more scheduler time to go to the overload cown and not any of the descheduled ones.

Similarly, cowns sending to cowns that have been muted will also be muted to prevent runaway queue growth while the muted cown in descheduled.

Muted cowns will be rescheduled after the receiver that triggered their muting (the "mutor") is no longer muted or overloaded.

src/rt/sched/backpressure.h
```diff
@@ -0,0 +1,249 @@
...
+ * to it will be temporarily descheduled ("muted") so that the overloaded cown
+ * can work through its queue. Similarly, cowns sending to muted receivers will
+ * also be muted so that cowns do not themselves become overloaded as a
+ * consequence of being muted. Muted cowns will be rescheduled after the
+ * receiver that triggered their muting (the "mutor") is no longer muted or
+ * overloaded.
+ *
+ * ## Detecting Overloaded Cowns
```
SeanTAllen, line 20: This section assumes that someone knows or would otherwise reach for measuring queue depth as a means of knowing that something is overloaded. That seems reasonable to me, however, I'm not sure it really would hold.

The problem as I see it is that we just threw a ton of information at someone in the previous sentence and they might not have absorbed it all. It's very dense and there is a lot to cover.

I think it would be better to be safe, and start with something about how we need to determine overload and that while the basic definition (unbounded queue growth) is straightforward, the implementation isn't.

Avoiding "unbounded queue growth" is important but because there are so many cowns, really we need a heuristic that will allow us to provide an arbitrary bounding by saying that a cown is overloaded.

One method of doing this would be count the number of items in the queue on a regular basis and compare those lengths over time to determine if it is growing in a fashion that looks unbounded.  However, counting every item in every queue for every cown adds a tremendous amount of overhead to the runtime and is thus avoided.

The current means of determining if a cown is overloaded is to use the following heuristic:

src/rt/sched/backpressure.h
```diff
@@ -0,0 +1,249 @@
...
+ *
+ * An overloaded cown is one that experiences message queue growth as it is
+ * processing messages. Since it would be prohibitively expensive to calculate
+ * the length of a message queue directly, a count of messages removed from the
+ * queue over some period is used as an approximation for message queue size.
+ * Like scheduler fairness, the period is determined using a token message that
```
SeanTAllen, line 26: I don't think you should assume that someone knows how scheduler fairness works here. I think you should detail how it works here. If someone hadn't previously explained scheduler fairness to me, I doubt I would understand this.

src/rt/sched/backpressure.h
```diff
@@ -0,0 +1,249 @@
...
+ * also be muted so that cowns do not themselves become overloaded as a
+ * consequence of being muted. Muted cowns will be rescheduled after the
+ * receiver that triggered their muting (the "mutor") is no longer muted or
+ * overloaded.
+ *
+ * ## Detecting Overloaded Cowns
+ *
+ * An overloaded cown is one that experiences message queue growth as it is
+ * processing messages. Since it would be prohibitively expensive to calculate
+ * the length of a message queue directly, a count of messages removed from the
```
SeanTAllen, line 24: I think you need to explain how "count of message removed from the queue" over time is different than counting the number of messages in the queue (in terms of performance).

This currently assumes that is obvious to the reader.

src/rt/sched/backpressure.h
```diff
@@ -0,0 +1,249 @@
...
+ * will occasionally fall out of the queue and then be inserted to begin the
+ * next period. The "load" of a cown at any time is the accumulation of messages
+ * processed over multiple periods.
+ *
+ * ## Muting Cowns
+ *
+ * All messages from one set of sending cowns to a set of receiving cowns are
+ * scanned to determine if the senders should be muted. The senders will be
+ * muted once their action completes if all of the following are true:
+ *   1. The receiver set does not contain any of the senders.
```
SeanTAllen, line 36: I'm not sure if it should be detailed here, but I'm having a really hard time imagining what the scenario is where a cown is in both the sending and receiving sets.

src/rt/sched/backpressure.h
```diff
@@ -0,0 +1,249 @@
+// Copyright Microsoft and Project Verona Contributors.
+// SPDX-License-Identifier: MIT
+#pragma once
+
+/**
+ * ## Introduction
```
SeanTAllen, line 6: @rengolin Renato, do you think this would have been easier for you if all the terminology like "sender", "receiver", "muted", "overloaded", "mutor" etc had been defined up front before the introduction?

src/rt/sched/backpressure.h
SeanTAllen: Going over yet again, I think that you should write out the rules for muting in one section and the rules for priority on another section. I think the mixing them together results in some interesting negations and double negations that are really hard to parse.

src/rt/sched/backpressure.h
Theodus: This would happen on a recursive behaviour call or some form of continuation for a set of cowns.

src/rt/sched/backpressure.h
```diff
@@ -0,0 +1,249 @@
...
+ * priority message will result in any muted receivers being unmuted so that the
+ * senders may make progress sooner.
+ *
+ * If all of the above conditions are met, the first receiver that is either
+ * overloaded or muted is identified as the mutor for the senders. First in this
+ * case means the least cown, determined by the mutlimessage cown order.
+ *
+ * ## Tracking Muted Cowns
+ *
+ * Once a scheduler thread completes a message action that would result in
```
SeanTAllen, line 50: Is it reasonable to assume that someone reading this will be familiar with scheduler threads and how they work? On my first read this didn't stand out to me, now, I think it might be problematic. 

Theo, Do you think it would be distracting as part of the intro to cover a bit more about the scheduler threads related to backpressure? I think it might help make the tracking (and maybe other aspects) more understable. 

I'm not sure about that. Maybe what is needed is to introduce some background of "must know about scheduler threads to understanding mute tracking" right here at the start of this section.

src/rt/sched/backpressure.h
```diff
@@ -0,0 +1,249 @@
...
+ * only exist in a mute set on the scheduler thread that muted it.
+ *
+ * A muted cown may not be scheduled or collected until they are removed from
+ * their scheduler thread's mute map and rescheduled by that thread ("unmuted").
+ *
+ * Each scheduler thread scans the mutors of its mute map on each iteration of
+ * the run loop. If any mutor is not muted and is not overloaded, all cowns in
+ * the corresponding mute set are unmuted and the entry is removed from the mute
+ * map.
+ */
```
SeanTAllen, line 68: I think you should add either here or near the beginning, some examples of cases where backpressure will work to control unbounded queue growth and example cases where it will not.

src/rt/sched/cown.h
SeanTAllen: I don't see how that is relevant. We have issues to track work that needs to be done. If it needs to be done, there should be an issue.

src/rt/sched/schedulerthread.h
Theodus: Ok, fixed.

src/rt/sched/backpressure.h
SeanTAllen: That definitely sounds like something that shouldn't be detailed right here as it would derail from the key point. However, I do think that going "a level deeper" would be helpful for people in the future, however, I don't see that as required for now. 

If you agree that additional deeper details would be valuable to help others work with the code in the future (I think it would) then I think an issue should be created (assigned to you) to address that.

Given you are far and away the best person to write that, I think we would want to make it enough of a priority to complete before your contract ends.

I imagine the "one level deeper" additions to be the sort of information like the above that you would give someone if you were handing off maintenance of this to them.

Information something like "great know you understand in general how it works, here's some nitty gritty things you need to know to understand why some things are the way they are".

src/rt/sched/backpressure.h
Theodus: I think that the scheduler threads, mutli-message acquisition, and actions should be described in more detail elsewhere and should be a prerequisite for explaining backpressure. So I've added to the intro a short list of expectations for the reader.

src/rt/sched/backpressure.h
Theodus: Updated.

src/rt/sched/backpressure.h
SeanTAllen: Letting people know what they need for background is an excellent approach.

src/rt/sched/cown.h
rengolin: I see. I think my thought process comes from the way I use the TODO comment.

To me, TODO and FIXME are at least potential issues, not necessarily real issues. So, for example, I could add two TODOs in one file, while I only need to implement one and the other is "done". Or, I could add a FIXME and later realise it wasn't important.

More commenly, though, is adding TODOs for ideas that could be relevant, but the code gets refactored and all of those lose meaning and become irrelevant.

Tracking all of those changes with all issues in Github is not a trivial amount of work and doesn't scale well with the number of issues. We'd need TODOs with #number identifiers to make sure we track them, etc.

My general approach to bureaucracy is to do exactly as much as necessary, not more, not less. With a small team, a fast moving target, and a lot of code being produced and changed every month, I think that creating a tracking system for everything we're doing is overkill and will likely increase our overhead.

src/rt/sched/backpressure.h
rengolin: Those terms specifically were pretty clear to me what they meant from the start, but perhaps I'm already biased by being at the meetings where theses things are being discussed. :)

Thogh, my personal preference would be to have a document explaining all of those things and not in a header comment. We have a docs directory for that, would be nice to populate it with some better documents aimed at developers and users.

src/rt/sched/cown.h
SeanTAllen: I'm strongly against this. I think we should discuss with the larger team.

src/rt/sched/backpressure.h
SeanTAllen: That makes sense @rengolin. I think then we should either include that document as part of this PR or open an issue for the creation of that document and our general expectations of what should be in it.

rengolin: Perhaps most of the discussion here and most of this comment could go inside such document, and here a quick explanation and a link to the document.

src/rt/sched/backpressure.h
Theodus: I've added a section for known limitations

src/rt/sched/backpressure.h
SeanTAllen: That sounds quite reasonable to me.

src/rt/sched/backpressure.h
```diff
@@ -0,0 +1,280 @@
...
+    /**
+     * Calculate the total load accumulated for this cown. This value will be
+     * in the range [0, 1215].
+     */
+    inline uint32_t total_load() const
+    {
+      // Add the current load to the 4 upper nibbles stored as load history. The
+      // load history could be more efficiently compressed, but the clang SLP
+      // vectorizer optimizes this nicely with AVX2 instructions.
+      const uint32_t h3 = (bits & 0x00'f000'00) >> 16;
```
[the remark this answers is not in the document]
Theodus: I did not find a bit extract template, so I added one to morebits.h.

src/rt/sched/backpressure.h
```diff
@@ -0,0 +1,280 @@
...
+    /**
+     * Reset the current load and store its upper nibble in the load history.
+     */
+    inline void reset_load()
+    {
+      const uint32_t hist = (bits & 0x00'0fff'f0) << 4;
+      bits = (bits & 0xff'0000'00) | hist;
+    }
+  };
+
```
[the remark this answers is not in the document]
Theodus: Added

src/rt/sched/multimessage.h
```diff
@@ -27,7 +27,7 @@ namespace verona::rt
     friend verona::rt::MPSCQ<MultiMessage>;
     friend class Cown;
 
-    std::atomic<MultiMessage*> next;
+    std::atomic<MultiMessage*> next = nullptr;
```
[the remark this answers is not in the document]
Theodus: fixed

src/rt/sched/backpressure.h
```diff
@@ -0,0 +1,286 @@
...
+ * Like scheduler fairness, the period is determined using a token message that
+ * will occasionally fall out of the queue and then be inserted to begin the
+ * next period. The "load" of a cown at any time is the accumulation of messages
+ * processed over multiple periods.
+ *
+ * ## Muting Cowns
+ *
+ * All messages from one set of sending cowns to a set of receiving cowns are
+ * scanned to determine if the senders should be muted. The senders will be
+ * muted once their action completes if all of the following are true:
```
mjp41, line 47: This is for the avoidance of deadlock.  But you don't mention deadlock at all in the description.  I think it would be good to put an example of how this breaks deadlocks here.  My understanding is:

> * Deadlock could occur if a cown 'A' is muted due to another cown 'B' being overloaded.
> * To process `B`'s messages requires `A`.
>
> Thus, B can no longer proceed.
>
>  In this algorithm, if you are attempting a acquire a muted cown, if you are overloaded, then you `unmute` the cown you are trying to acquire.

Now, there are bunch of awkward corner cases,
* `unmute` involves a message, can a cown be unmuted and remuted, before the `when` is processed.
* can multiple `unmute` messages to the same cown exist.
* What happens if a cown is unmuted, before the message arrives?
* What stops a cown being muted after a multimessage has been sent, that could cause a deadlock?

mjp41:
P.S.  I really like this design now.  It feels pretty solid to me.

src/rt/sched/cown.h
```diff
@@ -771,11 +771,191 @@ namespace verona::rt
...
+     */
+    static inline void
+    backpressure_scan(const MessageBody& senders, const MessageBody& receivers)
+    {
+      // Make receivers unmutable if any sender is overloaded.
+      for (size_t s = 0; s < senders.count; s++)
+      {
+        auto* sender = senders.cowns[s];
+        if (
+          !sender->backpressure.load(std::memory_order_relaxed).overloaded()
```
mjp41, line 835: Should you have a yield here?

src/rt/sched/cown.h
```diff
@@ -771,11 +771,191 @@ namespace verona::rt
...
+          if (senders.cowns[s] == receivers.cowns[r])
+            return;
+        }
+      }
+
+      // Mute senders if any receivers are overloaded/muted.
+      for (size_t r = 0; r < receivers.count; r++)
+      {
+        auto* receiver = receivers.cowns[r];
+        const auto bp = receiver->backpressure.load(std::memory_order_acquire);
```
mjp41, line 864: Should have a yield here.

src/rt/sched/cown.h
```diff
@@ -771,11 +771,191 @@ namespace verona::rt
...
+      }
+    }
+
+    /**
+     * Update backpressure based on the ocurrence of a token message. Return
+     * true if the current message is a token.
+     */
+    inline bool check_message_token(Alloc* alloc, MessageBody* curr)
+    {
+      auto bp = backpressure.load(std::memory_order_relaxed);
```
mjp41, line 888: Should have a yield here.

src/rt/sched/cown.h
```diff
@@ -797,15 +977,21 @@ namespace verona::rt
       auto until = queue.peek_back();
       yield(); // Reading global state in peek_back().
 
+      const auto bp = backpressure.load(std::memory_order_acquire);
```
mjp41, line 980: Should have a yield here.

src/rt/sched/cown.h
```diff
@@ -977,11 +1174,15 @@ namespace verona::rt
         }
       }
 
+      assert(!backpressure.load(std::memory_order_acquire).muted());
```
mjp41, line 1177: Should have a yield before the assert.

src/rt/sched/schedulerthread.h
```diff
@@ -168,6 +180,93 @@ namespace verona::rt
...
+      else
+      {
+        mutor->weak_release(alloc);
+      }
+      auto& mute_set = *it.value();
+
+      for (size_t i = 0; i < count; i++)
+      {
+        auto* cown = cowns[i];
+        auto bp = cown->backpressure.load(std::memory_order_relaxed);
```
mjp41, line 206: Should have a field here.

src/rt/sched/schedulerthread.h
```diff
@@ -168,6 +180,93 @@ namespace verona::rt
...
+        { // Messages in this cown's queue must be scanned.
+          cown->schedule();
+          continue;
+        }
+
+        yield();
+
+        auto bp_muted = bp;
+        bp_muted.set_state_muted();
+        if (!cown->backpressure.compare_exchange_weak(
```
mjp41, line 221: Should have a yield here.

src/rt/sched/schedulerthread.h
```diff
@@ -168,6 +180,93 @@ namespace verona::rt
...
+     * unmuted.
+     */
+    void mute_map_scan(bool force = false)
+    {
+      for (auto entry = mute_map.begin(); entry != mute_map.end(); ++entry)
+      {
+        auto* m = entry.key();
+        if (
+          force ||
+          m->backpressure.load(std::memory_order_acquire).triggers_unmuting())
```
mjp41, line 250: Should have a yield here.

mjp41:
@pdeligia this might be interesting to you.  A concurrent failure, that we are going to try to find in systematic testing.

pdeligia:
> @pdeligia this might be interesting to you. A concurrent failure, that we are going to try to find in systematic testing.

Thanks for the pointer @mjp41! This will be really useful, as we plan to resume work and experimentation on the native scheduler soon (we got sidetracked with the .NET side of coyote since we last chatted).

rengolin:
https://dev.azure.com/ProjectVeronaCI/Project%20Verona/_build/results?buildId=1311

Theodus:
Requires #212

rengolin:
If you rebase this branch (without the last identical commit), it should pass again.

Theodus:
Requires #216

PR #185, opened 2020-07-10
Add LLVM's handling to building instructions

rengolin:
This is a temporary work around for trying to cache the LLVM build and is how it's working on the CI and some developers' machines already.

If we want more automation or seamless "cmake" magic, we may have to define what's the kind of caching we really want, long term, and work towards that, or we revert to always building LLVM on every Verona build and it will be much simpler (but a lot slower).

Also, on developer machines, building LLVM only happens once, but on the CI, because the machines are not on 24/7, we need to rebuild the whole LLVM, making a simple pre-commit test take hours.

### DIFF
--- a/docs/building.md
+++ b/docs/building.md
@@ -11,7 +11,6 @@ git submodule update --init --recursive
 ```
 from the root of the checkout.
 
-
 ## Updating code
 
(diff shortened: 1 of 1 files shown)


rengolin:
Better fix the script so that the download becomes automatic from the right place.

PR #186, opened 2020-07-10
Fix llvm script to download from the blob, not universal package

rengolin:
Since our CI was downloading before running the script, this went unnoticed when we moved from Azure Universal Package to blob storage. I have tested on Linux, and there's no reason not to work on Mac, but I can't make it work on my Windows, even though it works on the CI's Windows.

### DIFF
--- a/docs/building.md
+++ b/docs/building.md
@@ -11,7 +11,6 @@ git submodule update --init --recursive
 ```
 from the root of the checkout.
 
-
 ## Updating code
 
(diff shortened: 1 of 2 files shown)


rengolin:
Note, "can't make it work" == bash problems, not broken script problems. So I'm blaming myself for that, not the script itself.

rengolin:
Simple fix change and better docs. I'll merge and if there's any worries we can fix later.

PR #188, opened 2020-07-13
Refactor MLIR generator to use ast_utils for all AST handling

rengolin:
This removes every dependency that the MLIR generator had on the
AST structure. All searches in the AST structure and accessors to names,
IDs, types are done via helper functions.

The only knowledge of the AST is via a NodeKind switch in parseNode,
which is a lot better than calling "isFoo" and "isBar" functons, but it
could be made so if people feel strongly inclined. Given that this is a
helper enum, not an AST enum, it should be ok.

The logic for accessing the AST structure is now 100% encoded in the
helper functions and comments were added where things were less than
clear. This makes it easier for us to exchange knowledge and refine the
API of both AST and MLIR to make things simpler.

### DIFF
--- a/src/mlir/ast-utils.cc
+++ b/src/mlir/ast-utils.cc
@@ -53,8 +53,7 @@ namespace mlir::verona::ASTInterface
       ptr = ptr->nodes[0];
 
     // Return the nodes
-    for (auto n : ptr->nodes)
-      nodes.push_back(n);
+    nodes.insert(nodes.end(), ptr->nodes.begin(), ptr->nodes.end());
     return nodes;
   }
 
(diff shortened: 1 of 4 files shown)


rengolin:
@sylvanc, pay close attention to `ast-utils.cc` to make sure all my AST-specific assumptions are correct.

src/mlir/ast-utils.cc
```diff
@@ -8,37 +8,103 @@
...
-    assert(ptr->is_token && "Bad node");
+    std::vector<::ast::WeakAst> nodes;
+
+    // Single node, but of type 'seq', descend
+    if (ptr->nodes.size() == 1 && isA(ptr->nodes[0], NodeKind::Seq))
+      ptr = ptr->nodes[0];
+
+    // Return the nodes
+    for (auto n : ptr->nodes)
+      nodes.push_back(n);
```
davidchisnall, line 57:
```suggestion
    nodes.insert(nodes.end(), ptr->nodes.begin(), ptr->nodes.end());
```

Issue #179, opened 2020-07-07
CI and Nightly failing with MacOS image timeout


rengolin:
https://developercommunity.visualstudio.com/content/problem/1104836/macos-images-timing-out.html

rengolin:
MacOS images seem more stable today, but no feedback on the ticket yet. If nothing happens for another day I'll close this issue.

rengolin:
Problem fixed, ticket closed.

rengolin:
Happened again today:
https://dev.azure.com/ProjectVeronaCI/Project%20Verona/_build/results?buildId=1202&view=results

rengolin:
Nope, the new timeouts are on tests, the RT tests run for too long (48min) and the build times out because the free Mac builders have a hard-coded timeout of 60 minutes.

PR #189, opened 2020-07-14
Force 2h timeout for all builds

rengolin:
Some builds recently failed because the Mac builders were too slow and the RT tests were too long. Ramping up timeouts to 2hs and hopping the free builders can cope with it.


rengolin:
Fixes #179

Issue #117, opened 2020-05-19
Convert new AST into MLIR


rengolin:
First iteration using opaque MLIR and see what comes out of it. We still won't be able to lower that in any helpful way before we have a minimal dialect and the type inference working.

Acceptance criteria:
 * Take examples that are already parsed by the front-end and convert the whole AST into MLIR
 * Create some tests to show the conversion into some "expected" format
 * Add necessary nodes and types to the Verona dialect that make the process simpler

Not included:
 * Working on all examples before the parser can generate an AST for them
 * A complete Verona dialect (#106)

rengolin:
Will continue working on https://github.com/rengolin/verona/tree/mlir-ast branch from #107.

rengolin:
Skeleton implementation of `verona-mlir` in https://github.com/rengolin/verona/tree/verona-mlir.

Doesn't parse any AST, but has MLIR -> {MLIR, LLVM} structure, uses verona dialect and has some tests.

Next step, start looking at some basic AST constructs and lower to dialect IR.

rengolin:
First prototype merged (#155) and improved (#175, #177). Next step: internationalisation.

rengolin:
Now that the MLIR generator is working on basic constructs, and we have started working on the Verona dialect (#106) and its type   system (#104), we should start moving from opaque lowering to Verona dialect lowering (#107). I'll close this ticket and start working on that one.

PR #191, opened 2020-07-16
Ramping up support for more language constructs

rengolin:
We're still using `"verona.opp"` function-like syntax before we develop a Verona dialect, but everything is being built to be interchangeable when we actually do. We'd just replace `genOperation` for `builder.create<mlir::VeronaOpX>`. The reason why I'm not doing that yet is that the types are still not well defined, and MLIR requires exact types to be compatible.

This PR introduces the following features:
* Constant literal support for `int` and `float`, working on expressions and assignments
* Basic arithmetic and comparison binary operators
* Moved variables to stack allocation, load/store, to avoid having to track them through multiple basic-blocks
* Nested conditionals (if/else if/else chains)

The alloca strategy above will be optimised by the MLIR/LLVM machinery early on, in a pass that creates the PHI nodes, so we don't have to care about doing that ourselves.

The type system and the casts are in line with what @sylvanc is preparing for the AST, so it should be a good long-term strategy.

The commits are well separated, best to review them individually.

Continues to implement #117.


src/mlir/generator.cc
```diff
@@ -356,24 +404,130 @@ namespace mlir::verona
...
       auto arg0 = parseNode(getOperand(ast, 0).lock());
       if (auto err = arg0.takeError())
         return std::move(err);
       auto arg1 = parseNode(getOperand(ast, 1).lock());
       if (auto err = arg1.takeError())
         return std::move(err);
-      auto dialect = Identifier::get("type", &context);
-      auto type = mlir::OpaqueType::get(dialect, "ret", &context);
-      return genOperation(getLocation(ast), "verona.add", {*arg0, *arg1}, type);
+      if (name == "+")
```
davidchisnall, line 421: Would this nor be simpler as a `StringSwitch` returning a pair, if it's just mapping from `name` to the op name and type?

src/mlir/generator.cc
```diff
@@ -213,9 +214,21 @@ namespace mlir::verona
     assert(args.size() == argVals.size() && "Argument mismatch");
     for (auto var_val : llvm::zip(args, argVals))
     {
+      // Get the argument name/value
       auto name = getID(std::get<0>(var_val).lock());
       auto value = std::get<1>(var_val);
-      declareVariable(name, value);
+      // Allocate space in the stack
+      auto alloca =
+        genOperation(getLocation(ast), "verona.alloca", {}, allocaTy);
```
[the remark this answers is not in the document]
rengolin: Allocas need a type and currently the opaque types don't match any of the MLIR native ones. In time, we will use native alloca/load/store for all types that have MLIR types.

rengolin:
@davidchisnall thanks for the comments, all addressed.

src/mlir/generator.cc
```diff
@@ -356,24 +405,101 @@ namespace mlir::verona
...
       auto arg1 = parseNode(getOperand(ast, 1).lock());
       if (auto err = arg1.takeError())
         return std::move(err);
-      auto dialect = Identifier::get("type", &context);
-      auto type = mlir::OpaqueType::get(dialect, "ret", &context);
-      return genOperation(getLocation(ast), "verona.add", {*arg0, *arg1}, type);
+      // Get op name and type
+      using opPairTy = std::pair<llvm::StringRef, mlir::Type>;
+      opPairTy op = llvm::StringSwitch<opPairTy>(name)
+                      .Case("+", std::make_pair("verona.add", unkTy))
```
davidchisnall, line 425: I believe you can use initialiser lists here, instead of make pair, because the argument of `.Case` is defined by the string switch.  That will make it easier to extend this to a tuple if you need more values later.

PR #192, opened 2020-07-16
Add perf-ubench runtime microbenchmark

Theodus:


winksaville:
As a future enhancement, have you considered permanently logging the measurments so changes in performance could be indentified?

davidchisnall:
I think there’s likely to be a lot of noise since CI runs in a VM, but enough samples may give us enough of a signal to spot regressions. It would be good to have a nightly or weekly benchmark run that Ran the Verona and snmalloc perf tests and recorded the results. Azure Pipelines can record perf numbers, I believe.

rengolin:
Preparing a machine to be stable enough to create metrics on benchmarks is not trivial. Relying on the free VMs we use to be that stable, even less so. If we do create such a nightly build we need to make sure we run them on something that we can control enough to make the numbers stable where there are no code changes.

We also need to run one benchmark at a time, which means we won't be able to run a lot of different benchmarks overnight, so we need to choose wisely which ones we do and how.

It's also advisable to only put on this build programs that were created with the intention of being benchmarks. Often people compile programs and run them as benchmarks, but their execution is dominated by setup/teardown or I/O issues.

Finally, we should be able to set different targets and tolerance to each different benchmark run. Even with the same binary, different arguments can yield very different deviation distributions. We could increase runtime of smaller benchmarks if we're forced to have a single threshold, but controlling deviation is usually a lot harder.

PR #193, opened 2020-07-17
Implement while blocks, improve conditional support and lexical contexts

rengolin:
While implementing simple while blocks, I realised the variable context for lexical blocks wasn't pushing new frames when entering if/else blocks, which were also needed when entering loop blocks. This PR fixes that problem and implements simple while loops with conditionals, continue and break.

Not supported yet: nested loops. I need to do the refactor of the BranchOp / Value return that I mention in the TODO. This will be done next.

The patches can be reviewed as one or individually.

Continues to implement #117.

### DIFF
--- a/src/mlir/ast-utils.cc
+++ b/src/mlir/ast-utils.cc
@@ -38,7 +38,7 @@ namespace mlir::verona::ASTInterface
   {
     auto nodes = ast.lock()->nodes;
     return std::find_if(nodes.begin(), nodes.end(), [&](::ast::Ast& node) {
-             return getKind(node) == kind && !node->nodes.empty();
+             return isA(node, kind);
            }) != nodes.end();
   }
 
@@ -311,12 +311,16 @@ namespace mlir::verona::ASTInterface
 
   bool hasElse(::ast::WeakAst ast)
   {
-    return isA(ast, NodeKind::If) && hasA(ast, NodeKind::Else);
+    // Else nodes always exist inside `if` nodes, but if there was no `else`
+    // block, they're empty. We should only return true if they're not.
+    return isA(ast, NodeKind::If) && hasA(ast, NodeKind::Else) &&
+      !findNode(ast, NodeKind::Else).lock()->nodes.empty();
   }
 
   ::ast::WeakAst getCond(::ast::WeakAst ast)
   {
-    assert(isAny(ast, {NodeKind::If, NodeKind::While}) && "Bad node");
+    // These are the nodes that could have conditions as subnodes
+    assert((isIf(ast) || isLoop(ast)) && "Bad node");
     // Cond nodes are always seq, even with just one cond
     auto cond = findNode(ast, NodeKind::Condition);
     return findNode(cond, NodeKind::Seq);
@@ -342,6 +346,12 @@ namespace mlir::verona::ASTInterface
     return isA(ast, NodeKind::While);
   }
 
+  bool isLoop(::ast::WeakAst ast)
+  {
+    // Add isFor when we support it
+    return isWhile(ast);
+  }
+
   bool isContinue(::ast::WeakAst ast)
   {
     return isValue(ast) && isA(ast, NodeKind::Continue);
@@ -354,7 +364,7 @@ namespace mlir::verona::ASTInterface
 
   ::ast::WeakAst getLoopBlock(::ast::WeakAst ast)
   {
-    assert(isWhile(ast) && "Bad node");
+    assert(isLoop(ast) && "Bad node");
     return findNode(ast, NodeKind::Block);
   }
 }
--- a/src/mlir/ast-utils.cc
+++ b/src/mlir/ast-utils.cc
@@ -342,6 +342,16 @@ namespace mlir::verona::ASTInterface
     return isA(ast, NodeKind::While);
   }
 
+  bool isContinue(::ast::WeakAst ast)
+  {
+    return isValue(ast) && isA(ast, NodeKind::Continue);
+  }
+
+  bool isBreak(::ast::WeakAst ast)
+  {
+    return isValue(ast) && isA(ast, NodeKind::Break);
+  }
+
   ::ast::WeakAst getLoopBlock(::ast::WeakAst ast)
   {
     assert(isWhile(ast) && "Bad node");
--- a/src/mlir/ast-utils.cc
+++ b/src/mlir/ast-utils.cc
@@ -34,6 +34,14 @@ namespace mlir::verona::ASTInterface
     return std::find(kinds.begin(), kinds.end(), getKind(ast)) != kinds.end();
   }
 
+  bool hasA(::ast::WeakAst ast, NodeKind kind)
+  {
+    auto nodes = ast.lock()->nodes;
+    return std::find_if(nodes.begin(), nodes.end(), [&](::ast::Ast& node) {
+             return getKind(node) == kind && !node->nodes.empty();
+           }) != nodes.end();
+  }
+
   ::ast::WeakAst findNode(::ast::WeakAst ast, NodeType type)
   {
     assert(!isValue(ast) && "Bad node");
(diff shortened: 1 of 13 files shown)


src/mlir/ast-utils.cc
```diff
@@ -34,6 +34,14 @@ namespace mlir::verona::ASTInterface
     return std::find(kinds.begin(), kinds.end(), getKind(ast)) != kinds.end();
   }
 
+  bool hasA(::ast::WeakAst ast, NodeKind kind)
+  {
+    auto nodes = ast.lock()->nodes;
+    return std::find_if(nodes.begin(), nodes.end(), [&](::ast::Ast& node) {
+             return getKind(node) == kind && !node->nodes.empty();
```
davidchisnall, line 41: Is there a reason that this can’t be implemented in terms of isa? 

What is the second check in this test for? Can you add a comment explaining it.

src/mlir/ast-utils.cc
```diff
@@ -301,9 +309,14 @@ namespace mlir::verona::ASTInterface
...
 
+  bool hasElse(::ast::WeakAst ast)
+  {
+    return isA(ast, NodeKind::If) && hasA(ast, NodeKind::Else);
+  }
+
   ::ast::WeakAst getCond(::ast::WeakAst ast)
   {
-    assert(isIf(ast) && "Bad node");
+    assert(isAny(ast, {NodeKind::If, NodeKind::While}) && "Bad node");
```
davidchisnall, line 319: We will probably end up with a bunch of conditional and loop constructs, it might be worth factoring this check out into an isConditional or similar.

src/mlir/ast-utils.cc
```diff
@@ -323,4 +336,25 @@ namespace mlir::verona::ASTInterface
...
+  }
+
+  bool isBreak(::ast::WeakAst ast)
+  {
+    return isValue(ast) && isA(ast, NodeKind::Break);
+  }
+
+  ::ast::WeakAst getLoopBlock(::ast::WeakAst ast)
+  {
+    assert(isWhile(ast) && "Bad node");
```
davidchisnall, line 357: Similarly here, it would be good to factor the check out into an isLoop check (so isConditional can be isLoop or isIf), the. This will work automatically when we add other loop kinds.

src/mlir/generator.cc
```diff
@@ -263,11 +264,12 @@ namespace mlir::verona
       return builder.getNoneType();
 
     // If type is in the alias table, get it
-    if (typeTable.inScope(desc))
-      return typeTable.lookup(desc);
+    auto type = typeTable.lookup(desc);
```
davidchisnall, line 267: Can this use operator[] and assign to the result on line 273, rather than doing the search twice?

src/mlir/generator.cc
```diff
@@ -329,7 +337,11 @@ namespace mlir::verona
     if (isLocalRef(ast))
     {
       // We use allocas to track location and load/stores to track access
-      auto var = symbolTable.lookup(getTokenValue(ast));
+      auto name = getTokenValue(ast);
+      auto var = symbolTable.lookup(name);
+      if (!var)
+        return parsingError(
```
davidchisnall, line 343: Not sure it matters at the moment, but this is probably the kind of error that shouldn’t stop parsing (and suppressing future errors). For the bootstrap compiler, the input should always have been type checked with the second compiler, so it might not matter...

src/mlir/generator.cc
```diff
@@ -470,28 +491,106 @@ namespace mlir::verona
     // Create basic-blocks, conditionally branch to if/else
     mlir::ValueRange empty{};
     auto ifBB = currentFunc.addBlock();
-    auto elseBB = currentFunc.addBlock();
+    mlir::Block* elseBB = nullptr;
+    if (hasElse(ast))
```
davidchisnall, line 495: Do we have a style rule relating to braces? I have a strong preference for braces around every conditional body.

src/mlir/ast-utils.cc
rengolin: No reason, changing. Looking at it again, the second check is more about the `else` node than the `hasA` functionality, so I'll move it down to the right place.

src/mlir/ast-utils.cc
rengolin: The `isConditional` checks if the node is `cond`. This check is to make sure the node "could possibly have conditionals". For now, `if` and loops can, so I'll change it to that.

src/mlir/generator.cc
rengolin: That is true, but I didn't want to have a potential error unchecked (and null pointers dangling). I could just change that to an assert, as the front-end should really have checked it.

src/mlir/generator.cc
rengolin: I'm trying to avoid the call to `getOpaqueType()`. Initially, types will search twice, but after the types are all declared, they'll only be searched once (to retrieve from the cache).

If I switch the order, I need to generate the opaque type, and then every call, including when it's in the cache, will create a new type and then realise it's already there, and dispose of the temp.

src/mlir/generator.cc
rengolin: This is a multi-line (single statement) if/else, so I'll add braces.

We discussed it and the conclusion is that it's generally good for most cases, but it can clutter in error treatment cases like:
```
    auto var = foo(bar);
    if (auto err = var.takeError())
        return err;
    ... continue;
```

My approach is to only refrain from braces if there is a single line on both `if` and `else` (if available) statements. 

But it would also be fine to add to *any* `if-with-else` statement and have it only on `single-line-only-if-block` case, which I have a strong preference for.

rengolin:
Comments addressed, let me know if the answers to the remaining two make sense.

src/mlir/generator.cc
davidchisnall: I meant something along the lines of:
```c++
auto &type = typeTable[desc];
if (!type)
  type = getOpaqueType(desc, context)
return type;
```
Would this not work?

src/mlir/generator.cc
davidchisnall: I'd really like to avoid omitting braces on anything with an else clause.  The compiler will warn in ambiguous cases, but the C grammar does not do what humans expect in this case.

rengolin: That was my point above, I agree with that. The only one I still think we should avoid braces is "if with single line statement and no else".

src/mlir/generator.cc
rengolin: First, this is a cache pattern. I need to insert it into the table before returning. I can't return straight from `getOpaqueType`.

Second, the table isn't a simple map, it's a stack of maps. There are two searches: `inScope` only looks at the current lexical frame while `lookup` looks at the whole stack. Overloading `[]` could be done to either of them and it's not clear to me which would make more sense.

Finally, before, I was doing a local search ('inScope') and then returning it only if it was in the current scope (not the lexical parents), which is wrong. I need to traverse all contexts. This is an expensive search for new types, but so is creating the type (it will get more complex in the future). The idea is that we'll use a lot more types than we'll create new ones.

src/mlir/generator.cc
rengolin: PS: You need to look at the new change to see that I have added braces on the block you asked. If you click on the link above the comment, you will see the old change, which is outdated.

src/mlir/generator.cc
rengolin: I have just pushed a new commit that considerably simplifies the symbol table. It was a hybrid of a few past attempts and extremely redundant. I think this fix and the new implementation will considerably reduce the complexity.

davidchisnall: I see.  If you miss in the first scope, you need to search the others anyway so that you can insert.  That can lead to quite expensive lookups for nested scopes, but it's probably fine.  In the past, I've implemented this with a map that does CoW for the buckets, so creating a new scope is a cheap refcounted 'copy', inserting new symbols may be relatively expensive (actually copies part of the map) and lookup cost is independent of nesting complexity and the lookup-and-insert-if-not-present operation is cheaper than lookup then insert.  

I don't think that's necessary here.

PR #195, opened 2020-07-20
Adding support for function calls and return values

rengolin:
Changing how to detect basic-block termination to account for all known
terminators and empty basic blocks, so we don't need to terminate blocks
that have already been handled.

Implementing return values (of single expressions) as well as function
calls with the appropriate casts for the arguments, given we don't yet
know all types at this time.

If Verona accepts multiple return values, we'll need to change the
return value of all `parseX` functions to a vector of values. This
should be done later, as a refactor PR with examples. It will also fix
the `return mlir::Value()` problem, as we can return an empty list.

(ref. #117)

### DIFF
--- a/src/mlir/ast-utils.cc
+++ b/src/mlir/ast-utils.cc
@@ -56,6 +56,13 @@ namespace mlir::verona::ASTInterface
     return *sub;
   }
 
+  ::ast::WeakAst getSingleSubNode(::ast::WeakAst ast)
+  {
+    auto ptr = ast.lock();
+    assert(ptr->nodes.size() == 1 && "Wrong number of nodes");
+    return ptr->nodes[0];
+  }
+
   std::vector<::ast::WeakAst> getSubNodes(::ast::WeakAst ast)
   {
     auto ptr = ast.lock();
@@ -237,6 +244,11 @@ namespace mlir::verona::ASTInterface
     return isA(ast, NodeKind::Call);
   }
 
+  bool isReturn(::ast::WeakAst ast)
+  {
+    return isA(ast, NodeKind::Return);
+  }
+
   bool isAssign(::ast::WeakAst ast)
   {
     return isA(ast, NodeKind::Assign);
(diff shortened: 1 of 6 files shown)


rengolin:
@plietar, I also can't add you as a reviewer.

plietar:
Looks like I don't have write access anymore, which would be why I can't be reviewer.
@mjp41 would you mind promoting me?

rengolin:
Regarding the comment from @davidchisnall in #196, I don't think I should move the implementation to the header of all list helpers. Keeping vector as a dependency, for now, until we reach a better implementation, should be fine.

davidchisnall:
I hadn't noticed until you pointed it out in the other review that all of these wrappers are in the .cc file.  Is there a reason for this?  Unless we're compiling with LTO, this means that they won't be inlined and most of them are trivial functions that would benefit a lot from inlining.  In particular, things like `isA` and `isReturn` are likely to have their performance dominated by the call overhead.

rengolin:
This interface will not be here in the long run, so I'm not worrying about how it performs right now. It will also change radically in the next month or so, with the AST changing so much and new MLIR dialects, the implementation and use of those helpers will change.

The main reason why I split is to reduce compilation time in development cycle, which right now is more important than runtime. I'd prefer to leave building performance analysis to after we have stable and good code.

davidchisnall:
Did you measure the built-time differences?  I'd expect them to be minimal (and basically zero if you enable C++ modules in the build).  At the moment, we're paying a cost from having hard-coded `std::vector` into the interface (and, more generally, from not being able to use templates in this layer).

rengolin:
Fair enough, I didn't have a strong reason to have it in a cpp file to begin with. It was more arc reflex, and then I rationalised it. Will merge them.

rengolin:
@davidchisnall, given that vector and separate file is a pattern that was already there, and that I have another PR with refactories, I'll prefer to add this change there (#196) than here, since this is only about adding calls.

rengolin:
Refactrory requested in #196

PR #196, opened 2020-07-20
Four unrelated refactories in MLIR Generator

rengolin:
Two non-functional-changes:
 * Add a comment with the str2tag of each node to ease debugging
 * Update comments to doc-style on MLIR Generator

One structural change:
 * Add support for zero, one or more return values with asserts to guarantee semantics

This commit needs PR #195 to be merged first.

(ref #117).


rengolin:
@plietar

src/mlir/ast-utils.cc
```diff
@@ -288,6 +300,15 @@ namespace mlir::verona::ASTInterface
...
 
+  std::vector<::ast::WeakAst> getAllOperands(::ast::WeakAst ast)
+  {
+    auto numOps = numOperands(ast);
+    std::vector<::ast::WeakAst> ops;
+    for (size_t i = 0; i < numOps; i++)
+      ops.push_back(getOperand(ast, i));
+    return ops;
+  }
+
```
davidchisnall, line 311: I'm a bit nervous about embedding `std::vector`s at this kind of level.  Could this be written as a template that takes a reference to the collection as a parameter, something like:
```suggestion
  template<class T>
  std::vector<::ast::WeakAst> getAllOperands(T &ops, ::ast::WeakAst ast)
  {
    auto numOps = numOperands(ast);
    for (size_t i = 0; i < numOps; i++)
      ops.push_back(getOperand(ast, i));
    return ops;
  }
```

src/mlir/generator.cc
```diff
@@ -397,25 +417,47 @@ namespace mlir::verona
...
     // All operations are calls, only calls to previously defined functions
-    // are function calls. FIXME: Is this really what we want?
-    if (functionTable.inScope(name))
+    // are function calls.
+    if (auto func = functionTable.lookup(name))
     {
-      // TODO: Lower calls.
-      return parsingError(
-        "Function calls not implemented yet", getLocation(ast));
+      auto argNodes = getAllOperands(ast);
```
davidchisnall, line 436: With the change above, this could then be:
```suggestion
      llvm::SmallVector<::ast::WeakAst, 8> argNodes;
      getAllOperands(argNodes, ast);
```

And now we can choose to use any vector-like type.

src/mlir/ast-utils.h
```diff
@@ -45,36 +45,37 @@ namespace mlir::verona::ASTInterface
...
-    Local = peg::str2tag("local"),
-    Localref = peg::str2tag("localref"),
-    Float = peg::str2tag("float"),
-    Condition = peg::str2tag("cond"),
-    If = peg::str2tag("if"),
-    Else = peg::str2tag("else"),
-    While = peg::str2tag("while"),
-    Continue = peg::str2tag("continue"),
-    Break = peg::str2tag("break"),
+    Module = peg::str2tag("module"), // 73005690
```
davidchisnall, line 48: What are these numbers?

rengolin: The integral value of calling `peg::str2tag("nodename")`, and consequently, the integral value of the enum item.

Perhaps I could make the comment `// = NNN` to make it clearer?

src/mlir/ast-utils.cc
rengolin: Sorry, this is a commit from a different PR, in which this change is based on (#195). I see what you mean and will look at the other cases, too, but will update the other PR with it, then rebase this one. This is for the three last commits only.

I don't know how to do multiple reviews in one without carrying the commits along in Github.

src/mlir/generator.cc
rengolin: Because of the template parameter, this can only work if the implementation is in the header, not the `cc` file.

Given this is only a prototype for the interface, I don't think we should worry about it that much.

rengolin:
Updated to `// = NNN` on the enum comments to make it clearer. Comments about call should be done in #195.

rengolin:
Note, two changes here are non-functional and one is preparation for a functional change (tuple return) but has yet not changed the behaviour at all, thus no new tests.

PR #197, opened 2020-07-20
Use CMake ExternalProject for LLVM

mjp41:
This PR uses CMake `ExternalProject_Add` to download or build LLVM.

By setting
```
-DVERONA_DOWNLOAD_LLVM=On
```
It will download the correct LLVM blob, and then build using that.

By setting
```
-DVERONA_DOWNLOAD_LLVM=Off
```
It will build the submodule as a subproject, and once that has completed build the Verona files.

The core complication is that Verona is  using CMake files produced by the LLVM build, so we need to schedule the build
of Verona after LLVM has completed.  This means we build Verona as a subproject, so that we can execute the cmake configure phase after the LLVM build has completed.


mjp41:
@plietar for some reason I can't add you as a reviewer.

Could everyone please try this to see if it works locally.

docs/building.md
```diff
@@ -24,19 +24,30 @@ git submodule update --recursive
 
 LLVM builds take a long time, so we cached the build directory from Linux, Windows and MacOS for our CI, that can also be used on developer's machines.
 
-Run the following script at the root of your checkout:
+The CMake flag `VERONA_DOWNLOAD_LLVM` is used to control whether a cached build 
+is used or not. 
+
 ```
-bash ./utils/llvm/setup-llvm-builddir.sh verona-llvm-x86_64-${OS}-${BuildType}-${LLVMCommit}.tar.gz
+cmake -DVERONA_DOWNLOAD_LLVM=On ..
```
Theodus, line 31: Minor nit: we should remain consistent about using all caps for `ON` and `OFF`.

docs/building.md
```diff
@@ -24,19 +24,30 @@ git submodule update --recursive
...
-With `${OS}` = { `linux`, `windows`, `macos` }, `BuildType` = { `release`, `debug` } and `${LLVMCommit}` being the same as the llvm-project submodule (currently @ `3c123acf57c`).
-
-On Windwows, run it on PowerShell, not `cmd`, to get access to unix-like tools. You can use LLVM release builds on Linux and MacOS debug builds of Verona, but on Windows, you must use the same build type, to avoid CRT incompatibility issues.
-
-If you want to compile LLVM directly, the options we used in the cache are available [here][]. For now, you'll have to build it before Verona in its own build directory. We're working to make that build straight from Verona's build directory, but it's not yet ready.
-
-[here]: https://github.com/microsoft/verona/blob/master/devops/llvm.yml
+If this flag is unset, 
+```
+cmake -DVERONA_DOWNLOAD_LLVM=On ..
```
Theodus, line 39: This part is ambiguous about whether this comment unsets the flag or whether it will result in compiling LLVM locally

CMakeLists.txt
```diff
@@ -1,9 +1,186 @@
...
+endmacro()
+
+verona_option(ENABLE_ASSERTS "Enable asserts even in release builds" OFF)
+verona_option(RT_TESTS "Including unit tests for the runtime" OFF)
+verona_option(USE_SCHED_STATS "Track scheduler stats" OFF)
+verona_option(USE_ASAN "Use address sanitizer" OFF)
+verona_option(VERONA_CI_BUILD "Disable features not sensible for CI" OFF)
+verona_option(USE_SYSTEMATIC_TESTING "Enable systematic testing in the runtime" OFF)
+verona_option(VERONA_EXPENSIVE_SYSTEMATIC_TESTING "Increase the range of seeds covered by systematic testing" OFF)
+verona_option(USE_CRASH_LOGGING "Enable crash logging in the runtime" OFF)
```
Theodus, line 22: Should these options include the `USE_MEASURE`, `USE_SCHED_STATS`, and `USE_QUICK_EXIT` options removed from the rt cmake?

CMakeLists.txt
```diff
@@ -1,9 +1,186 @@
...
+    list (APPEND LLVM_EXTRA_CMAKE_ARGS
+      -DCMAKE_INSTALL_PREFIX=${MLIR_INSTALL}
+      -DLLVM_ENABLE_PROJECTS=mlir
+      -DLLVM_TARGETS_TO_BUILD=X86
+      -DLLVM_ENABLE_ASSERTIONS=On
+      -DLLVM_ENABLE_RTTI=ON
+      -DLLVM_ENABLE_EH=ON
+    )
+
+    # TODO: remove before finished!
```
Theodus, line 111: Should this remain for now?

plietar:
With `VERONA_DOWNLOAD_LLVM=Off` it at first fails for me (on macOS):

```
> mkdir build && cd build
> cmake .. -GNinja -DCMAKE_BUILD_TYPE=Debug -DVERONA_DOWNLOAD_LLVM=Off
> ninja
[6/16] Performing configure step for 'verona'
[....]
-- Setting LLVM_DIR as /Users/paul/PhD/verona/build/mlir_install/lib/cmake/llvm
-- Setting MLIR_DIR as /Users/paul/PhD/verona/build/mlir_install/lib/cmake/mlir
-- Setting LLVM_EXTERNAL_LIT as /Users/paul/PhD/verona/build/mlir_install/bin/llvm-lit
CMake Error at src/mlir/CMakeLists.txt:20 (find_package):
  Could not find a package configuration file provided by "MLIR" with any of
  the following names:

    MLIRConfig.cmake
    mlir-config.cmake

  Add the installation prefix of "MLIR" to CMAKE_PREFIX_PATH or set
  "MLIR_DIR" to a directory containing one of the above files.  If "MLIR"
  provides a separate development package or SDK, be sure it has been
  installed.
[....]
ninja: build stopped: subcommand failed.
```

By adding `set(MLIR_TARGETS mlir)` at the end of the `!VERONA_DOWNLOAD_LLVM` case in `CMakeLists.txt`, I get a little bit further:
```
> ninja
[4/16] Performing configure step for 'mlir'
[....]
-- Performing Test CXX_SUPPORTS_CUSTOM_LINKER
-- Performing Test CXX_SUPPORTS_CUSTOM_LINKER - Failed
CMake Error at cmake/modules/HandleLLVMOptions.cmake:272 (message):
  Host compiler does not support '-fuse-ld=lld'
[....]
ninja: build stopped: subcommand failed.
```

I've then commented out the `-DLLVM_ENABLE_LLD=ON` line, and it gets me to the point where everything compiles.

-------

With `VERONA_DOWNLOAD_LLVM=On`, I'm getting linking issues (can't find `-lcurses`), but I was already getting that with the previous script. This probably comes from my nixpkgs-based setup being different from the one used to create the prebuilt packages, so I'm not too worried. Would be good if someone with a more conventional setup could try.

--------

For reasons\*, I would usually always pass `-DCMAKE_OSX_DEPLOYMENT_TARGET=10.14` (or higher) to cmake in order to compile Verona. Now the flag is lost in the recursive cmake invocation. I can work around it by instead using the `MACOSX_DEPLOYMENT_TARGET` environment variable, which cmake uses as a default value. However, I worry this will apply to any kind of local setup-specific cmake argument, such as CMAKE_CXX_COMPILER, ...

\* My compiler is installed with nixpkgs, which targets everything to 10.12 by default for binary portability or whatever.

CMakeLists.txt
mjp41: USE_SCHED_STATS has remained.  The others should have been removed already, as they don't control anything in Verona anymore.

CMakeLists.txt
mjp41: No as @seantallen would not be happy ;-) 
I am going to expose `LLVM_EXTRA_CMAKE_ARGS` for passing additional arguments to CMake, so this can pass the flag.  Will drop the comment.

docs/building.md
mjp41: Fixed

mjp41:
@plietar I have exposed `VERONA_EXTRA_CMAKE_ARGS` and `LLVM_EXTRA_CMAKE_ARGS`.  These pass the arguments through to the underlying two cmake builds. Does this work for you?

rengolin:
I'll be testing on my machine today...

davidchisnall:
Looks fine to me. @letmaik, if you have a few spare minutes it would be great to get your opinion on this.

rengolin:
> With `VERONA_DOWNLOAD_LLVM=On`, I'm getting linking issues (can't find `-lcurses`), but I was already getting that with the previous script. This probably comes from my nixpkgs-based setup being different from the one used to create the prebuilt packages, so I'm not too worried. Would be good if someone with a more conventional setup could try.

Unfortunately, this is bound to happen with any download. I get `libtinfo` issues because Arch has moved on, while Ubuntu hasn't. Curses is what its name imply and implementations can be wildly different even on similar platforms.

The download is a "best effort" thing to help CI and hopefully will help some developers. If that doesn't work, you can always build it locally (once) and not bother about that until we change the LLVM version (a month or so later).

rengolin:
First experience on Linux Arch, `cmake -G Ninja -DVERONA_DOWNLOAD_LLVM=On && ninja`, downloaded LLVM and built Verona with all tests passing.

The only thing I'd say is that the download progress is very verbose and unnecessary. If there's a way to be less verbose (say, every 10%) or some other type of progress, would be nice, but not "important".

rengolin:
First issue: When I set `-DCMAKE_CXX_COMPILER=clang++` in Verona, it doesn't propagate that decision to the LLVM build. It's not mandatory, but highly encouraged, that we use the same compiler for both builds, preferably without having to set them manually on the new `LLVM_EXTRA_CMAKE_ARGS` flag.

I would propagate at least `CMAKE_C_COMPILER` and `CMAKE_CXX_COMPILER` flags, but not their `_FLAGS` counterparts, as that wouldn't make sense.

Of course, passing them down via `LLVM_EXTRA_CMAKE_ARGS` would overwrite them.

rengolin:
I can't seem to pass both C and CXX argument to `-DLLVM_EXTRA_CMAKE_ARGS`...

CMakeLists.txt
```diff
@@ -1,9 +1,187 @@
...
+# artifacts, which are not available otherwise.
+include(ExternalProject)
+if (NOT DEFINED VERONA_LLVM_LOCATION)
+  project(verona CXX)
+  # Restrict types of builds to ones where we have cached LLVM
+  set(default_build_type "Release")
+  if (MSVC)
+    if (NOT CMAKE_CONFIGURATION_TYPES)
+      if (NOT CMAKE_BUILD_TYPE)
+        set(CMAKE_CONFIGURATION_TYPES "Debug;Release" CACHE STRING "Choose the types of build." FORCE)
```
rengolin, line 38: What about 'RelWithDbgInfo'?

CMakeLists.txt
```diff
@@ -1,9 +1,187 @@
...
+    find_package(Git)
+    execute_process(
+      COMMAND ${GIT_EXECUTABLE} rev-parse --short=11 HEAD
+      WORKING_DIRECTORY "${CMAKE_CURRENT_SOURCE_DIR}/external/llvm-project"
+      OUTPUT_VARIABLE LLVM_PACKAGE_GIT_VERSION
+    )
+
+    string(TOLOWER ${CMAKE_SYSTEM_NAME} PLATFORM)
+    if (${PLATFORM} STREQUAL "darwin")
+      set (PLATFORM "macos")
```
rengolin, line 65: We could rename the packages to 'darwin', too. Since we're creating entirely new packages, this would be a good time to do it. :)

CMakeLists.txt
```diff
@@ -1,9 +1,187 @@
...
+    endif ()
+
+    set(MLIR_TARGETS)
+    # Need loop to download all types for possible configurations
+    # Required by multi-build systems like msbuild
+    foreach (BUILD_TYPE_MIXED ${CMAKE_CONFIGURATION_TYPES})
+      string (TOLOWER ${BUILD_TYPE_MIXED} BUILD_TYPE)
+      if (${BUILD_TYPE} STREQUAL relwithdebinfo)
+        # RelWithDebInfo should use release LLVM.
```
rengolin, line 74: For context, we only do this in CI because the 'RelWithDbgInfo' package is too large for our builders. I don't think this decision should be made here for everybody else.

CMakeLists.txt
```diff
@@ -1,9 +1,187 @@
...
+    set (MLIR_TYPE mlir-$<LOWER_CASE:$<CONFIG>>)
+    set (MLIR_INSTALL 
+      ${CMAKE_BINARY_DIR}/${MLIR_TYPE}-prefix/src/${MLIR_TYPE}/install)
+  else()
+    set (MLIR_INSTALL ${CMAKE_BINARY_DIR}/mlir_install)
+
+    if (NOT DEFINED LLVM_EXTRA_CMAKE_ARGS)
+      set (LLVM_EXTRA_CMAKE_ARGS)
+    endif ()
+    list (APPEND LLVM_EXTRA_CMAKE_ARGS
```
rengolin, line 106: I think we should add `CMAKE_C_COMPILER` and `CMAKE_CXX_COMPILER` flags, too.

CMakeLists.txt
```diff
@@ -1,9 +1,187 @@
...
+      -DCMAKE_INSTALL_PREFIX=${MLIR_INSTALL}
+      -DLLVM_ENABLE_PROJECTS=mlir
+      -DLLVM_TARGETS_TO_BUILD=X86
+      -DLLVM_ENABLE_ASSERTIONS=On
+      -DLLVM_ENABLE_RTTI=ON
+      -DLLVM_ENABLE_EH=ON
+    )
+
+    if (NOT MSVC AND NOT APPLE)
```
rengolin, line 115: LLD is not mandatory for Linux either. I think this could be a thing for extra flags.

mjp41:
@rengolin, was that a completely clean build directory.  You have to get rid of subdirectories too, given the new cmake.

devops/ci.yml
```diff
@@ -100,19 +76,19 @@ jobs:
...
+      ninja
     workingDirectory: build
     displayName: 'Compile'
 
   - script: |
       set -eo pipefail
       export ASAN_SYMBOLIZER_PATH=/usr/lib/llvm-6.0/bin/llvm-symbolizer
       export ASAN_OPTIONS="alloc_dealloc_mismatch=0 symbolize=1"
-      ctest -j 4 -E "([1-9][0-9]00|[4-9]00)" --timeout 400 --output-on-failure
+      ninja check
```
rengolin, line 91: On Linux, just calling `ninja` already runs all the tests. Is this step really necessary?

devops/ci.yml
```diff
@@ -210,36 +166,21 @@ jobs:
...
     displayName: 'CMake'
     inputs:
       cmakeArgs: |
-        .. -DCMAKE_BUILD_TYPE=$(BuildType) -DENABLE_ASSERTS=ON -DVERONA_CI_BUILD=On -DSNMALLOC_CI_BUILD=On -DRT_TESTS=$(RTTests)
+        .. -DCMAKE_BUILD_TYPE=$(BuildType) -DENABLE_ASSERTS=ON -DVERONA_CI_BUILD=On -DRT_TESTS=$(RTTests)
 
   - script: |
       set -eo pipefail
-      make -j 4 install
+      make -j 4
```
rengolin, line 177: I know it was there already, but hard-coding `4` here isn't helpful. There's a way to get the number of cores on Mac.

docs/building.md
```diff
@@ -24,19 +24,36 @@ git submodule update --recursive
...
+```
+cmake -DVERONA_DOWNLOAD_LLVM=OFF ..
+```
+then the build will compile LLVM locally, and use that to build the Verona
+compiler.
 
-[here]: https://github.com/microsoft/verona/blob/master/devops/llvm.yml
+Additional, parameters can be passed to the LLVM build: e.g.
+```
+cmake -DVERONA_DOWNLOAD_LLVM=OFF .. -DLLVM_EXTRA_CMAKE_ARGS="-DLLVM_USE_SANITIZER=Address"
```
rengolin, line 46: Give an example with more than one flag, as I was tricked by the semi-colon.

src/mlir/CMakeLists.txt
```diff
@@ -1,19 +1,22 @@
 find_package(Threads REQUIRED)
 
 # We're not building LLVM, but downloading the build dir from an artefact
-set(LLVM_BUILD "${CMAKE_SOURCE_DIR}/external/llvm-project/build")
+set(LLVM_BUILD ${VERONA_LLVM_LOCATION})
+
+# std::iterator is used by MLIR/LLVM
```
rengolin, line 6: I'd expand on this a bit, saying a fix in LLVM is probably going to happen eventually and we need to revisit this warning when it does.

rengolin:
> @rengolin, was that a completely clean build directory. You have to get rid of subdirectories too, given the new cmake.

Completely new. My test cycle is:
```
10 rm -rf *
20 cmake ... 
30 ninja
40 GOTO 10
```

rengolin:
Hm, I got an LLVM build failure. My external project is in the right commit, `3c123acf57c` and I haven't set any extra flags this time. The error I get is in MLIR's test header:
```
home/rengolin/devel/msrc/verona/external/llvm-project/mlir/test/lib/Transforms/TestInlining.cpp:49:75:   required from here
/usr/include/c++/10.1.0/type_traits:909:52: error: static assertion failed: template argument must be a complete class or an unbounded array
  909 |       static_assert(std::__is_complete_or_unbounded(__type_identity<_Tp>{}),
      |                     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~
```

The MLIR build flags don't seem any different than the LLVM CI job, so I'm a bit lost. I'll check after lunch.

CMakeLists.txt
mjp41: This is not currently built by the llvm.yml.  I need to finish this today. So, can we leave expanding functionality to a later PR?

devops/ci.yml
mjp41: That should not be the case.  Except on MSVC, as I couldn't get `ExternalProject_Add_StepTargets` to work correctly with it.

rengolin: On my build, `ninja check` complained the target didn't exist, so I just ran `ninja` and it ran the tests.

CMakeLists.txt
rengolin: Sure, np. Just add a TODO saying this is about the CI.

rengolin:
To be clear, I'd already be happy with a change that downloads the package and build Verona, but required developers to build LLVM by hand before if they wanted it that way. Building LLVM automatically, for me personally, is a bonus that doesn't need to come in this PR.

The main problem with this is for the cases where the download image fails to build (library issues) and the developer is not familiar with LLVM. Giving we're still in a pre-prototype stage, I'd say both risk and impact are low.

rengolin:
If we want to make sure LLVM builds in tree, I'd add a new CI pipeline to do that on all supported configurations, raising the timeout and letting it run overnight every time we update LLVM.

devops/ci.yml
mjp41: TODO in #198

CMakeLists.txt
mjp41: TODO in #198

CMakeLists.txt
mjp41: TODO in #198

devops/ci.yml
mjp41: Okay.  Which type of build did you do for this.  Download, or build for llvm?  Any options?  Which version of CMake?

mjp41:
I have added an issue to track the further improvements to this, but I would like this to land before I stop working for 1.5 weeks.

> The only thing I'd say is that the download progress is very verbose and unnecessary. If there's a way to be less verbose (say, every 10%) or some other type of progress, would be nice, but not "important".

Unfortunately, you can have in on or off. Will add a quiet build option that doesn't show the CMake phases, or the download to the issue tracking TODOs.

mjp41:
> The only thing I'd say is that the download progress is very verbose and unnecessary. If there's a way to be less verbose (say, every 10%) or some other type of progress, would be nice, but not "important".

Added to the follow up issue.

Issue #198, opened 2020-07-21
Improvement to new CI Process


mjp41:
- [x] Enable more types of cached build, e.g. RelWithDbgInfo (https://github.com/microsoft/verona/pull/197#discussion_r457989495)
- [x] Expose more flags automatically (https://github.com/microsoft/verona/pull/197#discussion_r457989949)
- [x] Fix Mac CI to not use hardcoded 4 parallism
- [x] Automatically determine `-j` option for ctest and pass as part of the test target.
- [x] Add CI pipelines to check that building LLVM directly works (not currently tested).
- [x] Add quiet option to the not print the subtasks in details.
- [x] `compile_commands.json` doesn't get generated any more
- [x] LLVM downloads always happen again (even though file is still there) -> `URL_HASH` not set
- [x] C++ standard reverted to C++11
- [x] Failed to compile LLVM with `-DVERONA_DOWNLOAD_LLVM=Off` on Linux Arch
- [x] Cannot use `GOOD_COMMIT` flag with CI/Nightly builds and get a different LLVM build
- [x] `-DEXTRA_VERONA_CMAKE_FLAGS=` doesn't seem to work
- [x] `-DCMAKE_BUILD_TYPE=Debug` doesn't propagate to Verona
- [x] Multiple builds per invocation of `ninja` on Linux, always building from scratch when passing compiler flags
- [x] If LLVM archive is not available (new hash, etc) the failure is cryptic: add an error message to that
- [x] Both `interpreter` and `compiler` were left out of the build, building only on `make check` (#265)

rengolin:
- [ ] `compile_commands.json` don't get generated any more
- [ ] Removing `CMakeCache.txt` and running CMake again makes `ninja` download LLVM again (even though it's still there)

rengolin:
Build with existing LLVM also doesn't work:
```
verona/build$ rm -rf *; cmake .. -G Ninja -DVERONA_LLVM_LOCATION=../external/llvm-project/install/
-- The CXX compiler identification is GNU 10.1.0
-- Check for working CXX compiler: /usr/bin/c++
-- Check for working CXX compiler: /usr/bin/c++ - works
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Detecting CXX compile features
-- Detecting CXX compile features - done
-- Version: 6.2.0
-- Build type: 
-- CXX_STANDARD: 11
...
[...]

PR #200, opened 2020-07-21
Verona CMake options inside ExternalProject

rengolin:
Both C++17 and compile_options were being ignored, as Verona is now an
external project.

Addresses one item of #198


mjp41:
So I can see this add C++17 to Clang, but the other change is just moving the settings to the verona-lang project to a different place?  The recursive CMake trick might be confusing here.

rengolin:
Because the sub-project (Verona) is created before that settings is added, it doesn't get it. Moving them before creating both LLVM and Verona external projects didn't help either. This is the only thing that worked.

I'm not claiming I know how it works and this is the best solution, but it's the only one I could make it work.

The problem for me is that I don't get editor support and the build fails with C++17 errors about `constexpr` use. So I can't do anything if this isn't fixed.

If we find a better way later, I'd be happy to review, but this doesn't seem half bad to me to stay in until we find a better way.

mjp41:
It should get it on the recursive call, at least that was my intent, but if this works, and previously it didn't go for it.

PR #201, opened 2020-07-21
Move definition of clangformat target to toplevel

plietar:
With the introduction of ExternalProject, the clangformat target was
being picked for the outer build using a special `VERONA_MINIMAL_CMAKE`
flag.

In addition to being overly complicated, this did not expose the target
for builds that manually specified a `-DVERONA_LLVM_LOCATION` flag.

We can fix that by moving the target to the top-level CMakeLists file,
making it independent of the LLVM build strategy.


rengolin:
Similar issue in #200

Issue #106, opened 2020-05-07
Design a Verona MLIR dialect


rengolin:
Related to #104, this task is about lowering the operations, constructs (like if, when, where), lexical scopes, lambda logic, future registration, dispatch and completion, etc.

This can be a mix of existing dialects (std, loop) and Verona-specific constructs, but anything that touches specific Verona types must be Verona-specific nodes as well, so likely to be mostly Verona anyway.

Acceptance criteria:
* A consensus on the representation of each language construct
* A description of the steps needed to take from opaque MLIR to Verona dialect
* Examples in MLIR for each of the constructs and variations, with expected semantics

Not included:
* Implementation of the partial conversion process (has to be concurrent with #105)

plietar:
Two sources for the list of operations we need is the existing IR and Bytecode.
https://github.com/microsoft/verona/blob/master/src/compiler/ir/ir.h
https://github.com/microsoft/verona/blob/master/src/interpreter/bytecode.h

In the existing prototype, a number of runtime features were exposed as builtin functions rather than special syntax and ir statements (they do have special bytecode instructions though). I don’t know what makes more in MLIR.

mjp41:
I would be keen to keep them as builtin functions, and only do something special when lowering to LLVM.  I think this will help with experimentation (as long as the type system can express their requirements).

mjp41:
I believe `ponyc` allows builtins to be given an LLVM command to replace them.  This is very useful there for expanding the core libraries without doing anything at the front-end.

https://github.com/ponylang/ponyc/blob/eff764fb0173a1629d37ddc9a817d8fded5f36bd/packages/builtin/float.pony#L82

davidchisnall:
Pony has a much weaker isolation model for foreign code, so I'm a bit nervous about accidentally introducing an `unsafe` keyword via a mechanism like that.  I think I'd rather provide a small (well-documented) set of compiler builtins that the standard library can depend on.  Ideally this should be sufficient to allow different compiler implementations to use the same standard library (and vice versa), though I don't consider that to be anything that we need to care about in the short term (OpenJDK and .NET 5 both went through this process, which was quite a useful exercise for both but not something that should block anything else for us).

rengolin:
In MLIR, operations and function calls are treated equal (ish). For example, you can have an unnamed/undeclared "operator" by having its name in quotes and operands laid out like a function:

    %0 = "verona.myop" (%a, %b) : (i64, f32) -> i32

[...]

PR #202, opened 2020-07-21
Introduce a few operations to the Verona dialect.

plietar:
The level of abstraction described by the dialect matches roughly the the existing bytecode VM and Latex/Coq formalisms. At some later point, we may want to introduce a "low-Verona" dialect, with more primitive and explicit operations.

Basic verification of the operations is provided, mostly aimed at catching bugs in the frontend (eg. referencing an inexistent class) rather than type errors. All operations accept and return any Verona type. Language aware type-checking will be added later as a custom verifier / pass.


plietar:
I will add a couple of tests for the verifiers tomorrow.

src/mlir/dialect/VeronaOps.cc
```diff
@@ -4,7 +4,114 @@
...
+}
+
+/**
+ * AllocateRegionOp and AllocateObjectOp share similar features, which we verify
+ * here:
+ * - The class referenced by the operation must exist.
+ * - There must be as many field names as there are operands.
+ * - The set of field names must match those specified in the class definition (TODO: not implemented yet)
+ *
+ * TODO: Rather than a templated function, should this be a trait?
```
rengolin, line 52: This looks like a pretty standard verify pass, to me. A trait is more about the operation itself, not the state of the arguments outside the operation.

src/mlir/dialect/VeronaOps.cc
```diff
@@ -4,7 +4,114 @@
...
+  return verifyAllocationOp(op);
+}
+
+static LogicalResult verify(verona::ClassOp classOp)
+{
+  Block* body = classOp.getBody();
+  if (body->getNumArguments() != 0)
+    return classOp.emitOpError("expected body to have no arguments");
+
+  // Verify that classes contain only fields, and that there are no duplicates.
```
rengolin, line 93: Where do the class methods go?

src/mlir/dialect/VeronaOps.td
```diff
@@ -5,36 +5,240 @@
...
+      %x and %y.
+
+      ```mlir
+        %w = verona.new_region @C["f", "g"](%x, %y : !verona.A, !verona.B) : !verona.C
+      ```
+
+      TODO: the syntax is a little ugly, something that looks like a record
+      eg. `verona.new_region @C[ "f" = %x : !verona.A, "g" = %y : !verona.B ]`,
+      binding field names to SSA values might be nicer. Probably needs a custom
+      parser though.
```
rengolin, line 128: MLIR's default syntax is not super consistent. We'll have to use a custom parser at some point, but it should be fine now for the first pass.

src/mlir/dialect/VeronaOps.td
```diff
@@ -5,36 +5,240 @@
...
     }];
+
+    let arguments = (ins
+      SymbolRefAttr:$class_,
```
rengolin, line 132: Why the underscore in `class_`?

testsuite/mlir/mlir-parse/dummy.mlir
```diff
@@ -1,18 +1,26 @@
...
-  func @bar(%arg0: !verona.U64) -> !type<"U64 & imm"> {
-    // CHECK: %[[res:[0-9]+]] = verona.foo %arg0 : !verona.U64
-    %res = verona.foo %arg0 : !verona.U64
-    // CHECK: %[[cast:[0-9]+]] = "verona.cast"(%[[res]]) : (!verona.U64) -> !verona.S64
-    %cast = "verona.cast"(%res) : (!verona.U64) -> !verona.S64
-    // CHECK: %[[foo:[0-9]+]] = "verona.test"(%[[cast]]) : (!verona.S64) -> !type<"U64 & imm">
-    %foo = "verona.test"(%cast) : (!verona.S64) -> (!type<"U64 & imm">)
-    // CHECK: return %[[foo]] : !type<"U64 & imm">
-    return %foo : !type<"U64 & imm">
+  verona.class @C {
```
rengolin, line 5: It would be nice to have some Verona code that translates into this, so we have an idea of what in Verona translates to what in MLIR. At least for this one example, not necessarily for all the other tests (though it is always helpful).

testsuite/mlir/mlir-parse/dummy/out.mlir
```diff
@@ -1,10 +1,20 @@
...
-  func @bar(%arg0: !verona.U64) -> !type<"U64 & imm"> {
-    %0 = verona.foo %arg0 : !verona.U64
-    %1 = "verona.cast"(%0) : (!verona.U64) -> !verona.S64
-    %2 = "verona.test"(%1) : (!verona.S64) -> !type<"U64 & imm">
-    return %2 : !type<"U64 & imm">
+  verona.class @C {
+  }
+  verona.class @D {
+    verona.field "f" : !verona.U64
+    verona.field "g" : !verona.U64
```
rengolin, line 8: The previous test checks both `U64` and `S64`. We should keep both in the new test, too.

rengolin:
Note: this starts to implement #106.

src/mlir/dialect/VeronaOps.cc
plietar: There will eventually be a `verona.method` operation that will be part of classes. I have not included them yet because I want to figure out a few things first that methods need (generics, where-clauses, ...).

src/mlir/dialect/VeronaOps.td
plietar: Without the underscore TableGen generates identifiers called `class`, which is invalid in C++.
We'll need a general strategy for naming things that conflict with C++ keywords, and I find a trailing underscore the least noisy option.

testsuite/mlir/mlir-parse/dummy/out.mlir
plietar: Done.

src/mlir/dialect/VeronaOps.cc
rengolin: Right, I'd leave a TODO here, then.

Also, I don't know how MLIR handles functions inside regions. An alternative to functions inside classes would be name mangling of functions outside the classes.

src/mlir/dialect/VeronaOps.td
rengolin: Blergh, ok. I don't like underscores as they're too easy to add irregularly and makes things confusing, but it should be fine for now.

testsuite/mlir/mlir-parse/dialect.mlir
```diff
@@ -0,0 +1,51 @@
...
+// }
+// ```
+
+module {
+  verona.class @C {
+  }
+
+  verona.class @D {
+    verona.field "f" : !verona.U64
+    verona.field "g" : !verona.U64
```
rengolin, line 33: Don't forget the `S64` type test. Also helps looking at the lowering below.

src/mlir/dialect/VeronaOps.cc
plietar: Will do.

As far as I understand there is no “MLIR functions”, only Standard/core dialect functions. We can make a `verona.method` operation and give it whatever semantics we want. At some point I suspect we’ll lower everything to top level `func` with mangled names.

src/mlir/dialect/VeronaOps.td
plietar: Do you know if llvm or clang have a convention for dealing with this? A quick gren through the source code didn’t reveal anything obvious.

src/mlir/dialect/VeronaOps.cc
rengolin: Right, exactly. MLIR core is basically just the idea of SSA values, types, instructions and regions. The implementation of those are all in dialects, including the "standard" one. The benefit in using the standard ones (including loop, affine, etc) is twofold: first, the conversion to LLVM dialect, then LLVM IR, is done already; second, there are existing passes on those dialects that simplify and optimise, even if they have unknown nodes within.

I see the following lowering stages, long term:
1. AST -> Verona dialect with opaque types
2. Type check and inference -> Verona dialect with Verona types
3. Verona specific optimisations that use the dialect to infer high level semantics
4. Sequence of partial lowering to other dialects, dialect specific optimisations (affine, poly, loop)
5. Full lowering to LLVM dialect, then straight to LLVM IR
6. A select pipeline of LLVM passes, for different optimisation levels (perf, space, debug)
7. Code generation (MIR, ASM, OBJ, link, EXE)

We could have a Verona specific "method" construct, but we should always keep in mind that we will *have* to convert it down to an actual MLIR function with actual arguments and return values. The mangling could very well be done at that stage.

I don't have a strong opinion if this is better to be done at the AST level or MLIR level and I think assuming MLIR is a safer bet with the information we have right now.

src/mlir/dialect/VeronaOps.td
rengolin: Well, you could call it "ClassName" or something, giving that this is the attribute, not the class object.

plietar: Changed to `class_name`.

plietar:
Closing as CI and opening a new PR, as CI is being stroppy

Issue #203, opened 2020-07-23
Implement while/for loops in MLIR


rengolin:
With the dialect effort starting up, it's the right time to start using Verona dialect on the MLIR lowering, at least for the high level structure.

In Verona, `for` loops are syntactic sugar for `while` loops in the following way:
```
for (a in x) {
   foo(a);
}
```
equals:
```
{
  var $iter = x;
  while ($iter.has_value())
  {
    a = $iter.apply();
    $iter.next();
    foo(a);
  }
}
```

rengolin:
In the end, this did not bring the benefits we were expecting, so I'll revert the patch creating loop regions, and continue working on loops as control flow. This will remove the new operations (loop_exit, loop_return, continue, break) and the region based approach.

PR #204, opened 2020-07-23
[NFC] Test for nested while loops

rengolin:
Left over from last implementation.

This will change soon to use a custom dialect operation for while loops when we lower for loops as while loops, but it's good to have the test here to make sure we can still parse the AST correctly into the right MLIR structure. (See #203).


rengolin:
Ping

PR #205, opened 2020-07-27
Bail out on MLIR parsing errors.

plietar:
Otherwise we would be proceeding with a null pointer.

Additionally parseSourceFile already runs the verifier, so no point in running it again ourselves.

### DIFF
--- a/src/mlir/generator.cc
+++ b/src/mlir/generator.cc
@@ -68,17 +68,14 @@ namespace mlir::verona
       return runtimeError(
         "Cannot open file " + filename + ": " + err.message());
 
-    // Setup source manager and parse
+    // Setup source manager and parse the input. This includes verification of
+    // the IR.
     llvm::SourceMgr sourceMgr;
     sourceMgr.AddNewSourceBuffer(std::move(*srcOrErr), llvm::SMLoc());
     module = mlir::parseSourceFile(sourceMgr, builder.getContext());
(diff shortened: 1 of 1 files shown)


src/mlir/generator.cc
```diff
@@ -68,17 +68,14 @@ namespace mlir::verona
       return runtimeError(
         "Cannot open file " + filename + ": " + err.message());
 
-    // Setup source manager and parse
+    // Setup source manager and parse the input. This includes verification of
+    // the IR.
     llvm::SourceMgr sourceMgr;
     sourceMgr.AddNewSourceBuffer(std::move(*srcOrErr), llvm::SMLoc());
     module = mlir::parseSourceFile(sourceMgr, builder.getContext());
+    if (!module)
```
rengolin, line 76: Dumping the IR on failure is quite helpful, especially at this stage.

Also, no reason to change the error message.

plietar: We don't have access to the IR is any of the verifiers fail, `parseSourceFile` return null. There's nothing we could dump.

I've changed the error message intenitonally because this case happens on both parsing and verification errors, and there's not much we can do to distinguish the two. "Load" felt like a generalization of the two steps.

rengolin: Right, this is reading MLIR...

PR #206, opened 2020-07-27
Initial type checking of MLIR

plietar:
This adds join, meet and capability types, with the usual algebraic rules for subtyping (distributivity is not supported yet).

A verifier is added to `verona.copy`, checking that the input is a subtype of the output. A `verona.return` opcode is added that behaves just like the standard `return`, but allows for subsumption: the return value must be a subtype of the function return type.

Other operations have more complicated typing rules, so these are left for a later PR.


plietar:
As usual, GitHub can't cope with chained PRs. This depends on #202.

davidchisnall:
FYI: There is a process for doing chained PRs on GitHub.  You need to set the target of the PR to the version that this depends on, not the main branch, and then merge PRs from leave PRs inwards until you hit trunk.

rengolin:
Also, please add descriptive comments to the methods and classes as well as descriptions to the table-gen operations and types.

plietar:
@davidchisnall AFAIK that only works if the base PR is on a branch on the upstream repo, but I've put the PR on my fork.

 @rengolin Yes of course, this is not intended to be ready, but more to get the discussion going, as there are a few points I want to discuss.

-----

## Should we use the meet/join terminology, or intersection/union, or somethings else?
meet/join is nice and short, and both names have the same number of characters.
intersection/union is probably more familiar to non- mathematicians / type-theorists.

## Should `verona.meet<>` be allowed, with no elements?
At the moment, both this syntax, and the equivalent `verona.top` can be parsed. It always gets printed as `verona.top`.
Same applies to `verona.join<>`/`verona.bottom`.

## Should Verona types be closed?
In other words, can you have a `verona.meet<f32>`, where `f32` is the core MLIR type, or should it only allow Verona types.

## Should type-checking happen in MLIR verifiers.
It currently is, because that was the easiest route.
I'd however be inclined to have our own traversal of the IR, distinct from the builtin verifier:
- The base MLIR parsing functions (eg. `parseSourceFile`) return null when verifiers fail, so we can't even dump the module (See discussion in #205).
- Verifiers are (AFAIK) not configurable. We want to initial type checking to be quite permissive (eg. allow elided types), but later stages should be stricter.
- Verifiers stop at the first error. We instead will want to catch and report as many errors as possible.

In short, I'd suggest verifiers are only used to catch bugs in the frontend, not programmer errors. This moves typechecking and region checking to a separate custom traversal.

## Should the type hierachy be stored in MLIR types.
At the moment, "composed" types, like meet/join, store an array of `mlir::Type`. An alternative would be to define a custom storage and representation, independent of MLIR, and have a thin mlir::Type to wrap it at the top-level only. This would be similar to the LLVM dialect type, `mlir::LLVM::LLVMType`, which only stores a pointer to a `llvm::Type*`.

rengolin:
> ## Should we use the meet/join terminology, or intersection/union, or somethings else?
> meet/join is nice and short, and both names have the same number of characters.
> intersection/union is probably more familiar to non- mathematicians / type-theorists.

This is not supposed to be be exposed to the average user, it's just an IR construct, so I don't think the name matters much if they're clearly documented what their meanings are.

> ## Should `verona.meet<>` be allowed, with no elements?
> At the moment, both this syntax, and the equivalent `verona.top` can be parsed. It always gets printed as `verona.top`.
> Same applies to `verona.join<>`/`verona.bottom`.

I would avoid redundancy in the IR representation.

> ## Should Verona types be closed?
> In other words, can you have a `verona.meet<f32>`, where `f32` is the core MLIR type, or should it only allow Verona types.

It's usually profitable if dialects can interoperate directly with other dialects, but the type system is particularly sensitive. We want to use generic types (like `f32`), for example, when calling an external function that returns it, but set operations on those types won't be equivalent to Verona types (like `F32`) unless we make them so.

A trivial solution to that problem is to create automatic casts from standard types to Verona types on return, so that we're always operating on Verona types, then remove them once we lower the Verona dialect into standard ones.

A less trivial (but hopefully pretty simple) solution would be to use the MLIR standard dialect types for our "core" types, so no conversions are needed, but we need to make sure the semantics is the same (or that we wrap them safely).

> ## Should type-checking happen in MLIR verifiers.
> It currently is, because that was the easiest route.
> I'd however be inclined to have our own traversal of the IR, distinct from the builtin verifier:
> 
> * The base MLIR parsing functions (eg. `parseSourceFile`) return null when verifiers fail, so we can't even dump the module (See discussion in #205).

This is for MLIR sources, not Verona sources. However, the MLIR could be in *any* state, including opaque representation with opaque Verona types, and that has to pass, too.

> * Verifiers are (AFAIK) not configurable. We want to initial type checking to be quite permissive (eg. allow elided types), but later stages should be stricter.

The idea we discussed was to make Verona types more flexible and then the verifier won't break on type problems until we have done the type check (by making pre-check types all compatible). Post-check types must fail the verifier if broken.

> * Verifiers stop at the first error. We instead will want to catch and report as many errors as possible.

Overcoming that problem may be more complex than needed, unless we want to completely bypass the verifier and do our own verify passes altogether (I don't think we should).

There are other ways to do incremental compilation, for instance, splitting the IR into functions or regions and only verifying those.

> In short, I'd suggest verifiers are only used to catch bugs in the frontend, not programmer errors. This moves typechecking and region checking to a separate custom traversal.

Bugs in the front-end will be caught by the AST generator. We should not have front-end bugs in MLIR. IR verification is also *very* important to guarantee transformation semantics of the passes we add, including type inference.

The LLVM IR lowering will run verify passes on the lowering stage and we don't want to only get errors then.


> ## Should the type hierachy be stored in MLIR types.
> At the moment, "composed" types, like meet/join, store an array of `mlir::Type`. An alternative would be to define a custom storage and representation, independent of MLIR, and have a thin mlir::Type to wrap it at the top-level only. This would be similar to the LLVM dialect type, `mlir::LLVM::LLVMType`, which only stores a pointer to a `llvm::Type*`.

While this works, it looks to me like a hack to make it work. LLVM type system was well defined and had its own semantics well before MLIR, and MLIR has a simplistic type system which doesn't wrap into LLVM's semantics too well.

It's possible that we'll have to do a similar trick, but I'd rather try in pure-MLIR type system first, and only go there when we have no other option.

It would, however, make the type verification in the naive verifiers a lot simpler by always returning `true` respective of the contained types.

plietar:
So I've moved typechecking out of the verifier into its own traversal. It uses an interface for operations to opt-in to custom type-checking. The typechecker traverses the whole tree looking for operations that implement that interface.

I'm not really happy with adding more stuff to generator, but I think this needs a bigger refactor later. In particular I think it should only be responsible for AST/source- > in-memory MLIR, not for loading and emitting textual MLIR, nor lowering to LLVM. That should probably be fixed in a later PR though.

plietar:
Interestingly, there is a proposal for an MLIR-native type hierachy for the LLVM dialect: https://llvm.discourse.group/t/rfc-first-party-support-for-llvm-types/1295
What I'm doing ressembles it quite a lot.

plietar:
> I'm not really happy with adding more stuff to generator, but I think this needs a bigger refactor later. In particular I think it should only be responsible for AST/source- > in-memory MLIR, not for loading and emitting textual MLIR, nor lowering to LLVM. That should probably be fixed in a later PR though.

I've created #215 to fix this and based this PR on it.

rengolin:
> Interestingly, there is a proposal for an MLIR-native type hierachy for the LLVM dialect: https://llvm.discourse.group/t/rfc-first-party-support-for-llvm-types/1295
> What I'm doing ressembles it quite a lot.

I think the main difference there is that the LLVM types are the end-point, while for Verona types are transient. We *want* to convert to MLIR types soon, to benefit from generic optimisations and we *need* to convert to LLVM types, for lowering.

We'll eventually translate Verona types into two main categories: things that have size and ISA semantics (int, float, bool, struct) and things that only have language semantics (iso, mut, etc), and will be lowered as thunks with the code, operating on sized types. Regions that only define the latter types will be versioned to work with the former types soon enough.

Even if we redesign our pipeline to only apply Verona-specific optimisations, we still have to lower them to LLVM types, which will have to be concrete, as this is the only thing LLVM will understand. But the sooner we get there, the more we'll benefit from translating our MLIR constructs into existing (optimiseable) dialects (like linalg, affine, async).

So, I believe our complex somewhat opaque type system should be transient, mostly for type inference and checks and Verona-specific passes. Once we start converting Verona dialect into others, we'll need to make use of MLIR/LLVM types directly. A problem that the LLVM type system doesn't have.

plietar:
Alright, I think this is (finally) ready for review now.

src/mlir/dialect/Typechecker.cc
```diff
@@ -0,0 +1,162 @@
...
+
+  namespace detail
+  {
+    /// A Rule wraps a callable object with arguments of type `Left` and
+    /// `Right`.
+    ///
+    /// Whenever the Rule is applied to arbitrary Type values, the left-hand
+    /// side and right-hand side Type are respectively dyncasted to `Left` and
+    /// `Right`.
+    //
```
rengolin, line 40: Nit: `///` or this will break docs.

src/mlir/dialect/Typechecker.cc
```diff
@@ -0,0 +1,162 @@
...
+    });
+
+  bool isSubtype(Type lhs, Type rhs)
+  {
+    // TODO: Applying the syntax-directed rules enumerated above is not enough
+    // to recognize certain relationships, such as distributivity:
+    //
+    // (A & (B | C)) <: (A & B) | (A & C)
+    //
+    // For this to work, we would need to apply normalization to the types
```
rengolin, line 146: We definitely want to normalise types, or chasing patterns can get quite complicated.

src/mlir/dialect/VeronaOps.cc
```diff
@@ -113,6 +114,25 @@ static LogicalResult verify(verona::ClassOp classOp)
...
+    return checkSubtype(getLoc(), input().getType(), output().getType());
+  }
+
+  LogicalResult ReturnOp::typecheck()
+  {
+    auto fn = cast<FuncOp>(getParentOp());
+    const auto& results = fn.getType().getResults();
+    if (results.size() != 1)
+    {
+      // TODO: The `verona.return` op could easily be extended to support
```
rengolin, line 128: `return` can actually return 0 and multiple values.

The generator has a class `ReturnValue` that handles that, too.

src/mlir/dialect/VeronaOps.td
```diff
@@ -183,6 +186,42 @@ def Verona_AllocateObjectOp : Verona_Op<"new_object"> {
     }];
 }
 
+def Verona_ReturnOp : Verona_Op<"return", [
```
rengolin, line 189: I'm still not convinced we need a special return. Sub-typing can be done as a pattern matching of cast+op.

Sure, it's easier for type checking if we bundle everything in one op, but it's harder for optimising and lowering if we have to treat standard operations as special.

src/mlir/dialect/VeronaTypes.cc
```diff
@@ -0,0 +1,334 @@
...
+
+  Type parseVeronaType(DialectAsmParser& parser)
+  {
+    MLIRContext* ctx = parser.getBuilder().getContext();
+
+    StringRef keyword;
+    if (parser.parseKeyword(&keyword))
+      return Type();
+
+    if (keyword == "meet")
```
rengolin, line 226: In the parser and generator, we're using the `str2tag` to make that into a trivial integer switch.

src/mlir/dialect/VeronaTypes.h
```diff
@@ -0,0 +1,108 @@
...
+    size_t getWidth() const;
+    bool getSign() const;
+
+    static bool kindof(unsigned kind)
+    {
+      return kind == VeronaTypes::Integer;
+    }
+  };
+
+  enum class Capability
```
rengolin, line 89: Why is this enum here and not inside the storage class, like Integers?

src/mlir/dialect/Typechecker.cc
plietar: Yes, I've got a follow up PR to do just that. I just thought it would be better to break things up.

src/mlir/dialect/VeronaTypes.h
plietar: Unlike Integer's `SignType` (which ends up being exposed as a bool anyway), the capability inside the type is very much part of its public interface. In various places during typechecking we'll need to look at the capability to make decisions.

rengolin: Integer sign and width are also part of their public interfaces, that you expose using `getImpl()`. What's the difference?

src/mlir/dialect/VeronaTypes.cc
plietar: Interesting, thanks. I don't really want to tie the `dialect/` folder to the peg library, and duplicating the implementation feels overkill for just this switch statement.

There's also `StringSwitch` from LLVM (which I haven't used before), but it seems to require values to be associated to each case, not blocks of code. Obviously function pointers and lambdas are values, so it could work, but it feels a bit too complicated.

rengolin: > Interesting, thanks. I don't really want to tie the `dialect/` folder to the peg library, and duplicating the implementation feels overkill for just this switch statement.

Good point.

> There's also `StringSwitch` from LLVM (which I haven't used before), but it seems to require values to be associated to each case, not blocks of code. Obviously function pointers and lambdas are values, so it could work, but it feels a bit too complicated.

You could outline a function that turns strings into enum values and use the switch on those values.

src/mlir/dialect/VeronaTypes.h
plietar: The sign is exposed as a boolean, so the definition of its enum, `SignType`, doesn't need to be public.
For capability, we definitely want to expose it as an enum, not as a (set of) boolean.

src/mlir/dialect/VeronaOps.cc
plietar: Done.

src/mlir/dialect/Typechecker.cc
plietar: Done.

src/mlir/dialect/VeronaTypes.h
rengolin: What about integer width? That's not a boolean either.

plietar: I'm sorry I don't understand your point.

You were asking why the `Capability` enum is defined in the header and not in the source file, as part of the storage class. The answer is because it is part of method signatures of `CapabilityType` (both `get` and `getCapability`). Therefore it needs to be exposed to everyone that uses `CapabilityType`, hence must be in the header.

All methods on `IntegerType` use standard C++ types. The sign is exposed as a boolean and the width as an integer. `SignType` is an implementation detail of the storage class, hence it can happily live in the `.cpp` file.

rengolin: D'oh! Now I get it, sorry. Friday.

src/mlir/dialect/VeronaTypes.cc
rengolin: Are you doing the `StringSwitch` on this PR or leaving for later?

rengolin: anyway is fine. This PR looks good to me now, thanks!

PR #207, opened 2020-07-27
Introduce a few operations to the Verona dialect.

plietar:
The level of abstraction described by the dialect matches roughly the the existing bytecode VM and Latex/Coq formalisms. At some later point, we may want to introduce a "low-Verona" dialect, with more primitive and explicit operations.

Basic verification of the operations is provided, mostly aimed at catching bugs in the frontend (eg. referencing an inexistent class) rather than type errors. All operations accept and return any Verona type. Language aware type-checking will be added later as a custom verifier / pass.


rengolin:
Github -> Devops integration is broken, but the test passed. I'll merge this now, so Paul can continue his work, and look what the cause of the problem is in parallel.

PR #212, opened 2020-07-28
Fix iterator returned from ObjectMap insert

Theodus:


Theodus:
CI: https://dev.azure.com/ProjectVeronaCI/Project%20Verona/_build/results?buildId=1314&view=results

mjp41:
https://github.com/microsoft/verona/blob/6f276efa676f4b8d721aed68fffdaa67b11c6a77/src/rt/object/object.h#L222

Can you make the object map use this for the key. Then it should function correctly with systematic testing.

Theodus:
@mjp41 that would require any set to be a map with systematic testing. Also, the `id()` wouldn't have the lower bits available for storing the probe length. Would it be enough to just to insert into the table based on `hash(key->id())` instead of `hash(key)`?

rengolin:
Proposed changes from Matt implemented in #216

PR #213, opened 2020-07-29
Handle macOS backtrace symbol format

Theodus:
Unexpected symbol format would cause assertions on macOS to hang the process indefinitely rather than dump the flight recorder.


Theodus:
Closing,

PR #215, opened 2020-07-29
Add a driver class responsible for running passes.

plietar:
This moves the MLIR/LLVM pipeline logic out of the generator and into
a new class. The Generator is now only responsible for AST->MLIR
conversion.

This will be helpful as we start adding more passes, and the logic to
handle those becomes more complicated.

### DIFF
--- a/src/mlir/driver.h
+++ b/src/mlir/driver.h
@@ -1,0 +1,62 @@
+#pragma once
+
+#include "ast/ast.h"
+#include "error.h"
+#include "mlir/IR/Module.h"
+#include "mlir/Pass/PassManager.h"
+
+namespace mlir::verona
+{
+  /**
+   * Main compiler API.
+   *
+   * The driver is used by calling three methods in succession:
+   * - `readXXX` is used to load the MLIR module. It can either be loaded from
+   *    textual MLIR or from the AST.
+   * - `process` runs the MLIR lowering pipeline. Passes are selected by the
+   *    Driver's constructor's arguments.
+   * - `emitXXX` writes the result to a text file, either in MLIR or in LLVM IR
+   *    format.
+   *
+   * For now, the error handling is crude and needs proper consideration,
+   * especially aggregating all errors and context before sending it back to
+   * the public API callers.
+   */
+  class Driver
+  {
+  public:
+    Driver(unsigned optLevel = 0, bool lowerToLLVM = false);
+
+    // TODO: add a readSource function that parses Verona source code.
+    // this might be more thinking about the error API of the Driver.
+
+    /// Lower an AST into an MLIR module, which is loaded in the driver.
+    llvm::Error readAST(const ::ast::Ast& ast);
+
+    /// Read textual MLIR into the driver's module.
+    llvm::Error readMLIR(const std::string& filename);
+
+    // Run the MLIR pipeline on the module. This uses the passes that were
+    // configured by the driver's constructor.
+    llvm::Error process();
+
+    /// Emit the module as textual LLVM IR.
+    /// This will fail if the module is not in LLVM dialect yet.
+    llvm::Error emitLLVM(const llvm::StringRef filename);
+
+    /// Emit the module as textual MLIR.
+    llvm::Error emitMLIR(const llvm::StringRef filename);
+
+  private:
+    /// MLIR context.
+    mlir::MLIRContext context;
+
+    /// MLIR module.
+    /// It gets modified as the driver progresses through its passes.
+    mlir::OwningModuleRef module;
+
+    /// MLIR Pass Manager
+    /// It gets configured by the constructor based on the provided arguments.
+    mlir::PassManager passManager;
+  };
+}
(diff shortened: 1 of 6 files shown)


src/mlir/driver.h
```diff
@@ -0,0 +1,62 @@
...
+    // TODO: add a readSource function that parses Verona source code.
+    // this might be more thinking about the error API of the Driver.
+
+    /// Lower an AST into an MLIR module, which is loaded in the driver.
+    llvm::Error readAST(const ::ast::Ast& ast);
+
+    /// Read textual MLIR into the driver's module.
+    llvm::Error readMLIR(const std::string& filename);
+
+    // Run the MLIR pipeline on the module. This uses the passes that were
```
rengolin, line 39: Nit: triple slashes here, too.

src/mlir/generator.h
```diff
@@ -104,55 +104,39 @@ namespace mlir::verona
...
-    /// Transform the Verona dialect into LLVM IR.
-    llvm::Error emitLLVM(const llvm::StringRef filename, unsigned optLevel = 0);
-
     using Types = llvm::SmallVector<mlir::Type, 4>;
 
-  private:
     /// MLIR module.
     mlir::OwningModuleRef module;
+    /// MLIR context.
+    mlir::MLIRContext* context;
```
rengolin, line 137: Aren't you leaking this context, now that it's just a pointer and you don't destroy it?

src/mlir/driver.h
```diff
@@ -0,0 +1,62 @@
...
+   *    format.
+   *
+   * For now, the error handling is crude and needs proper consideration,
+   * especially aggregating all errors and context before sending it back to
+   * the public API callers.
+   */
+  class Driver
+  {
+  public:
+    Driver(unsigned optLevel = 0, bool lowerToLLVM = false);
```
rengolin, line 28: Having `lowerToLLVM` as a flag here will create problems if we don't set it and then call `emitLLVM`. This is a side-effect of separating `process` in its own step.

I recommend making `process` a private method, and making `emitMLIR` and `emitLLVM` call it with different values of the flag, so the public API is just: `readXXX -> emitYYY`.

src/mlir/generator.h
plietar: The context is owned by the driver (or whoever uses the generator). It's their responsibility to manage it.

I'll add a note to the constructor of Generator.

src/mlir/driver.h
plietar: I think being able to `emitMLIR` with the lowering to LLVM is still a useful feature. It allows us to view the LLVM MLIR dialect.

I could store the parameters and delay the construction of the pass manager to the `emitXXX` method, and `lowerToLLVM` is overriden in the `emitLLVM` case. Delaying the construction of the pass manager also means I can turn the settings into methods on the driver, which is nicer I guess.

Does that sound good?

src/mlir/generator.h
rengolin: Ah, makes sense!

src/mlir/driver.h
rengolin: What if you don't pass `lowerToLLVM` and call `emitLLVM`?

That's not a good long term approach. The MLIR will change considerably throughout the pipeline and having a flag for one particular dialect or point in time doesn't make sense.

The current meaning of `emitMLIR` is "the first lowering", after AST, before optimisation. For now, `emitLLVM` means "LLVM IR". Adding more "special states" will make picking places within the pipeline harder and less meaningful.

The expected meaning of `emitMLIR` will only be truly complete when we add optimisation options (see LLVM's `opt` tool), in which case you can add `-lower-to-llvm-dialect` as an option.

Given that we don't have neither LLVM dialect nor LLVM IR, I wouldn't take this decision in this patch, and do that work on another one. For now, I'd go with the simple semantics of "first" and "last" stops for MLIR and LLVM respectively.

plietar: > What if you don't pass lowerToLLVM and call emitLLVM?

What I’m proposing is that emitLLVM ignores the value of lowerToLLVM and always includes that pass.

I don’t think emitMLIR should be tided to a particular spot in the pipeline. Otherwise we’ll end up with loads of ‘emitInferredMLIR’, 'emitOptimizedMLIR’, etc methods. MLIR in the function name essentially refers to the output text format, like emitLLVM says we’re using the LLVM text format. The set of passes depends on how the driver is configured, not on which emit method is called.

Similarly, in the future we can have the driver load MLIR source at various points in the lowering. We don’t want a ‘readMLIR’ and ‘readTypedMLIR’ and ‘readExplicitRegionsMLIR’ (a hypothetical intermediary dialect). We should have ‘readMLIR’, and configuration knobs to enable various passes.

In any case, it’s very early in the design of this class, so any decision will probably be overridden later when we have more context. I’m happy to stick with the existing semantics if that’s preferable.

rengolin: I think this is the wrong time to have that discussion. We largely agree with each other, so why don't we just remove completely the "emit LLVM" functionality and just emit MLIR for now.

Once we actually can lower to either LLVM dialect or IR, we can revisit this option (and the driver will look nothing like it does now).

plietar: Alright, sounds good. I agree this conversation is a bit premature anyway.

Done the change.

PR #216, opened 2020-07-30
Insert into ObjectMap based on hash of object id

Theodus:


src/rt/ds/hashmap.h
```diff
@@ -308,7 +308,7 @@ namespace verona::rt
 
       assert(key_of(entry) != 0);
       const auto key = unmark_key(key_of(entry));
-      const auto hash = bits::hash((void*)key);
+      const auto hash = bits::hash(((const Object*)key)->id());
```
rengolin, line 311: I assume key is guaranteed to be an Object, then a c++ static cast would be more appropriate.

Is there any compile time expression (template parameter, type trait, etc) checking to make sure it is?

src/rt/ds/hashmap.h
```diff
@@ -355,7 +355,9 @@ namespace verona::rt
 
       // Maximum probe length reached, resize and retry.
       resize(alloc);
-      return insert(alloc, std::forward<E>(entry));
+      // Entry may have been swapped prior to resize.
+      insert(alloc, std::forward<E>(entry));
+      return std::make_pair(true, find((const KeyType*)key));
```
rengolin, line 360: This will incur in two map searches, right? Perhaps check the result of `insert` and only do that if it actually changes?

src/rt/ds/hashmap.h
```diff
@@ -424,7 +426,7 @@ namespace verona::rt
           out << " ∅";
           continue;
         }
-        out << " (" << (const KeyType*)unmark_key(key_of(slots[i]))
+        out << " (" << ((const KeyType*)unmark_key(key_of(slots[i])))->id()
```
rengolin, line 429: I'd use c++ casts instead of C casts everywhere, and deal with the errors, if any.

src/rt/ds/hashmap.h
Theodus: Yes, there are: https://github.com/microsoft/verona/blob/master/src/rt/ds/hashmap.h#L54

src/rt/ds/hashmap.h
Theodus: Fixed

src/rt/ds/hashmap.h
Theodus: I'll open a separate issue for using C++ casts in the runtime. Currently it's almost exclusively C-style casts everywhere.

Theodus: https://github.com/microsoft/verona/issues/217

PR #218, opened 2020-07-30
LLVM update to 6ce97eee354

rengolin:
Other than updating CMake, the change brings 4000-ish commits closer to trunk and breaks nothing else.

### DIFF
--- a/devops/ci.yml
+++ b/devops/ci.yml
@@ -69,6 +69,8 @@ jobs:
       set -eo pipefail
       sudo apt-get update
       sudo apt-get install -y ninja-build
+      sudo apt-get remove --purge cmake
+      sudo snap install cmake --classic
       sudo pip install wheel OutputCheck
     displayName: 'Dependencies'
 
--- a/devops/llvm.yml
+++ b/devops/llvm.yml
@@ -43,7 +43,8 @@ jobs:
   - script: |
       set -eo pipefail
       sudo apt-get update
-      sudo apt-get install -y clang ninja-build cmake lld
+      sudo apt-get install -y clang ninja-build lld
+      sudo snap install cmake --classic
       curl -sL https://aka.ms/InstallAzureCLIDeb | sudo bash
     displayName: 'Install Build Dependencies'
 
--- a/devops/nightly.yml
+++ b/devops/nightly.yml
@@ -71,6 +71,8 @@ jobs:
       set -eo pipefail
       sudo apt-get update
       sudo apt-get install -y ninja-build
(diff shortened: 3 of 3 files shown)


rengolin:
[LLVM Build](https://dev.azure.com/ProjectVeronaCI/Project%20Verona/_build/results?buildId=1334&view=results)
[Verona Nightly](https://dev.azure.com/ProjectVeronaCI/Project%20Verona/_build/results?buildId=1337&view=results)

plietar:
LGTM.

In the future, should we have an issue to track features in upstream we wish we could have used but had to work around? As MLIR is quite a fast moving target, I suspect this will happen quite often. That way once time comes to update llvm we can go through the list and clean up the hacks?

For example, the other day I wanted to use the `NoRegionArguments` op trait, but it wasn’t available in our commit, so I had to reimplement it (it’s literally 2 lines). It’s too small a todo to have its own tracking issue, so having a common place would be nice.

I’m of course not suggesting that the person and PR that does the LLVM update is responsible for fixing those. It could be done at a later point by others.

rengolin:
> In the future, should we have an issue to track features in upstream we wish we could have used but had to work around? As MLIR is quite a fast moving target, I suspect this will happen quite often. That way once time comes to update llvm we can go through the list and clean up the hacks?

That's a great idea.


> For example, the other day I wanted to use the `NoRegionArguments` op trait, but it wasn’t available in our commit, so I had to reimplement it (it’s literally 2 lines). It’s too small a todo to have its own tracking issue, so having a common place would be nice.

I'd just put a `TODO` on a comment before the line. We can agree on a text that can mark this, like `TODO LLVM:` and then we go through them and solve the issue. If they're explicit enough (for instance mention the trait you need), it should be easy to anyone doing the merge to do. If it isn't, you can always submit a quick fix after the rebase.

> I’m of course not suggesting that the person and PR that does the LLVM update is responsible for fixing those. It could be done at a later point by others.

This could actually be another way we do the merge. You really need something, or the merge is so simple it's worth doing just for that, and then we move on like that. The only problem I see with this is for work in progress that hasn't been PR'd yet and may be affected.

davidchisnall:
At some point we will want to move to using https://github.com/CTSRD-CHERI/llvm-project, so we can get the Morello bits.  They're about a month behind LLVM a the moment, we probably don't want to get too close.  WE should get the Morello simulator in September, so will aim to add it to CI soon after that.

rengolin:
Perhaps for the next move, then, we should aim at having the same base as CHERI. It would save us the trouble of scanning commits and buildbot status at least. :)

PR #221, opened 2020-08-03
Make Quiet LLVM download the default option

rengolin:
We can't control how much to log, the options are either every percent
or nothing. 100 lines on the build is annoying enough, but the lines
go to stderr and the CI logs them as "errors" polluting the reports.

Another #198 check.


rengolin:
Also changed "true" and "false" to "ON" and "OFF" to keep the same pattern as others.

PR #222, opened 2020-08-03
Detect cores on MacOs CI

rengolin:
Another check from #198


rengolin:
All jobs running with similar compile times, option works.

PR #223, opened 2020-08-03
Use MD5 hash to avoid downloading LLVM every time

rengolin:
LLVM Build: https://dev.azure.com/ProjectVeronaCI/Project%20Verona/_build/results?buildId=1364&view=results

All MD5 files are in the container, Verona build works nicely with and without the archive.


rengolin:
Partially addresses #198

Issue #181, opened 2020-07-07
Split Nightly test in groups


rengolin:
In order to have nice test banners on the README, we could have multiple nightly jobs instead of just one.

Right now, if anything fails, the banner is red and there's no way to know what happened.

If we had a nightly job for, say, Linux, Windows and Mac separate, we could see what's actually failing, and external users wouln't feel "all is broken" just by looking at the front-page.

With all the instability we're having with Mac builders, this could mitigate the apparent bad quality until we have more stable CI.

SeanTAllen:
What's the value of the banners for nightlies?

rengolin:
Mostly psychological.

External users that won't be getting failure emails or looking at our build infrastructure can look at the front-page and see that the tests are passing and commits coming frequently, so quality is measured and the project is healthy. This increases the chances that people will be interested in our project, perhaps decide to contribute, or use for different purposes than we intended and we get a healthy community from it.

If we keep all information inside unknown places (even if public), people don't get to be aware of what we're doing and lose interest.

The banners are part of that strategy, but also public PRs, public documents, public CI, public issues, etc.

But the point of this issue is not about the banner itself, we already have it. It's to split them into categories. This achieves two goals:
1. It tells the users what combinations of hardware/os/environment we care about. Having a matrix of architectures by os with different compilers, for example, would show which ones we keep green, which ones are sometimes red and which ones are non-existent.
2. If there are instabilities in one architecture (as recently we had MacOS failing every time), it doesn't make others red. Right now there's only one banner and if a tiny thing fails, it turns red, giving the impression that we don't care about testing.

So, in the end, it's not important per se. But it would be nice to have a better public reporting system for testing quality, which we do care a lot internally, but are currently failing to demonstrate upstream. 

That's why it's an "enhancement". If we end up having it, when we have it, it would be nice to have.

rengolin:
No need, just learnt we can get tags to specific builds inside a pipeline.

PR #227, opened 2020-08-05
Update build matrix with specific types of builds

rengolin:
As mentioned in #181 we don't need to have different pipelines to show different badges for each build type.

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,7 +15,13 @@ This project is at a very early stage, parts of the type checker are still to be
 implemented, and there are very few language features implemented yet. This will
 change, but will take time.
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -15,7 +15,13 @@ This project is at a very early stage, parts of the type checker are still to be
...
+Windows | [![Build Status](https://dev.azure.com/ProjectVeronaCI/Project%20Verona/_apis/build/status/Verona%20Nightly?branchName=master&jobName=Windows&configuration=Windows%20Release)](https://dev.azure.com/ProjectVeronaCI/Project%20Verona/_build/latest?definitionId=6&branchName=master) | [![Build Status](https://dev.azure.com/ProjectVeronaCI/Project%20Verona/_apis/build/status/Verona%20Nightly?branchName=master&jobName=Windows&configuration=Windows%20RelWithDebInfo)](https://dev.azure.com/ProjectVeronaCI/Project%20Verona/_build/latest?definitionId=6&branchName=master) |
```
mjp41, line 23: RelWithDebInfo versus Debug?  Should we complicate the table here?

rengolin: I thought better not... it's close enough. Same with ASAN (which I only show Debug+ASAN, not Rel+ASAN).

PR #228, opened 2020-08-05
disable -Werror outside of CI

Theodus:
This sets the `-Wno-error` compiler option when not compiling for CI. This is something I always do while debugging so that I don't spend so much time dealing with `warning: unused parameter` preventing the program from compiling.


Theodus:
Ok, I suppose it's easy enough to change manually when necessary.

PR #230, opened 2020-08-05
First stage of loop lowering into Verona dialect

rengolin:
This is the first stage, just changing the basic-block CFG style for a construct with a region and four new operations:
* `loop_exit`: an exit condition that breaks the loop when `%cond` evaluates to false.
* `loop_return`: function return inside loops (to differentiate from region return).
* `break`: breaks the loop unconditionally.
* `continue`: continues the loop unconditionally.

Variable access is still using load/store like before. This is a change for the syntax of loop representation in MLIR, not semantics.

The second stage will be to start variable tracking, so we can isolate the while region with arguments and return values, and use the `IsolatedFromAbove` trait that allows parallel execution in the MLIR pipeline. This will come in a later patch.


rengolin:
Partially implements #203.

rengolin:
The main problem of using a new region as a generic loop construct is that it becomes really hard to `return` values from inside the loop.

Basically, the loop region's CFG is completely isolated from the encompassing region, and to return values or control from inside a loop, we'd have to `break` the loop (and all nested loops around) and branch to a basic block that will return the value.

If we don't track variables (arg, ret), then we need to store the returned value on some global alloca (not in the while region), set a condition to be true (immediate return), break all loops, then `cond_br` to the exit branch.

If we do track variables, it's slightly simpler but still not trivial. We'd need to return from the loop immediately, and on every encompassing nested loop check if the return is a function return (still needs a function global alloca for the flag), and trickle down to the return of the outer-most loop, then check the flag and return the value of the while loop.

If the loop's return values are not the same as the function's (which is possible when we track multiple local variables inside the loop), we'd have to know which value to return, which will likely be a special value for every loop, taking place of the global alloca for the return value.

Multiple function return values complicate the scenario, but not tremendously so.

I'm starting to think we better only use closed loop forms for special loops we know don't return and continue with BB-CFG implementation for the generic case.

plietar:
With a custom verona.return operation, we can set as a verifier that it must appear within a function or a verona.while block. At some point in the lowering process, after we’ve lowered while ops to a flat CFG, we can get replace verona.return with the standard one. 

We need this special return anyway because we need it to support subsumption (see my type checking PR).

Can the `loop_exit` operation be replaced by a conditional branch plus a break instead?

rengolin:
> We need this special return anyway because we need it to support subsumption (see my type checking PR).

I see. I'll add the operation, then.


> Can the `loop_exit` operation be replaced by a conditional branch plus a break instead?

We can, but I wanted to mark the division between the condition block and the body block. This will make further analysis easier by allowing us to calculate range limits, boundaries, etc. Another option I thought of was to lower the condition block in its own sub-region, returning a single boolean value, and then having a conditional branch, but that sounded more complicated...

src/mlir/dialect/VeronaOps.cc
```diff
@@ -111,6 +133,32 @@ static LogicalResult verify(verona::ClassOp classOp)
...
+    if (verona::LoopExitOp exit_op = dyn_cast<verona::LoopExitOp>(op))
+    {
+      hasExit = true;
+      break;
+    }
+  }
+
+  // And the terminator of the last block must be a continue.
+  bool hasCont =
+    isa<verona::ContinueOp>(body.getBlocks().back().getTerminator());
```
plietar, line 154: This seems very restrictive, and I don't really see what it's trying to prevent.

The following should be valid IMO:
```
verona.while {
    jump ^bb1
^bb0:
    verona.continue
^bb1:
    jump ^bb0
}
```

src/mlir/dialect/VeronaOps.td
```diff
@@ -244,4 +245,117 @@ def Verona_FieldOp : Verona_Op<"field", [InClassScope]> {
...
+    let builders = [
+      OpBuilder<"OpBuilder &builder, OperationState &state, bool cond">
+    ];
+    let arguments = (ins BooleanType:$cond);
+    let results = (outs);
+    let assemblyFormat = [{ $cond `:` type($cond) attr-dict }];
+    let verifier = [{ return success(); }];
+}
+
+def Verona_LoopReturnOp : Verona_Op<"loop_return", [Terminator, InLoopScope]> {
```
plietar, line 278: What I had in mind was a single `Verona_ReturnOp` operation, which can be used either in loops or directly in function. Rather than `InLoopScope`, it would have trait `HasParent<"FuncOp", "WhileOp">`. The AST lowering would only ever emit `verona.return`, never `return`.

Additionally you need a verifier that checks that the argument type matches the function's return type (just like the standard `return` does). Since my PR #206 already does that (as a verifier), I'm happy for you to just leave a TODO.

src/mlir/dialect/VeronaOps.cc
rengolin: That's certainly possible, and we even have a test for that. 

The line is only trying to add a default continue on the last block if none was added by the code. So, if the code is:
```
while (...) {
  if (...) continue;
  break;
}
```

You don't add an additional `continue` after the break. But most loops will look like:
```
while (...) {
  if (...) ...;
}
```
And no terminator will be added to the last block, thus needing a continue.

src/mlir/dialect/VeronaOps.td
rengolin: I know but that approach has several drawbacks.

First, there are many types of returns. Return from function, return from regions (into their parent region) and return from function inside a region (which is what `loop_return` does). Both function return and region return work as expected even with Verona types, so we only need to add the third type.

Second, once we add free variable analysis (args, rets) and track all region captures and returns, we could just return to the encompassing region with a flag (saying immediate return) but that would be hard if the loop could function-return either of multiple values.

Third, eventually, loops will be lowered to CFG blocks, in which case, `verona.loop_return` is identical to just `return` and can be converted without any additional semantic checks. We get that for free. So if we complicate the lowering now, we'll complicate the lowering later. If we keep it simple now, we'll get it for free.

Fourth, I'm still not convinced we need a special return for sub-typing. If we do, then we'll need a special "everything", and that won't scale. As we discussed earlier, as reasonable implementation is to add casts until the types are known, and even keep the casts for generics up until the final lowering. The semantics is complete and the pattern is simple.

Finally, my own experience lowering different kinds of IR, including at least two other MLIR projects, is that the more special your dialect is, the harder it is to work with because you need to imbue special meaning to everything. The worst thing to make special is control flow, because that knowledge is often embedded into the language itself, with the assumption that no one in their right minds would re-implement branches and returns.

I'd ask that we keep it like that, for now at least, until we can have a proper conversation about sub-typing and its effects in the dialect.

rengolin:
> > Can the `loop_exit` operation be replaced by a conditional branch plus a break instead?
> 
> We can, but I wanted to mark the division between the condition block and the body block. This will make further analysis easier by allowing us to calculate range limits, boundaries, etc. Another option I thought of was to lower the condition block in its own sub-region, returning a single boolean value, and then having a conditional branch, but that sounded more complicated...

To expand on this a bit, what I mean is that I want to separate, in a clear way, between:
```
while (a > b && c < d) {
    foo();
}
```
and
```
while (a > b) {
    if (c < d)
        foo();
}
```
The current implementation transforms the original loop as:
```
while (true) {
    if (a > b && c < d)
        foo();
    else
        break;
}
```
which in basic-block lowering is too similar to the alternative loop. It's not identical, but I want to make clear and easy to the loop analysis what is inside the condition and what is the loop body.

An alternative implementation would be:
```
verona.while {
    %cond = {
        %0 = gt a,b
        %1 = lt c, d
        return and %0, %1;
    }
    cond_br %cond, ^exit, ^continue
  ^exit:
    break;
  ^continue:
   ...
}
```

which would be slightly more convoluted to generate, but not much so. I'm happy either way.

src/mlir/dialect/VeronaOps.cc
plietar: > That's certainly possible, and we even have a test for that.

Do we? I don't see any test where the last block does not end with a `continue`. Only a test which has a `continue` in both in the middle and at the end.

> The line is only trying to add a default continue on the last block if none was added by the code

That line doesn't add anything as far as I can tell. It only ensures it exists. It's the builder that adds the missing `continue` (and that's totally fine).

Here's another example that, I believe would be rejected by this verifier, even though it shouldn't:
```
while (...) {
    return;
}
```

src/mlir/dialect/VeronaOps.cc
```diff
@@ -111,6 +133,32 @@ static LogicalResult verify(verona::ClassOp classOp)
   return success();
 }
 
+static LogicalResult verify(verona::WhileOp whileOp)
+{
+  auto& body = whileOp.body();
+
+  // While blocks must have a condition on the first block
+  // with a LoopExit operation.
```
plietar, line 141: Similarily, this is quite a broad requirement.
What if the program looks like, using a ternary operator:
```
while (x ? y : z) { ... }
```

Then the IR will look like
```mlir
while {
  cond_br %x, ^bb1, ^bb2
^bb1:
  br ^bb3(%y)
^bb2:
  br ^bb3(%z)
^bb3(%0):
  loop_exit %0
}
```

Then the first basic block does not contain a `loop_exit`.

(We might not have this form of ternary operator, but I'm sure we can have expression that generate more than one basic block)

You could, as a requirement, have that all paths through the loop must go through a `loop_exit`, but I'm not even sure I see the point of this requirement.

src/mlir/generator.cc
```diff
@@ -564,14 +569,7 @@ namespace mlir::verona
...
     assert(isContinue(ast) && "Bad node");
-    // Nested loops have multiple heads, we only care about the last one
-    if (!loopTable.inScope("head"))
-      return parsingError("Continue without a loop", getLocation(ast));
-    auto head = loopTable.lookup("head");
-    mlir::ValueRange empty{};
-    // We assume the continue is the last operation in its basic block
-    // and that was checked by the parser
-    builder.create<mlir::BranchOp>(getLocation(ast), head, empty);
+    builder.create<mlir::verona::ContinueOp>(getLocation(ast));
```
plietar, line 572: This will allow, at the parser level, a `continue` outside of any loop, which will trip MLIR verifiers. I think we should check the presence of a loop here, such that we never generate syntactically invalid MLIR.

Same for break.

src/mlir/dialect/VeronaOps.cc
```diff
@@ -111,6 +133,32 @@ static LogicalResult verify(verona::ClassOp classOp)
   return success();
 }
 
+static LogicalResult verify(verona::WhileOp whileOp)
+{
+  auto& body = whileOp.body();
```
plietar, line 138: I think you need to verify that the region has no parameters (eg. using NoRegionArguments).
It's not allowed by the parser, by the generic syntax would let you construct one (or through builders).

I guess with value tracking this requirement will change, but we should still encode it now.

src/mlir/dialect/VeronaOps.td
```diff
@@ -244,4 +245,117 @@ def Verona_FieldOp : Verona_Op<"field", [InClassScope]> {
...
+    let results = (outs);
+    let assemblyFormat = [{ attr-dict }];
+    let verifier = [{ return success(); }];
+}
+
+def Verona_LoopExitOp : Verona_Op<"loop_exit", [InLoopScope]> {
+    let summary = "breaks any loop block if condition is `false`";
+
+    let builders = [
+      OpBuilder<"OpBuilder &builder, OperationState &state, bool cond">
```
plietar, line 270: Is this builder actually used / defined anywhere?

src/mlir/dialect/VeronaOps.cc
rengolin: > ```
> while (...) {
>     return;
> }
> ```

Ah, I see what you mean! The back-branch example earlier threw me off the track. Just checking for all basic blocks having terminators should be enough.

The version with `cond_br` instead of `loop_exit` has a tail block that was the `continue` and why the check remained like this.

src/mlir/dialect/VeronaOps.cc
rengolin: `loop_exit` is not required to be in the first basic block, but it separates the basic blocks that compose the condition from those that compose the body. It's precisely because both condition and body could have multiple basic blocks that I want to split the region.

src/mlir/generator.cc
rengolin: The AST will not allow `continue` or `break` out of loops in the first place, the `isContinue()` does the trick. The `verify` is there mostly to check that behaviour is not broken by further passes.

src/mlir/dialect/VeronaOps.td
rengolin: Hmm, good question. I've added and removed a lot of code, this could be a left over. Will check.

src/mlir/dialect/VeronaOps.cc
rengolin: Updated the check to get this one right, too.

rengolin:
New update fixes all pending issues. The only remaining discussion is about `verona.return`.

src/mlir/dialect/VeronaOps.cc
```diff
@@ -111,6 +133,36 @@ static LogicalResult verify(verona::ClassOp classOp)
...
+        if (verona::LoopExitOp exit_op = dyn_cast<verona::LoopExitOp>(op))
+        {
+          hasExit = true;
+          break;
+        }
+      }
+    }
+    // All blocks must terminate
+    if (!b.back().isKnownTerminator())
+      return failure();
```
plietar, line 158: I don't think this is necessary, the builtin MLIR verifier should already be taking care of this.

rengolin: Good point

src/mlir/dialect/VeronaOps.td
rengolin: Let's leave this discussion for the next step, when we discuss functions and returns, if they have to be special or not.

PR #232, opened 2020-08-05
Move RNG and Scrambler to systematic namespace

Theodus:
This change is made primarily to allow Object IDs to be a scrambled
permutation when systematic testing is enabled.


src/rt/sched/cown.h
```diff
@@ -745,7 +737,7 @@ namespace verona::rt
 
 #ifdef USE_SYSTEMATIC_TESTING
       std::sort(&sort[0], &sort[count], [](Cown*& a, Cown*& b) {
-        return Scheduler::get_scrambler()(a->id(), b->id());
+        return Systematic::get_scrambler()(a->id(), b->id());
```
mjp41, line 740: This is double scrambled now as `id()` is already permuted.

PR #233, opened 2020-08-05
Improvements to CMake for different build types

mjp41:
Uses different directories for each type of MLIR it downloads. This
interoperates better with MSbuild. It also correctly passes the
build type through on both Ninja and MSBuild now.


external/CMakeLists.txt
```diff
@@ -0,0 +1,84 @@
+# This layer of cmake is used to convert multi-configuration systems,
```
rengolin, line 1: other projects (like LLVM) use directory `cmake` and the name of the file is meaningful, like `MultiSystem.cmake` or `AddMLIR.cmake`, with separate concerns per file, which makes it easier to maintain long term.

mjp41: It is the only cmake in a directory `external` it is just for building the external dependencies.  The comment explains who it is achieving it.  I am not sure giving the file a specific name would help with understanding the crazy that is cmake.  If we factor the CMake into many files in a single directory that makes sense.

This might make sense to do at the top-level where the recursively calling the same CMake file occurs. Perhaps having two files here would make more sense.  I can do that as a separate PR.

rengolin: Fair enough. Once we have more files, we can move them around.

PR #235, opened 2020-08-07
Fixes to CI

mjp41:
Address some sub-issues in #198.

* Automatically pass compiler to subprojects
* Fixes bug introduced in #233 that didn't correctly generate subprojects for Verona (RelWithDebInfo, Debug).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,8 +1,16 @@
 cmake_minimum_required(VERSION 3.10.0)
 
-option (VERONA_DOWNLOAD_LLVM "Download cached version of LLVM" ON)
-option (VERBOSE_LLVM_DOWNLOAD "Verbose LLVM/MLIR download step" OFF)
+# CMAKE_HOME_DIRECTORY should not be passed down to another project.
+set (VERONA_DONT_PASS CMAKE_HOME_DIRECTORY)
+macro (option_top VAR HELP DEFAULT)
+  option (${VAR} ${HELP} ${DEFAULT})
+  list (APPEND VERONA_DONT_PASS ${VAR})
+endmacro ()
 
+option_top (VERONA_DOWNLOAD_LLVM "Download cached version of LLVM" ON)
+option_top (VERBOSE_LLVM_DOWNLOAD "Verbose LLVM/MLIR download step" OFF)
+
+message (STATUS "Download LLVM: ${VERONA_DOWNLOAD_LLVM}")
 # Lifted from snmalloc. Hard to include with external projects, so copied
 macro(clangformat_targets)
   # The clang-format tool is installed under a variety of different names.  Try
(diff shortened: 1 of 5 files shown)


mjp41:
[Nightly](https://dev.azure.com/ProjectVeronaCI/Project%20Verona/_build/results?buildId=1406&view=logs&j=b319cb12-f070-52ae-e2a6-04d190fd2bf4) in progress, but looks like it should pass.

devops/nightly.yml
```diff
@@ -40,24 +40,35 @@ jobs:
...
         BuildType: Debug
         Asan: On
+        LLVMDownload: On
       Clang Release (ASAN):
         CC: clang
         CXX: clang++
         CXXFLAGS: -stdlib=libstdc++
         BuildType: Release
         Asan: On
+        LLVMDownload: On
```
rengolin, line 64: I would create another set, also ran at night, with all three platforms: Win,Lin,Mac on Release mode.

I often use the Nightly build to make sure a change of mine isn't breaking the build, especially infrastructure changes.

Adding the long build to it will make it really hard to test those changes.

devops/nightly.yml
```diff
@@ -80,7 +91,7 @@ jobs:
     displayName: 'CMake'
     inputs:
       cmakeArgs: |
-        .. -GNinja -DCMAKE_BUILD_TYPE=$(BuildType) -DVERONA_EXTRA_CMAKE_ARGS="-DCMAKE_C_COMPILER=$(CC);-DCMAKE_CXX_COMPILER=$(CXX);-DCMAKE_CXX_FLAGS=$(CXXFLAGS)" -DENABLE_ASSERTS=ON -DUSE_ASAN=$(Asan) -DVERONA_CI_BUILD=On -DRT_TESTS=ON
+        .. -GNinja -DCMAKE_BUILD_TYPE=$(BuildType) -DVERONA_EXTRA_CMAKE_ARGS="-DCMAKE_CXX_FLAGS=$(CXXFLAGS)" -DENABLE_ASSERTS=ON -DUSE_ASAN=$(Asan) -DVERONA_CI_BUILD=On -DRT_TESTS=ON -DVERONA_DOWNLOAD_LLVM=$(LLVMDownload)
```
rengolin, line 94: `CXX_FLAGS` should also have been propagated in the same way. We shouldn't need `EXTRA_ARGS` for anything trivial.

mjp41: Should CXX_FLAGS be consistent across LLVM and Verona?

CMakeLists.txt
```diff
@@ -68,13 +69,43 @@ if (NOT DEFINED VERONA_LLVM_LOCATION)
   endif()
 
   set (MLIR_INSTALL ${CMAKE_BINARY_DIR}/$<CONFIG>/mlir)
+
+  set (EXTERNAL_EXTRA_CMAKE_ARGS)
```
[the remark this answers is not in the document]
mjp41: We probably need to be able to supply Verona with different `CXX_FLAGS` to LLVM. 

It does only pass the flags to LLVM if it is not being downloaded. https://github.com/microsoft/verona/blob/ab40a91d6ca8856bf3c3d176e4714864866f7ebc/CMakeLists.txt#L79

You can do
```
cmake -DLLVM_EXTRA_CMAKE_ARGS=-DUSE_LLD=On;-DCXX_FLAGS=-O3 ..
```

I don't know how to pass all the flags that are supplied at the command line to an `externalproject`.

devops/nightly.yml
mjp41: I can add another one, but if you change the infrastructure, you are probably going to have to run this other build too.

rengolin: I do phased testing. First I run the CI, which is quick and checks the basic. When that's passing, I run the nightly, which can take up to an hour. The build with LLVM would probably add at least one hour on top of that, which then means I can only test 2-3 times a day.

Also, because we're on the free builders, we can only use 10 of them at the same time. The nightly build already has 8 builds, if we add LLVM to three of them (Win/Lin/Mac), it's more than 10 and one build will have to wait for another to finish, and block any other testing done by other people.

devops/nightly.yml
rengolin: That's a good question. Probably not consistently, but if the user adds the option on the top-level project, I'd say "yes".

It depends on what semantics do we want for those CMake flags. I'd assume "top-level flags propagate across all sub-projects" and "project-specific flags go straight to them, and override top-level flags". But that's easier said than done.

CMakeLists.txt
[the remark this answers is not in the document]
mjp41: What does overridden mean here?
```
-DCXX_FLAGS=-O3
-DLLVM_EXTRA_CMAKE_ARGS=-DCXX_FLAGS=-O2
```
How about the following semantics?

Nothing is passed to LLVM, except what is supplied with LLVM_EXTRA_CMAKE_ARGS, and everything is passed to the Verona sub build?  Most people will not build LLVM, so this is not a common case.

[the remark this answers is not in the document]
rengolin: Makes sense. Simpler to understand and easier to implement.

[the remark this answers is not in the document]
mjp41: So 

   https://stackoverflow.com/questions/25533831/passing-cmake-variables-to-externalproject-add

implies you can pass everything through.  Adding code for routing variables to only one build seems hard, and I don't know how we would want to do overriding.

devops/nightly.yml
rengolin: Let's leave the LLVM build for later, it's not as simple with the too-small MS-hosted builders.

devops/nightly.yml
rengolin: I like your idea on the other question, let's go with that.

CMakeLists.txt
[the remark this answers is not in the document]
rengolin: How about we don't do overriding and only have LLVM's `EXTRA_ARGS`? With your idea to route flags to Verona only, that'd work, no? ie. get rid of `VERONA_EXTRA_CMAKE_ARGS`.

rengolin:
The snmalloc changes leaked out again

mjp41:
> The snmalloc changes leaked out again
Sorry

PR #237, opened 2020-08-07
rename action to behaviour

Theodus:
Replace the term "action" with "behaviour", since there is no longer a meaningful distinction between the two. Also add some documentation to `Cown` and `SchedulerThread`.


src/rt/sched/cown.h
```diff
@@ -25,6 +25,20 @@ namespace verona::rt
...
+   * exactly one of the following states:
+   *   1. Unscheduled
+   *   2. Scheduled, in the queue of a single scheduler thread
+   *   3. Running on a single scheduler thread
+   *
+   * Once a cown is running, it executes a batch of multi-message behaviours.
+   * Each message may either acquire the running cown for participation in a
+   * future behaviour, or execute the behaviour if it is the last cown to be
+   * acquired. If the running cown is acquired for a future behaviour, it will
+   * be descheduled until that behaviour has completed.
```
rengolin, line 40: Nice! Very clear!

src/rt/sched/multimessage.h
```diff
@@ -19,7 +19,7 @@ namespace verona::rt
       size_t index;
       size_t count;
       Cown** cowns;
-      Action* action;
+      Behaviour* be;
```
rengolin, line 22: Do we have to shorten it to `be`? It's harder to read like that. I'd leave as `behaviour`.

Theodus: fixed

PR #238, opened 2020-08-07
Add a diagnostic handler for prettier error messages.

plietar:
This gives us a bit of colour, prints the relevant piece of code, and
displays notes attached to diagnostics.

### DIFF
--- a/src/mlir/driver.cc
+++ b/src/mlir/driver.cc
@@ -16,7 +16,8 @@
 
 namespace mlir::verona
 {
-  Driver::Driver(unsigned optLevel) : passManager(&context)
+  Driver::Driver(unsigned optLevel)
+  : passManager(&context), diagnosticHandler(sourceManager, &context)
   {
     // Opaque operations and types can only exist if we allow
     // unregistered dialects to co-exist. Full conversions later will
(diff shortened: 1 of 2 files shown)


rengolin:
Nice! How does this interoperate with `main`'s `InitLLVM`? Is it a replacement for that initialisation, or is that still required?

mjp41:
Does this give colour error messages on Windows?  For the prototype I implemented 
  
  https://github.com/microsoft/verona/blob/master/src/ds/console.h

to set the appropriate state for the console on Windows.

plietar:
> How does this interoperate with main's InitLLVM? Is it a replacement for that initialisation, or is that still required?

I think they are orthogonal. `InitLLVM` seems to be (among other things) about catching aborts and segfaults, whereas the diagnostic handler is for all the MLIR `emitError` calls.

`mlir-ops` does both, it calls `InitLLVM` in `mlir-opt.cpp` and sets up a diagnostic handler in `MlirOptMain.cpp`.

plietar:
> Does this give colour error messages on Windows?

I don't really know to be honest. A quick look through the llvm code base, it seems like it ends up relying on stuff in `llvm/lib/Support/Windows/Process.inc`, where colours are set by calling native Windows APIs (rather than emitting ANSI code).
It seems like it might "just work", or at least work as well as llvm and clang do.

Would you mind giving it a try @mjp41? You can pass `verona-mlir` a garbage file like:
```
module {
  abc
}
```

I get this output (notice `error` is in red, and the error message is in bold):

Screenshot 2020-08-07 at 15 11 23

PR #241, opened 2020-08-11
Backpressure: set unmutable on fast_send

Theodus:
On `fast_send` the unacquired cowns required for a multi-message are made unmutable if any are overloaded. This continues to cover cases where an overloaded cown, `a`, mutes another cown, `b`, and then requires `b` in a future message. But now cowns are set unmutable only prior to a fast acquire loop.

### DIFF
--- a/src/rt/sched/backpressure.h
+++ b/src/rt/sched/backpressure.h
@@ -66,7 +66,7 @@
  * State Properties:
  * - Normal: The inital state of all cowns.
  * - Muted: A muted cown may not be scheduled or deallocated.
- * - Unmutable: An nnmutable cown may not be muted.
+ * - Unmutable: An unmutable cown may not be muted.
  *
  * State Transition Rules:
  * - Cowns sending to overload/muted cowns may be muted.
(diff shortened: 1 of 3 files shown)


src/rt/sched/backpressure.h
```diff
@@ -166,7 +166,7 @@ namespace verona::rt
      */
     uint32_t _has_token : 1;
     /**
-     * Indicates if a cown is currently muted (1), unmutable (2), or
+     * Indicates if a cown is currently normal (0), muted (1), unmutable (2), or
```
rengolin, line 169: nit: `normal` can mean many things, easily overloaded term. Perhaps `unmuted` would be better in this context?

src/rt/sched/cown.h
```diff
@@ -901,6 +881,31 @@ namespace verona::rt
...
+      yield();
+      if (
+        !requires_unmute
+#ifdef USE_SYSTEMATIC_TESTING
+        && !Systematic::coin(3)
+#endif
+      )
+        return;
+
+      for (size_t i = body->index; i < body->count; i++)
```
rengolin, line 903: nit: range loop?

Theodus: `body->cowns` is a raw pointer. A container should be added eventually, like vector but with snmalloc.

src/rt/sched/backpressure.h
Theodus: Normal is the term used in the big comment at the top of the file. It will be rewritten soon, so I'll make a note to myself to come up with a better name.

PR #244, opened 2020-08-11
Normalize types before applying subtyping rules.

plietar:
This helps recognise distributivity of joins over meets, such as:

  meet<A, join<B, C>> <: join<meet<A, B>, meet<A, C>>

While distributivity of joins over meets alone is quite niche, and
many languages get away without supporting it, distributivity over
viewpoint types (which will be added soon) is a lot more critical.
This PR lays the ground work for that.

### DIFF
--- a/src/mlir/dialect/VeronaTypes.cc
+++ b/src/mlir/dialect/VeronaTypes.cc
@@ -334,4 +334,124 @@ namespace mlir::verona
   {
     return llvm::all_of(types, isaVeronaType);
   }
+
+  /// Distribute a lattice type (join or meet) by applying `f` to every element
+  /// of it. Each return value of the continuation is added to `result`.
+  ///
+  /// Assuming `type` is in normal form, this method will process nested `T`s
+  /// as well.
+  ///
+  /// For example, given `join<A, join<B, C>>`, this method will add
+  /// `f(A), f(B), f(C)` to `result`.
+  template<typename T>
+  static void distributeType(
+    SmallVectorImpl<Type>& result, T type, llvm::function_ref<Type(Type)> f)
+  {
+    for (Type element : type.getElements())
+    {
+      if (auto nested = element.dyn_cast<T>())
+        distributeType<T>(result, nested, f);
+      else
+        result.push_back(f(element));
+    }
+  }
+
+  /// If the argument `type` is of kind `T` (where `T` is a lattice type, ie.
+  /// JoinType or MeetType), distribute it by applying `f` to every element of
+  /// it. The return values are combined to form a new lattice type of the same
+  /// kind. If `type` is not of kind `T`, it is directly applied to `f`.
+  ///
+  /// Assuming `type` is in normal form, this method will process nested `T`s
+  /// as well.
+  ///
+  /// For example, given `join<A, join<B, C>>`, this method will return
+  /// `join<f(A), f(B), f(C)>`.
+  template<typename T>
+  static Type
+  distributeType(MLIRContext* ctx, Type type, llvm::function_ref<Type(Type)> f)
+  {
+    if (auto node = type.dyn_cast<T>())
+    {
+      SmallVector<Type, 4> result;
+      distributeType<T>(result, node, f);
+      return T::get(ctx, result);
+    }
+    else
+    {
+      return f(type);
+    }
+  }
+
+  /// Normalize a meet type.
+  /// This function returns the normal form of `meet<normalized..., rest...>`,
+  /// distributing any nested joins.
+  ///
+  /// Types in `normalized` must be in normal form and not contain any joins.
+  /// Types in `rest` may be in any form.
+  ///
+  /// This method uses `normalized` as scratch space; it recurses with more
+  /// elements pushed to it. When it returns, `normalized` will always have its
+  /// original length and contents.
+  ///
+  /// TODO: this function uses recursion to iterate over the `rest` array,
+  /// because that works well with normalizeType. It could be rewritten to use
+  /// loops, which is probably more efficient and doesn't risk blowing the
+  /// stack.
+  Type normalizeMeet(
+    MLIRContext* ctx, SmallVectorImpl<Type>& normalized, ArrayRef<Type> rest)
+  {
+    if (rest.empty())
+      return MeetType::get(ctx, normalized);
+
+    Type element = normalizeType(rest.front());
+    return distributeType<JoinType>(ctx, element, [&](auto inner) {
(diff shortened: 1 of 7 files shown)


src/mlir/dialect/VeronaTypes.h
```diff
@@ -16,6 +16,18 @@ namespace mlir::verona
   /// dialect.
   bool areVeronaTypes(llvm::ArrayRef<Type> types);
 
+  /// Normalize a type by distributing unions and intersections, putting the
+  /// type in disjunctive normal form. This is a necessary step in order for
+  /// subtyping to recognise certain relations.
+  ///
+  /// The amount of normalization done is quite limited. In particular it does
```
rengolin, line 23: I would move this type of comment inside the implementation, as a TODO maybe.

src/mlir/dialect/VeronaTypes.cc
```diff
@@ -334,4 +334,124 @@ namespace mlir::verona
...
+        return type;
+
+      case VeronaTypes::Join:
+        return normalizeJoin(ctx, type.cast<JoinType>().getElements());
+
+      case VeronaTypes::Meet:
+        return normalizeMeet(ctx, type.cast<MeetType>().getElements());
+
+      default:
+        abort();
```
rengolin, line 454: An error message would be more friendly here

testsuite/mlir/mlir-parse/subsumption.mlir
```diff
@@ -33,4 +33,11 @@ module {
 
     return
   }
+
+  func @test_distributivity(%a: !verona.meet<U64, join<iso, mut>>) {
```
rengolin, line 37: This is only one of many possibilities of the new code. Please add tests for at least the major variations (join of meets, meets of joins, simple types, etc).

plietar:
> Is this only meant as verification? If so, then we'll have to normalise every time we verify. I was expecting an transform pass that normalises all types, so that we don't need to worry about it from then on, only when creating new operations, in which case we call normalizeType again, just in case.

Say you have a type A with a field of type `(S|T)`, and a type B with a field of type `U`. (Ignoring viewpoint adaptation), the result of reading the field from `A&B` will be `(S|T) & U`, which normalized into `(S & U) | (T & U)`. The line of code below is fully normalized, like a normalization pass would generate.

```mlir
verona.field_read %x["f"] : !verona.meet<A, B> -> !verona.join<meet<S, U>, meet<T, U>>
```

However, the type-checking rule for field reads will be (naively) generating the non-normalized type  `(S|T) & U`, and will reject the operation above. We could of course add normalization to just this rule, however I think we will soon end up in the situation where we apply normalization everywhere we use subtyping, so subtyping is a pretty natural unique place to do this.

In practice, I expect type inference will generate normalized types, and we could also have a normalization pass if it helps for performance. But if our language's semantics allow for normalization in subtyping, then we should match that.

testsuite/mlir/mlir-parse/subsumption.mlir
plietar: This (and obviously variants of it) is, I strongly believe, the only new subtyping pattern that is now recognized but wasn't before.

While the normalization code have to handle all other sorts of types, you can't really observe them just from the subtyping rules. If we want to test the normalization process in more details we should IMO do it with unit testing around just the `normalizeType` function.

src/mlir/dialect/VeronaTypes.h
plietar: I consider what this function does or does not normalize part of its contract, and not an implementation detail, so I think it is relevant to have it in the header file.

rengolin: The first paragraph, yes. The second, sounds like a TODO to me.

If not, then we won't do any flattening, and we'll have the undesirable property of: `join<A, join<B, C>> != join<join<A, B>, C> != join<A, B, C>`. Same for simplification, we'll end up having to do that kind of match elsewhere.

I'd rather normalise them all in a single place.

testsuite/mlir/mlir-parse/subsumption.mlir
rengolin: Testing for the variants is a good exercise in understanding the domain (TDD), as well as increased coverage for future changes, as well as expressing all possible cases to future developers. So I'd strongly advocate for adding them all.

Meets of joins, joins of meets, unchanging transforms (like no combination, only joins, only meets), leaving the unimplemented normalisations for later (flattening, deduplication, etc).

plietar: What I mean is that from a subtyping POV, which is the only way normalization is invoked at the moment, for all these cases, it doesn't make a difference whether normalization was applied.

For instance, `join<A, A>` is a subtype of A, just by virtue of the subtyping rules of `join`. Similarily, `join<meet<U64, iso>, meet<U64, mut>>` is actually already a subtype of `verona.meet<U64, join<iso, mut>>`.

Therefore if we want to write any meaningful test for those, it cannot be just by testing subtyping, since the tests would have passed regardless of normalization.

src/mlir/dialect/VeronaTypes.h
plietar: I didn't mean we shouldn't change that contract ever, only that this is merely its current contract. But I guess it's a minor detail of it, so I've moved it out of the header.

src/mlir/dialect/VeronaTypes.cc
plietar: Done.

testsuite/mlir/mlir-parse/subsumption.mlir
rengolin: Ah, right, we're not talking about the same thing, then. :)

My point is that the normalisation code can do a lot more than this example, regardless of how it integrates with the sub-typing. I guess that's why you mentioned about unit tests, but creating a unit test infrastructure for this patch doesn't make sense.

I was just thinking one function with unrelated instructions on various types and then the result of those types to be printed back, but you don't actually change the type, just check it, so it's harder to test that way.

My main worry is that there is something in the code that is broken but we can't see because there are no tests for it. I guess for this patch, focused on sub-typing, not normalisation, we could let it be for now.

I really want to see the types normalised in the IR as soon as possible, in which case it would be easy to test. But that's for another patch. For now, looks ok.

rengolin (approved):
This change looks great, thanks!

Please create an issue to handle normalisation as a transformation as well as analysis, and we shall discuss the merits of each side as well as the testing there.

PR #248, opened 2020-08-14
Add cast on return

rengolin:
This makes sure the returned value is of the same type as the region
needs it to be. This is a copy of the implicit function return, and we
need to common them up, but they come from different AST nodes, where
one is implicit.

Given we're not yet completely sure on what we'll do with returns, this
refactory can be done later, with the risk of being irrelevant soon.

### DIFF
--- a/src/mlir/generator.cc
+++ b/src/mlir/generator.cc
@@ -426,9 +426,40 @@ namespace mlir::verona
     auto expr = parseNode(getSingleSubNode(ast).lock());
     if (auto err = expr.takeError())
       return std::move(err);
+
+    // Check which type of region we're in
     auto region = builder.getInsertionBlock()->getParent();
+    mlir::Type retTy;
     auto op = region->getParentOp();
-    if (isa<mlir::verona::WhileOp>(op))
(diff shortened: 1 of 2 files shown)


src/mlir/generator.cc
```diff
@@ -426,9 +426,40 @@ namespace mlir::verona
     auto expr = parseNode(getSingleSubNode(ast).lock());
     if (auto err = expr.takeError())
       return std::move(err);
+
+    // Check which type of region we're in
     auto region = builder.getInsertionBlock()->getParent();
+    mlir::Type retTy;
     auto op = region->getParentOp();
-    if (isa<mlir::verona::WhileOp>(op))
+    bool isInLoop = isa<mlir::verona::WhileOp>(op);
```
Theodus, line 434: Can this variable binding be removed? I don't think it's used until the if on line 462

rengolin: It cannot because the `op` changes if the ast node is deeply nested.

If I let the `isa` check happen after the `while` loop, it will always return `false` when nested.

PR #249, opened 2020-08-17
Add class types, field typing rules and viewpoint operator.

plietar:


src/mlir/dialect/VeronaTypes.cc
```diff
@@ -510,4 +511,114 @@ namespace mlir::verona
...
+    {
+      auto [readType, writeType] = lookupFieldType(op, origin, name);
+      assert((readType == nullptr) == (writeType == nullptr));
+      if (readType != nullptr)
+      {
+        readElements.push_back(readType);
+        writeElements.push_back(writeType);
+        found = true;
+      }
+      else if (exhaustive)
```
rengolin, line 645: What's the semantics of exhaustive?

plietar: It's described in the header file. It indicates whether we want the field to be present in all elements of `origins`, or only in at least one. I don't really like the name, but I couldn't come up with anything better.

rengolin: Right, I did read the header comment and that didn't clarify much (to me, personally). Perhaps I don't quite get what the semantics of the field write is to have a good understanding of why only the last one matching would make sense. Do you have some code example that could help?

plietar: > only the last one

It's not the last, it's at *least* one

> I don't quite get what the semantics of the field write is

The flag does not control whether this is a read or a write, it controls whether we are looking up in a join or meet type.

> Do you have some code example that could help?

When looking up a field in `join<T, U>`, the origin could be *either* a T or a U, so the field is required to be present in both types. When lookin up in `meet<T, U>`, the origin is *both* a T and a U. Therefore the field being present in just one of them is enough for us.

The degenerate case, when `origins` is empty, works as well. An empty meet (ie. top), is a type that allows any value. It therefore gives us no information about the contents, hence lookups always fail.

An empty join (ie. bottom) is a bit trickier. It is a type that has no allowed value, therefore any operation on it is allowed (since the code is unreachable anyway).

---

Tbf I'm not set on the utility of merging these two cases into one function, so if it causes too much confusion I'm happy to separate it into two different functions.

plietar: > I'm not set on the utility of merging these two cases into one function, so if it causes too much confusion I'm happy to separate it into two different functions.

Alright, I've split it and I think it makes things much clearer. Let me know what you think.

plietar:
I'll a couple more tests (should do that in a standalone file, rather than cram everything in `dialect.mlir`), but overall I think this looks ready for review.

This encodes the type of fields directly in the type of the class, eg. `!verona.class<"D", "f" : meet<U64, imm>, "g" : meet<class<"C">, mut>>` for a class D with two fields of type `meet<U64, imm>` and `meet<class<"C">, mut`. This follows the approach taken both by LLVM and by other MLIR dialects (eg. [flang](https://github.com/flang-compiler/f18-llvm-project/blob/33490acf241a487028bfee73c0bd9e8026fcdb5c/flang/lib/Optimizer/Dialect/FIRType.cpp#L237) or the [new LLVM dialect](https://reviews.llvm.org/D84339)).

Things that need to be addressed at some later point:
- There is no relation whatsover between a `verona.class @C { ... }` operation and the `!verona.class<"C", ...>` type. It's not clear where this connection should be established, if at all. Maybe in the typechecker for allocation operations?

- Recursive types can be represented in memory and constructed programmatically, but not parsed nor printed. This is pretty straightforward to fix, thanks to the LLVM dialect having done it just a few weeks ago. I already have a follow up PR for that ready.
- At the moment, the syntax for classes is actually `verona.class<"D", "f" : i1 : meet<U64, imm>>`, where i1 can be any type and is ignored. This is an artefact of the restricted API available for parsing; the only way to parse a string is to parse an attribute and cast it, but parsing an attribute consumes the colon and a type. This can be fixed by updating LLVM (there is now a `parseOptionalString` method), or better by adding a `parseString` method upstream (it's a trivial and hopefully uncontroversial change, but I need to look into the LLVM/MLIR contribution process).
- Class types can quite long and repetitive. This is obviously only a problem in the text representation, not in memory. Type aliases would help, but due to how the parser is set up we cannot have aliases inside another Verona type. The parser should be fixed to allow this.

src/mlir/dialect/VeronaTypes.cc
rengolin: I was going to suggest `any|all` flag, but split looks cleaner. Thanks!

rengolin:
> I'll a couple more tests (should do that in a standalone file, rather than cram everything in `dialect.mlir`), but overall I think this looks ready for review.

Agree.

> * There is no relation whatsover between a `verona.class @C { ... }` operation and the `!verona.class<"C", ...>` type. It's not clear where this connection should be established, if at all. Maybe in the typechecker for allocation operations?

I don't see the need of a class operation when the type describes everything we need. It's perfectly fine to have hidden contracts in the IR (ex. how to read/write to fields from a class type by directly lowered ops to IR) if there are no more benefits of encoding it for analysis and transformation purposes.

Simply put, the IR needs to be as high-level as it needs to be, not more.

> * Recursive types can be represented in memory and constructed programmatically, but not parsed nor printed. This is pretty straightforward to fix, thanks to the LLVM dialect having done it just a few weeks ago. I already have a follow up PR for that ready.

SGTM.

> * At the moment, the syntax for classes is actually `verona.class<"D", "f" : i1 : meet<U64, imm>>`, where i1 can be any type and is ignored. This is an artefact of the restricted API available for parsing; the only way to parse a string is to parse an attribute and cast it, but parsing an attribute consumes the colon and a type. This can be fixed by updating LLVM (there is now a `parseOptionalString` method), or better by adding a `parseString` method upstream (it's a trivial and hopefully uncontroversial change, but I need to look into the LLVM/MLIR contribution process).

The LLVM contribution process is very simple and straightforward. You should definitely try to upstream that.

> * Class types can quite long and repetitive. This is obviously only a problem in the text representation, not in memory. Type aliases would help, but due to how the parser is set up we cannot have aliases inside another Verona type. The parser should be fixed to allow this.

We shouldn't care too much about the textual representation of the IR. Verbose IR is better than implicit IR.

rengolin (approved):

plietar:
> At the moment, the syntax for classes is actually verona.class<"D", "f" : i1 : meet<U64, imm>>, where i1 can be any type and is ignored. This is an artefact of the restricted API available for parsing; the only way to parse a string is to parse an attribute and cast it, but parsing an attribute consumes the colon and a type. This can be fixed by updating LLVM (there is now a parseOptionalString method), or better by adding a parseString method upstream (it's a trivial and hopefully uncontroversial change, but I need to look into the LLVM/MLIR contribution process).

Fixed that thanks to the MLIR update.
I've rebased and added a couple of tests.

rengolin:
I'm touching similar files in #276 and will create a conflict. You should merge this one first, and I'll rebase the other one once you're done it.

PR #252, opened 2020-08-21
MLIR function refactory

rengolin:
These two changes implement function prototypes and refactor the type handling of both arguments and return values and fixes the 'void return' TODO.

### DIFF
--- a/src/mlir/generator.cc
+++ b/src/mlir/generator.cc
@@ -121,7 +121,10 @@ namespace mlir::verona
     if (auto err = proto.takeError())
       return std::move(err);
     auto& func = *proto;
-    auto retTy = func.getType().getResult(0);
+
+    // If just declaration, return the proto value
+    if (!hasFunctionBody(ast))
+      return func;
 
     // Create entry block
     auto& entryBlock = *func.addEntryBlock();
@@ -164,6 +167,7 @@ namespace mlir::verona
 
     // Return last value (or none)
     // TODO: Implement multiple return values for tuples
+    auto retTy = func.getType().getResult(0);
     if (last->hasValue() && last->get().getType() != retTy)
     {
       // Cast type (we trust the ast)
--- a/src/mlir/generator.cc
+++ b/src/mlir/generator.cc
@@ -101,10 +101,14 @@ namespace mlir::verona
     getFunctionArgs(args, ast);
     for (auto arg : args)
       types.push_back(parseType(getType(arg).lock()));
-    auto retTy = parseType(getFunctionType(ast).lock());
-    auto funcTy = builder.getFunctionType(types, retTy);
+
+    // Return type is nothing if no type
+    llvm::SmallVector<mlir::Type, 1> retTy;
+    if (hasType(getFunctionType(ast)))
+      retTy.push_back(parseType(getFunctionType(ast).lock()));
 
     // Create function
+    auto funcTy = builder.getFunctionType(types, retTy);
     auto func = mlir::FuncOp::create(getLocation(ast), name, funcTy);
     functionTable.insert(name, func);
     return func;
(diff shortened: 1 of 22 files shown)


rengolin:
ping. This is needed by two other PRs.

testsuite/mlir/mlir-parse/function.verona
```diff
@@ -1,23 +1,22 @@
 // Copyright Microsoft and Project Verona Contributors.
 // SPDX-License-Identifier: MIT
 
-// RUN: verona-mlir %s -o - | FileCheck %s
+empty_declaration();
```
mjp41, line 4: Why have you dropped the FileCheck stuff?

testsuite/mlir/mlir-parse/function.verona
```diff
@@ -1,23 +1,22 @@
...
+
+// Is this really valid?
+single_arg(x);
```
mjp41, line 7: I thought we required types on signatures? (Or at least should)

Theodus, line 7: My understanding is that this is not valid, unless x is a type name, since top level declarations should not require inference. I may be incorrect though.

src/mlir/generator.cc
```diff
@@ -164,21 +171,35 @@ namespace mlir::verona
...
+      {
+        last = genOperation(
+          last->get().getLoc(),
+          "verona.cast",
+          {last->get()},
+          retTy.getResult(0));
+        if (auto err = last.takeError())
+          return std::move(err);
+      }
+      builder.create<mlir::ReturnOp>(getLocation(ast), last->getAll());
```
mjp41, line 191: Cast only if different feels like a common pattern, should this be factored out?

mjp41 (approved):
LGTM. Couple of minor questions.

testsuite/mlir/mlir-parse/function.verona
rengolin: right, so I'd need to declare the type every time? I have changed the remaining tests to reflect that, but this is the `function` test, so needs to stress a bit more.

However, if I understand correctly, this is a parsing error, not a lowering error, so the AST should never have allowed this to be here in the first place.

If that's true, then I'll remove this test, but not raise an error in that case, assuming a future version of the parser will do so.

testsuite/mlir/mlir-parse/function.verona
rengolin: Because we're not using it, we're comparing directly with the output, like the other tests. This was a left-over from the move that I only did now.

testsuite/mlir/mlir-parse/function.verona
mjp41: I don't want to do interprocedural type inference.  So that is my view too.  We may relax that later for inference within a module for private members.

src/mlir/generator.cc
rengolin: It should, but not in this commit. I'm planning a bigger refactory to add more MLIR helpers, so that existing lowering functions (`parse...`) can reuse more code, and this will go into the pool. See #258.

src/mlir/generator.h
```diff
@@ -153,8 +152,6 @@ namespace mlir::verona
     mlir::Type allocaTy;
     /// Unknown types, will be defined during type inference.
     mlir::Type unkTy;
-    /// Temporary type to hold no types at all (ex: return void).
```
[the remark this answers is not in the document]
rengolin: None here was a replacement for `void`. Now we treat void as no values, rather than void values. This is the way MLIR understands void returns, too.

testsuite/mlir/mlir-parse/function.verona
rengolin: Ok, I'll drop this line from the test and check all others for similar patterns.

PR #254, opened 2020-08-24
Prevent teardown with external event source count

Theodus:
This change replaces the boolean that enables runtime teardown with an
atomic counter of external event sources. This is necessary to support
asynchronous I/O events in the runtime.


Theodus:
This is the first step for #255

mjp41 (approved):
LGTM

Issue #253, opened 2020-08-21
Add 'bool' type


rengolin:
We do have `true` and `false` on the grammar, but no `bool` type to indicate a return from a function, like:

```
check(a : S32) : bool;
...
if (check(x)) { ... }
```

Even though `bool` parses as a custom type and gives me `bool` in the MLIR (so I can correctly parse it), I think it should be a defined type for conditional construction among other things.

@sylvanc, does it make sense? Am I missing something?

rengolin:
Turns out, typing doesn't have special rules at all, only constants, so #260 should take care of the inconsistency for now, and we can think about `bool` versus `i1` when we get to type checking and inference.

PR #256, opened 2020-08-24
Implement for loops in MLIR

### LINKED ISSUE #203: Implement while/for loops in MLIR

rengolin:
This introduces `for` loops as MLIR constructs, similar to `while` but with the additional iteration handling.

A few following commits fix some issues found when calling functions (extra casts, no return values, arg types without args).

Depends on #252.

Fixes #203.

### DIFF
--- a/src/mlir/generator.cc
+++ b/src/mlir/generator.cc
@@ -668,8 +668,9 @@ namespace mlir::verona
     auto list = parseNode(seqNode);
     if (auto err = list.takeError())
       return std::move(err);
-    declareVariable("$iter", list->get());
-    llvm::SmallVector<mlir::Value, 1> iter{symbolTable.lookup("$iter")};
+    declareVariable(ABI::iteratorHandler, list->get());
+    llvm::SmallVector<mlir::Value, 1> iter{
+      symbolTable.lookup(ABI::iteratorHandler)};
 
     // Create the head basic-block, which will check the condition
     // and dispatch the loop to the body block or exit.
@@ -683,7 +684,7 @@ namespace mlir::verona
     // First node is a check on the lists `has_value` boolean return.
     builder.setInsertionPointToEnd(headBB);
     auto condLoc = getLocation(seqNode);
-    auto has_value = functionTable.lookup("has_value");
+    auto has_value = functionTable.lookup(ABI::iteratorHasValue);
     auto cond = builder.create<mlir::CallOp>(condLoc, has_value, iter);
     builder.create<mlir::CondBranchOp>(
       condLoc, cond.getResult(0), bodyBB, empty, exitBB, empty);
@@ -698,11 +699,11 @@ namespace mlir::verona
     // val must have been declared in outer scope
     builder.setInsertionPointToEnd(bodyBB);
     auto indVarName = getTokenValue(getLoopInd(ast));
-    auto apply = functionTable.lookup("apply");
+    auto apply = functionTable.lookup(ABI::iteratorApply);
     auto value = builder.create<mlir::CallOp>(condLoc, apply, iter);
     declareVariable(indVarName, value.getResult(0));
     //  %iter.next();
-    auto next = functionTable.lookup("next");
+    auto next = functionTable.lookup(ABI::iteratorNext);
     builder.create<mlir::CallOp>(condLoc, next, iter);
 
     // Loop body, branch back to head node which will decide exit criteria
--- a/src/mlir/generator.cc
+++ b/src/mlir/generator.cc
@@ -457,11 +457,16 @@ namespace mlir::verona
   llvm::Expected<ReturnValue> Generator::parseReturn(const ::ast::Ast& ast)
   {
     assert(isReturn(ast) && "Bad node");
+
     // TODO: Implement returning multiple values
-    // FIXME: Functions that don't return anything won't have a return value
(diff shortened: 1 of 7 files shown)


src/mlir/generator.cc
```diff
@@ -621,6 +652,75 @@ namespace mlir::verona
...
+    // All of that is within the scope of the loop, so we need to create a
+    // scope now, to drop them from future ops.
+    SymbolScopeT var_scope{symbolTable};
+
+    // First, we identify the iterator. Nested loops will have their own
+    // iterators, of the same name, but within their own lexical blocks.
+    auto seqNode = getLoopSeq(ast).lock();
+    auto list = parseNode(seqNode);
+    if (auto err = list.takeError())
+      return std::move(err);
```
mjp41, line 670: Do you need `std::move` here, can't it be elided?

src/mlir/generator.cc
```diff
@@ -621,6 +652,75 @@ namespace mlir::verona
...
+    mlir::ValueRange empty{};
+    auto headBB = addBlock(region);
+    auto bodyBB = addBlock(region);
+    auto exitBB = addBlock(region);
+    builder.create<mlir::BranchOp>(getLocation(ast), headBB, empty);
+
+    // First node is a check on the lists `has_value` boolean return.
+    builder.setInsertionPointToEnd(headBB);
+    auto condLoc = getLocation(seqNode);
+    auto has_value = functionTable.lookup("has_value");
```
mjp41, line 686: Mildly worried about hard coded strings here `"has_value"`

testsuite/mlir/mlir-parse/for/mlir.txt
```diff
@@ -0,0 +1,27 @@
...
+    %3 = "verona.constant(0)"() : () -> !type.int
+    %4 = "verona.store"(%3, %2) : (!type.int, !type.alloca) -> !type.unk
+    %5 = "verona.load"(%0) : (!type.alloca) -> !type.unk
+    br ^bb1
+  ^bb1:  // 2 preds: ^bb0, ^bb2
+    %6 = call @has_value(%5) : (!type.unk) -> i1
+    cond_br %6, ^bb2, ^bb3
+  ^bb2:  // pred: ^bb1
+    %7 = call @apply(%5) : (!type.unk) -> !type.S32
+    call @next(%5) : (!type.unk) -> ()
```
mjp41, line 21: Should `next` occur before or after the loop body? If it has side-effects then the two semantics are observable. Personally, I think after the body would lead to fewer confusions if it has side-effects.

mjp41 (approved):

src/mlir/generator.cc
```diff
@@ -216,6 +216,11 @@ namespace mlir::verona
     if (type)
       return type;
 
+    // Special treatment for type name 'bool'
+    // TODO: Should this be a special type in the grammar?
```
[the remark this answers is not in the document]
rengolin: See #253

src/mlir/generator.cc
rengolin: This is the idiom, repeated throughout the code. `err` is an `Error`, the return value is an `Expected`, I believe the elision only happens when the types are the same or a sub-type of, which is not the case.

src/mlir/generator.cc
rengolin: Me too. But I don't have a better idea for the moment. This is part of the language ABI, so it cannot be different (for whatever name we choose), but this knowledge should not be here, and should be in an ABI handler.

This ABI handler would not only know what the function names and signatures be, it would also declare them beforehand and return the signatures when requested by the lowering mechanism.

I haven't thought too much about such a mechanism and would like a larger discussion around this, as we'll need to add many more things around language ABI.

See #259.

For now, I'll define them as constants in the Generator class (like the types and other ABI stuff). We can move everything out later in a different PR.

testsuite/mlir/mlir-parse/for/mlir.txt
rengolin: I see what you mean, and I agree. One example is logging...
```
has_value -> true
apply -> 42
next -> 44
a = a + 42
has_value -> false
...

I'll change that.
```

src/mlir/generator.cc
mjp41: Thanks, I just remember having 
```
warning: moving a local object in a return statement prevents copy elision
```
messages with similar code.

testsuite/mlir/mlir-parse/for/mlir.txt
rengolin: Found a problem with this: if the block has an explicit terminator (cont, break, ret), I'd be emitting the "next" after that, which breaks the syntax (basic block with terminator in the middle and not at the end) and semantics (I could end up in a loop that never increments).

I'll keep this behaviour for now and we can worry about the side-effects when we get there.

PR #257, opened 2020-08-24
[NFC] Add pragma once to ast-utils.h

rengolin:
The header has no guard and would break compilation if used twice.
However, we also need to move it to a static struct, as the functions
would be defined twice, and break linkage.

This wasn't a problem before, when the header only declared the
functions (and ast-utils.cc would define them), but since we moved
implementation to the header (due to the use of templates on array
filling functions), we now have all definitions in the header.

This is a non-functional-change, that just makes it safer to use the
ast-utils on other files, when needed.

Needs to be applied on top of #256 for ordering issues.

### DIFF
--- a/src/mlir/ast-utils.h
+++ b/src/mlir/ast-utils.h
@@ -1,6 +1,8 @@
 // Copyright Microsoft and Project Verona Contributors.
 // SPDX-License-Identifier: MIT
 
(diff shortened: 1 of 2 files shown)


rengolin:
@theodus Thanks! I can only merge after #256 is merged.

PR #260, opened 2020-08-26
Add bool type in grammar, constants in MLIR

### LINKED ISSUE #253: Add 'bool' type

rengolin:
This makes the bool type like the others for constants, changing the AST
node from "true (true)" to "bool (true)", which makes more sense.

Also adding parsing rules to the MLIR generator to accept those constant
types.

Fixes #253

### DIFF
--- a/src/mlir/ast-utils.h
+++ b/src/mlir/ast-utils.h
@@ -65,6 +65,9 @@ namespace mlir::verona
       Type = peg::str2tag("type"), // = 4058008
       TypeRef = peg::str2tag("type_ref"), // = 2115269750
       TypeBody = peg::str2tag("typebody"), // = 2117081800
+      Bool = peg::str2tag("bool"), // = 3565102
+      Hex = peg::str2tag("hex"), // = 110293
+      Binary = peg::str2tag("binary"), // = 3884723791
       Params = peg::str2tag("params"), // = 4271185724
       NamedParam = peg::str2tag("namedparam"), // = 1544471404
       ID = peg::str2tag("id"), // = 3565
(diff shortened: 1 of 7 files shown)


src/mlir/ast-utils.h
```diff
@@ -151,7 +154,13 @@ namespace mlir::verona
     /// Return true if node is a value
     static bool isConstant(::ast::WeakAst ast)
     {
-      return isValue(ast) && isAny(ast, {NodeKind::Integer, NodeKind::Float});
+      return isValue(ast) &&
```
Theodus, line 157: Shouldn't string literals be included? Or are those not fully implemented yet?

rengolin: Not implemented yet. Strings in MLIR are a major PITA, I'll leave it for when we really need it...

davidchisnall: Strings are also complex to do right in the language. We want to avoid baking the standard library string type into the language if at all possible. I would like to have a sequence-of-UTF8-code-units literal that can be converted to string types, in the same way that we have a number literal that can be converted to any kind of integer, but I’m not sure how feasible that is. I’d also like to have a string interface in the standard library and avoid taking a dependency on a concrete string type anywhere. For low-level code, we will want a string that is a simple array of 8-bit values, for higher-level code we will want a rich Unicode string representation. The C++ standard library bakes a load of locale stuff in, which makes it very hard to run in embedded environments (e.g. in a kernel) where you expect to generate things that another component will localise if necessary. C goes in the opposite direction and gives a string that is unusable in any higher level applications (and is hard to optimise). Rust has many strings but doesn’t make it easy to mix them.

Issue #259, opened 2020-08-25
Create a language ABI handler


rengolin:
The language has some behaviour that needs a common place to live. The first example is `for`'s handling for iterative types (`has_value`, `apply`, `next`) in #256. Other examples will be that lambdas, constants, when blocks will need special anonymous objects with pre-defined shape, which the lowering needs to know how to lower both sides,

Another option would be to do that at the AST level (like current empty function names are converted to `apply`), and just let the MLIR lowering be as dumb as possible in regards to ABI.

We need to discuss how this will be implemented and at which level (AST, MLIR).

rengolin:
One of the problems raised in #261 is that ABI/Lowering handlers don't know where they're implemented. 

If they're implemented by the *compiler*, we need to define (not just declare) the function directly into the module.

If they're implemented by the *library*, we need do declare them in a header that is always included and make sure linking finds the runtime library.

If they're implemented by the *user*, we don't need to do anything and should just rely on the user having done the right thing, and erroring out if they didn't.

The ABI/Lowering handlers must do the right thing, independent of how it's implemented.

davidchisnall:
I think there are a few things conflated here that we might want to track separately:

 - Desugaring-style lowerings (e.g. expanding closures to objects with an `apply` method, calling `apply` for function-like invocations), where we're not locally defining the binary layout, but we are defining the equivalence of two source-level constructs.  These will be 100% shared between all targets.
 - Traditional ABI concerns, such as the layout of vtables (with selector colouring), packing of structures, and so on, which may vary between targets (some may be mostly generic, but depend on things like the representation of a pointer / virtual address for the target).
 - Contracts between the language and user code, such as how iteration loops get the range, how string literals turn into a concrete type, and so on (personally, I'd like as many of these as possible to be entirely library constructs, but that depends on sufficient source-level reflection working that we can implement for-in loops the standard library).
 - Contracts between the language and the standard library, in the form of built-in types and functions that are used to implement things that can't be expressed in Verona.

There are some overlaps between them but it's probably good to consider them separately.

rengolin:
Some de-sugaring happens at the AST level, for instance, replacing empty functions with `apply`. Some de-sugaring doesn't, for example, the `for` loop. We need to agree on what happens where. Without evidence, I think we should do **all** de-sugaring at the AST level. Happy to be proven wrong, but I'd like to treat the corner cases as special cases, not the rule.

[...]

PR #261, opened 2020-08-28
Generator refactory, splitting AST parsers from MLIR generators

### LINKED ISSUE #258: Refactor MLIR generator to add pure-MLIR builders
Currently, every `parse...` function takes in an `AST` node and spits out `mlir::Value`s. This is a simple pattern that works well initially, but soon enough (right about now, actually) it shows its main weakness: sharing code.

Common patterns, for example "casting when types differ" or "take the first result of a call" are obvious candidates, but there's also some reuse between `parseFor` and `parseWhile` that should really be one and only, called with MLIR nodes and types instead of AST nodes.

Another huge candidate will be for building lambdas (as an anonymous object with captures as fields, a constructor and a method `apply`, etc). This pattern will be reused for constants and `when` blocks at the very least, and the code is bound to be complicated.

We still need the `parse...(AST)` functions to prepare the MLIR values and then call the builder patterns, so that's not going to change. But once we get those builders working, we'll be able to reuse a lot more code.

rengolin:
This is a large change, moving a lot of code from repeated patterns in the parsers into more generic functions creating the MLIR nodes. But the functionality is still identical, and therefore, no new tests. The only change in the output is the unnecessary declaration of "for" loop iteration methods that are now automatically declared.

The purpose of this change is to:
 * Allow code reuse for compiler generated sequences, like `for` loops, `lambda`s, etc.
 * Allow code reuse across AST parsers (ex. `break` & `continue`, `load` and `store`, `constant`s, etc).
 * Introduce an ABI+Lowering layer to carry language-specific concepts independent of the lowering.

The ABI side isn't complete, by far. It's just a prototype of what we want to do. When we have a more complete Verona dialect (with types), we can start looking at defining the constructs with real types. We'll also need to encode the interaction between the concepts somehow (see #259).

Fixes #258.

### DIFF
--- a/src/mlir/generator.cc
+++ b/src/mlir/generator.cc
@@ -447,7 +447,7 @@ namespace mlir::verona
         "Return operation without parent function", getLocation(ast));
     }
 
-    // Either retrurns empty or auto-cast'ed value
+    // Either returns empty or auto-cast'ed value
     if (expr.hasValue())
     {
       assert(retTy && "Return value from a void function");
--- a/src/mlir/generator.cc
+++ b/src/mlir/generator.cc
@@ -68,6 +68,21 @@ namespace mlir::verona
     return llvm::Error::success();
   }
 
+  llvm::Expected<mlir::FuncOp> Generator::generateProto(
+    mlir::Location loc,
+    llvm::StringRef name,
+    llvm::ArrayRef<mlir::Type> types,
+    llvm::ArrayRef<mlir::Type> retTy)
+  {
+    assert(!functionTable.inScope(name) && "Duplicated function declaration");
+
+    // Create function
+    auto funcTy = builder.getFunctionType(types, retTy);
+    auto func = mlir::FuncOp::create(loc, name, funcTy);
+    functionTable.insert(name, func);
+    return func;
+  }
+
   llvm::Expected<mlir::FuncOp> Generator::parseProto(const ::ast::Ast& ast)
   {
     assert(AST::isFunction(ast) && "Bad node");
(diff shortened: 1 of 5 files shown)


rengolin:
Note: 'for' and 'while' lowering are still separate, as it doesn't seem particularly straightforward to do that (see #262).

rengolin:
GCC didn't warn me of a copy&paste error leading to a function without explicit return. Long live clang!

rengolin:
Gentle ping

src/mlir/generator.cc
```diff
@@ -42,14 +43,53 @@ namespace mlir::verona
...
+    symbolTable.update(name, val);
+  }
+
+  void Generator::declareFunction(
+    llvm::StringRef name,
+    llvm::ArrayRef<llvm::StringRef> types,
+    llvm::StringRef retTy)
+  {
+    // If already declared, ignore
+    if (functionTable.inScope(name))
```
mjp41, line 72: Should we check the declaration matches?

src/mlir/generator.cc
```diff
@@ -480,19 +447,13 @@ namespace mlir::verona
         "Return operation without parent function", getLocation(ast));
     }
 
+    // Either retrurns empty or auto-cast'ed value
```
mjp41, line 450: returns

src/mlir/generator.cc
```diff
@@ -716,36 +697,162 @@ namespace mlir::verona
...
+  }
+
+  llvm::Error
+  Generator::generateLoopBranch(mlir::Location loc, llvm::StringRef blockName)
+  {
+    if (!loopTable.inScope(blockName))
+      return parsingError("Loop branch without a loop", loc);
+    auto block = loopTable.lookup(blockName);
+    mlir::ValueRange empty{};
+    // We assume the branch is the last operation in its basic block
```
mjp41, line 798: Can we check this assumption in debug?

src/mlir/generator.h
```diff
@@ -46,6 +46,10 @@ namespace mlir::verona
     /// Default constructor, builds an empty return value.
     ReturnValue() {}
     /// Single value constructor.
```
mjp41, line 48: Comment moved function body? Is that correct, should both functions have the same comment.

src/mlir/generator.cc
```diff
@@ -480,19 +447,13 @@ namespace mlir::verona
...
     if (expr.hasValue())
     {
       assert(retTy && "Return value from a void function");
       // Emit the cast if necessary (we trust the ast)
-      if (expr.get().getType() != retTy)
-      {
-        auto op =
-          genOperation(expr.get().getLoc(), "verona.cast", {expr.get()}, retTy);
-        if (auto err = op.takeError())
-          return std::move(err);
```
mjp41, line 492: The error handling code here is removed by the refactor.

mjp41 (approved):
LGTM, couple of minor queries.

src/mlir/generator.cc
rengolin: This is intentional. The code could never return an error.

`llvm::Expected<>` and `llvm::Error` assert if the errors are not treated, so there's no way we can use a value that *could* error without handling it.

src/mlir/generator.cc
rengolin: We check this by running the `mlir::verify()` function at the end of the module generation, on both debug and release builds.

src/mlir/generator.h
rengolin: Actually, that was meant to be a replacement. `mlir::Value` is meant to always be passed by value. Removed the reference version.

But yes, the comment is still relevant.

src/mlir/generator.cc
rengolin: oops

Issue #226, opened 2020-08-05
Improvements to backpressure


Theodus:
- [ ] 
- [ ] 
- [ ] 
- [ ]

PR #263, opened 2020-08-31
Split backpressure bits into status and state

Theodus:
This change splits the `Backpressure` bits for each cown into two parts:
1. `BackpressureState` - may be modified from multiple threads, and therefore requires compare exchanges for modification
2. `Status` - storage bits for load heuristics,  token message tracking, and future I/O related things.

The large piece of documentation for backpressure has been removed, in favor of a complete rewrite. Some major changes are likely to come soon, and the documentation has diverged quite a bit from both the model and the implementation already. #226 has been updated to include new documentation.


src/rt/sched/status.h
```diff
@@ -0,0 +1,123 @@
...
+      case BackpressureState::Normal:
+        return os << "Normal";
+      case BackpressureState::Muted:
+        return os << "Muted";
+      case BackpressureState::Unmutable:
+        return os << "Unmutable";
+      case BackpressureState::MaybeUnmutable:
+        return os << "MaybeUnmutable";
+      default:
+        return os << "??";
```
rengolin, line 40: According to the enum, this would be `IsUnmutable`. Why `??`?

src/rt/sched/status.h
```diff
@@ -0,0 +1,123 @@
...
+    }
+  }
+
+  /// Tracks status information for a cown. This class may only be modified by
+  /// the scheduler thread running a cown.
+  /// TODO: possibly I/O thread when cown is uncheduled
+  class Status
+  {
+    // Load at which cown is overloaded.
+    static constexpr uint32_t overload_threshold = 800;
```
rengolin, line 50: I imagine this could be dynamic depending on hardware specs, types of queues (ex. I/O vs. msg). This should probably be a multiplier `8:1` over a formula for the types, which could be hard-coded to `100` for now.

src/rt/sched/schedulerthread.h
```diff
@@ -216,22 +216,21 @@ namespace verona::rt
...
-        auto bp = cown->backpressure.load(std::memory_order_relaxed);
+        auto bp = cown->bp_state.load(std::memory_order_relaxed);
         yield();
-        assert(!bp.muted());
+        assert(bp != BackpressureState::Muted);
 
         if (
-          (bp.unmutable()) || (state == ThreadState::PreScan) ||
-          (state == ThreadState::Scan) || (state == ThreadState::AllInScan))
+          (bp & BackpressureState::IsUnmutable) ||
```
rengolin, line 224: This is not clear from the enum declaration and is left as an implementation detail for anyone to either abuse or get it wrong. The lack of comments on the enum makes this issue worse.

Either hide the implementation on static methods and opaque enum (aka. unsigned) or make it explicit on the enum declaration that this is intentional and assert on places that write to `bp_state` that it cannot ever be `0b10`.

src/rt/sched/status.h
rengolin: Now I get it, `IsImutable` is just a flag test and not an actual value. This is confusing, as I explained below. The fact that this isn't an assert adds up to the confusion with the lack of documentation and opaque implementation.

src/rt/sched/schedulerthread.h
Theodus: I've added doc strings to the enum members. The `operator<<` on `BackpressureState` now aborts on the default branch, and the `check_backpressure_transition` would return false when either state is `IsUnmutable`.

rengolin:
LGTM, I'll let @mjp41 review and approve it.

src/rt/sched/cown.h
```diff
@@ -906,79 +934,51 @@ namespace verona::rt
...
                            << std::endl;
-        assert(bp.has_token());
-        token_reached = true;
+        assert(stat.has_token == 1);
+        stat.has_token = 0;
+        status.store(stat, std::memory_order_release);
+
+        auto bp = bp_state.load(std::memory_order_acquire);
+        if (bp == BackpressureState::Unmutable)
+          backpressure_transition(BackpressureState::Unmutable);
```
mjp41, line 953: Bit confused by
```C++
 backpressure_transition(BackpressureState::Unmutable);
```
Shouldn't that be
```
 backpressure_transition(BackpressureState::MaybeUnmutable);
```
otherwise, I don't see how we get back to `Normal`?

Theodus: You are correct, that was a typo

PR #266, opened 2020-09-02
WIP Refactor of C++ API for RT unit tests.

mjp41:
The C++ unit tests were using `new` to allocate Verona objects, but
were partially initialising the object using the Verona RT, and then
letting the C++ constructor fill in the rest.  On GCC, this was
triggering dead-store elimination for the Verona RT initialisation and
causing crashes.

This change makes a collection of static methods that build the Verona
object, with an embedded C++ object.

This commit only updates one of the unit tests as an experiment to see
how it looks with this alternative API.

### DIFF
--- a/src/rt/cpp/vobject.h
+++ b/src/rt/cpp/vobject.h
@@ -21,161 +21,239 @@ namespace verona::rt
   // This is better than ignoring methods with the right name but the wrong
   // signature.
   template<class T, class = void>
-  struct has_notified : std::false_type
+  struct has_notified_cpp : std::false_type
   {};
   template<class T>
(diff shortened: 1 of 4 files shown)


mjp41:
Could I get feedback on this proposed change to the C++ unit tests for the Verona runtime.  I only updated a single unit test as a confirmation that the refactor works.  If you are happy, I will push through the full change.

rengolin:
Awesome! When you finish, can you try and re-enable the GCC tests? If they pass, we should merge all together and make sure we don't regress in the future.

mjp41:
This didn't work out nicely enough.  Currently exploring another refactor.  Will close this for now.

PR #268, opened 2020-09-03
Update backpressure progress

Theodus:
This brings the backpressure implementation up to date with the model, where overloaded or unmutable message participants result in the other participants becoming unmutable as well. This only applies when the fast send is interrupted, since that is the only time a recipient may be muted prior to running the message.

### DIFF
--- a/src/rt/sched/cown.h
+++ b/src/rt/sched/cown.h
@@ -540,7 +540,6 @@ namespace verona::rt
       assert(body->index <= last);
 
       const auto make_unmutable = backpressure_ensure_progress(body);
-
       for (; body->index < body->count; body->index++)
       {
         MultiMessage* m = MultiMessage::make_message(alloc, body, epoch);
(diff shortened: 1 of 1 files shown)


src/rt/sched/cown.h
```diff
@@ -535,42 +535,40 @@ namespace verona::rt
...
+        bool needs_scheduling = next->send<YesTransfer, YesTryFast>(m);
         if (!needs_scheduling)
         {
           // Case 1: target cown was already scheduled.
+          // The message may have beeen freed at this point.
           Systematic::cout()
-            << "MultiMessage " << m << " fast send interrupted" << std::endl;
+            << "MultiMessage " << m << ": fast send interrupted" << std::endl;
+          if (make_unmutable)
+            next->unmute(true);
```
mjp41, line 559: I'm worried `next` could have been deallocated here.

mjp41: Holding an epoch from 551 -> 559 should fix the issue.

mjp41 (changes requested):

src/rt/sched/cown.h
Theodus: fixed

mjp41 (approved):
LGTM

PR #269, opened 2020-09-03
Ci fixes

mjp41:
This fixes some issues for CI.  See individual commits for details.


rengolin (approved):
LGTM, thanks!

Just a question: what's the point of excluding the external projects from the targets? Aren't they going to be picked up anyway as dependency of the Verona targets?

mjp41:
> Just a question: what's the point of excluding the external projects from the targets? Aren't they going to be picked up anyway as dependency of the Verona targets?

It means that any unit tests or other target that is there, but we don't use are not built.  Pegamatite was building a static lib, we weren't using.  The others didn't make a difference, but it seemed good for hygiene.

PR #276, opened 2020-09-08
Generating verona types from AST

### LINKED ISSUE #274: Lower types to Verona dialect (incl. meet/join)
Move away from `!type.<str>` into `!verona.<str>` and use the validation baked in those types already to check basic syntax.

Need to lower `&` and `|` into `meet` and `join`, too.

rengolin:
Not yet converting `where` types to proper Verona types, but all others (integer, float, join, meet) including nested operations.

Added Float and Bool types in the dialect to complete the support needed for this change.

Fixes #274

### DIFF
--- a/src/mlir/ast-utils.h
+++ b/src/mlir/ast-utils.h
@@ -63,7 +63,10 @@ namespace mlir::verona
       Constraints = peg::str2tag("constraints"), // = 4070926742
       Constraint = peg::str2tag("constraint"), // = 2466070853
       Type = peg::str2tag("type"), // = 4058008
+      TypeTuple = peg::str2tag("type_tuple"), // = 1428239871
       TypeRef = peg::str2tag("type_ref"), // = 2115269750
+      TypeOp = peg::str2tag("type_op"), // = 4098764984
+      TypeOne = peg::str2tag("type_one"), // = 2115257603
       TypeBody = peg::str2tag("typebody"), // = 2117081800
       Bool = peg::str2tag("bool"), // = 3565102
       Hex = peg::str2tag("hex"), // = 110293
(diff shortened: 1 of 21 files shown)


rengolin:
@plietar, merged and rebased. Good to go.

plietar:
In the long run, I'm not sure we should be treating Bool/Integer/Float as special kinds of types in the early stages of the compiler. Instead they should be considered "classes" with no fields. For the vast majority of the compiler they look just like that. In particular, it means we can define methods on primitives. It's only when we start to lower to concrete bits and structs that it matters. 

The old compiler has a notion of "primitive" types, such as [`U64`](https://github.com/microsoft/verona/blob/c325e82b4305803ff9805ab1348f79ea4e45610e/src/stdlib/builtin.verona#L95-L111). Classes and Primitive types are all described as "entities" in the implementation and treated, for the most part, uniformly.

I was happy with IntegerType because it started off as a demo of what an MLIR type looks like. I can get behind having a BoolType as a temporary measure, since we want bools working so we can have control flow. I really don't see the point of FloatType.

rengolin:
We still need to lower types to their respective counterparts in machine code, and matching "a class that has a name that sounds like floating point and a field that looks like a bit width" doesn't really cut it. If we try to go further and define a class that is actually a float with a specified bit width, then we have just created a float type. Might just as well have a native float type.

In discussions with @sylvanc, the idea is that concrete numeric types and boolean really are what they are. If we try to make them too fancy (like Rust) too soon, we may end up with with a compiler that has lots of potentially great ideas but none of them actually working.

Again, I think that's a conversation we should have later. Right now we're lowering different parts of the compiler together and I need to lower more constructs before I can start lowering classes with all the constructs that are required. We still haven't defined class layout, name mangling, and a lot of other things that we have issues for but haven't reached consensus. I can't wait until all of those things are defined to implement now the infrastructure that the compiler will rely on in the future.

We have to start with something, and this is something that works now. You have just started working with classes now and it will take a while for the implementation to settle. I still have a lot of other things to do until then.

plietar (approved):

PR #277, opened 2020-09-08
Improve parsing and printing of types

plietar:
Struct types can now be recursive, following the approach taken by the LLVM dialect.
Additionally, Verona types may now refer to type aliases, which helps keep test cases concise.

For clarity, the parsing and printing code is moved to a separate file, which makes the PR a little harder to review. The first commit does the move with no functional change. Later commits add the new features.


testsuite/mlir/mlir-parse/dialect.mlir
```diff
@@ -23,38 +23,40 @@
 // }
 // ```
 
+!C = type !verona.class<"C">
+!D = type !verona.class<"D", "f" : meet<U64, imm>, "g" : meet<!C, mut>>
```
rengolin, line 27: If the `!verona.class<>` completely defines the class, why do we need the dialect `verona.class { }` construct with the same info? Seems to me that if we don't keep them in sync, weird stuff is going to happen.

testsuite/mlir/mlir-parse/dialect.mlir
```diff
@@ -23,38 +23,40 @@
...
 module {
   verona.class @C {
   }
 
   verona.class @D {
     verona.field "f" : !verona.meet<U64, imm>
-    verona.field "g" : !verona.meet<class<"C">, mut>
+    verona.field "g" : !verona.meet<!C, mut>
```
rengolin, line 34: From the lack of changes in the output, I imagine this is identical to the previous syntax, which reinforces my point above.

testsuite/mlir/mlir-parse/dialect.mlir
```diff
@@ -23,38 +23,40 @@
...
   }
 
   func @bar(%x: !verona.meet<U64, imm>, %y: !verona.meet<U64, imm>) {
-    %a = verona.new_region @C [ ] : !verona.meet<class<"C">, iso>
-    %b = verona.view %a : !verona.meet<class<"C">, iso> -> !verona.meet<class<"C">, mut>
+    %a = verona.new_region @C [ ] : !verona.meet<!C, iso>
+    %b = verona.view %a : !verona.meet<!C, iso> -> !verona.meet<!C, mut>
```
rengolin, line 39: This is **so** much nicer than having to redefine `class<"C">` and `class<"D">` everywhere.

rengolin (commented):
I think we could deprecate the previous class declaration (`verona.class { }`) and have their types only (`!verona.class<>`).

For methods (static and dynamic) we'll use name mangling anyway, which won't need to be inside the class regions to work.

plietar:
> I think we could deprecate the previous class declaration (verona.class { }) and have their types only (!verona.class<>).

At the moment, the `verona.class` operations and types are completely unrelated. Originally I was going to have the type point to the operation, but this turned out to be messy because MLIR isn't designed to have symbols in types.

The two things have different purposes. The type only describes the class' signature, whereas the operation describes its implementation. We could skip the operation, and rely solely on name mangling, but by doing that we miss a thing to group all the mangled methods together.

Some very vague plan I have in mind is to use the class operation as what eventually lowers to the type descriptor

```mlir
verona.class @A {
  verona.method m(x: T): T { /* method body */ }
}

func @main() {
  verona.allocate @A : meet<!verona.class<"A", "m": T -> T>, iso>
}
```

Lowers to

```mlir
verona.descriptor @A {
  verona.vtable_entry "m" @A_m
}
func @A_m(x: T): T { /* method body */ }
func @main() {
  verona.allocate @A : meet<!verona.class<"A", "m": T -> T>, iso>
}
```

Lowers to (in C syntax)

```c
struct Descriptor {
  const char* name;
  Method* methods;
};

Method[] A_methods = { { "m", &A_m } };
Descriptor A_desc = {
  .name = "A";
  .methods = A_methods;
};
T A_m(T x) { /* method body */ }

```

testsuite/mlir/mlir-parse/dialect.mlir
plietar: The allocate operation is what will tie them together. You can only allocate an object by giving it an implementation that has the right signature.

rengolin: That's what I don't like. If we change the implementation of the class (say extract fields to create lists), we'll have to change the operations and the types in the exact same way. It's twice the amount of work for no additional benefit.

rengolin:
> The two things have different purposes. The type only describes the class' signature, whereas the operation describes its implementation. We could skip the operation, and rely solely on name mangling, but by doing that we miss a thing to group all the mangled methods together.

Why do we need a thing to group things together?

We have a type that describes the shape. This allows us to create a structure with the right fields (including the vtable) for allocation. All methods have the same mangling "classX.classY....method". Once lowered, all calls will be to the mangled names, so we don't care which class they were originally from.

`when` blocks will tie cowns (regions that are parametrised by types, including classes), so we'll know what goes together.


> Method[] A_methods = { { "m", &A_m } };
> Descriptor A_desc = {
>   .name = "A";
>   .methods = A_methods;
> };
> T A_m(T x) { /* method body */ }
> ```

These can still be lowered at compile time from the type information. Once that's done, the IR doesn't need to know what was where originally. For source-to-source transformations we still have the source location being propagated, so nothing is lost.

plietar:
I think it's nice to have "something" that represents the class definition and groups the method bodies together, but I'm not that attached to it, so I wouldn't mind getting rid of it. It was more an experiment than anything else.

On the other hand, I think that discussion is orthogonal to this PR (I had already raised this point in #249). The PR just made it more obvious through the introduction of type aliases.

rengolin (approved):
> I think it's nice to have "something" that represents the class definition and groups the method bodies together, but I'm not that attached to it, so I wouldn't mind getting rid of it. It was more an experiment than anything else.

Right. I strongly believe we should remove anything that feels unnecessary at the first opportunity. If turns out we needed, we can always add later, with 20/20 hind-sight and access to the previous code in git's history. The cost is much lower than keeping multiple semantics alive and have to work around them.

> On the other hand, I think that discussion is orthogonal to this PR (I had already raised this point in #249). The PR just made it more obvious through the introduction of type aliases.

Agreed. This can come as a follow up PR. Thanks!

PR #278, opened 2020-09-09
WIP: Initial implementation of the region checker.

plietar:
Still pretty early days. At the moment it only computes the `alias` relation, and doesn't use the result at all.

It introduces a pretty generics datalog-like engine in `Query.h`.
The engine is used in `RegionChecker.cc` to implement the fact derivation rules.

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/verona/278)

### DIFF
--- a/src/mlir/Query.h
+++ b/src/mlir/Query.h
@@ -1,0 +1,556 @@
+// Copyright Microsoft and Project Verona Contributors.
+// SPDX-License-Identifier: MIT
+
+#pragma once
+
+#include "llvm/ADT/ArrayRef.h"
+
+#include <optional>
+#include <tuple>
+
+namespace mlir::verona
+{
+  /// Index provides a comparator that orders tuples in lexicographical order,
+  /// but following a given field sequence. For example, when applied to pairs,
+  /// `Index<_, 1, 0>` will first compare the second elements, and only if they
+  /// match will it compare the first ones. This is used to provide custom
+  /// orderings for tuples stored in Relations.
+  ///
+  /// The first type parameter, `Compare`, is used as the underlying comparator
+  /// for elements of the tuples. The same comparator type must be compatible
+  /// with all constituent types.
+  ///
+  /// Index supports both `std::tuple` and any class which defines a
+  /// `data() const` which returns a tuple. This allows wrappers around tuples
+  /// to be used with Relation.
+  template<typename Compare, size_t... Keys>
+  struct Index;
+
+  /// This is the base case, for when no key is provided. Since there is nothing
+  /// to compare, all values are considered equal, ie. (left < right) == false.
+  template<typename Compare>
+  struct Index<Compare>
+  {
+    template<typename T>
+    bool operator()(const T& left, const T& right) const
+    {
+      return false;
+    }
+  };
+
+  template<typename Compare, size_t Head, size_t... Tail>
+  struct Index<Compare, Head, Tail...>
+  {
+    template<typename... Ts>
+    bool
+    operator()(const std::tuple<Ts...>& left, const std::tuple<Ts...>& right)
+    {
+      const auto& left_value = std::get<Head>(left);
+      const auto& right_value = std::get<Head>(right);
+      if (Compare()(left_value, right_value))
+        return true;
+      else if (Compare()(right_value, left_value))
+        return false;
+      else
+        return Index<Compare, Tail...>()(left, right);
+    }
+
+    /// Retrieve the "primary" element of the tuple, that is the one that is
+    /// used first during comparisons.
+    template<typename... Ts>
+    static const auto& get_primary(const std::tuple<Ts...>& value)
+    {
+      return std::get<Head>(value);
+    }
+
+    template<typename T>
+    bool operator()(const T& left, const T& right)
+    {
+      return (*this)(left.data(), right.data());
+    }
+
+    template<typename T>
+    static const auto& get_primary(const T& value)
+    {
+      return get_primary(value.data());
+    }
+
+    /// Underlying comparator used on the contents of the tuple.
+    using value_compare = Compare;
+  };
+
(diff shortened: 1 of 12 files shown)


src/mlir/Query.h
```diff
@@ -0,0 +1,556 @@
...
+    static constexpr bool is_index_v = is_index<T>::value;
+  }
+
+  struct QueryEngine;
+
+  /// Type-erased version of a Relation, used to be stored in a QueryEngine.
+  struct AbstractRelation
+  {
+    AbstractRelation(QueryEngine& engine);
+    virtual bool iterate() = 0;
```
rengolin, line 103: Avoid `virtual` classes as they require RTTI and we'll disable RTTI+EH as soon as we don't depend on it in peglib.

src/mlir/Query.h
```diff
@@ -0,0 +1,556 @@
...
+  };
+
+  template<typename Compare, size_t Head, size_t... Tail>
+  struct Index<Compare, Head, Tail...>
+  {
+    template<typename... Ts>
+    bool
+    operator()(const std::tuple<Ts...>& left, const std::tuple<Ts...>& right)
+    {
+      const auto& left_value = std::get<Head>(left);
```
rengolin, line 48: use `get_primary`?

rengolin (commented):
Just some quick code comments.

rengolin:
Just to understand, this is a "Verona Region" checker, not an "MLIR Region" checker, right?

When will this checker pass? Every time `verify` is called? Every time those instructions are created? Only once, as a pass in the pipeline?

To me it looks like this is a very expensive check and we'd profit from not running it all the time. Perhaps, once validated, we can force all Verona passes to never break the contract? Or add some annotation to detect breakages?

src/mlir/Query.h
```diff
@@ -0,0 +1,556 @@
...
+  struct AbstractRelation
+  {
+    AbstractRelation(QueryEngine& engine);
+    virtual bool iterate() = 0;
+    virtual ~AbstractRelation(){};
+
+    // We don't allow copying nor moving an AbstractRelation, because the
+    // QueryEngine holds a pointer to it.
+    AbstractRelation(const AbstractRelation&) = delete;
+    AbstractRelation& operator=(const AbstractRelation&) = delete;
```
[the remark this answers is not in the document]
plietar: I believe move constructors are implicitely deleted when copy constructors are.

plietar:
> Just to understand, this is a "Verona Region" checker, not an "MLIR Region" checker, right?

Yes, it's "Verona Region". The name collision between the two is very unfortunate.

> When will this checker pass? Every time verify is called? Every time those instructions are created? Only once, as a pass in the pipeline?
> To me it looks like this is a very expensive check and we'd profit from not running it all the time. Perhaps, once validated, we can force all Verona passes to never break the contract? Or add some annotation to detect breakages?

The pass is intentionally not efficient yet, since I expect we'll be making a lot of changes to it in the weeks to come. There's a number of low hanging fruits that would help performance immensely (eg. use a union-find structure to track aliases) but make changes more complicated. Nevertheless, this will never be considered a "cheap" pass.

In a release build, this should only be run once, quite early in the pipeline (probably just after type inference). In a debug build, I would leave it enabled after every pass for now. If we notice that it significantly hinders development time we can reconsider that choice.

src/mlir/Query.h
plietar: Does `virtual` really require RTTI? I thought that was only for `dynamic_cast<>`. MLIR is full of virtual methods.

src/mlir/Query.h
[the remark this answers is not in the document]
rengolin: You're right. You have to implement the move c-tor if you delete the copy-ctor.

src/mlir/Query.h
rengolin: Interesting. My impression from many years ago is that compilers would indeed generate RTTI tables from virtual methods. Either this was a part of the standard that was never defined and compilers got smarter, or my memory is just misplaced.

rengolin:
> The pass is intentionally not efficient yet, since I expect we'll be making a lot of changes to it in the weeks to come. There's a number of low hanging fruits that would help performance immensely (eg. use a union-find structure to track aliases) but make changes more complicated. Nevertheless, this will never be considered a "cheap" pass.

Right, I was referring to the fastest possible version. It will still be an expensive pass.

> In a release build, this should only be run once, quite early in the pipeline (probably just after type inference). In a debug build, I would leave it enabled after every pass for now. If we notice that it significantly hinders development time we can reconsider that choice.

Ack.

src/mlir/Query.h
davidchisnall: `virtual` doesn't require RTTI.  Both GCC and clang will still generate the two RTTI-related vtable entries (for the offset to the superclass vtable and the `type_info` structure) but will set them both to `nullptr` when RTTI is disabled.

rengolin:
@plietar, this review is a month old and very outdated. Also, it seems we're moving towards a region check pass to happen at the same time (or even on the same) type checking pass. Is this PR still relevant?

plietar:
> @plietar, this review is a month old and very outdated.

Sorry I've gone on a few tangents over the past few weeks. I do intend to make progress at some point, it's just taken some time.

>  Also, it seems we're moving towards a region check pass to happen at the same time (or even on the same) type checking pass.

I don't think there's any proper consensus on this. I have a pretty clear idea of how I'd do things as separate passes, just need more work and time to implement. The best we have for a combined type- and region-checking is hand waving examples.

rengolin:
Fair enough. I just wanted to make sure you don't work towards a goal that may eventually be rejected. Make sure you and @sylvanc are on the same page on this.

plietar:
Alright, I'm closing this for now until we figure out the details

Issue #165, opened 2020-06-30
GCC built runtime tests fail with corruption


rengolin:
After cleaning up for warnings (#164), the GCC build still fails. The following tests crash:
	  7 - func-con-ext_ref (Child aborted)
	 11 - func-con-finalisers (Child aborted)
	 14 - func-con-memory (Child aborted)
	 25 - perf-con-linkedlist (Child aborted)

This has been present since the first commit, so not a "regression" per se, but a bug nonetheless. With the tests broken, we'll have to disable the GCC build on the CI, which is a bad thing.

Fixing this problem while we're still fresh would be better than leaving this to hang for months and then deciding to never build with GCC any more.

@theodus @mjp41

rengolin:
I'm using GCC 10.1 on Arch and I cannot make it fail... :/

Theodus:
Have you tried with `-DRT_TESTS=ON`?

rengolin:
yeah:
```
verona/build$ cmake .. -GNinja -DRT_TESTS=On -DEXTRA_VERONA_CMAKE_FLAGS="-DCMAKE_C_COMPILER=gcc;-DCMAKE_CXX_COMPILER=g++;-DCMAKE_CXX_FLAGS=-stdlib=libstdc++"
verona/build$ ninja
verona/build/verona-prefix/src/verona-build$ ctest -j 32 -R -con-
```

rengolin:
@theodus, I thought you had fixed the build error, but left the runtime test error.

A recent [nightly build](https://dev.azure.com/ProjectVeronaCI/Project%20Verona/_build/results?buildId=1286&view=logs&j=58c83696-1038-5e38-29b3-1a10c75b2938&t=ea4bd8d5-ac33-5318-3408-4424411dc055&l=280) still shows build errors.

That build uses GCC 7.5.

Theodus:
Those changes were never merged, since the CI could never pass

PR #282, opened 2020-09-10
Refactor to store meta-data before Object*

mjp41:
GCC was applying dead-store-eliination to the region-meta data
as it was part of Object, and assigned before the C++ constructor.
This changes the representation of Object, to not contain the meta
data, but access the meta-data directly before the Object.

There are some defensive code paths to try to ensure that each Object*
actually has the meta-data before it.

### DIFF
--- a/src/rt/cpp/vobject.h
+++ b/src/rt/cpp/vobject.h
@@ -79,12 +79,17 @@ namespace verona::rt
       ((T*)o)->~T();
     }
 
-    static constexpr Descriptor desc = {
-      vsizeof<T>,
-      gc_trace,
-      has_finaliser<T>::value ? gc_final : nullptr,
-      has_notified<T>::value ? gc_notified : nullptr,
-      has_destructor<T>::value ? gc_destructor : nullptr};
+    static Descriptor* desc()
+    {
+      static Descriptor desc = {
+        vsizeof<T>,
+        gc_trace,
+        has_finaliser<T>::value ? gc_final : nullptr,
+        has_notified<T>::value ? gc_notified : nullptr,
+        has_destructor<T>::value ? gc_destructor : nullptr};
+
+      return &desc;
+    }
 
     void trace(ObjectStack&) {}
 
@@ -99,33 +104,33 @@ namespace verona::rt
     void* operator new(size_t)
     {
       if constexpr (std::is_same_v<Base, Object>)
-        return RegionClass::template create<desc.size>(
-          ThreadAlloc::get(), &desc);
+        return RegionClass::template create<vsizeof<T>>(
+          ThreadAlloc::get(), desc());
       else
         return Object::register_object(
-          ThreadAlloc::get()->alloc<desc.size>(), &desc);
+          ThreadAlloc::get()->alloc<vsizeof<T>>(), desc());
     }
 
     void* operator new(size_t, Alloc* alloc)
     {
       if constexpr (std::is_same_v<Base, Object>)
-        return RegionClass::template create<desc.size>(alloc, &desc);
+        return RegionClass::template create<vsizeof<T>>(alloc, desc());
       else
-        return Object::register_object(alloc->alloc<desc.size>(), &desc);
+        return Object::register_object(alloc->alloc<vsizeof<T>>(), desc());
     }
 
     void* operator new(size_t, Object* region)
     {
       static_assert(std::is_same_v<Base, Object>, "Don't call on cowns.");
-      return RegionClass::template alloc<desc.size>(
-        ThreadAlloc::get(), region, &desc);
+      return RegionClass::template alloc<vsizeof<T>>(
+        ThreadAlloc::get(), region, desc());
     }
 
     void* operator new(size_t, Alloc* alloc, Object* region)
     {
       static_assert(std::is_same_v<Base, Object>, "Don't call on cowns.");
       if constexpr (std::is_same_v<Base, Object>)
-        return RegionClass::template alloc<desc.size>(alloc, region, &desc);
+        return RegionClass::template alloc<vsizeof<T>>(alloc, region, desc());
     }
 
     void operator delete(void*)
--- a/src/rt/cpp/vobject.h
+++ b/src/rt/cpp/vobject.h
@@ -57,12 +57,22 @@ namespace verona::rt
     {
       if constexpr (has_notified<T>::value)
         ((T*)o)->notified(o);
+      else
+      {
+        UNUSED(o);
+      }
     }
 
     static void gc_final(Object* o, Object* region, ObjectStack& sub_regions)
     {
       if constexpr (has_finaliser<T>::value)
         ((T*)o)->finaliser(region, sub_regions);
+      else
+      {
+        UNUSED(o);
+        UNUSED(region);
+        UNUSED(sub_regions);
+      }
     }
 
     static void gc_destructor(Object* o)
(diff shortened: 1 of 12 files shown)


src/rt/cpp/vobject.h
```diff
@@ -99,40 +94,38 @@ namespace verona::rt
...
     }
 
     void* operator new(size_t, Object* region)
     {
-      if constexpr (std::is_same_v<Base, Object>)
-        return RegionClass::template alloc<sizeof(T)>(
-          ThreadAlloc::get(), region, desc());
-      else
-        return ThreadAlloc::get()->alloc<sizeof(T)>();
+      static_assert(std::is_same_v<Base, Object>, "Don't call on cowns.");
```
davidchisnall, line 119: I don't understand how the expression relates to the message, can you add a comment?  (Or extend the comment at the top of the class to explain the valid values of `Base`)

src/rt/sched/cown.h
```diff
@@ -50,9 +50,31 @@ namespace verona::rt
       YesTryFast
     };
 
-    Cown(const Descriptor* desc) : Object(desc)
+    Cown(bool initialise = true)
```
davidchisnall, line 53: Can this be a template parameter and then line 57 be a `constexpr` if?

davidchisnall (approved):
Looks fine, a few small comments.

src/rt/cpp/vobject.h
mjp41: Base is either Cown or Object (not enforced).  This can, and will, be significantly improved for clarity.

src/rt/sched/cown.h
mjp41: I was thinking about that, and then forgot.

src/rt/object/object.h
```diff
@@ -150,6 +150,10 @@ namespace verona::rt
     YesTransfer
   };
 
+  /// No fields as part of the C++ representation all meta-data stored before
+  /// object in Verona runtime header. Object should not be allocated directly,
```
Theodus, line 154:
```suggestion
  /// The C++ representation of objects has no fields. All meta-data for the
  /// object is in the `Header` struct. Object should not be allocated directly,
```

src/rt/object/object.h
```diff
@@ -150,6 +150,10 @@ namespace verona::rt
...
+  /// but instead should be allocated as part of the runtime.
+  /// In systematic testings contains an ID.
```
Theodus, line 156:
```suggestion
  /// Contains a unique ID in systematic testing.
```

rengolin:
Does this mean we can turn on the GCC tests again?

src/rt/object/object.h
```diff
@@ -201,35 +205,114 @@ namespace verona::rt
...
+    byte* real_start() const
+    {
+      return ((byte*)this) - sizeof(Header);
+    }
+
     Object(const Object&) = delete;
     Object& operator=(const Object&) = delete;
 
-    Object(const Descriptor* desc) : descriptor(desc) {}
+    /// Used to check type has been allocated by the runtime
```
Theodus, line 249:
```suggestion
    /// Used to check type has been allocated by the runtime.
```

src/rt/object/object.h
```diff
@@ -201,35 +205,114 @@ namespace verona::rt
...
+      if (previous == value)
+      {
+        if (value)
+        {
+          puts(
+            "Internal error! Runtime allocation in progress on this thread.");
+          // This error can occur because
+          //  * a type was allocated by the runtime that does not inherit from
+          //  Object.
+          //  * ...
```
Theodus, line 266: Should this `...` be removed?

src/rt/object/object.h
```diff
@@ -201,35 +205,114 @@ namespace verona::rt
...
+        }
+      }
+
+      previous = value;
+#endif
+    }
+
+    // Should be called by the region allocator prior to initialising an
+    // object as part of the runtime.  This is used to ensure that all
+    // subclasses of rt::Object are actually part of the runtime.
```
Theodus, line 285:
```suggestion
    /// Should be called by the region allocator prior to initialising an
    /// object as part of the runtime.  This is used to ensure that all
    /// subclasses of rt::Object are actually part of the runtime.
```

src/rt/object/object.h
```diff
@@ -201,35 +205,114 @@ namespace verona::rt
...
+      obj->get_header().descriptor = desc;
+#ifdef USE_SYSTEMATIC_TESTING
+      obj->sys_id = ++id_source;
+#endif
+      runtime_alloc(true);
+      return obj;
+    }
+
+    // Given a pointer to the start of the header return pointer to the
+    // start of the object.
```
Theodus, line 298:
```suggestion
    /// Given a pointer to the start of the header, return a pointer to the
    /// start of the object.
```

src/rt/cpp/vobject.h
plietar: Can we entirely remove this for non-Object bases? Might require a VBase with the common stuff and a V with two specializations.

src/rt/object/object.h
```diff
@@ -201,35 +205,114 @@ namespace verona::rt
...
+
+      // This should have already been set up.
+      assert(get_descriptor() != nullptr);
+    }
 
     inline const Descriptor* get_descriptor() const
     {
-      return (const Descriptor*)((uintptr_t)descriptor.load() & ~MARK_MASK);
+      return (
+        const Descriptor*)((uintptr_t)get_header().descriptor.load(std::memory_order_relaxed) & ~MARK_MASK);
```
Theodus, line 315: :+1:

Theodus (approved):
LGTM. For the GCC build errors in `ObjectMap`: if I remember correctly, `(uintptr_t&)entry.first` -> `(uintptr_t&)std::get<0>(entry)`

src/rt/cpp/vobject.h
mjp41: I have done this now.  Not quite what I think you proposed, but hopefully simpler.

mjp41:
This requires https://github.com/microsoft/snmalloc/pull/246.  Once that has landed on `snmalloc`, then hopefully CI will pass. 🤞

src/rt/object/object.h
```diff
@@ -201,35 +205,114 @@ namespace verona::rt
...
+#endif
+      runtime_alloc(true);
+      return obj;
+    }
+
+    // Given a pointer to the start of the header return pointer to the
+    // start of the object.
+    static Object* object_start(void* p)
+    {
+      return (Object*)((char*)p + sizeof(Object::Header));
```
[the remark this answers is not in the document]
mjp41: It is no more broken than it was before.

rengolin:
Did you run nightly with GCC, too?

mjp41:
> Did you run nightly with GCC, too?

Not yet.  Still trying to get it through CI.yml, will run a manual nightly.yml, before merging.

rengolin (approved):

rengolin:
Fixes #165

plietar:
@mjp41 How did you update the snmalloc sub-module?

It seems like it now points to commit https://github.com/microsoft/snmalloc/commit/faece6a2fafe1c5d1bc2a66e5fff4e06a3b2300f, which is an equivalent but different commit to snmalloc master (https://github.com/microsoft/snmalloc/commit/d3ecd66f732e7923b6ff939ad95be6601b70c367).

Since `faece6a` does not belong to any branch I worry it could get GC'ed by GitHub.

mjp41:
@plietar, I think it must be the commit from my PR, rather the squashed commit.  I'll submit a new PR to fix this.

Issue #44, opened 2020-01-18
DSL for "Calibre Check"


prateekrastogi:
Looking for ways to inject in Verona a DSL enabling concurrent neural network executions for:
https://github.com/jgraph/drawio/issues/715

mjp41:
Verona is not at a stage where we can support this kind of scenario.  If you specify the requirements you would need for a DSL, then when we plan language features we will can consider those requirements.

prateekrastogi:
My requirements is to support large-scale system bootstrap backpropagation through unfold and recurse, but with automatic concurrent integration of FFI for software dependencies with a system level DSL such as UML or json.

Issue #275, opened 2020-09-08
Error handling, the next level


rengolin:
This is a long story, not a quick-fix issue. It's an overall design "document".

In #168 / #177 we introduced LLVM's `Error` and `Expected` patter for error handling. This works well in the MLIR generator, but we still have a few problems:

First, only the MLIR generator is using, while the Parser is using its own handler. This makes it harder to collate errors into a single hierarchical structure.

Second, the error messages are baked in (in English) inside the generating code. This makes it harder to display error in different settings (console, pop-ups, status bars).

Finally, the errors are final and can't be composed or returned as a list. This makes it harder to do partial compilation or to ignore errors in order to get as many as possible to show the user.

These three problems are important to get right early on, as it gets harder to do the more code we have. They're also an important part of our goal for the Verona compiler (partial compilation, helper tools, etc).

rengolin:
To fix the first problem, we just need to take a decision on which error handling we'll use for the project as a whole, and then propagate that to all interacting sub-projects. For now, that's mostly the AST parser and MLIR generator. But we need to take a decision that makes sense for both projects, and all future integrations later (driver, LLVM handler).

rengolin:
To fix the second problem, the common and simple approach is to make `ParsingError` and `RuntimeError` base classes for derived behaviour, with constructors that don't propagate the meaning as text, but as which derived class we use. The front-ends then can choose how to report them, using the internal structure to populate the error messages.

For example:
```
IncompatibleType : public ParsingError {
  AST type1, type2;
  ...
};
llvm::make_error<IncompatibleType>(type1, type2, loc);
```
The front-end would receive that object and use `type1`, `type2` and `loc` directly instead of printing an already rendered string.

We'd probably need some object-to-text conversion that returns some UTF-8 string, but since those will match the source code, we don't need to worry about internationalisation in the error objects themselves.

rengolin:
The third problem is harder. We'll need to change the whole structure of the code. 

The parser should be simple, as it can "skip tokens" until some closing token is found (ex. `}`) to continue.

The MLIR generator, however, assumes that the AST is well formed, so any error in nested blocks will likely be type errors, or worse, transform errors, and will need to "skip the block", which may not be semantically valid to do at that late point (or easy to report back, with source code location).

[...]

PR #284, opened 2020-09-14
Add flags to help with writing negative tests 

plietar:
Two flags are added to the verona-mlir binary to help with writing
negative tests, where we expect an error to be raised.

`--verify-diagnostics` allows error messages to be written inline within
MLIR files. The compiler succeeds only if all errors listed in the
file get emitted.

`--split-input-file` allows for a single input file to be split into
pieces, where each piece is treated independently. This works around
the fact that MLIR verifiers abort on the first error they find. By
splitting the input file we can include multiple related test cases
and expected errors in a single file.

----

Based on some discussion we've had today, we discussed the fact that this may not be the way we want to proceed.
However I had the PR almost ready, so I figured I may put in out for others to see.

Alternative options include:
1) Using FileCheck/OutputCheck to check for diagnostics. This is what the old compiler did.
2) Configure the test pipeline to split the input before feeding it to the compiler.
3) Use two separate binaries, a fully configurable `verona-opt` which would include these two (and more, such as a comfigurable pipeline), and a end-user binary.

I've realized since discussing these options that
1) gives us poorer integration with the compiler, eg. FileCheck will only check that all expected diagnostics are emitted. It won't complain if the compiler emits a diagnostic that was not expected by the test case
2) will cause line numbers to be mumbled up, which could cause confusion. The `--split-input-file` tries to handles this, although it could probably do a better job at it (eg. an error at line 50, with a split at 44 is rendered as `dialect.mlir split at line #44:6:52: error: [...]`


plietar:
I did a bit of refactoring, replacing `llvm::Error` by `LogicalResult` in the driver, because 1) `mlir::splitAndProcessBuffer` requires it and 2) I got annoyed you can't discard `llvm::Error`s.

I'm still very unhappy about it (eg. you now can't control how the error message is printed by the driver), so I'm happy to hear alternative opinions.

cc #275

rengolin:
> 2) I got annoyed you can't discard `llvm::Error`s.

This is a feature, not a bug, and we shouldn't discard it.

plietar:
> This is a feature, not a bug, and we shouldn't discard it.

I now it's a feature of `llvm::Error`, but it's not a feature I want in this case. Essentially with `--verify-diagnostics`, I don't want to fail if the pass manager fails (it almost certainly will). I could have refactored the driver to not create an `llvm::Error` in this situation, but do in the normal case.

I'm happy to rework this PR and figure out something cleaner for error handling, assuming we want this PR (or part of it) at all.

src/mlir/verona-mlir.cc
```diff
@@ -104,50 +227,32 @@ int main(int argc, char** argv)
   // Set up pretty-print signal handlers
   llvm::InitLLVM y(argc, argv);
 
+  // Register some generic MLIR command line options
+  // mlir::registerAsmPrinterCLOptions();
```
mjp41, line 231: Intended to commit this?

mjp41 (commented):
Is the `--split-input-file` illustrated in this PR?  Shouldn't you have somewhere in the test?
```
// -----
```

I like the compiler checking in this way, I think it is quite nice.  Checking files directly as output would be annoying for line numbers, which this handles, and seems nicer than using FileCheck/OutputCheck.  As the infrastructure is in MLIR already it seems a good approach.

rengolin:
> I now it's a feature of `llvm::Error`, but it's not a feature I want in this case. Essentially with `--verify-diagnostics`, I don't want to fail if the pass manager fails (it almost certainly will). I could have refactored the driver to not create an `llvm::Error` in this situation, but do in the normal case.

We want the driver to return the errors instead of handling them, because we want whoever uses the driver (front-end, UI, debugger) to handle the printing themselves. This is discussed in #275.

`LogicalError` is extremely subpar in that sense, as all we get is a glorified boolean return value.

Right now, we only allow one error to occur per run, that's an intentional over-simplification. Implementing #275 will mean composing `llvm::Error` objects in a stack-like structure, with possible multiple & nested "exceptions" being handled at the same time, all being returned to the caller to handle.

Once that happens, the driver will need to know how to split the range into sub-tokens of some form (ex. find the next function signature, continue from there) and save the error stack associated with the token it failed from (ex: function symbol in addition to file location of the actual error).

Before that happens, we can also pass to the driver the range that we're interested in (for UI methods) and still parse all previous declarations (or following, in case of pre-declared types, for instance), but ignore the errors from anything outside of the range passed. This is nice for code manipulation tools, etc.

We should not print errors in the driver nor the generator at all.

rengolin (changes requested):
I like the idea of verify diagnostics and split the files, but not the move from `llvm::Error` to `LogicalResult`. We'd be going backwards, not forwards that way.

src/mlir/verona-mlir.cc
plietar: Nope. Thanks.

plietar:
> Is the --split-input-file illustrated in this PR? Shouldn't you have somewhere in the test?

I've added a proper example in `testsuite/mlir/mlir-fail/subtyping.mlir`

src/mlir/driver.cc
```diff
@@ -47,65 +58,71 @@ namespace mlir::verona
...
     // Add the input to the source manager and parse it.
     // `parseSourceFile` already includes verification of the IR.
-    sourceManager.AddNewSourceBuffer(std::move(*srcOrErr), llvm::SMLoc());
+    sourceManager.AddNewSourceBuffer(std::move(buffer), llvm::SMLoc());
     module = mlir::parseSourceFile(sourceManager, &context);
+
     if (!module)
-      return runtimeError("Can't load MLIR file");
+    {
+      llvm::errs() << "Can't load MLIR file\n";
```
rengolin, line 91: Does `parseSourceFile` log the errors using the diagnostic handler, too?

src/mlir/driver.cc
```diff
@@ -47,65 +58,71 @@ namespace mlir::verona
...
     assert(module);
 
-    if (filename.empty())
-      return runtimeError("No output filename provided");
-
     if (failed(passManager.run(module.get())))
     {
       module->dump();
-      return runtimeError("Failed to run some passes");
+      llvm::errs() << "Failed to run some passes\n";
```
rengolin, line 105: What about all pas manager errors? How do we capture them, for instance, if the type inference is a pass and it has a user-facing error, like "these two types are incompatible"?

src/mlir/driver.cc
```diff
@@ -47,65 +58,71 @@ namespace mlir::verona
...
-    // Write to the file requested
-    std::error_code error;
-    auto out = llvm::raw_fd_ostream(filename, error);
-    if (error)
-      return runtimeError("Cannot open output filename");
+    module->print(os);
+    return success();
+  }
+
+  LogicalResult Driver::verifyDiagnostics()
```
rengolin, line 113: This method is called asynchronously from the user (`verona-mlir`) and so is independent on the kind of errors they emit. It also shouldn't print error messages and should return the errors to the caller to handle.

src/mlir/verona-mlir.cc
```diff
@@ -52,16 +53,32 @@ namespace
...
+             "process each chunk independently"),
+    cl::init(false));
 
-  // Set defaults form command line arguments
-  void cmdLineDefaults()
-  {
-    // Default input is stdin
-    if (inputFile.empty())
-      inputFile = "-";
+  cl::opt<bool> verifyDiagnostics(
```
rengolin, line 62: there are too many things called `verifyDiagnostics`. This global variable, the driver's method, the driver's flag (which uses underscore to differentiate). This makes for confusing names and get hard to follow.

src/mlir/verona-mlir.cc
```diff
@@ -92,10 +109,113 @@ namespace
...
+    if (!*output)
+    {
+      llvm::errs() << err << "\n";
+      return mlir::failure();
+    }
+
+    return mlir::success();
+  }
+
+  mlir::LogicalResult processVeronaInput(llvm::raw_ostream& output)
```
rengolin, line 162: This pattern now solidifies the clear separation of error handling, not only in different components, but based on different input arguments, which creates really confusing behaviour as well as making it impossible for any tooling using the API to have a sane error handling mechanism.

plietar:
Alright, I've switched the Driver to only use `llvm::Error`.

`verona-mlir` still makes heavy use of `LogicalResult`, because it is required by the API of `mlir::splitAndProcessBuffer`. Unless we rewrite that function to use something else, there isn't much we can do about it.

In "verify diagnostics" mode, there is a case where we need to ignore the error result, which I do with:
```
handleAllErrors(std::move(error), [](const llvm::ErrorInfoBase& e) {})
```

It's not super elegant since it explicitely works around a feature of `llvm::Error`, but at least avoids the mess in the driver.

@rengolin does this look more acceptable to you?

src/mlir/verona-mlir.cc
```diff
@@ -92,10 +109,136 @@ namespace
...
+    if (!error)
+      error = driver.emitMLIR(output);
+
+    if (error)
+    {
+      driver.dumpMLIR(llvm::errs());
+      logAllUnhandledErrors(std::move(error), llvm::errs());
+      return mlir::failure();
+    }
+    else
```
rengolin, line 189: confusing swap of meaning. a simpler sequence would be:
```
error = readAST();
if (!error)
  error = emitMLIR();
if (!error)
  return success;

dumpMLIR()
logErrors()
return failure;
```

same applies to mlirBuffer below

rengolin (commented):
So, I like the simplification of the driver (move of opening files to the user) as well as the new dumpMLIR for debug purposes. But the use `llvm::Error` is just a gimmick replacement for `LogicalResult`, instead of propagating the errors through it.

The main problem I have with this change is that it's self-contained to the driver, but still doesn't integrate with the generator's error handling (currently in `llvm::Error`).

This means that if you input MLIR, the parser will use the handler to print specific error messages and verify their existence in the tests. But if you input Verona, none of that will happen, and we'll still rely on the existing `llvm::Error` handling. In the end, we still have two completely orthogonal error handling mechanisms.

Given the push to use more Verona source files and less intermediate MLIR state in testing, we may not care much what the MLIR reader does in the end. We do, however, care a lot about what the AST reader does and how it report errors.

Note that I am not saying the current situation is good, or even better than the diagnostic handler. It's quite likely that if we used the handler in the AST lowering, we'd get better reporting. And if that's so, this would be a good thing to do. But this is not what you propose in this PR.

So, assuming you're right, and the diagnostic handler is a superior error handling for the AST, our current situation is not good. But your PR would make it worse. To be in a strictly better situation, we would need to replace the AST handling to the diagnostic handler before this move.

If we do that, then this PR become obviously correct.

Does that make sense?

plietar:
> To be in a strictly better situation, we would need to replace the AST handling to the diagnostic handler before this move.

I'd love to be able to use `-verify-diagnostics` with Verona source code, but that is a significant change that I'm not willing do to in this PR.

We (probably) can't use the MLIR diagnostic handler to emit errors in the Generator. What we really need is our own version of a diagnostic handler (eg. `src/compiler/source_manager.h`), and reimplement the verify diagnostic mechanism.

> But your PR would make it worse

I don't really see how it makes things worse. We already had an MLIR diagnostic handler that just printed everything to stderr before. The flag just helps use a different one.

rengolin:
> I'd love to be able to use `-verify-diagnostics` with Verona source code, but that is a significant change that I'm not willing do to in this PR.

Are you planning to add diagnostic handlers to the generator yourself soon?

> I don't really see how it makes things worse. We already had an MLIR diagnostic handler that just printed everything to stderr before. The flag just helps use a different one.

Errors from the generator are not being picked by the diagnostic handler nor being printed by the main file any more.

`LogicalResult` is still being used in addition to the two diagnostic handlers and `llvm::Error`.

plietar:
> Errors from the generator are not being picked by the diagnostic handler

They never were.

> nor being printed by the main file any more.

They are. They get embedded in an `llvm::Error` and printed by the `logAllUnhandledErrors` call in `processVeronaInput`. The `readAST` flow has barely changed.

> LogicalResult is still being used in addition to the two diagnostic handlers and llvm::Error.

Not in the driver. In `verona-mlir.cc` we can't get around using `LogicalResult` for `processMLIRBuffer` if we want to use `mlir::splitAndProcessBuffer`. I guess `openInput` and others could use `llvm::Error`, but I figured it was simpler to stick to `LogicalResult` for now, given this isn't intended to be reusable.

> llvm::Error is being used as a gimmick for LogicalResult.

I don't understand what that means. `llvm::Error` is used in exactly the same way as before.

----

> Are you planning to add diagnostic handlers to the generator yourself soon?

No. I'm trying to add minimum set of features I need so I can better test other pieces (eg. typechecking and #278). I'm not going to design and implement a whole diagnostic infrastructure when it is a significant amount of work that is low priority and unrelated to what I'm doing. Until someone else does that, I consider a solution that only covers the MLIR pipeline better than nothing.

I thought this would have been a reasonably straightforward change. Admittedly my first version was very much a mess, but I've done my best to fix it since and touch as little as possible. I don't understand what you want me to do now, apart from designing and implementing the whole diagnostics and error handling. I'll hapilly close the PR if you believe we can't add this simple feature without first redoing everything.

rengolin:
> No. I'm trying to add minimum set of features I need so I can better test other pieces (eg. typechecking and #278). I'm not going to design and implement a whole diagnostic infrastructure when it is a significant amount of work that is low priority and unrelated to what I'm doing. Until someone else does that, I consider a solution that only covers the MLIR pipeline better than nothing.

This is exactly the point.

We're trying to build a compiler that has a set of features, but also that has a common structure, including error handling. If each one of us implement only the parts of the features we need without thinking about the long term impact of those changes, we'll get to a situation where we can't integrate the parts together without significant refactoring.

Like you, I don't like the current state. I don't like the CMake test runner, I don't like the disparate error handling, but I'm also not happy with adding more complication. The AST has yet another type of error handling.

You can already test your type checking without this, I think. For example, splitting the tests into small files and just checking stderr. The parser does that, so it shouldn't be hard.

The diagnostic handler is a good idea, but I don't want to add a piece of it for some parts of the code and not others, and then just leave it there. It may make your type checking test easier, but it will make the error handling refactory harder in the future. And we do need that refactory.

Issue #280, opened 2020-09-09
Lower class to class types


rengolin:
Use the `!verona.class<>` type, and `verona.field_read` and `verona.field_write` operations to access fields.

Methods will be handled in a separate issue.

rengolin:
WIP: https://github.com/rengolin/verona/tree/mlir-class

rengolin:
Before I get to field access, I need to lower `new` at least for POD classes. Classes with other classes as field types could need some ABI support (struct layout, padding, packing).

rengolin:
WIP: field read/write: https://github.com/rengolin/verona/tree/mlir-field

rengolin:
After discussions during the MLIR lowering efforts, there are a number of issues with the current implementation, and the proper fix is to have two new operators: `call` and `static-call` (WIP in the [mlir-clas](https://github.com/rengolin/verona/tree/mlir-class) branch).

The issues are:
1. Types are mostly unknown at this time. Past efforts to _guess_ them were fraught with problems because we don't want to implement type inference _before_ lowering, but as a pass, after it.
2. Arithmetic operations aren't really machine instructions at this point, and numeric types aren't really native types. They're classes with dynamic methods where the LHS is the object and the RHS is the argument to the call. So, ``a + b` is the same as `a.+(b)`.
3. Static calls need the type to resolve the function to call. For example, two classes with method `foo` would need to expose their unique implementations, but a symbol table for functions would need to differentiate them apart. But with unknown types, this is impossible before type inference.
4. Compiler generated functions (like `for` loop de-sugaring) can call functions (like `next`) that haven't been declared yet for the particular type, especially if that type is parametric or union, which will only be known after reachability analysis and reification.

So we need a set of `call` operators that will not fail is the type is unknown or if the symbol it calls can't yet be resolved. The standard MLIR `call` operation will fail under these conditions. Given that we have two clear types of calls, static and dynamic, we need to clearly different dialect operators that after type inference (and possibly reification) will lower to either standard dialect or LLVM dialect calls, whichever makes more sense at the time.

Also, following Pony's example, at the LLVM lowering time, we'll treat _known numeric arithmetic_ (ex: `U32 a + b`) as native LLVM instructions (or calls to the correct intrinsics, or custom lowering). But during the MLIR lifetime, those calls will remain in the Verona dialect, and reachability analysis can use those instead of standard calls without issue.

[...]

PR #286, opened 2020-09-15
Lowers classes as ops/types

rengolin:
This lowers classes as the existing dialect operator and type. I'm following the existing `dialect.mlir` file for syntax and I can make it agree on the class definition.

Part of this work was also lowering `new`, `view` and fields access, but that may need changes to the dialect representation, so I'll leave it for the next PR, after we have discussed it properly.

Working towards #280

### DIFF
--- a/src/mlir/ast-utils.h
+++ b/src/mlir/ast-utils.h
@@ -394,19 +394,6 @@ namespace mlir::verona
       }
     }
 
-    /// Get the hierarchy of types from class fields without the field info
-    /// T must be a list<ast> type with push_back.
-    template<class T>
-    static void getClassTypeElements(T& nodes, ::ast::WeakAst ast)
-    {
-      auto body = getClassBody(ast).lock();
-      for (auto field : body->nodes)
-      {
-        auto oftype = findNode(field, NodeKind::OfType);
-        nodes.push_back(oftype);
-      }
-    }
-
     // ================================================= Function Helpers
     /// Get the string name of a function node
     static llvm::StringRef getFunctionName(::ast::WeakAst ast)
--- a/src/mlir/ast-utils.h
+++ b/src/mlir/ast-utils.h
@@ -392,6 +392,18 @@ namespace mlir::verona
       }
     }
 
+    /// Get the hierarchy of types from class fields without the field info
+    template<class T>
+    static void getClassTypeElements(::ast::WeakAst ast, T& nodes)
+    {
+      auto body = getClassBody(ast).lock();
+      for (auto field : body->nodes)
+      {
+        auto oftype = findNode(field, NodeKind::OfType);
+        nodes.push_back(oftype);
+      }
+    }
+
     // ================================================= Function Helpers
     /// Get the string name of a function node
     static llvm::StringRef getFunctionName(::ast::WeakAst ast)
--- a/src/mlir/ast-utils.h
+++ b/src/mlir/ast-utils.h
@@ -303,7 +303,8 @@ namespace mlir::verona
       return ptr->nodes[0];
     }
 
-    /// Return a list of sub-nodes
+    /// Get a list of sub-nodes.
+    /// T must be a list<ast> type with push_back.
     template<class T>
     static void getSubNodes(T& nodes, ::ast::WeakAst ast)
     {
@@ -361,7 +362,8 @@ namespace mlir::verona
     }
 
     /// Get the list of types grouped by the same operator (| or &) or
-    /// a single type, if no grouping. T must be a list type with push_back.
+    /// a single type, if no grouping.
+    /// T must be a list<ast> type with push_back.
     template<class T>
     static void getTypeElements(::ast::WeakAst ast, char& sep, T& nodes)
     {
@@ -393,8 +395,9 @@ namespace mlir::verona
     }
 
     /// Get the hierarchy of types from class fields without the field info
+    /// T must be a list<ast> type with push_back.
     template<class T>
-    static void getClassTypeElements(::ast::WeakAst ast, T& nodes)
+    static void getClassTypeElements(T& nodes, ::ast::WeakAst ast)
     {
       auto body = getClassBody(ast).lock();
       for (auto field : body->nodes)
@@ -425,7 +428,8 @@ namespace mlir::verona
       return getType(sig);
     }
 
-    /// Get the ast nodes for the function arguments
+    /// Get the ast nodes for the function arguments.
+    /// T must be a list<ast> type with push_back.
     template<class T>
     static void getFunctionArgs(T& args, ::ast::WeakAst ast)
     {
@@ -439,6 +443,7 @@ namespace mlir::verona
     }
 
     /// Get the ast nodes for the function constraints
+    /// T must be a list<ast> type with push_back.
     template<class T>
     static void getFunctionConstraints(T& constraints, ::ast::WeakAst ast)
     {
@@ -551,7 +556,8 @@ namespace mlir::verona
       return args.lock()->nodes[n - 1];
     }
 
-    /// Return all operands of the operation
+    /// Get all operands of the operation.
+    /// T must be a list<ast> type with push_back.
     template<class T>
     static void getAllOperands(T& ops, ::ast::WeakAst ast)
     {
--- a/src/mlir/ast-utils.h
+++ b/src/mlir/ast-utils.h
@@ -479,6 +479,24 @@ namespace mlir::verona
       return findNode(ast, NodeKind::TypeBody);
     }
 
+    /// Get the body of a class/module declaration
+    static bool isClassType(::ast::WeakAst ast)
+    {
+      if (!isTypeHolder(ast))
+        return false;
+
+      auto typeID = findNode(
+        findNode(
+          findNode(findNode(ast, NodeKind::Type), NodeKind::TypeOne),
+          NodeKind::TypeRef),
+        NodeKind::ID);
+      auto ptr = typeID.lock();
+      auto def = ast::get_def(ptr, ptr->token);
+      if (!def)
+        return false;
+      return isClass(def);
+    }
+
     // ================================================= Operation Helpers
     /// Return the number of operands in an operation
     static size_t numOperands(::ast::WeakAst ast)
--- a/src/mlir/generator.cc
+++ b/src/mlir/generator.cc
@@ -131,15 +131,13 @@ namespace mlir::verona
     typeTable.insert(name, type);
 
     // Field names and types
-    llvm::SmallVector<::ast::WeakAst, 1> names;
-    AST::getSubNodes(names, AST::getClassBody(ast));
-    llvm::SmallVector<::ast::WeakAst, 1> types;
-    AST::getClassTypeElements(types, ast);
+    llvm::SmallVector<::ast::WeakAst, 1> nodes;
+    AST::getSubNodes(nodes, AST::getClassBody(ast));
     llvm::SmallVector<std::pair<StringRef, mlir::Type>, 4> fields;
-    for (auto name_type : llvm::zip(names, types))
+    for (auto node : nodes)
     {
-      auto fieldName = AST::getID(std::get<0>(name_type));
-      auto fieldType = parseType(std::get<1>(name_type).lock());
+      auto fieldName = AST::getID(node);
+      auto fieldType = parseType(AST::getType(node).lock());
       fields.push_back({fieldName, fieldType});
     }
     type.setFields(fields);
--- a/src/mlir/generator.cc
+++ b/src/mlir/generator.cc
@@ -124,6 +124,9 @@ namespace mlir::verona
   {
     assert(AST::isClass(ast) && "Bad node");
 
+    // Add the type first (allows recursive declaration)
+    parseClassType(ast);
+
     // Create the class operation
     auto name = AST::getID(ast);
     auto loc = getLocation(ast);
@@ -147,7 +150,6 @@ namespace mlir::verona
     builder.create<mlir::verona::ClassEndOp>(loc);
     builder.setInsertionPointToEnd(prev);
 
-    // TODO: Add the type/alias
     return c;
   }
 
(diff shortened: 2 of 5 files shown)


plietar:
I'm somewhat confused by the way `parseClassType` is used.

Does the implementation support classes defined in reverse order, where `A` is defined before `B`, but `A` has a field of type `B`? Seems like the implementation relies on `parseClassType` having a side-effect of adding to the cache, which would break in this case. As a more complex example, does it support two mutually recursive classes?

I'm somewhat uncomfortable about the `String -> Type` map to lookup types. How well is that going to work with multiple modules? A mapping from AST node (the class definition one) to Type would be more reliable in my opinion. I don't think there's any need to cache non-class types. That being said, I'm not familiar with the frontend nor with the way name resolution is done, so I could be missing something.

rengolin:
> Does the implementation support classes defined in reverse order, where `A` is defined before `B`, but `A` has a field of type `B`? Seems like the implementation relies on `parseClassType` having a side-effect of adding to the cache, which would break in this case. As a more complex example, does it support two mutually recursive classes?

It does not support any of those cases yet. Nor it can with the current infrastructure. We'd need to leave type holders and treat them as valid Verona types, or do type inference right after declarations, none of which exists yet.

If class A has a field of undeclared class B, the type of that field will be `!type.B` which is fine for the IR, but not for the current validation, so the "test fails". This is work in progress.


> I'm somewhat uncomfortable about the `String -> Type` map to lookup types. How well is that going to work with multiple modules? A mapping from AST node (the class definition one) to Type would be more reliable in my opinion.

Multiple modules and classes will use mangling and no two sub-classes can have the same name in the same module/class.

The AST uses tokens (strings) to refer to variables, functions, types. 


> I don't think there's any need to cache non-class types. That being said, I'm not familiar with the frontend nor with the way name resolution is done, so I could be missing something.

I wasn't caching them, but since I'm using the same type parser for all types, I need to cache the classes and opaque ones, so they end up cached, too. There are no side-effects as far as I can see, and it would be more confusing not to cache them.

plietar:
> The AST uses tokens (strings) to refer to variables, functions, types.

Do we not have a name resolution pass that converts from string identifiers to pointers to AST nodes? I would have thought `src/ast/sym.h` would take care of it.

> It does not support any of those cases yet. Nor it can with the current infrastructure. We'd need to leave type holders and treat them as valid Verona types, or do type inference right after declarations, none of which exists yet.

You can decouple the type and class definition generation. As soon as you see a reference to a class name you generate the corresponding ClassType, even if you haven't seen the definition yet:
- Look up the corresponding AST node for the class definition.
- Create a ClassType and add it to the cache
- Generate the types of each field. If these types mention a class, try to use the cache, if not found.
- Finish the ClassType definition.

Obviously the first step requires on a preliminary name resolution pass to make it possible to find the definition from the reference.

> Multiple modules and classes will use mangling and no two sub-classes can have the same name in the same module/class.

I don't see the point of mangling in early stages of the compiler when we can use structured data and pointer identity instead. In my opinion we should rely on mangling only when it comes to emitting symbols for LLVM, purely because LLVM / ELF / other binary formats don't support namespaces.

plietar:
Okay so looking into the `ref.cc` file, it seems like we perform name resolution only for the sake of changing the kind of ast node, from `ref` to `classdef`/`local`/..., but we don't actually save a pointer to the referenced node anywhere.

@sylvanc what's the status of this? Is there a reason why the pointer isn't persisted into the AST, or just "it wasn't needed at the time".

rengolin:
> You can decouple the type and class definition generation. As soon as you see a reference to a class name you generate the corresponding ClassType, even if you haven't seen the definition yet:

How do I know it's a class and not some type I don't know about?

Should I assume that *any* type I don't know about *is* a class? If so, then this becomes trivial.

> > Multiple modules and classes will use mangling and no two sub-classes can have the same name in the same module/class.
> 
> I don't see the point of mangling in early stages of the compiler when we can use structured data and pointer identity instead. In my opinion we should rely on mangling only when it comes to emitting symbols for LLVM, purely because LLVM / ELF / other binary formats don't support namespaces.

We'll need mangling to other MLIR dialects, too, as they won't understand our class regions. We may get away if we can isolate the methods inside the classes, so that other dialect passes can ignore where they are (hopefully, this is true).

However, I'm still not yet convinced having the class operator, in addition to their types, has any effect on the lowering. Using annotations to define regions, class membership, aliasing is probably as easy (if not easier) than using MLIR regions (that have a specific semantics that doesn't always match what we want).

I think trying to shoe-horn MLIR regions into our dialect may look nice as text but doesn't make it easier to work with through the APIs.

plietar:
> How do I know it's a class and not some type I don't know about?

There is no such thing as a type you don't know about. The name resolution pass should be storing a pointer on the type reference to the AST node that defined it, which should tell you what kind of type it is (class, type parameter, ...). Sadly it doesn't yet (as I mentioned in my other comment), but you can probably work around it. See the `def.cc` file. If `ast` is the type reference, `ast::get_def(ast, ast->token)` will give you its definition.

We're not doing the C/C++ thing where definition order matters.

> Should I assume that any type I don't know about is a class? If so, then this becomes trivial.

Some types will be generic type parameters, but we don't have a way to represent those in MLIR so it can just abort for now.

> We'll need mangling to other MLIR dialects, too
Okay sure, depending on the structure of the MLIR pipeline we might need to do some mangling at some point. But the AST should definitely not be using mangled names is what I was getting at.

> I'm still not yet convinced having the class operator
Me neither. I still think we need *something*, but I don't know what yet.

rengolin:
> > How do I know it's a class and not some type I don't know about?
> 
> There is no such thing as a type you don't know about. The name resolution pass should be storing a pointer on the type reference to the AST node that defined it, which should tell you what kind of type it is (class, type parameter, ...). Sadly it doesn't yet (as I mentioned in my other comment), but you can probably work around it. See the `def.cc` file. If `ast` is the type reference, `ast::get_def(ast, ast->token)` will give you its definition.

For now, I can probably assume it's a class and declare it as so. If I declare as `!type.Name` it would be difficult to change to a class later without doing type inference. 

We don't have examples where we need more than that yet, so once we have the type references, and more complex cases, we can change the implementation.

> Okay sure, depending on the structure of the MLIR pipeline we might need to do some mangling at some point. And yeah, I'm quite annoyed about the way the `verona.class` type creates a global namespace and presumably requires some form of mangling so early on. But what I was getting at is that the AST should definitely not be using mangled names, and a `map<Ast*, ClassType>` will be better on every aspect than a string map.

AST not mangled: wholehearted agreed! :)

rengolin:
> But what I was getting at is that the AST should definitely not be using mangled names, and a `map<Ast*, ClassType>` will be better on every aspect than a string map.

To this specific point: I partially disagree.

The AST has "the source of truth", so it seems obvious that referencing it directly will be unique and unequivocal. However, given the number of compiler generated constructs (lambda capture structures, iterator functions, etc.) that have no node in the AST, it becomes impossible to create those constructs and add to the same map.

To solve this separation, we could have different maps, then search in all of them, in some predefined order. Or we could have a map with a more generic key (like their names), and mangle-as-necessary. Probably other solutions as well.

For now, that we don't have name conflicts, I'm using a string map, because it's cheap and it works for the current purpose without any additional complexity.

For now, I'm also relying on dozens of helper functions in `ast-utils.h`, which are my way of exposing the AST's semantics into an MLIR point of view. Many of those methods could be wrong, but if so, they're explicitly wrong, which is good. The same thing applies to the current maps and builder functions.

We could build yet another layer of abstraction on top of the symbol table, so that we can search for both strings and ast nodes, but that is certainly not for this PR. And probably not going to mean much before the AST is finalised and canonical.

rengolin:
Addressing review comments:
* Added support for pre-declaration of classes + tests
* Refactored `parseClass` + `parseClassType` into one to simplify type logic

src/mlir/generator.cc
```diff
@@ -93,21 +93,81 @@ namespace mlir::verona
...
+
+    // Add the type first
+    // Declare before building fields to allow for recursive declaration
+    auto name = AST::getID(ast);
+    auto type = ClassType::get(context, name);
+    typeTable.insert(name, type);
+
+    // Field types
+    llvm::SmallVector<::ast::WeakAst, 1> nodes;
+    AST::getClassTypeElements(ast, nodes);
```
mjp41, line 135: `getClassTypeElements` is only used in one place, where you then `zip` something based on the same.  Would it be better to build this more directly.  The function `getClassTypeElements` doesn't seem generally useful, and producing the post-zip version seems to have a more sensible semantics to me.

src/mlir/generator.cc
```diff
@@ -199,15 +259,25 @@ namespace mlir::verona
 
   mlir::Type Generator::generateType(llvm::StringRef name)
```
mjp41, line 260: Method name does not indicate caching.  Perhaps at least add a comment.

src/mlir/generator.cc
```diff
@@ -246,11 +316,17 @@ namespace mlir::verona
     if (nodes.size() == 1)
     {
       auto name = AST::getID(nodes[0]);
-      if (auto type = typeTable.lookup(name))
+      if (AST::isClassType(ast))
+      {
+        // Forward declare a class
```
mjp41, line 321: Bit worried about the forward declare comment?  We do not expect classes to need forward declaration.

src/mlir/ast-utils.h
```diff
@@ -392,6 +392,18 @@ namespace mlir::verona
       }
     }
 
+    /// Get the hierarchy of types from class fields without the field info
+    template<class T>
```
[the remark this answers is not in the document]
rengolin: Right, this is a common pattern, but you're right, we should at least explain what's expected.

I don't personally like this pattern for this very reason, but it does have its benefits, so I'm ok with it.

src/mlir/generator.cc
rengolin: Makes sense.

src/mlir/generator.cc
rengolin: Ah, yes, I didn't mean "forward declaration" as in the language, but I agree the name overlap is unfortunate. I'll reword it.

src/mlir/generator.cc
rengolin: Right, the problem is that the zip auto loop makes use of `parseType`, which is not an AST functionality, but a Generator one, so only available here. I don't want to change the `class T` type to something other than `list<ast>` because all others are like that and, in this case, the best I can do is to have a `list<ast, ast>`.

The whole point of the `ast-utils` is to have very simple (almost useless) helper methods, so we can make the AST manipulation explicit. They will invariably all be inlined now, and most likely replaced by simpler methods in a better AST API, so I'm not worrying if they're not super useful right now, as long as they explicitly state the semantics of the AST traversal.

I will, however, change that code a bit, as it does looks confusing. Hopefully it will be enough.

rengolin:
Addressing review comments

 * Adding comments to all methods that use <class T> for list push_back
 * Simplifying getClassTypeElements use in parseClass
 * Adding a new case for multiple pre-declaration uses + test
 * Update pre-declaration comment (was forward declaration)

src/mlir/generator.cc
```diff
@@ -93,21 +93,82 @@ namespace mlir::verona
...
+
+    // Add the type first
+    // Declare before building fields to allow for recursive declaration
+    auto name = AST::getID(ast);
+    auto type = ClassType::get(context, name);
+    typeTable.insert(name, type);
+
+    // Field names and types
+    llvm::SmallVector<::ast::WeakAst, 1> names;
+    AST::getSubNodes(names, AST::getClassBody(ast));
```
mjp41, line 135: Sorry isn't the following equivalent.  It doesn't require a zip or a method call that returns the types of sub nodes, and then reconstructs their pairing?

```C++
llvm::SmallVector<::ast::WeakAst, 1> ast_fields;
AST::getSubNodes(ast_fields, AST::getClassBody(ast));
llvm::SmallVector<std::pair<StringRef, mlir::Type>, 4> fields;
for (auto field : ast_fields)
{
   auto field_name = AST::getID(field);
   auto field_type = parseType(AST::getType(field).lock());
   fields.push_back({fieldName, fieldType});
}
```

mjp41: This is also much closer to what you do below in this function with the builder?

src/mlir/generator.cc
```diff
@@ -246,11 +317,20 @@ namespace mlir::verona
     if (nodes.size() == 1)
     {
       auto name = AST::getID(nodes[0]);
-      if (auto type = typeTable.lookup(name))
+      if (AST::isClassType(ast))
```
mjp41, line 320: Sorry, shouldn't this just go inside `generateType`?

mjp41:
```C++
auto name = AST::getID(nodes[0]);
return generateType(name);
```
and check `AST::isClassType(ast)` inside `generateType`

src/mlir/generator.cc
rengolin: Ah, yes. This is indeed, simpler.

My aim on creating them was to make sure all my expectations for building the MLIR nodes from AST were as explicitly as possible, even if that meant duplicating logic. So I took the task of exposing every "semantic extraction" as a new method, but I think your proposal is indeed simpler and still "semantically" equivalent.

The reason why I did this is first to make sure I got the right semantics, and second to clearly see what methods are the same and what are not. I guess it's already working. :)

src/mlir/generator.cc
rengolin: I know what you mean. That was my first attempt. I didn't do it for some reason that I can't remember. I'll have another go.

rengolin: Ah, yes, so, `generateType` receives a name, not an AST node. This is because we want to be able to fetch/generate types from the compiler implementation as well as from the AST. This is the discussion I was having with Paul above.

My reasoning was either of:
1. Implement the logic in `parseType`, even if slightly redundant, to make the behaviour explicitly misplaced (ie, easy to spot to change later).
2. Pass a second argument to `generateType`, an AST pointer, that defaults to nullptr. If there is one, and it's a class, lower as a class. This looks to me as orthogonal specialisation anti-pattern and not one I'd like to introduce.
3. Create a wrapper `generateType(ast)` that takes the name, checks the table, checks for class type, then returns `generateType(name)` if not. This is slightly better than (2) and probably (1), but creates a new layer of abstraction that may confuse and ultimately lead to misuse of that function.

For the sake of simplicity, I chose (1). It's not the best, but it won't lead to misuse of functions or anti-patterns.

I'd be happy to apply some other solution if you think they're better than the reasoning above.

mjp41: Makes, sense.

mjp41 (approved):

PR #289, opened 2020-09-18
Show object ID in systematic logs

Theodus:
All systematic logs would previously show the address of objects, instead of the object identifier generated under systematic testing. This PR adds an overload for the stream operator so that `const Object*` is displayed as the ID under systematic testing.

### DIFF
--- a/src/rt/object/object.h
+++ b/src/rt/object/object.h
@@ -193,7 +193,7 @@ namespace verona::rt
 #ifdef USE_SYSTEMATIC_TESTING
     inline friend std::ostream& operator<<(std::ostream& os, const Object* o)
     {
-      return os << o->sys_id;
+      return os << o->id<false>();
     }
 #endif
 
@@ -246,10 +246,14 @@ namespace verona::rt
     }
 #endif
 
+    template<bool scramble = true>
     inline uintptr_t id() const
     {
 #ifdef USE_SYSTEMATIC_TESTING
-      return Systematic::get_scrambler().perm(sys_id);
+      if constexpr (scramble)
+        return Systematic::get_scrambler().perm(sys_id);
+      else
+        return sys_id;
 #else
       return (uintptr_t)this;
 #endif
--- a/src/rt/object/object.h
+++ b/src/rt/object/object.h
@@ -140,7 +140,7 @@ namespace verona::rt
       case EpochMark::SCANNED:
         return os << "SCANNED";
       default:
-        return os;
+        abort();
     }
   }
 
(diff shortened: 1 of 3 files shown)


src/rt/sched/schedulerthread.h
```diff
@@ -356,16 +356,16 @@ namespace verona::rt
...
-                           << cown->get_epoch_mark() << ")" << std::endl;
-
         // Administrative work before handling messages.
         if (!prerun(cown))
         {
           cown = nullptr;
           continue;
         }
 
+        Systematic::cout() << "Schedule cown: " << cown << " ("
```
rengolin, line 366: Nit: This change is not described in the commit message.

src/rt/sched/schedulerthread.h
```diff
@@ -898,8 +898,8 @@ namespace verona::rt
           {
             count++;
             *p = c->next;
-            c->dealloc(alloc);
             Systematic::cout() << "Stub collected: " << c << std::endl;
+            c->dealloc(alloc);
```
rengolin, line 902: ouch

src/rt/object/object.h
```diff
@@ -140,7 +140,7 @@ namespace verona::rt
       case EpochMark::SCANNED:
         return os << "SCANNED";
       default:
-        return os;
+        abort();
```
rengolin, line 143: Nit: this change is not described in the commit message

rengolin (commented):
The changes that are described in the commit message are obvious to me and looks good.

I don't understand what the other two (abort, prerun) have to do with it, but I'm assuming they're either required side-effect or they were issues discovered by doing so.

It would be nice to have a small phrase in the commit message describing which of them (or otherwise explaining them).

src/rt/sched/schedulerthread.h
mjp41: Was fine before, as it printed the pointer as a hex string, but not after this.  Glad we always run ASAN.

mjp41 (approved):
LGTM

src/rt/sched/schedulerthread.h
rengolin: as long as dealloc doens't null the pointer

src/rt/sched/schedulerthread.h
Theodus: This message would previously give garbage information in the case that the `cown` is actually the token cown used for fairness. This message should be after the prerun step that checks for the token.

src/rt/object/object.h
Theodus: This is consistent with a previous decision to abort on a missed case in the stream operator overload for `BackpressureState`. This would have caught the error of logging the garbage epoch mark of the token cown before prerun.

PR #290, opened 2020-09-18
Make `next` call at end of `for` loop

mjp41:
Changes for loop from
```
   while ($iter.has_value())
     val = $iter.apply()
     $iter.next()
     [body]
```
to
```
   while ($iter.has_value())
     val = $iter.apply()
     [body]
     $iter.next()
```
There is additional complexity to deal with continue, which must call `next`.


rengolin (approved):
LGTM, thanks!

PR #293, opened 2020-09-22
Remove bit fields from cown status

Theodus:
This PR replaces the use of bit fields for tracking cown status. The added separation of the `_load_hist` and `_current_load` fields prevents the clang SLP vectorizer from optimizing the `total_load()` operation into AVX2 instructions. Based on some profiling of the backpressure system, this optimization has a negligible impact on overall runtime/backpressure performance over the non-AVX2 version.

### DIFF
--- a/src/rt/sched/status.h
+++ b/src/rt/sched/status.h
@@ -49,30 +49,26 @@ namespace verona::rt
   /// Tracks status information for a cown. This class may only be modified by
   /// the scheduler thread running a cown.
   /// TODO: possibly I/O thread when cown is uncheduled
-  class Status
+  class alignas(4) Status
   {
     // Load at which cown is overloaded.
     static constexpr uint32_t overload_threshold = 800;
     // Load at which a cown is no longer overloaded.
     static constexpr uint32_t unoverloaded_threshold = 100;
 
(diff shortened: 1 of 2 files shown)


src/rt/sched/status.h
```diff
@@ -119,10 +113,18 @@ namespace verona::rt
...
+    /// new token message should be added if the cown runs another message.
+    inline bool has_token()
+    {
+      return bits::extract<0, 0>(_misc) == 1;
+    }
+
+    /// Set whether the cown has a token in its queue.
+    inline void set_has_token(bool value)
+    {
+      _misc |= (uint8_t)value;
```
mjp41, line 127: So
```
set_has_token(false)
```
does nothing?  Shouldn't this unset the bottom bit, but `|` is not your friend.

mjp41 (changes requested):

src/rt/sched/status.h
Theodus: Ah, thank you! I was busy debugging something else that this bug triggered and forgot to push the correct version.

mjp41 (approved):
LGTM

PR #294, opened 2020-09-22
Fix log message for rescheduled cown

Theodus:
This PR removes prevents a log message in the scheduler from attempting to show an invalid epoch mark in the case that the cown popped from the scheduler queue is the token.


Theodus:
Required by #293

mjp41:
Could we make the message conditional on it not being a token?  I have just this message for debugging before.

Theodus:
@mjp41 updated

mjp41 (approved):
LGTM

PR #295, opened 2020-09-28
Add initial stab at a strings design doc.

davidchisnall:


davidchisnall:
Straw-man for discussion.  Lots of incomplete parts!

docs/strings.md
```diff
@@ -0,0 +1,464 @@
...
+	match (str)
+	{
+		AsciiString => (if (c > 127) return NotFound) else ...
+		...
+	}
+}
+```
+
+This should be equivalent to a generic where the type used for `str` is a generic paramter.
+If the compiler knows that the argument is an `AsciiString` then it should be free to substitute a version that only includes the `AsciiString` version of the `match`.
```
rengolin, line 450: This is what I think @sylvanc was referring to earlier. If we convert that to a template parameter, restricted to implement the `String` interface, it becomes easier to match the type at compile time:
```
func(a: T & readonly, c: Unichar) : U64 | NotFound
    where T: String
{
    ...
}
```

docs/strings.md
```diff
@@ -0,0 +1,464 @@
...
+Iteration over strings is therefore very complex.
+Should the unit of iteration be a code unit, a code point, or a grapheme cluster?
+Should iteration canonicalise any of these?
+For example, é can be represented as either a single code point or as an e followed by with an accent combining diacritic, but both representations are regarded as equivalent by unicode.
+
+Mutable and immutable strings
+-----------------------------
+
+A lot of programs depend on immutable strings.
+The Java security model depends on strings being immutable and the `SecurityManager` can be bypassed if a vulnerability allows immutable string objects to be modified.
```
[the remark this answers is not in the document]
davidchisnall: If you can modify a `String` in Java (using a vulnerability in the implementation) then you can trick the `SecurityManger` into authorising things that you shouldn't be able to do.

docs/strings.md
```diff
@@ -0,0 +1,464 @@
...
+C++, C#, and other languages have recently introduced the idea of a view within a string, similar to a Module-2 slice.
+A string view gives a range within a string but does not duplicate the underlying storage.
+If a string view is mutable then changes in it will be reflected into the underlying storage.
+
+String views in Verona can use the region abstraction to ensure correct lifetime management.
+A string view on a mutable string will be in the same region as the underlying string and it is the responsibility of that region's memory management to ensure that the region does not outlive the string view.
+Alternatively, string views may be stack types that are bound to a specific region.
+
+String views that refer to immutable strings do not introduce memory management difficulties in Verona because immutable objects are reference counted.
+Even if the string view is mutable, it can be in any region without causing problems.
```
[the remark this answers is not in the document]
davidchisnall: I think we need some extra words here.  An `imm & StringView` can only refer to an `imm & String` because an `imm` type can't refer to a mutable object, but you may want to have a read-only view of a portion of a string.  A read-only view of a mutable string is actually pretty problematic because someone else can mutate the string out from under you.

You could have an `iso` (or `mut`) `StringView` that is a mutable view (i.e. you can change the bounds) of an immutable string (i.e. the string is `String & imm`).  That can be useful if you want to have a moving window over a string, where something captures the `StringView` and you update it.  I don't really like this though, I'd much rather that `StringView`s were treated as value types and can modify a string if the underlying string is mutable.

docs/strings.md
```diff
@@ -0,0 +1,464 @@
...
+-------------------------------------
+
+Baking a representation of string literals into a language causes problems in AoT-compiled languages (this is less of a problem for JIT'd or interpreted languages, where the representation can be changed when the runtime environment is upgraded).
+For example, NeXT defined the Objective-C constant string representation to be an instance of the `NXConstantString` class, with a structure that contained the `isa` (class) pointer, a pointer to a C string, and the length.
+This caused two significant problems over the next decade.
+First, when source code stopped being ASCII, the length would be the number of bytes but the number of UTF-16 code units may differ and so implementing the `-length` method on `NXConstantString` became nontrivial.
+Second, constant strings were increasingly used as dictionary keys and so needed a fast `-hash` method.
+The `-hash` method on `NXConstantString` was always O(n), whereas a constant `NSString` had a field to store a fixed hash value.
+
+In the specific case of a hash, it's not clear that it's best to have a hash for a string or a wrapper type that stores the hash.
```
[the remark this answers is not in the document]
davidchisnall: Yup.  I think the most common thing we'll actually want is an `InternedString` wrapper that guarantees identity and caches a hash.  Expensive to create (requires communicating with a cown that manages the set), but very cheap to compare.

docs/strings.md
```diff
@@ -0,0 +1,464 @@
...
+
+Necessary operations on strings
+-------------------------------
+
+Optimisation for strings can mean different things.
+For immutable strings, iterating over characters (whatever that means), equality comparison, and extracting substrings form the basis of most operations.
+
+# Iteration
+
+ICU provides a number of kinds of iteration over strings, as does OpenStep's `NSString`.
```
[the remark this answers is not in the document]
davidchisnall:
```suggestion
The [International Components for Unicode Library](http://site.icu-project.org/) (ICU) provides a number of kinds of iteration over strings, as does OpenStep's `NSString`.
```

docs/strings.md
```diff
@@ -0,0 +1,464 @@
...
+	/**
+	 * Access a specific character.  Returns a unicode code point for the
+	 * specific index in the string.
+	 */
+	func apply(self: readonly & String, index: U64) : Unichar;
+
+	/**
+	 * Access a specific character.
+	 */
+	func apply(self: readonly & String, index: U64) : Unichar;
```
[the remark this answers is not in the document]
davidchisnall: Typo:
```suggestion
	func update(self: readonly & String, index: U64) : Unichar;
```

docs/strings.md
```diff
@@ -0,0 +1,464 @@
...
+}
+```
+
+For discussion, the above proposal is somewhat simplified:
+The `update` operation should take a `StringView`, a non-owning view of a substring.
+The `String` interface should also include an operation for providing a `StringView` on a specific range.
+The Verona region system makes tracking ownership of string views relatively simple:
+
+ - By default, all mutable strings are `iso`.
+ - Mutable string views are `mut`, in the same region as their owning string.
```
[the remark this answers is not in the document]
davidchisnall: Right, if they were iso they couldn't refer to the string because it would be in a different region (or, rather, they would be the only thing that could refer to the string).

docs/strings.md
```diff
@@ -0,0 +1,464 @@
...
+ - Immutable strings are globally immutable (`imm`).
+ - Immutable string views are immutable.
+
+This makes it easy to ensure that the lifetime of a string view is bound to the lifetime of the underlying string.
```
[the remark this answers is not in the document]
davidchisnall: When `blergh` goes out of scope, the region is destroyed.  The `return v` fails the region checker.

docs/strings.md
```diff
@@ -0,0 +1,464 @@
...
+Requirements on the language
+----------------------------
+
+Supporting a rich Verona string library imposes a few requirements on the language.
+Most of these are already in progress but are captured here for completeness.
+
+# Literal syntax
+
+String literals in the source code should be an opaque immutable type that implements an interface that allows bytes to be iterated.
+We then need lightweight syntax for passing an instance of this type to the constructor for a concrete string type.
```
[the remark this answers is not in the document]
davidchisnall: I think we want the standard library to do so.  If source code is always UTF-8, it can expand to UTF-32 and then pick the right encoding (or, at least, scan the input and calculate the best size for each encoding).  We actually do this in the compiler for ObjC: https://github.com/llvm/llvm-project/blob/31b3f32104db1293aec74ce865f7891ec8019090/clang/lib/CodeGen/CGObjCGNU.cpp#L1046

I want to move all of this logic to the standard library so that programmers can easily substitute their own policy.

docs/strings.md
davidchisnall: Yes, exactly.  This is a write-up of what @syvanc and I discussed.

docs/strings.md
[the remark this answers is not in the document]
rengolin: Right, I didn't know that, so the phrase was confusing. Perhaps rewording with a bit more context could help other people that don't know about this.

docs/strings.md
```diff
@@ -0,0 +1,464 @@
...
+Encodings
+---------
+
+Unicode defines three common serialisations, with different properties:
+
+ - UTF-8 is the most space-efficient for most Latin alphabet languages (an Emoji) and encodes every Unicode code point in 1-4 bytes.
+   The variation in length can make processing and random access indexed by code point expensive.
+ - UTF-16 is the most space-efficient for CJK languages and encodes every Unicode code point to 2 or 4 bytes.
+   As with UTF-8, random access requires additional metadata, though this can typically be omitted for strings in European languages, where every non-Emoji character will be 2 bytes.
+ - UTF-32 is the least space-efficient encoding, using 4 bytes for every character, but has the advantage of being a fixed-length encoding.
```
nwf, line 217: Maybe add "(despite fixed-length of codepoints, grapheme cluster indexing remains challenging)"?

docs/strings.md
```diff
@@ -0,0 +1,464 @@
...
+ - UTF-EBCDIC is equivalent to UTF-8 but stores the EBCDIC character set in the 7-bit space, rather than the ASCII set.
+
+There is no single encoding that is the best fit for all applications.
+There are two key design questions related to encodings:
+
+ - Should strings that use a non-Unicode internal encoding ever expose that across the API boundary?
+   For example, should Japanese text in Big5 encoding be required to convert to Unicode for all operations that are not part of the class's internal implementation?
+ - Should strings that are using a Unicode encoding expose interfaces in a *specific* unicode encoding?
+
+*Proposal*: Standard library strings should expose interfaces in 32-bit Unicode code points.
```
nwf, line 238: And sequences thereof for grapheme clusters?

docs/strings.md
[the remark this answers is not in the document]
rengolin: Right, so here you refer to the string views that are mutable (offset, range) on an immutable string (contents).

Suppose I create a function that takes in a mutable `StringView` to change the underlying data. But if I pass a mutable view on an immutable string, I'll trigger an error trying to write to it (if I don't catch it at compile time).

How do I differentiate, in the syntax of the argument's type, between a "mutable string view" over a "mutable string" from one over an "immutable string"?

rengolin (commented):
.

docs/strings.md
```diff
@@ -0,0 +1,464 @@
...
+}
+
+/**
+ * The set of Unicode encodings.  All standard library functions that have
+ * special cases for specific encodings should handle these.
+ */
+typedef UnicodeStringEncoding = StringEncodingASCII |
+                                StringEncodingUTF8 |
+                                StringEncodingUTF16 |
+                                StringEncodingUTF32;
```
nwf, line 297: Verona doesn't have open type unions, I take it?

docs/strings.md
```diff
@@ -0,0 +1,464 @@
...
+class StringEncodingUTF32
+{
+	typedef CharType = U32;
+}
+/**
+ * Interface for other string encodings.  Any other type can be used as a valid
+ * string encoding but code that does not have any special cases for that
+ * encoding will be able to treat is as unknown.
+ */
+interface StringEncodingCustom
```
nwf, line 285: Why `Custom` and not just `interface StringEncoding`?  The `class`es above ascribe to this `interface`, yes?

docs/strings.md
```diff
@@ -0,0 +1,464 @@
...
+ */
+interface String
+{
+	/**
+	 * Return the length of the string in Unicode code points.
+	 */
+	func length(self: readonly & String) : size_t;
+
+	/**
+	 * The encoding that this string can most easily convert to.  This will
```
nwf, line 331: "most easily convert" makes it sound like this is a property of the pair of the string and its current encoding?

nwf (commented):
Nits and some citations off to Haskell, mostly for additional precedents.

docs/strings.md
```diff
@@ -0,0 +1,464 @@
...
+
+In OpenStep, `NSMutableString` is a subclass of (immutable) `NSString`, which allows APIs to be implemented as methods on `NSString` or as functions that take `NSString` arguments.
+This can lead to unsafe down-casts and the fact that a mutable string is a kind of immutable string causes bugs when mutable strings are accidentally used as immutable strings and modified unexpectedly.
+
+In Verona, we can differentiate between `imm` parameters, where the callee may capture the argument and requires that it is never mutated, and `readonly` parameters, where the string may be mutable but the caller promises not to modify them.
+Collections using strings as keys should always take an `imm & String`, whereas functions that query properties of a string should take a `readonly & String`.
+
+Existing string implementations
+-------------------------------
+
```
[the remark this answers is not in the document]
davidchisnall: I thought about it, but they don't really have anything that I want us to copy.

docs/strings.md
sylvanc, line 297: Do we want an closed world (i.e. union type, this is all the possible types) or an open world (i.e. interface type, other possible types may be added) here?

docs/strings.md
```diff
@@ -0,0 +1,464 @@
...
+                                StringEncodingUTF16 |
+                                StringEncodingUTF32;
+/**
+ * The full set of possible string encodings, can be extended by third-party code.
+ */
+typedef StringEncoding = StringEncodingASCII |
+                         StringEncodingUTF8 |
+                         StringEncodingUTF16 |
+                         StringEncodingUTF32 |
+                         StringEncodingCustom;
```
sylvanc, line 305: I think, from "can be extended by third-party code", we want an open world here, and so an interface type?

docs/strings.md
```diff
@@ -0,0 +1,464 @@
...
+
+/**
+ * Character type, stores a unicode code point.
+ */
+typedef Unichar = U32;
```
sylvanc, line 310: Pure bikeshedding, so please ignore: I actually like the Go use of "Rune", so maybe something like `type Rune = U32;`.

docs/strings.md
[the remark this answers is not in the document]
davidchisnall: Right, I think this is an open question and one I'd be interested in feedback on.  My guess is a mutable `StringView` should always refer to mutable data, my use case for a mutable view of an immutable string feels contrived.

docs/strings.md
```diff
@@ -0,0 +1,464 @@
...
+
+// Placeholder until we have syntax for value types.
```
sylvanc, line 312: The plan is for value types to not be a separate type universe, but, like `imm` vs `mut`, be instance based. So I think we can write:

```
class Range
{
  index: USize;
  length: USize;
}
```

docs/strings.md
```diff
@@ -0,0 +1,464 @@
...
+In `UText`, the generic text object has three fields containing a pointer to a buffer, the character at the start of the buffer, and the length of the buffer, along with a virtual function that can fill the buffer.
+The generic `UText` allocator also provides an option to allocate space at the end of the object that instances can use.
+The fill function can either set the data field to point to its internal storage, copy into the space at the end, or allocate additional memory to store a temporary buffer.
+
+It is worth noting that `NSString` / `NSMutableString` and `UText` provide sufficiently similar abstractions that it is fairly trivial (under 500 lines of code) to create [bidirectional bridges between the two](https://github.com/gnustep/libs-base/blob/master/Source/GSICUString.m).
+
+String views
+------------
+
+C++, C#, and other languages have recently introduced the idea of a view within a string, similar to a Module-2 slice.
```
nwf, line 118:
```suggestion
C++, C#, and other languages have recently introduced the idea of a view within a string, similar to a Modula-2 slice.
```

docs/strings.md
```diff
@@ -0,0 +1,464 @@
...
+	func length(self: readonly & String) : size_t;
+
+	/**
+	 * The encoding that this string can most easily convert to.  This will
+	 * typically be the encoding that the string uses internally.
+	 * Note that a return value of `StringEncodingASCII` implies
+	 * `StringEncodingUTF8` for all read operations because ASCII is a strict
+	 * superset of UTF-8.
+	 */
+	func fastest_encoding(self: readonly & String) : StringEncoding;
```
sylvanc, line 337: Are we returning the encoded string, or is the intention to return the type as a value?

docs/strings.md
```diff
@@ -0,0 +1,464 @@
...
+
+	/**
+	 * The standard encoding that this string can most easily convert to.  This
+	 * will typically be the encoding that the string uses internally but may
+	 * not be for custom encodings.  For all standard-library string types,
+	 * this will return the same value as `fastest_encoding`.
+	 */
+	func fastest_unicode_encoding(self: readonly & String) : UnicodeStringEncoding;
```
sylvanc, line 345: Same question here.

docs/strings.md
```diff
@@ -0,0 +1,464 @@
...
+
+	/**
+	 * The standard encoding that will use the smallest amount of space to
+	 * store the characters.  For example, most English strings will return
+	 * `StringEncodingASCII`, French will typically return
+	 * `StringEncodingUTF8`, whereas CJK language strings will likely return
+	 * `StringEncodingUTF16`.
+	 */
+	func smallest_unicode_encoding(self: readonly & String) : UnicodeStringEncoding;
```
sylvanc, line 354: And here.

docs/strings.md
nwf, line 217:
```suggestion
 - UTF-32 is the least space-efficient encoding, using 4 bytes for every character, but has the advantage of being a fixed-length encoding (but grapheme cluster indexing remains challenging).
```

docs/strings.md
davidchisnall: I don't know, this was a bit of a stab in the dark. @sylvanc can tell me the idiomatic way of spelling it.

docs/strings.md
davidchisnall: I think we want to differentiate between text stored internally as unicode and something that is custom.  My rationale for this was to make it easy to pattern match on 'shrug, go to generic slow paths' vs 'unicode internally, we can do something fast here'.  This may not actually be useful.

docs/strings.md
davidchisnall: Yes, assuming that a string is internally represented as some encoding.

docs/strings.md
```diff
@@ -0,0 +1,464 @@
...
+	 * that may point to the internal storage for the string.
+	 *
+	 * If the caller first calls `fastest_unicode_encoding` and uses the result
+	 * as the generic parameter then it will return an internal buffer for
+	 * immutable strings.  This is intended to be used to build fast iterators.
+	 */
+	func copy_or_view_data[StringEncoding Enc](self: readonly & String,
+	                                           range: Range,
+	                                           buffer: Optional[Array[Enc.CharType] & mut])
+	     : Array[Enc.CharType] & (mut | imm);
```
sylvanc, line 380: The use of a path dependent type here is interesting. This seems like an especially interesting function to look at from a type system point of view, to get the best specialisation we can based on the actual type of `self`. It feels like the `buffer` being optional means it should be a separate method if you want to provide it?

Ill-thought out attempt:

```
view[S, Enc](self: S, range: Range & readonly): S~>(Slice[Enc.CharType] & mut)
  where S: String & readonly
  where Enc: StringEncoding
```

My vague idea there is that the Slice (placeholder name for a a view onto an Array) is `mut` if self is `mut` (that is, the Slice can then be used to change the underlying data) and `imm` if self is `imm`. Not sure if that either works or expresses what you want.

docs/strings.md
```diff
@@ -0,0 +1,464 @@
...
+
+	/******************************************************************************
+	 * Operations on mutable strings
+	 *****************************************************************************/
+
+	/**
+	 * Updates a specific character. 
+	 */
+	func update(self: (mut | iso) & String, index: U64, c: Unichar);
```
sylvanc, line 389: I think maybe:

```
update(self: mut, index: USize, c: Unichar): Unichar
```

If `self` is `iso` at the call-site, we'll be able to use it here as `mut` anyway. If we take something that may be `iso`, we can't let the call-site keep it! Returning the previous value at that index, just because that's convenient for most `update` calls.

docs/strings.md
sylvanc: Oops, didn't see this when I left my comment. Yeah, an interface type here gives us open world.

docs/strings.md
```diff
@@ -0,0 +1,464 @@
...
+
+This basic pattern was extended for collections in the `NSFastEnumeration` protocol and ICU's `UText` abstraction provides a similar model.
+The optimization goals for `UText` are slightly different: all of the operations on `UText` work with UTF-16 data and so the responsibility for converting to UTF-16 is delegated to individual string objects.
+In `UText`, the generic text object has three fields containing a pointer to a buffer, the character at the start of the buffer, and the length of the buffer, along with a virtual function that can fill the buffer.
+The generic `UText` allocator also provides an option to allocate space at the end of the object that instances can use.
+The fill function can either set the data field to point to its internal storage, copy into the space at the end, or allocate additional memory to store a temporary buffer.
+
+It is worth noting that `NSString` / `NSMutableString` and `UText` provide sufficiently similar abstractions that it is fairly trivial (under 500 lines of code) to create [bidirectional bridges between the two](https://github.com/gnustep/libs-base/blob/master/Source/GSICUString.m).
+
```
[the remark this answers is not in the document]
davidchisnall: It's not 100% clear to me whether the extra computational complexity involved in handling both UTF-8 and UTF-16 costs more than the space overhead of the fact that one is less efficient than the other for some languages.  I hope that, with the amount of compile-time specialisation within Verona will mean that software can mostly eliminate the code paths for the other.

docs/strings.md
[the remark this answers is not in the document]
davidchisnall: Can you have a go?  To me, that is exactly what the sentence says.

docs/strings.md
```diff
@@ -0,0 +1,464 @@
...
+char * strchr(const char *s, int c);
+```
+
+This takes a `const char*` (C string) as an argument and returns a pointer into that string.
+The returned pointer may point to an immutable object but it lacks a `const` qualifier.
+
+In OpenStep, `NSMutableString` is a subclass of (immutable) `NSString`, which allows APIs to be implemented as methods on `NSString` or as functions that take `NSString` arguments.
+This can lead to unsafe down-casts and the fact that a mutable string is a kind of immutable string causes bugs when mutable strings are accidentally used as immutable strings and modified unexpectedly.
+
+In Verona, we can differentiate between `imm` parameters, where the callee may capture the argument and requires that it is never mutated, and `readonly` parameters, where the string may be mutable but the caller promises not to modify them.
```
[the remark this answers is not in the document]
davidchisnall:
```suggestion
In Verona, we can differentiate between `imm` parameters, where the callee may capture the argument and requires that it is never mutated, and `readonly` parameters, where the string may be mutable but the caller promises not to modify them and the compiler enforces that guarantee.
```

docs/strings.md
```diff
@@ -0,0 +1,464 @@
...
+# Random access
+
+Random access in a string is a complex concept.
+Random access in C strings or C++ `std::string` is easy because they are simply arrays of code units but it is also meaningless in the context of variable-length encodings (byte *n* in a UTF-8 string, for example, may not be the start of a valid code point, let alone a grapheme cluster).
+Most random access wants to, at a minimum, index on code points and often on grapheme clusters.
+
+# Modification
+
+For mutable strings, inserting, deleting, and replacing substrings often need to be fast.
+This typically leads to mutable strings using a twine or similar representation.
```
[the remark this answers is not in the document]
davidchisnall: I'd imagine so, yes, though as long as they don't leak that across the interface boundary, I don't mind too much if people choose to implement things in a way that I'd consider very odd.

docs/strings.md
```diff
@@ -0,0 +1,464 @@
...
+Some frameworks provide operations on strings for handling path and URL data.
+These are increasingly being deprecated and being moved to specific path and URL types that enforce validity and use structured storage until they must be exported to a string.
+It is unlikely that a Verona string design will want to include any of these operations in the core string APIs.
+
+Encodings
+---------
+
+Unicode defines three common serialisations, with different properties:
+
+ - UTF-8 is the most space-efficient for most Latin alphabet languages (an Emoji) and encodes every Unicode code point in 1-4 bytes.
```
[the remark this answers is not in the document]
davidchisnall:
```suggestion
 - UTF-8 is the most space-efficient for most Latin alphabet languages (and Emoji) and encodes every Unicode code point in 1-4 bytes.
```

docs/strings.md
```diff
@@ -0,0 +1,464 @@
...
+   The variation in length can make processing and random access indexed by code point expensive.
+ - UTF-16 is the most space-efficient for CJK languages and encodes every Unicode code point to 2 or 4 bytes.
+   As with UTF-8, random access requires additional metadata, though this can typically be omitted for strings in European languages, where every non-Emoji character will be 2 bytes.
+ - UTF-32 is the least space-efficient encoding, using 4 bytes for every character, but has the advantage of being a fixed-length encoding.
+
+UTF-8 is common on a lot of Internet protocols and so may be the input or output format.
+UTF-8 is typically used for system call or C library arguments on UNIX systems, whereas UTF-16 is common on Windows.
+
+7-bit ASCII is sufficient to encode anything that can be encoded as a fixed-width UTF-8 string and is commonly sufficient for path and URL encodings and so can be worth special-case handling.
```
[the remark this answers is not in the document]
davidchisnall: As I recall (and it's been a long time since I read the relevant RFCs), URLs are actually quite a small subset of ASCII.  We definitely want to be able to convert URLs to and from strings, but I think it's a mistake to use strings to encode URLs for processing.  You want structured data that contains the schema, password, domain, port, path, and so on, rather than a string that you do operations that match on character ranges on.

docs/strings.md
```diff
@@ -0,0 +1,464 @@
...
+Note that, because 7-bit ASCII was mapped directly into Unicode, 7-bit ASCII is a valid Unicode encoding, though one that can store only a small (but common in English) subset of Unicode.
+Lower-level code is likely to want to opt into using 7-bit ASCII by default (this is sufficiently common in C that Sun's libc has special fast-path operations that can be selected at compile time for programs that don't care about localisation).
+
+There are two other uncommon Unicode encodings that are worth mentioning but probably not worth supporting as an internal representation for standard library strings:
+
+ - UTF-7 is an encoding into a 7-bit character space for use with old email systems that cannot process 8-bit text.
+ - UTF-EBCDIC is equivalent to UTF-8 but stores the EBCDIC character set in the 7-bit space, rather than the ASCII set.
```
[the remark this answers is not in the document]
davidchisnall: I am very happy for someone to implement a concrete class that uses UTF-EBCDIC as the encoding and implements that as a custom encoding.  I do want them to be able to do this and have it work with all of the standard library string functions.  I don't want to ever have to see their code.

docs/strings.md
```diff
@@ -0,0 +1,464 @@
...
+
+There is no single encoding that is the best fit for all applications.
+There are two key design questions related to encodings:
+
+ - Should strings that use a non-Unicode internal encoding ever expose that across the API boundary?
+   For example, should Japanese text in Big5 encoding be required to convert to Unicode for all operations that are not part of the class's internal implementation?
+ - Should strings that are using a Unicode encoding expose interfaces in a *specific* unicode encoding?
+
+*Proposal*: Standard library strings should expose interfaces in 32-bit Unicode code points.
+They should also be able to export or provide views in their native encoding so that operations can be optimised based on pattern matching.
```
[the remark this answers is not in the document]
davidchisnall: Comparison in the unicode  senses would iterate over them using character ranges.  The accessor mechanism is designed to make it easy to build fast iteration on top of (doing a small number of calls to either copy or get access to characters in a string).  

If the two strings happen to be the same encoding, and a comparison operator that just needs to give some stable sort could do a bytewise comparison on them (`strcmp` in the C locale vs `strcmp` in a real locale where humans live).

docs/strings.md
davidchisnall: POSIX also spells it Rune, so that might make sense here too, but I always think of [Hugo Rune](https://en.wikipedia.org/wiki/Hugo_Rune) when I read 'rune'.

docs/strings.md
davidchisnall: There's a comment above about this, but for the standard encodings I think we want closed world (though we may want LE and BE variants of UTF-16 and UTF-32).  We then want open world for the full set of allowed things.  I wrote it like this to capture the idea that we expect things from a closed set to be fast and things from an open set to be slower but I am not convinced that this is the right approach and just making StringEncoding an interface may be the right thing to do.

docs/strings.md
```diff
@@ -0,0 +1,464 @@
...
+ * Character type, stores a unicode code point.
+ */
+typedef Unichar = U32;
+
+// Placeholder until we have syntax for value types.
+value Range
+{
+	U64 index;
+	U64 length;
+}
```
davidchisnall, line 317:
```suggestion
class Range
{
	index: USize;
	length: USize;
}
```

docs/strings.md
davidchisnall: I was thinking of encodings as being akin to Pony primitives but have almost certainly got the syntax wrong for how we will do this in Verona.

docs/strings.md
davidchisnall: The idea here, which I don't think I captured well, is to support a fast enumeration pattern where the callee has three choices:

 - Return an internal view of the underlying data (probably only used on immutable instances, might be used on mutable string types that are some complex structure of immutable character runs)
 - Copy data into the buffer provided by the caller.
 - Allocate a new buffer to copy a large chunk into it (if the buffer is not large enough, for example if the caller asks for 128 code points as UTF-8 and provides a 128-byte buffer).

It might be better to make the optional buffer into two different calls.  I don't have a strong opinion either way.  I think your signature captures the variant that doesn't take a buffer correctly.

docs/strings.md
sylvanc: That makes sense. We get the instance of the singleton type back. Perfect.

docs/strings.md
sylvanc: Same answer @davidchisnall gave above.

docs/strings.md
sylvanc: Same answer @davidchisnall gave above.

docs/strings.md
davidchisnall: Do we yet have syntax for the equivalent of Pony's primitive?

rengolin:
@davidchisnall, is this ready for a new round of reviews?

davidchisnall:
Probably, yes.

docs/strings.md
```diff
@@ -0,0 +1,529 @@
...
+----------------------------------------------------------
+
+Strings are, informally, sequences of characters.
+This is not always a useful abstraction as the definition of a character can, in different contexts, mean:
+
+ * A code unit, the smallest indivisible unit of a single or multi-byte encoding (e.g. `char` or `wchar_t` in C, `unichar` in Objective-C)
+ * A Unicode code point, which may be stored as 1-4 octets, depending on the encoding.
+ * A [grapheme cluster](https://www.unicode.org/reports/tr29/), a sequence of code points that are combined to display something that a human will consider to be a character.
+
+Iteration over strings is therefore very complex.
```
rengolin, line 53: Iteration is hard, but random access is much harder and users also expect that to just work. I'm not sure it's worth mentioning here.

docs/strings.md
```diff
@@ -0,0 +1,529 @@
...
+Should the unit of iteration be a code unit, a code point, or a grapheme cluster?
+Should iteration canonicalise any of these?
+For example, é can be represented as either a single code point or as an e followed by with an accent combining diacritic, but both representations are regarded as equivalent by unicode.
```
rengolin, line 56: Just thinking out loud, I think we should canonicalise everything we possibly can, using the standard's own rules or best practices, and documenting when it's not clear or when we pick a specific default. This may result in similar strings being iterated or accesses in different ways for different standards, but in the vast majority of cases people use only one standard and convert everything else into it.

docs/strings.md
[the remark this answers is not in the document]
rengolin: I think I was just missing the context. Perhaps don't use that as the first example. Immutable strings being keys in collections seem much more important to me.

docs/strings.md
```diff
@@ -0,0 +1,529 @@
...
+char * strchr(const char *s, int c);
+```
+
+This takes a `const char*` (C string) as an argument and returns a pointer into that string.
+The returned pointer may point to an immutable object but it lacks a `const` qualifier.
+
+In OpenStep, `NSMutableString` is a subclass of (immutable) `NSString`, which allows APIs to be implemented as methods on `NSString` or as functions that take `NSString` arguments.
+This can lead to unsafe down-casts and the fact that a mutable string is a kind of immutable string causes bugs when mutable strings are accidentally used as immutable strings and modified unexpectedly.
+
+In Verona, we can differentiate between `imm` parameters, where the callee may capture the argument and requires that it is never mutated, and `readonly` parameters, where the string may be mutable but the caller promises not to modify them and the compiler enforces that guarantee.
```
rengolin, line 81: perhaps "the compiler enforces that promise"?

docs/strings.md
```diff
@@ -0,0 +1,529 @@
...
+Collections using strings as keys should always take an `imm & String`, whereas functions that query properties of a string should take a `readonly & String`.
```
rengolin, line 82: I'm still unsure of the difference between a `readonly` variable and an `immutable` view.

They're both a reference to a variable where you can't change the value of that variable through this reference.

I think this is a discussion for another thread, but perhaps this section is a bit loose in that it sounds important to consider but is not really linked with the surrounding sections...

rengolin (commented):
A few comments on the first half. Will continue the review later.

docs/strings.md
```diff
@@ -0,0 +1,529 @@
...
+------------------------
+
+Verona aims to be an *infrastructure programming language*.
+This category encompasses target use cases such as device drivers, distributed databases, online translation engines, and so on.
+Some of these will not have any user-facing component and so will use strings as identifiers (for example, dictionary keys, path components, error indicators) without the need for any localisation.
+Others will use strings to present text to a human and will need rich unicode processing.
+Our goal should be to support both use cases with the same set of abstractions but avoid projects in the former category needing to pay any performance or code-size penalty relative to a simple array-of-probably-ASCII-bytes string representation.
+
+A lot of complex Verona systems are likely to include components that are at both extremes in terms of their use of strings.
+We should not require a global decision about how strings are represented, nor should we penalise modules that require only simple strings.
```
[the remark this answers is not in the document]
davidchisnall: I think it's not quite that.  Core standard library things should all take a `String` interface.  We will provide an `AsciiString` concrete class and one or more unicode packages that provide richer strings.  If you don't care about anything other than ASCII, you're free to avoid using the other packages and just use ASCII everywhere.

docs/strings.md
davidchisnall: Agreed.  You can view random access on a string as partially cached iteration though (unless the string is ASCII, then the caching is free).

docs/strings.md
davidchisnall: Unfortunately, it's not always clear that this is a good idea.  If all you want to do is stream processing on a string, canonicalization is unnecessary and expensive.  One of the reasons strings are so annoying is that simple decisions like this can have huge performance implications on the macro scale, both positive and negative, depending on the workload.

docs/strings.md
davidchisnall: I tried factoring out a `StringView` from `String`, but I think a string view needs basically all of the methods of a string.  You can't have an `imm & String` (or `StringView`) that refers to a range in a larger string.  You can have a `readonly & String` that has storage that aliases another string.

docs/strings.md
```diff
@@ -0,0 +1,529 @@
...
+   Do these two strings contain the same set of Unicode code points?
+ - Canonicalisation-aware equality.
+   Do these two strings contain a sequence of Unicode code points that are equal under canonicalisation.
+ - Locale-unaware ordered comparison.
+   What is a stable ordering between these two strings that is fast to compute?
+ - Locale-aware collation.
+   What is the ordering between these two strings, with the comparison computed using the Unicode collation rules?
+
+The first and third of these depend only on the strings being able to expose length and iterators that export a common character encoding.
+The second and fourth require a complex unicode library and should probably be in a separate package.
```
rengolin, line 214: Are these covered in the ICU, for at least one type of encoding?

docs/strings.md
```diff
@@ -0,0 +1,529 @@
...
+ - Mutable string views are `mut`, in the same region as their owning string.
+ - Immutable strings are globally immutable (`imm`).
+ - Immutable string views are immutable.
+
+This makes it easy to ensure that the lifetime of a string view is bound to the lifetime of the underlying string.
+Once stack types are better defined, it should also be possible for string views to be stack types associated with the region of the string.
+
+From the perspective of reasoning about mutation, there is no difference between a mutable string that has mutable string views and a mutable string view: if one or more string views exist on a `mut` string then that string is subject to mutation via other objects.
+If a function is passed an `iso & String` then the region guarantees are sufficient to know that the only other string views that may allow modification are `mut` parameters that are explicitly constrained to be allowed to be within the same region as the isolated string.
+Any function that takes two or more `mut & String` parameters is responsible for ensuring that they do not share the same backing store.
```
rengolin, line 449: This refers to Matt's discussion about `when (A, B)` where `A` and `B` could be references to the same region, if we want the users to check or if we can make the type system guarantee. Whatever that ends up being, this should be updated.

docs/strings.md
```diff
@@ -0,0 +1,529 @@
...
+
+An immutable string and an immutable string view are entirely interchangeable from any perspective other than memory consumption.
+The backing store for an immutable string cannot be deallocated until all immutable string views that reference the string are deallocated.
+`StringView` should implement a deep-copy operation that allows the programmer to explicitly avoid this problem if memory overhead is a problem in a particular use.
```
rengolin, line 453: Returning a `String & iso | mut`, right? Ie. a new mutable object in its own region, or if requested, in another owned region.

docs/strings.md
```diff
@@ -0,0 +1,529 @@
...
+This may be sufficient to motivate adding a `StringView` interface, so that programmers can provide a type that supports this and is optimised for 
+
+The minimal standard library should also include:
+
+ - Concrete instantiations of the `String` interface, storing ASCII, UTF-8, and UTF-16.
+ - A generic `StringView` that takes a `String` as a generic parameter and gives a view on that string as specified by a `Range`.
+ - Operations for finding substrings and extracting substrings as string views.
+ - Operations for constructing the most size or speed-efficient concrete string type from a string literal or string view.
+ - Operations for computing efficient hashes on strings.
```
rengolin, line 462: Are these provided by ICU?

rengolin (commented):
I have some final comments. The proposal looks sensible to me, to the reach of my knowledge of string libraries, which isn't much. It's also too high level to think about implementation details now, but the overall picture looks good. Thanks!

rengolin (approved):
With the disclaimer, I think the value of this document in the repo is higher than as a PR.

We can amend it via further PRs later.

docs/strings.md
rengolin: Not defined syntax yet, but I'm working with string names for intrinsics, ex. Verona's `"std.addf"(x, y)` ->  MLIR's `%n = std.addf %x, %y`.

rengolin:
Reading back again, this might not be what we end up implementing in the end but there's a lot of good stuff in there that I think should be in the repo, not in a dangling PR. I'd propose we merge this now and let people push PRs to fix the document later, if there's any disagreements or better ideas. I've just rebased the branch and will merge later today if there are no objections.

PR #298, opened 2020-09-29
Add unknown type to dialect, implement `tidy` and `drop`.

rengolin:
Adding the unknown type to the Verona dialect to allow use in dialect operations. The opaque types could only really work with opaque "operations" and as we phase them out, we'll need a proper unknown type to run type inference over.

All opaque types removed, including literals and compile generate functions.

This refactoring was necessary to add support for new dialect constructs: `tidy` and `drop`. Given they're special dialect operations and not function calls, we need to special treat them early on.

Continues implementing #280.

### DIFF
--- a/src/mlir/generator.cc
+++ b/src/mlir/generator.cc
@@ -510,54 +510,87 @@ namespace mlir::verona
     }
 
     // Else, it should be an operation that we can lower natively
+    // TODO: This will change when we implement a Verona library, where
+    // most of the native ops will still be dynamic calls and resolved
+    // to native when converting to LLVM.
     if (AST::isUnary(ast))
-    {
-      return parsingError(
-        "Unary Operation '" + name.str() + "' not implemented yet",
-        getLocation(ast));
-    }
+      return parseUnop(ast);
     else if (AST::isBinary(ast))
-    {
-      // Get both arguments
-      // TODO: If the arguments are tuples, do we need to apply element-wise?
-      auto arg0 = parseNode(AST::getOperand(ast, 0).lock());
-      if (auto err = arg0.takeError())
-        return std::move(err);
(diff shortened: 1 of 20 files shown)


src/mlir/generator.cc
```diff
@@ -267,26 +267,41 @@ namespace mlir::verona
       return cacheReturn(getImm(context));
     else if (name == "bool")
       return cacheReturn(BoolType::get(context));
+    else if (name == "unk")
+      return cacheReturn(UnknownType::get(context));
 
     // Numeric
     size_t bitWidth = 0;
     name.substr(1).getAsInteger(10, bitWidth);
```
mjp41, line 275: Doesn't `getAsInteger` return false if it fails to parse?  Wouldn't it be better to use that to validate the width?

```C++
if (!substr(1).getAsInteger(10,bitWidth))
  bitWidth = 64;
```
Or if you envisage not having default as always `64`, then you could do
```C++
bool validWidth = substr(1).getAsInteger(10,bitWidth);
assert((!validWidth) || (bitWidth >= 8 && bitWidth <=128));
```

mjp41 (approved):
LGTM

src/mlir/generator.cc
rengolin: Interesting. I'm not fully happy with having a default, but this looks like a cleaner implementation.

plietar:
Could we use `verona.unknown` rather than the abbreviated `unk` form?

rengolin:
> Could we use `verona.unknown` rather than the abbreviated `unk` form?

Sure. I'll update that soon.

rengolin:
> > Could we use `verona.unknown` rather than the abbreviated `unk` form?
> 
> Sure. I'll update that soon.

Done in #299

PR #299, opened 2020-09-30
More class operations lowered from Verona

rengolin:
This PR has a lot of major changes, but in a direction to simplify code generation. After the several design meetings I am more confident in how to represent class operations in MLIR using what we can of the dialect. However, this means we don't really need the `class` operator, as previous changes reflected.

These are the following changes:
1. Removal of the `class` operation (and the `class_end` node). We really didn't need it and probably won't need in the future, so no point in keeping it. The other class operators relied on the symbols being declared by the class operator when we already have them in the `class` type.
2. Adding support for the `in` keyword in the AST. This wasn't implemented before and is required to define the region to allocate on, if necessary.
3. Implemented the `new` operator for both empty and POD classes, as both new regions and existing ones.
4. Refactored how the current use of `alloca` works. It's not perfect, but we'd need to implement region arguments (PHIs) to get rid of them, and this is a work bigger than the current one warrants. It will be done later.
5. Renamed `verona.unk` to `verona.unknown`, as requested by @plietar 

Each commit passes all tests. They're mostly independent, so it's probably easier to review them individually.

Continues to implementing #280.


plietar:
Thanks for the verona.unknown change.

Thinking more about it, one issue I have with the class type vs the operation is that types are interned, immutable and pointed at throughout the IR.

This means that if a pass needs to modify the structure or details of a class (and I can imagine a number use cases for this), it would have to walk through the entire IR to update pointer to the new version of the class. Unlike MLIR values, there isn’t even an easy way to iterate over all uses of a type to replace them.

To make things even worse, because of the way I implemented recursive classes, they are interned by their name rather than their contents. This means every pass that modifies classes has to come up with a new name for them.

rengolin:
> This means that if a pass needs to modify the structure or details of a class (and I can imagine a number use cases for this), it would have to walk through the entire IR to update pointer to the new version of the class. Unlike MLIR values, there isn’t even an easy way to iterate over all uses of a type to replace them.

While lowering the Verona source or reading the MLIR file, types can (and are) cached. There is no reason why we could not keep additional maps around to avoid walking the whole graph.

> To make things even worse, because of the way I implemented recursive classes, they are interned by their name rather than their contents. This means every pass that modifies classes has to come up with a new name for them.

I'm not sure I understand. If the type change needs to be propagated to the whole source, you *must* change them all, and their names must remain the same. If you want to change a local type (for example, extract fields as scalars), then the type *must* change, and so must its name.

Remember, the names should have full context (parent class/module), and are therefore, unique. So far, the names are not unique, but they have the `$parent` information in which we can extract full context. Once we agree *when* to lower the mangling, we can cache those types by their full history and then there's no conflict.

mjp41 (approved):
LGTM.  I still worry about generics and whether the class operation will make that easier or not.

rengolin:
If we can't do it with just types (and I think we can), we re-introduce the classes then, with a purpose.

PR #301, opened 2020-10-01
Stack refactor

mjp41:
This commit is a precursor to adding a stack of roots to the region meta-data.  It refactors the stack to be contained in a single pointer.


src/rt/ds/stack.h
```diff
@@ -7,128 +7,302 @@
...
-  class Stack
+  class StackBase
   {
   private:
     static constexpr size_t STACK_COUNT = 63;
 
-    struct Block
+    /**
+     * The assumes that the allocations are aligned to the same threshold as
+     * their size The blocks contain one previous pointer, and 63 pointers to
```
Theodus, line 23:
```suggestion
     * their size. The blocks contain one previous pointer, and 63 pointers to
```

src/rt/ds/stack.h
```diff
@@ -7,128 +7,302 @@
...
+     * Ts.  This is a pointer of two, so we can use the bottom part of the
```
Theodus, line 24: What do you mean by pointer of two? Do you mean power of two?

src/rt/ds/stack.h
```diff
@@ -7,128 +7,302 @@
...
 
-      if (index == 0)
-        pop_slow_path();
+    /// Call this to pop an element from the stack.  Only
+    /// correct to call this if pop_is_fast just returned
+    /// false.  This returns a pair of the popped element
+    /// and the block that the client must dispose of.
+    std::pair<T*, Block*> pop_slow()
+    {
+      auto item = peek();
```
Theodus, line 131: `assert(!pop_is_fast());`?

src/rt/ds/stack.h
```diff
@@ -7,128 +7,302 @@
...
+   * This class uses the block structured stack with extra fields
+   * for the allocator, and a backup block, so that the common case
+   * of 0-1 elements can be fast, and any other block boundrary case.
+   */
+  template<class T, class Alloc>
+  class Stack
+  {
+    using Block = typename StackBase<T, Alloc>::Block;
+    StackBase<T, Alloc> stack;
+    Alloc* alloc;
```
Theodus, line 202: We should add a note to the docstring that this is not expected to change threads

Theodus (approved):
LGTM

src/rt/ds/stack.h
```diff
@@ -7,128 +7,302 @@
...
 
-  private:
-    void pop_slow_path()
+  /**
+   * This class uses the block structured stack with extra fields
+   * for the allocator, and a backup block, so that the common case
+   * of 0-1 elements can be fast, and any other block boundrary case.
+   */
+  template<class T, class Alloc>
+  class Stack
```
davidchisnall, line 205: The logic for `Stack` and `StackSmall` is very similar.  Would `Stack` be simpler if it had an inner class that wrapped an `Alloc` and managed the `backup` pointer and then used a `SmallStack` that used this as the `Alloc` parameter?  Something like:

```c++
class BackupAlloc
{
  Block* backup = nullptr;
  Alloc *alloc;
  BackupAlloc(Alloc *a) : alloc(a) {}
  template<size_t Size>
  void * alloc()
  {
    static_assert(Size == sizeof(Block), "Allocating something not the size of a block");
    if (backup)
    {
      return std::exchange(backup, nullptr);
    }
    return alloc->template alloc<Size>();
  }
  template<size_t Size>
  void * dealloc(Block *b)
  {
    static_assert(Size == sizeof(Block), "Deallocating something not the size of a block");
    if (backup == nullptr)
    {
      backup = b;
    }
    else
    {
      alloc->template dealloc<Size>(b);
     }
  }
}
```

src/rt/ds/stack.h
```diff
@@ -7,128 +7,302 @@
...
+  };
+
+  /**
+   * Block structured stack, that is a single pointer in size.
+   *
+   * Operations require an explicit Alloc parameter in case more/less is
+   * required.
+   */
+  template<class T, class Alloc>
+  class StackSmall
```
rengolin, line 279: `SmallStack`?

rengolin (commented):
Don't take any of my comments as blockers, I'm just curious. :)

src/rt/ds/stack.h
mjp41: This change would require the BackupAlloc to be passed each time.

I guess there are two aspects that differ between the two wrappers
* caching the alloc structure
* using a backup for a mini pool (this has a pretty big effect in some of the microbenchmarks for the runtime.

I'll see if I can pull the two aspects apart to make it clearer.

src/rt/README.md
```diff
@@ -66,7 +66,7 @@ trace for that seed.
 These can be added to your cmake command line.
 
 ```
--DUSE_MALLOC=ON // Use malloc
+-DSNMALLOC_PASS_THROUGH=ON // Use malloc
```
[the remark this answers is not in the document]
mjp41: Yes, but could be clearer

src/rt/ds/stack.h
```diff
@@ -7,128 +7,302 @@
...
-    Alloc* alloc;
-    Block* block;
-    Block* backup;
-    size_t index;
+    inline static Block null_block{};
+
+    // Index of a full block allocated
+    // Due to pointer arithmetic with nullptr being undefined behaviour
+    // we use a statically allocated null block.
+    static constexpr T** null_index = &(null_block.data[STACK_COUNT - 1]);
```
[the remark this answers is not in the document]
mjp41: This is just to access the last element of the array.  The full block triggers a slow path, so then we can check for references to the `null_block` not on the common path.

src/rt/ds/stack.h
```diff
@@ -7,128 +7,302 @@
...
+
+  private:
+    /// Mask to access the index component of the pointer to a block.
+    static constexpr uintptr_t INDEX_MASK = STACK_COUNT * sizeof(T*);
+
+    /// Pointer into a block.  As the blocks are strongly aligned
+    /// the bits 9-3 represent the element in the block, with 0 being
```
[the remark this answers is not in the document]
mjp41: I'll add some asserts to tie things together better.

src/rt/ds/stack.h
```diff
@@ -7,128 +7,302 @@
...
+    /// a pointer to the `prev` pointer, and implying the empty block.
+    T** index;
+
+  private:
+    /// Takes an index and returns the pointer to the Block
+    static Block* get_block(T** ptr)
```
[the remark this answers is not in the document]
mjp41: Yes, but this is `private` so I think the risk is low.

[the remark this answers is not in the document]
mjp41: I could type def T** to make the intent more explicit?

src/rt/ds/stack.h
plietar: More bikeshedding, I'm not a fan of "small" because it reminds me of the `SmallVec`/`SmallSet` found in LLVM and other projects, but small has a completely different meaning there.

How about "ThinStack", from the thin/fat pointer terminology?

src/rt/ds/stack.h
plietar: With David's `BackupAlloc` you should be able to get rid of the distinction between StackSmall and StackBase.

`Stack<Alloc>` has a field `BackupAlloc<Alloc>` and `StackSmall<BackupAlloc<Alloc>>`, and it's implementation is limited to just augmenting the method calls with the extra BackupAlloc argument. I think that would also reduce the complexity of slow/fast path handling.

mjp41: You can't cache the Allocator if this is part of a region as it can be sent in a message to another thread, or an owning cown could move scheduler thread. So the reason for creating two types for the region tidy semantics we want, needs to be able to flow between scheduler threads.  The complexity was the only way I could see to share the code, and maintain the performance.

src/rt/ds/stack.h
mjp41: Shall I create a Doodle Pool? :-)

I think `ThinStack` is good.  I didn't like `StackSmall`, so thank you both for the bike shedding.

src/rt/ds/stack.h
plietar: This is what I had in mind: https://github.com/plietar/verona/commit/948431ee917da593582d1a0f43d3448d9f4dd7c3

As far as I can tell it behaves identically, and it keeps the control-flow more contralized.

src/rt/ds/stack.h
davidchisnall: I was planning on raising this in the context of the standard library, because at some point we will need to define naming conventions for that.  My suggestion would be to use suffixing rather than prefixing because it plays nicer with auto-complete.  Any identifier that can be used in an arbitrary context should start with the noun and add adjectives.  Apple introduced this rule about 15 years ago and renamed all of the OpenStep enums from things like `NSBoldTextFormat` to `NSTextFormatBold` but they didn't do it with classes.  This means you can type `NSTextFormat` and get autocompletions of all of the kinds of text format but you can't type `NSArray` and get all of the completions of class types.  Here, you can type `Stack` and get completions of the different kinds of stack, which is a bit more human friendly.  It's not useful for things that you'll use in a context where an editor (or, rather, LSP implementation) can refine the list of possible identifiers from some type information, but when naming types that isn't generally possible.

src/rt/ds/stack.h
mjp41: This is great, I tried something like this, but with the `snmalloc::function_ref` and it didn't work well.  Using the template works very nicely. Thanks

src/rt/ds/stack.h
mjp41: A lot of editors use substring rather than prefix to filter, then you get nice behaviour, but I am happy to adopt this kind of convention for other editors

src/rt/ds/stack.h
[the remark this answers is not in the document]
mjp41: I had an experiment, and the amount of code to nicely encapsulate this, beceome almost as big and complex as the `StackThin`, so I have not done it.

src/rt/ds/stack.h
davidchisnall: Yup, @plietar's commit is exactly what I had in mind, sorry for not explaining it better.

src/rt/ds/stack.h
```diff
@@ -7,128 +7,306 @@
...
+      assert(is_empty(iter));
+      auto next = get_block(iter);
 
-      backup = block;
-      block = prev;
-      index = STACK_COUNT;
+      assert(index != (T**)&null_block);
+      next->prev = index;
+      index = &(next->data[0]);
+      next->data[0] = item;
```
plietar, line 180: I've just realized block and next are the same, so actually this could be simplified quite a bit.
```c++
Block* next = (Block*)alloc->template alloc<sizeof(Block)>();
next->prev = index;
next->data[0] = item;
index = &(next->data[0]);
```

plietar (approved):

src/rt/ds/stack.h
mjp41: That was done to check alignment assumptions, but I can do that much more cleanly

PR #302, opened 2020-10-02
Adding field_read / field_write

rengolin:
Implementing field read/write for some simple cases. Not all options covered
yet. We have a few examples that show the mechanism working, but there
are a few cases that are not covered. The main impediment right now is
the viewpoint validation that fails to check some cases.

Before we enbark on a full check of the actual pass, we need to make
sure the types we're generating early (including unknown) will ever be
seen by these passes or the type inference (not implemented yet) will have
cleaned enough of them.

Additional work:
 * Adding assignment type from let's declaration, when available by
   processing the RHS first and only updating the type if the
   declaration has one.
 * Created a class-type extraction utility to field_read from lets so we
   can find the class extraction point even from complex type structure
   (meets of joint, joins of meets, etc)
 * Propagate types through opaque operations (see below).

Propagate types through opaque operations

This propagates types through most operations to avoid the type check
from failing at unknown types. Perhaps we should change the unknown type
to pass any initial type check, perhaps we won't evern get to a type
check with unknowns, but right now, we do and it breaks.

So, to avoid having to handle (and decide) it in this commit, here's
what I'm doing:
 * Arithmetic return value is a join<lhs, rhs>.
 * Load/store returns the type of the pointer

Since we don't have a pointer/reference type yet, and it's not clear
what new region/object operators will do regarding allocation, I'm
leaving the "alloca" nodes as the same type as their objects (no ptr).

This is clearly wrong, but will have to do until we define what's the
actual behaviour. This series is only about implementing the code for
accessing fields and their "addresses".

### DIFF
--- a/src/mlir/generator.cc
+++ b/src/mlir/generator.cc
@@ -398,6 +398,8 @@ namespace mlir::verona
         return parseBreak(ast);
       case AST::NodeKind::New:
         return parseNew(ast);
+      case AST::NodeKind::Member:
+        return parseFieldRead(ast);
     }
 
     if (AST::isValue(ast))
@@ -442,20 +444,29 @@ namespace mlir::verona
 
     // The right-hand side can be any expression
     // This is the value and we update the variable
-    // We parse it first, to get its type
+    // We parse it first to lower field-write correctly
     auto rhs = parseNode(AST::getRHS(ast).lock());
     if (auto err = rhs.takeError())
       return std::move(err);
-    auto type = rhs->get().getType();
 
-    // Can either be a let (new variable) or localref (existing variable).
-    auto var = AST::getLHS(ast);
-    auto name = AST::getLocalName(var);
+    auto lhs = AST::getLHS(ast);
+    // If the LHS is a field member, we have a Verona operation to represent
+    // writing to a field without an explicit alloca/store.
+    if (AST::isMember(lhs))
+      return parseFieldWrite(lhs.lock(), rhs->get());
+
+    // Else, it can either be a let (new variable)
+    // or localref (existing variable).
+    auto name = AST::getLocalName(lhs);
 
     // If the variable wasn't declared yet in this context, create an alloca
-    // TODO: Implement declaration of tuples (multiple values)
-    if (AST::isLet(var))
+    if (AST::isLet(lhs))
     {
+      auto type = unkTy;
+      // If type was declared, use it
+      auto declType = AST::getType(lhs);
+      if (AST::hasType(declType))
+        type = parseType(declType.lock());
       auto alloca = generateAlloca(getLocation(ast), type);
       declareVariable(name, alloca);
     }
@@ -463,7 +474,7 @@ namespace mlir::verona
     if (!store)
       return parsingError(
         "Variable " + name.str() + " not declared before use",
-        getLocation(var));
+        getLocation(lhs));
 
     // Store the value in the alloca
     return generateStore(getLocation(ast), rhs->get(), store);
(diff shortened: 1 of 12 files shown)


src/mlir/generator.cc
```diff
@@ -28,6 +28,29 @@ namespace
     region->push_back(new mlir::Block());
     return &region->back();
   }
+
+  /// Extracts a class type from join/meet
+  mlir::verona::ClassType extractClassType(mlir::Type type)
```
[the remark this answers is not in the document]
mjp41: I would sooner you return unknown if you encounter meet or join in the first cut.  You are trying to implement a form of type checking, and I think that should be done later, when we actually have information.

src/mlir/generator.cc
```diff
@@ -551,21 +585,28 @@ namespace mlir::verona
...
     if (auto err = arg0.takeError())
       return std::move(err);
+    auto lhs = arg0->get();
     auto arg1 = parseNode(AST::getOperand(ast, 1).lock());
     if (auto err = arg1.takeError())
       return std::move(err);
+    auto rhs = arg1->get();
+    // We don't know what the final type will be so we assume it's either of the
+    // input types, for now, if they're different.
+    // If the functions are known (ex. comparisons) we set the known retTy.
```
mjp41, line 598: Again, I would sooner we not try to do type checking/inference at this point.  We should do that in a single pass later. If you don't know the type use unknown.

src/mlir/generator.cc
```diff
@@ -915,6 +955,55 @@ namespace mlir::verona
...
+    // Find the variable to extract from
+    auto ref = AST::getLocalRef(ast);
+    auto var = symbolTable.lookup(ref);
+
+    // Get the field name
+    auto field = AST::getID(ast);
+    auto fieldAttr = StringAttr::get(field, context);
+
+    // Find the field type
+    auto type = extractClassType(var.getType());
```
mjp41, line 971: This won't work with unknown

src/mlir/generator.cc
```diff
@@ -915,6 +955,55 @@ namespace mlir::verona
...
```
mjp41, line 994: I think this should handle unknown. Or use unknown.

mjp41 (changes requested):
Overall, I think you are trying to generate types here, and that is going to complicate this code, and a lot of cases are going to be missed.  So I agree about not throwing away information we know, but I think trying to do it at this level is going to complicate code, and we will not manage to get the types in a lot of cases, so I don't see the benefit.

src/mlir/generator.cc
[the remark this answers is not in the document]
rengolin: This is not about type checking, but getting the class for the field access. If I leave it unknown, then I can't know if the type is of a class and which field it is.

I'm not sure how this would work, if at all:
```
class C { a }
class D { b }
foo(x: C | D) { return x.a }
```
but I'm expecting this is invalid syntax.

So, for now, I'm only supporting the cases where there's only one class.

I could add an error if there are more than one class type in the collection?

src/mlir/generator.cc
rengolin: There are two problems this avoid (not solves):
1. Unknown types are invalid in the type checker for anything. Sub-types, viewpoint adaptation, etc. The more unknowns I have the less I can do.
2. If I continue to propagate unknowns as if they were "poison", then eventually I'd only have unknown, and would not be able to find what class type a value has to be able to extract its field information.

The first is probably fixed by improving the type checking. The latter would have to follow the type and all the paths across the CFG to "know" (or best guess) what the class type is.

src/mlir/generator.cc
rengolin: Right. I can't "extract" a field from an unknown class type. The dialect constructs require the types to be compatible and the field type to be the same as the extracted type.

I could, obviously, remove that requirement on both `field_read` and `field_write` and leave everything unknown, then this whole type propagation would disappear.

src/mlir/generator.cc
rengolin: Same answer as above.

src/mlir/generator.cc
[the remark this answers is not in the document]
mjp41: That is not a syntax error.  It is type checking error.  You need to pass it through to type checking.  
```
class A 
{
  a
  b
}
class B 
{
  a
  c
}
foo(x: A|B) { return x.a }
```
Should type check. So unless you want to implement type checking either in this pass, or on the AST (which will also involve inference).  You are just going to have to let a later pass check the types.

src/mlir/generator.cc
mjp41: That is what I think should be done.  After inference and checking, we will remove unknowns.

src/mlir/generator.cc
mjp41: "best guess" should be "unknown"

src/mlir/generator.cc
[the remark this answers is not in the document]
rengolin: So, should this pass the initial lowering, too?
```
class C { a }
class D { b }
foo(x: C | D) { return x.a }
bar() { let d: D = new D; foo(d) }
```
but fail some later type check?

Then we'll have to change the validation on all dialect operations to not care at all about anything and let the type check pass to do all the right things and make sure none of the passes after the type check can ever add unknown types.

src/mlir/generator.cc
rengolin: We'll probably have to do that with every dialect construct, not just a few.

It was my original plan, I was just trying to adapt to the dialect constructs.

rengolin (commented):
Ok, I'll rewrite this back to using mostly unknowns and disable all the checks for now.

I'll also have to turn off the existing type check pass, because none of that will be able to verify.

I could add a flag to enable it for tests... but that won't work with our current testing arrangement (CMake+ctest), as the flags are hard-coded on the CMake files.

src/mlir/generator.cc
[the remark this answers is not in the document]
mjp41: Yes the type check for `foo` will fail.  But this could be pretty complex given how we want to infer generic parameters.

```
class C { a }
make[X](): X{ ... }
use_C(x: C) { ... }

bar()
{
  var x = make(); // Generic to be inferred based on use
  x.a = ...;
  use_C(x); // Constrains x to be C. Inference will back propagate this.
}
```
Unknown is all you can do unless you implement the inference algorithm.

rengolin:
All types reverted to unknown, and `Verona.UnknownType` added to the list of "always pass" types in the type check.

mjp41 (approved):

PR #305, opened 2020-10-05
Simplify Symbol Table Impl/API

rengolin:
Simplifying the implementation by only having one lookup and one
update/insert method and making "the other" a wrapper with a default
boolean flag.

Also simplifying the API by making all methods return the element if
inserted/updated/found and use that to avoid multiple redundant lookups
throughout the code.

The main difference is the use of `update` when declaring classes, so
that it is clear we can pre-declare them before they're fully defined
(for example, as a type for some earlier class' member).

Before that logic was hidden through a lookup + insert, and now it
simplifies as just an update.

In the end we have a lot less redundant map searches and a simpler code
to follow.

### DIFF
--- a/src/mlir/generator.cc
+++ b/src/mlir/generator.cc
@@ -51,18 +51,6 @@ namespace mlir::verona
       Identifier::get(path.file, context), path.line, path.column);
   }
 
-  void Generator::declareVariable(llvm::StringRef name, mlir::Value val)
-  {
-    assert(!symbolTable.inScope(name) && "Redeclaration");
-    symbolTable.insert(name, val);
-  }
-
-  void Generator::updateVariable(llvm::StringRef name, mlir::Value val)
-  {
-    assert(symbolTable.inScope(name) && "Variable not declared");
-    symbolTable.update(name, val);
-  }
-
   FuncOp Generator::genIntrinsic(
     llvm::StringRef name, llvm::ArrayRef<mlir::Type> types, mlir::Type retTy)
   {
@@ -103,9 +91,11 @@ namespace mlir::verona
     assert(AST::isClass(ast) && "Bad node");
 
     // Declare before building fields to allow for recursive declaration
+    // If class is used in definitions before, it has been declared empty
+    // already, so we use `update` to fetch it.
     auto name = AST::getID(ast);
     auto type = ClassType::get(context, name);
-    typeTable.insert(name, type);
+    typeTable.update(name, type);
 
     // Nested classes, field names and types, methods, etc.
     llvm::SmallVector<::ast::WeakAst, 4> nodes;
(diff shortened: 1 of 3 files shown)


src/mlir/generator.cc
```diff
@@ -313,13 +298,9 @@ namespace mlir::verona
...
-        // Pre-declaration multiple uses
-        if (auto type = typeTable.lookup(name))
-          return type;
-        // Class pre-declaration, cache it for multiple uses
-        auto type = ClassType::get(context, name);
-        typeTable.insert(name, type);
-        return type;
+        // Class pre-declaration, use `update` it for multiple uses
+        // before complete definition (done in `parseClass`).
+        return typeTable.update(name, ClassType::get(context, name));
```
mjp41, line 303: This is changing the behaviour a little as it keeps calling `ClassType::get` previously it would only be called once.

Would adding a `GetOrAdd` be more precise here, e.g. like in .NET 
[ConcurrentDictionary.getOrAdd](https://docs.microsoft.com/en-us/dotnet/api/system.collections.concurrent.concurrentdictionary-2.getoradd?view=netcore-3.1#System_Collections_Concurrent_ConcurrentDictionary_2_GetOrAdd__0_System_Func__0__1__).

mjp41 (approved):
LGTM to me, minor comment only.

src/mlir/generator.cc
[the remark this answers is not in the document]
rengolin: revisit what? the comment sounds odd, is that it?

rengolin: `ClassType::get` is a cheap constructor, similar to smart pointers, so should not be a problem calling it multiple types. It's also O(1). Previously, the additional search for an existing element would be at least O(logn).

It's not just "get or add" either, it's actually "insert or update with get at the end". I fear "GetOrAdd" would be misleading on updates.

rengolin:
I'll merge for now, we can discuss the name of the function later. Thanks!

mjp41:
I actually meant to add a `getOrInsert` which that bit would use, and leave the updates actually as `update`s.  This would allow the symbol table to do hashconing, and if that `ClassType::get` ever involved allocation.  

I was mainly noting the original code would give the result of the first call to `ClassType::get`.  After this PR, it is the `n`th call to it.

rengolin:
> I was mainly noting the original code would give the result of the first call to `ClassType::get`. After this PR, it is the `n`th call to it.

After I merged this I realised I had changed the nature of the function when I merged it with `insert`, and have made it behave like `GetOrAdd`. The call to `emplace` will not insert if there is one already and it will return the element stored previously.

The previous version used to `update` the `n`th copy using the brackets operator. But the previous function was unused (due to the numerous calls to `lookup` before). So, this is new behaviour within the actual usage of the API.

In the new version, the lookup object constructor will still run and allocate, which could be large if the type is complete, but we're still not using it in that way (`ClassType::get(context, name)` is _really_ cheap, but the alloc is not necessarily).

In the end, I don't think this class is worth long considerations, as I want to replace it with better versions for each concept we use. For example, there is already a symbol table for values in MLIR that we could use, there is a scoped map in LLVM that the Toy example uses and we could get rid of the "head"/"tail" basic block maps if we add those as metadata to some loop construct.

But for now, that class is helping me understand what we really need by being extremely simple, even if a bit inefficient. The changes I make to them are mostly to adapt to my understanding and help the process of finding a new structure for each current usage.

PR #308, opened 2020-10-07
Add additional roots to traced region

mjp41:
This adds an extra field to the region meta-data for keeping a stack of additional roots.  These are intended to be references from higher stack frames that must be kept alive during a region gc operation.

The additional roots must be empty if the region is dropped or merged into another region.  The target region of a merge may contain additional roots.

### DIFF
--- a/src/rt/region/region_trace.h
+++ b/src/rt/region/region_trace.h
@@ -287,8 +287,8 @@ namespace verona::rt
       reg->additional_entry_points.push(o, alloc);
     }
 
-    /// Remove root to the stack.
-    /// Must be called in reservse order with respect to push_additional_root.
+    /// Remove root from the stack.
+    /// Must be called in reverse order with respect to push_additional_root.
     static void pop_additional_root(Object* root, Object* o, Alloc* alloc)
     {
       RegionTrace* reg = get(root);
--- a/src/rt/region/region_trace.h
+++ b/src/rt/region/region_trace.h
@@ -68,6 +68,9 @@ namespace verona::rt
     // Compact representation of previous memory used as a sizeclass.
     snmalloc::sizeclass_t previous_memory_used = 0;
 
+    // Stack of stack based entry points into the region.
+    StackThin<Object, Alloc> additional_entry_points{};
+
     explicit RegionTrace()
     : RegionBase(), next_not_root(this), last_not_root(this)
     {}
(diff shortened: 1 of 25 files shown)


mjp41:
Manually triggered build passes:
https://dev.azure.com/ProjectVeronaCI/Project%20Verona/_build/results?buildId=1615&view=results

src/rt/region/region_trace.h
```diff
@@ -259,6 +279,25 @@ namespace verona::rt
...
+    /// Add root to the stack.
+    /// Preserves for object for a GC.
+    static void push_additional_root(Object* root, Object* o, Alloc* alloc)
+    {
+      RegionTrace* reg = get(root);
+      reg->additional_entry_points.push(o, alloc);
+    }
+
+    /// Remove root to the stack.
+    /// Must be called in reservse order with respect to push_additional_root.
```
plietar, line 291: reservse -> reverse

src/rt/region/region_trace.h
```diff
@@ -259,6 +279,25 @@ namespace verona::rt
...
 
+    /// Add root to the stack.
+    /// Preserves for object for a GC.
+    static void push_additional_root(Object* root, Object* o, Alloc* alloc)
+    {
+      RegionTrace* reg = get(root);
+      reg->additional_entry_points.push(o, alloc);
+    }
+
+    /// Remove root to the stack.
```
plietar, line 290: to -> from

src/rt/test/func/memory/memory_gc.h
```diff
@@ -426,47 +426,97 @@ namespace memory_gc
...
+    check(Region::debug_size(nnroot) == 8);
     RegionTrace::gc(alloc, nnroot);
-    assert(Region::debug_size(nnroot) == 1);
+    check(Region::debug_size(nnroot) == 1);
 
     Region::release(alloc, nnroot);
     snmalloc::current_alloc_pool()->debug_check_empty();
   }
 
+  void test_additional_roots()
```
plietar, line 467: Could you extend this test / add another one that checks that objects reachable from an additional root (but which aren't roots themselves) are preserved. It would make sure we not only mark the roots, but we trace them too.

plietar (approved):
LGTM, just some minor stuff. Out of curiosity, what's the difference between check and assert?

Are there some extra places we should assert that there are no additional roots, for instance when scanning for cowns? Presumably any region reachable from a non-executing cown shouldn't have any. I don't know whether regions reachable from messages should be allowed to have additional roots?

mjp41 (commented):
Check is run independent of Debug.  For tests, it doesn't make sense to assert, we want to check in all build types.  It is was the test is trying to do.

Not sure how to add the other checks for when we shouldn't have additonal roots. I'll raise an issue to track this for the future as we have more examples running through the code.  I think we will want a more comprehensive system of invariants on messages etc.

plietar (approved):

PR #310, opened 2020-10-09
Replace backpressure state with priority

Theodus:
This is the first step towards implementing the full backpressure system, as described by the TLA+ spec. This is mostly cosmetic, except that prioritization via overloading is a state transition that may occur before a message is processed by a cown. Also, cowns are now lowered to a normal priority when their queue becomes empty and goes to sleep.

### DIFF
--- a/src/rt/sched/cown.h
+++ b/src/rt/sched/cown.h
@@ -322,7 +322,7 @@ namespace verona::rt
      **/
     void weak_release(Alloc* alloc)
     {
-      Systematic::cout() << "Weak release " << this << std::endl;
+      Systematic::cout() << "Cown " << this << " weak release" << std::endl;
       if (weak_count.fetch_sub(1) == 1)
       {
         auto* t = owning_thread();
@@ -350,8 +350,8 @@ namespace verona::rt
 
     void weak_acquire()
     {
+      Systematic::cout() << "Cown " << this << " weak acquire" << std::endl;
       assert(weak_count > 0);
-
       weak_count++;
     }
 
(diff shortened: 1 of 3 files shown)


mjp41 (approved):
Looks great to me.  I think this has really cleaned up the code and made it much easier for someone to come in to this and understand what is going on.  Great job!

src/rt/sched/cown.h
```diff
@@ -534,19 +527,21 @@ namespace verona::rt
                            << next << ", index " << body->index << std::endl;
 
         {
-          // Hold an epoch so that a scheduled `next` cown is not collected by
-          // the time that we must set it unmutable.
-          Epoch e(alloc);
+          if (make_unmutable)
+            next->unmute(true);
+
```
mjp41, line 532: If we sleep at this point for a long time, is it possible for the unmutable state to be lost before we send the message?

Theodus: 1. I'll update to remove the term "unmutable" from the code.
2. It is possible that the the required cowns will lower their priority before they are acquired. But it's ok for priority to be raised when it is not strictly necessary, as long as it's not too frequent.

mjp41: But can't the priority go back down before the send, and then we reach a deadlock?

Theodus: The priority of the running cown will not go down. Priority can only be raised from another scheduler thread, not lowered

mjp41: But `next` is not running on this scheduler thread yet, it could be running on another scheduler thread, that processes some messages and empties its queue, and puts it into `Normal` priority due to processing its message queue be empty, or the token pops out a couple of times, and then a behaviour on that cown  sends something to an overloaded cown, and goes to `Low` priority, and then this message send occurs?

Theodus: You are correct. This is an issue. I'll go back to holding an epoch on each attempt to fast-send and raising priority if interrupted.

mjp41 (approved):
LGTM. If you send a message to a sleeping cown, and it thinks it is high priority, should you reduce the priority to MaybeHigh, and then put the token in and start processing messages? I think that would bring the system back to stable a bit quicker, but probably not necessary for correctness.

PR #311, opened 2020-10-09
Improve code quality around systematic logging

mjp41:
There are various log messages that are called that go to either the
flight recorder or the systematic logging. In Release builds, the
compiler should erase all the logging code, so it doesn't affect
performance.

This change makes that more true.  It removes various static
initialisers that were not optimised away by Clang.


src/rt/test/systematic.h
```diff
@@ -303,8 +313,8 @@ namespace Systematic
 #endif
 
   private:
-    std::ostream& o;
-    bool first = true;
+    std::ostream* o;
+    bool first ;
```
Theodus, line 317: Is there a reason not to default these members to `nullptr` and `true`?

Theodus (approved):
LGTM

src/rt/test/systematic.h
mjp41: It lead to a constructor and codegen.

src/rt/test/systematic.h
```diff
@@ -219,14 +219,24 @@ namespace Systematic
...
     friend class SysLog;
 
-    LocalLog* log;
-
-    ThreadLocalLog() : log(global_logs().acquire()) {}
+    LocalLog* log = nullptr;
+#ifdef USE_FLIGHT_RECORDER
+    ThreadLocalLog()
+    {
+      if constexpr (flight_recorder)
```
davidchisnall, line 226: In what situation can `flight_recorder` be true but `USE_FLIGHT_RECORDER` not be defined?

src/rt/test/systematic.h
```diff
@@ -303,8 +313,8 @@ namespace Systematic
 #endif
 
   private:
-    std::ostream& o;
```
davidchisnall, line 306: Slightly off-topic, but this direct dependency on iostream is the thing that makes it hardest to run Verona bare-metal.  It would be nice to encapsulate this somewhere (though possibly this class is sufficient).

davidchisnall (approved):

src/rt/CMakeLists.txt
```diff
@@ -120,8 +120,11 @@ if(NOT DEFINED VERONA_RT_ONLY_HEADER_LIBRARY)
           target_compile_definitions(${TESTNAME} PRIVATE USE_SYSTEMATIC_TESTING)
         else()
           add_test(${TESTNAME} ${TESTNAME})
-          target_compile_definitions(${TESTNAME} PRIVATE USE_FLIGHT_RECORDER)
+          if (VERONA_CI_BUILD)
+            target_compile_definitions(${TESTNAME} PRIVATE USE_FLIGHT_RECORDER)
+          endif ()
         endif ()
+        target_compile_options(${TESTNAME} PRIVATE -g)
```
rengolin, line 127: Windows builds don't like this...

src/rt/test/systematic.h
mjp41: None.  I initially tried to have an constructor with the `constexpr` but that didn't work, so I _changed_ to the `ifdef`, but didn't remove the original try.

src/rt/test/systematic.h
mjp41: I think this class is the right place to encapsulate that.  This could then be a ring buffer out of an enclave, or other minimal system. I think everything higher level can be unchanged (probably).

src/rt/CMakeLists.txt
mjp41: Whoops.

PR #312, opened 2020-10-09
Refactor class/type lowering, add static calls

### LINKED ISSUE #280: Lower class to class types

rengolin:
This is a massive change, but it's one that moves towards simpler code
and IR. The main changes are:

1. There are no special types any more

Integer, float, bool are all "classes" and they get casted to whatever
needed for arguments and return values. All numeric types will be coded
like classes with dynamic calls in the future, so that fits better with
the model.

Lowering of native arithmetic will be done when lowering to LLVM, with
special lowering for known patterns.

2. The DescriptorType was created

That's a handler to a class for static calls. Unlike dynamic calls, we
don't have an object to infer what the type is, so we create this type
that will hold a (global, unique) handler to lower a call in the same
way as dynamic calls, but with that handler instead of an object.

3. Two new operations were created: Call and Static

Call implements both dynamic and static calls, by taking the handlers
(descriptor or object) and call a method by name and arguments.

Static gets the descriptor of a type, to be able to call its static
methods.

4. Add qualifiers to functions as attributes

We should start annotating our nodes with the extra information that
comes from the AST and the difference between static and dynamic
calls is an important one.

The implementation is not limited to it, but so far, we only have examples
to that one.

5. Added comments to the remaining types (`VeronaTypes.h`)

Please check to make sure my comments are accurate.

Fixes #280


src/mlir/dialect/VeronaTypes.cc
```diff
@@ -316,8 +317,11 @@ namespace mlir::verona
 
   ClassType::FieldsRef ClassType::getFields() const
   {
-    assert(getImpl()->isInitialized);
-    return getImpl()->fields;
+    // We may not have a full declaration available
+    if (getImpl()->isInitialized)
+      return getImpl()->fields;
+    else
+      return ClassType::FieldsRef();
```
davidchisnall, line 324: How does a caller differentiate between not initialised yet and has no fields?

src/mlir/dialect/VeronaTypes.h
```diff
@@ -41,11 +41,15 @@ namespace mlir::verona
...
     struct IntegerTypeStorage;
     struct FloatTypeStorage;
+    struct DescriptorTypeStorage;
     struct CapabilityTypeStorage;
     struct ClassTypeStorage;
     struct ViewpointTypeStorage;
   }
 
+  /**
+   * Meet types are intersections between types (A & B).
```
davidchisnall, line 51: Is there a reason to invent a new word for them here?

src/mlir/dialect/VeronaTypes.h
```diff
@@ -55,6 +59,9 @@ namespace mlir::verona
     llvm::ArrayRef<mlir::Type> getElements() const;
   };
 
+  /**
+   * Meet types are unions between types (A | B).
```
davidchisnall, line 63: Copy-and-paste error and the same question regarding defining a new word.

src/mlir/dialect/VeronaTypes.h
```diff
@@ -64,41 +71,42 @@ namespace mlir::verona
...
   };
 
+  /**
+   * Capability types represents traits that other types possess.
+   *
+   * Isolated: An entry point to a new region, can only have one reference to
+   * it at any time (ownership model). Can also be created when returning a
+   * sub-region that needs to be isolated.
+   *
+   * Mutable/Immutable: Read-Write/Read-Only variables and viewpoints.
```
davidchisnall, line 108: `readonly` is distinct from immutable.  An `imm` capability refers to an immutable object, a `readonly` capability refers to an object that cannot be modified via this pointer (similar to C/C++ `const`, though stricter) but may or may not actually be immutable.

src/mlir/generator.cc
```diff
@@ -130,6 +130,14 @@ namespace mlir::verona
           return std::move(err);
         // Associate function with module (late mangling)
         func->setAttr("class", TypeAttr::get(type));
+        // Add qualifiers as attributes
```
davidchisnall, line 133: None of the tests has a non-empty list of qualifiers.  What is a qualifier in this context?

src/mlir/generator.cc
```diff
@@ -130,6 +130,14 @@ namespace mlir::verona
...
+        llvm::SmallVector<::ast::WeakAst, 4> quals;
+        AST::getFunctionQualifiers(quals, node);
+        llvm::SmallVector<mlir::Attribute, 4> qualAttrs;
+        for (auto qual : quals)
+          qualAttrs.push_back(
+            StringAttr::get(AST::getTokenValue(qual), context));
+        func->setAttr("qualifiers", ArrayAttr::get(qualAttrs, context));
```
davidchisnall, line 140: Shouldn't this be set only if the set of qualifiers is non-empty?

src/mlir/generator.cc
```diff
@@ -247,39 +255,19 @@ namespace mlir::verona
...
-
-    // We need better error handling here, but this should never happen
-    llvm_unreachable("Cannot lower type");
+    // Every other type is just a class that we don't know yet
+    return typeTable.insert(name, ClassType::get(context, name));
   }
 
   mlir::Type Generator::parseType(const ::ast::Ast& ast)
   {
+    // Qualified types are references to classes with typeargs
```
davidchisnall, line 267: What are typeargs?

davidchisnall (approved):
Some minor nits, nothing that should block a merge.  This looks like a big improvement and will make it fairly trivial to support things like U27 as storage-only types and U32 as types with operations defined on them.

There are a few bits where the terminology in the compiler appears to be diverging from the terminology in the language, we should try to unify those.

src/mlir/dialect/VeronaTypes.cc
rengolin: It does not. This is one of the things that will need to be changed later.

Right now, we're lowering "classes" that have no fields declared (like U32) mostly because we're not declaring it at all. This is the main reason why we need this change.

With time, and only after we implement modules/includes, we'll be able to assert on initialisation. Without this, no class type can be printed back.

src/mlir/dialect/VeronaTypes.h
rengolin: @plietar chose the name. I don't mind either way.

src/mlir/dialect/VeronaTypes.h
rengolin: oops

src/mlir/dialect/VeronaTypes.h
rengolin: That's what I meant as the difference between "variables and viewpoints". IIUC, an `imm` variable is always immutable, while a variable seen from an `imm` viewpoint is only read-only through that "pointer".

Perhaps I should be clearer...

src/mlir/generator.cc
rengolin: some do have the "static" qualifier, but it's hard to see in the sea of red and green... :)

they're not used yet by the compiler, but we'll use them later to check if a call is static or dynamic. 

I could have checked the type of the first argument (`DescriptorType` or not) but attributes are more stable than values (casts, etc), and we must support them for every operation, so doesn't cost to have them, since the info is in the AST anyway.

src/mlir/generator.cc
rengolin: Good idea, will do.

src/mlir/generator.cc
rengolin: Sorry, some comments refer to their AST nomenclature. I believe `typeargs` are the parameters, for example, generics. These are irrelevant in this context, I'll clean up the comment.

src/mlir/dialect/VeronaTypes.h
rengolin: I hope by being less verbose it got clearer...

src/mlir/dialect/VeronaTypes.cc
```diff
@@ -316,8 +317,11 @@ namespace mlir::verona
 
   ClassType::FieldsRef ClassType::getFields() const
   {
-    assert(getImpl()->isInitialized);
-    return getImpl()->fields;
+    // We may not have a full declaration available
```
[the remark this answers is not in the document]
rengolin: There was a resolved comment that explained it. Right now, we're lowering incomplete types, like S32 that haven't been declared. Until we have modules/includes, we can't declare all types and this fails to print the results.

It will be a compiler error once we have more implemented in the compiler.

src/mlir/dialect/VeronaTypes.h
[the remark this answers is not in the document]
rengolin: updated

src/mlir/dialect/VeronaTypes.cc
rengolin: I've added a TODO to remind us that this needs to be an assert in a complete compiler.

src/mlir/dialect/VeronaTypes.cc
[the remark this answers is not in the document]
rengolin: Added a `TODO` line to remind us of that later.

src/mlir/dialect/VeronaTypes.cc
```diff
@@ -106,6 +106,27 @@ namespace mlir::verona::detail
     }
   };
 
+  struct DescriptorTypeStorage : public TypeStorage
+  {
+    Type descriptor;
```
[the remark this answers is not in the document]
rengolin: Renamed to `StaticClassType`. I don't think we should bikeshed too much this name, but now it's very specific to what it means with enough comment on it to explain it further. This type may fuse with some other concepts in the future and we'll probably need to rename it, or if we just want to rename it as in it would be a simple commit.

src/mlir/dialect/VeronaTypes.h
rengolin: If we want to choose a new name, it shouldn't be done in this PR anyway, so I'll mark this as resolved.

src/mlir/dialect/VeronaTypes.h
```diff
@@ -64,41 +71,42 @@ namespace mlir::verona
...
 
-    static BoolType get(MLIRContext* context);
+    static UnknownType get(MLIRContext* context);
   };
 
-  struct UnknownType : public Type::TypeBase<UnknownType, Type, TypeStorage>
+  /**
+   * A class descriptor type, used for access to static members of the class,
+   * including fields and methods. The type pointed to could be unknown before
+   * type inference, but once known, should point to a fully qualified class.
```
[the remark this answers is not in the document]
plietar: It depends by what you mean by "unknown". I am intepreting it as "an instance of `UnknownType`", which are holes that the type inference pass will fill.

Generic type parameters, such as `X`, are different. These will still be present after type inference, and up until reification, *if* reification happens.

src/mlir/dialect/VeronaTypes.cc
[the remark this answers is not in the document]
plietar: Sorry I didn't mean renaming the struct, but the field and getter. I actually liked `DescriptorType`.

src/mlir/dialect/VeronaTypes.cc
```diff
@@ -244,39 +264,19 @@ namespace mlir::verona
...
 
-  BoolType BoolType::get(MLIRContext* ctx)
+  StaticClassType StaticClassType::get(MLIRContext* ctx, Type descriptor)
   {
-    return ::mlir::detail::TypeUniquer::get<BoolType>(ctx);
+    return Base::get(ctx, descriptor);
   }
 
-  UnknownType UnknownType::get(MLIRContext* ctx)
+  Type StaticClassType::getTypes() const
```
plietar, line 277: This shouldn't be in the plural form.

src/mlir/dialect/VeronaTypes.cc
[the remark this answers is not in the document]
plietar: I'm quite unhappy with the notion of "incomplete types". We do not want to follow the C++ model of forward declarations at all. If a type doesn't have a definition it can't be used, end of it.

Having some boilerplate empty definitions `class I32 {}` at the top of the test files is IMO a better workaround for the lack of modules/includes than introducing broken concepts into the compiler.

src/mlir/dialect/VeronaTypes.h
plietar: I don't know if David's "new word" was referring to Meet or to Intersection.

I chose meet/join over intersection/union for a few reasons, none of which are particularily strong:
- Union is a C++ keyword
- Meet and Join are short, and of the same length
- Intersection is hard to type
- (Maybe the most valid reason) there are various interpretations of the words "union" and "intersection", used in different languages. There's the C/C++/Rust unions, there's discriminated unions (ie. Haskell like data types), there's the ["implictly projected pairs of values" intersections](0), and there's the Scala-like intersection. None of these have the Pony/Verona semantics.
On the other hand, Meet and Join have, I believe, precise and unambiguous semantics. They are respectively the greatest lower bound and least-upper bound in the subtyping lattice.

[0]: https://dl.acm.org/doi/pdf/10.1145/2951913.2951945

src/mlir/dialect/VeronaTypes.cc
rengolin: good catch, I'll fix it.

src/mlir/dialect/VeronaTypes.cc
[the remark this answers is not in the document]
rengolin: This is a temporary work around and not an established behaviour. What you describe will be enforced in a more complete compiler. Trying to make the compiler perfect on day one is counter-productive.

src/mlir/dialect/VeronaTypes.h
rengolin: I personally like `Meet` and `Join` for the same reasons @plietar explain above.

src/mlir/dialect/VeronaTypes.h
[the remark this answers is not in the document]
rengolin: I mean the English sense of "unknown" == "not known". If it was specific to `UnknownType` I would have called it so explicitly.

PR #313, opened 2020-10-12
Make cown logs more consistent

Theodus:
The systematic log entries now refer to cowns in the form "cown <id>". This makes it easier to search through logs for all of the events that have been logged for a specific cown.


rengolin (approved):
Makes sense to me. Using semi-formatted output (like `text : cown`) only makes sense if we're expecting to parse it with some simple tool. But the new format doesn't seem that more complicated anyway, so LGTM.

PR #316, opened 2020-10-12
Changes how classes are represented: MLIR modules

### LINKED ISSUE #306: Function scope and name mangling
The current implementation "bypasses" the need for name mangling by not having tests with conflicting function names in different contexts (classes). But to add type classes (like `U32` and `U64`), needing the exact same functions (ex. `add`), we'll need to resolve the conflict.

There are two ways we can do that:
1. Mangling the name as soon as possible, as "class__function" (more underscores for nested classes) and storing the function symbols in a flat map. Lookup "resolves" the right symbol by mangling the called function name before searching.
2. Add context to the symbol table. This is more than the current scope, as we don't want to `pop` the contexts as we leave, or we'll lose the symbol. This is similar to the existing MLIR symbol table, so perhaps we should move to it?

We can still keep the existing simplified symbol table for the other uses (like types, basic blocks, etc).

rengolin:
This PR has two cleanups and a major change: classes are now MLIR modules to allow declaration of functions with the same name on different classes without polluting the dialect with new operations and having to implement fields and functions lists on the operation.

It seems to me that modules do that job natively and without any extra work. It should also be easy to merge other modules if we keep their names unique.

Note, the three patches are separate, but the second is a required fix for the third.

The two cleanups from #312:
1. Final requests (comments, class name, etc) from @plietar
2. Clean up from the `for` loop calls as Verona dialect calls

Fixes #306.


plietar:
While I'm not against this change, I have to ask, how is this any different from the old `verona.class` operation that we removed?

plietar:
Modules are barely a "native MLIR" construct. They are just an operation defined by the standard dialect, and is no more special than an operation we define ourself. To cite the [documentation of `ModuleOp`](https://mlir.llvm.org/doxygen/classmlir_1_1ModuleOp.html):
> ModuleOp represents a module, or an operation containing one region with a single block containing opaque operations.

There is no reason why we couldn't put functions inside a `verona.class` operation. It would have been, for the most part, equivalent to a module, but with added flexibilty. Of course we can store everything in a module and put a bunch of stuff into attributes and call it a day, but MLIR is intentionally designed to make new operations cheap to define, so why not take advantage of it?

The reason I had duplicated fields in both the operation and the type is that I needed to look up the type of fields when type checking a field read/write, and the semantics of referencing an operation from within a type is not very well defined in MLIR. I do not see how methods will be any simpler, and if we solve the problem for methods then we can apply that solution for fields, and the `class` type can return to being a flat identifier, with no "keeping in sync" problem.

rengolin:
> Modules are barely a "native MLIR" construct. They are just an operation defined by the standard dialect, and is no more special than an operation we define ourself.

Semantics. 

It's an existing operation that is already defined and does *exactly* what I need. Why would I create a whole new one and all the boilerplate it comes with it?

I am not claiming that `Module`s are better than a custom `ClassOp`. Without further evidence, they're basically equivalent for the purposes this PR is concerned about. Occam's razor tells me to not create new things that already exist.

Once we go further into the compiler, more evidence may accumulate in favour of a `ClassOp`, in which case, we'll just switch. It's not that hard, as you said yourself.

> There is no reason why we couldn't put functions inside a `verona.class` operation. It would have been, for the most part, equivalent to a module, but with added flexibilty.

Right now, I need no more flexibility than `Module`s can give me.

If by the time we get to do name mangling we need something else, creating a new similar operation would be simple enough.

> The reason I had duplicated fields in both the operation and the type is that I needed to look up the type of fields when type checking a field read/write, and the semantics of referencing an operation from within a type is not very well defined in MLIR.

The field read/write operations have the class type, with the field (and its type) in the operation. There's no need to look it up anywhere else.

> I do not see how methods will be any simpler,

I don't think adding methods to types would make any sense.

> and if we solve the problem for methods then we can apply that solution for fields,

Modules "solve" the "problem" for methods with zero added code. Until we have a concrete example in the compiler that makes this not true, I see no reason to create anything new. But, once we do, I'll be the first to propose we create it.

> and the `class` type can return to being a flat identifier, with no "keeping in sync" problem.

If by "flat identifiers" you mean names only (and context, obviously), like `class<"A", "$parent" : class<"B", "$parent" : class<"C"...>...>...>` then I guess you don't have a sync problem, but you moved the field name/type from the type (that we have in hand for field read/write) to an operation (that needs a table lookup).

I don't see how that is better.

plietar:
> I don't think adding methods to types would make any sense.

Why does it make more sense than adding fields to them? How will you type check method calls? eg.

```
  verona.call "foo"[%1 : !verona.class<"A">](): !verona.class<"B">
```

This needs to look at the list of methods that the class A has to find one called "foo", and make sure that method return a (subtype of) B.

If you can solve the problem of checking method calls, then you can solve the problem of checking fields in the exact same way, by encoding the list of fields in the class operation/module.

-----

I'm happy to go ahead and just use modules, but I'm 90% sure we'll go full circle and add a special class operation in a few weeks.

rengolin:
> If you can solve the problem of checking method calls, then you can solve the problem of checking fields in the exact same way, by encoding the list of fields in the class operation/module.

I see what you mean. Though this is up ahead, not right now.

> I'm happy to go ahead and just use modules, but I'm 90% sure we'll go full circle and add a special class operation in a few weeks.

It's very likely you are right. But there are too many moving parts right now to create a lot of stuff we don't use yet assuming they'll be the right thing in the future.

It's much easier and faster to create partial support using existing components and only change when the features you're implementing clearly need it. I spent some time going round in circles with the previous implementation and could only unwind when I simplified everything. Then the ideas became simpler and easier to extend. This is just another step. It will most certainly not be the final state.

mjp41:
While I am in favour of only doing things where necessary, I am also concerned churn is going to make it harder to parallise the down stream phases.

PR #317, opened 2020-10-12
Add support for the shadow root stack in the VM

plietar:
The VM gains two new opcodes, PROTECT and UNPROTECT, which respectively add and remove objects from the shadow root stack of the runtime.

The compiler uses liveness information to insert the operations around every method call. Note that this is very conservative. The compiler makes no attempt to guess whether a reference is in fact in the same region as one of the arguments or not. It just assumes it could and protects everything.

Couple of cleanups and fixes:
- Clarify parameter names of the shadow stack API.
- Fix a bug caused by finalisers, where the value stack grows at unexpected times.
- Add proper support for opcodes with a variable number of arguments. The Print opcode already did this, but in a very ad-hoc way.

### DIFF
--- a/src/compiler/codegen/builtins.cc
+++ b/src/compiler/codegen/builtins.cc
@@ -85,26 +85,24 @@ namespace verona::compiler
     assert(abi_.arguments >= 2);
     assert(abi_.returns == 1);
 
-    size_t value_count = abi_.arguments - 2;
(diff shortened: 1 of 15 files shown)


mjp41 (approved):
LGTM.  Thanks for doing this.  Gives us a nice way to test how the additional roots works.

Issue #303, opened 2020-10-02
Fix up alloca for local variable lookup


rengolin:
In the current implementation, local variables are all allocas and we load/store from them when we need to read/write from/to them.

This is a problem because we need to make sure we actually load/store from everywhere, and we need to alloca type to be a pointer type to the data (so we can infer the loaded/stored type), for example, when accessing fields from local variables.

We don't have a pointer type yet (neither does [MLIR](https://mlir.llvm.org/docs/LangRef/#standard-types) natively), so we hacked with a non-pointer type to get the type check to validate.

But this is broken. There are two ways to fix this:

1. We fix all places with loads/stores and create a Verona dialect pointer type, but make sure we never look at the allocas, but at their loaded/stored values.
2. We finally implement free variable analysis and get rid of the need for allocas in the first place. This analysis has been done in [AST](https://github.com/rengolin/verona/tree/zzz/freevar), and we could use that to generate MLIR directly with PHI nodes (MLIR block arguments) instead of alloca/load/store inside the CFG regions.

The former is probably "simpler" but gets us in an intermediate state that will need to change anyway. The latter will likely be necessary soon, so perhaps we can wait until we get to that point to fix the alloca problem.

Note: this has been exacerbated by #302

rengolin:
I'll close this issue and open a new one just for finally removing allocas.

PR #318, opened 2020-10-13
Fix alloca values not being loaded

rengolin:
During the type refactoring, the comparison to load a value moved to
`unkTy` which was the wrong way of querying an alloca.

While the real fix is to not need those in the first place, they help
simplify condition/loop lowering. Once we simplify `for` loops, for
example, and we get to a single loop lowering, we should expand it to
carry free variables as arguments to basic blocks, and then we can get
rid of the alloca/store/load mode.

Most test changes are the introduction of loads before calls, but the
re-numbering of SSA values make it look like it's a big thing.

Partially implements #303.

Based on #316.

### DIFF
--- a/src/mlir/generator.cc
+++ b/src/mlir/generator.cc
@@ -21,6 +21,14 @@ namespace
     return !bb->getOperations().empty() && bb->back().isKnownTerminator();
   }
 
+  /// Check if value is from an alloca operation
+  /// TODO: This will probably disappear entirely once we have free var analisys
+  bool isAlloca(mlir::Value val)
+  {
+    // For now, allocas are opaque operations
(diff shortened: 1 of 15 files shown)


mjp41 (approved):
Looks fine to me.  
I think we should prioritise moving away from alloca.  I have been looking at doing some of the later passes, and where we assume explicit type information is available is going to change quite a bit when we move away from alloca.

rengolin:
> Looks fine to me.
> I think we should prioritise moving away from alloca. I have been looking at doing some of the later passes, and where we assume explicit type information is available is going to change quite a bit when we move away from alloca.

Agreed. I'm looking at it now. There is an old free variable analysis in the AST branch of mine, but I want to do the de-sugaring of for loops first, otherwise we'll have to patch two loop lowerings.

PR #323, opened 2020-10-14
Clangformat use git ls-files.

### LINKED ISSUE #319: Running clangformat rule touches snmalloc
For some reason, snmalloc is in `src/rt/external` instead of just `external` and that's messing with the clang format rule that globs `*.cc` and `*.h` inside `src`.

I don't see the problem in the CI, but when I run it on my machine, I get:
```
$ ninja clangformat
[1/1] cd /home/rengolin/devel/msrc/verona/build && /opt/clang-format-static/clang-format-9 -i /home/rengolin/devel/ms.../ubench.cc /home/rengolin/devel/msrc/verona/src/rt/test/systematic.h /home/rengolin/devel/msrc/verona/src/rt/verona.h
$ git diff
diff --git a/src/rt/external/snmalloc b/src/rt/external/snmalloc
--- a/src/rt/external/snmalloc
+++ b/src/rt/external/snmalloc
@@ -1 +1 @@
-Subproject commit 923705e514e850ee0070cef470fca9d9c47ca3da
+Subproject commit 923705e514e850ee0070cef470fca9d9c47ca3da-dirty
```

Which then I have to fix with:
```
$ git -C ../src/rt/external/snmalloc/ checkout .
Updated 1 path from the index
```

There are two ways to go about this and I'm not sure which one is best:
1. Move snmalloc to `external`, like the other submodules
2. Patch the `GLOB` expression to avoid `src/rt/external` (or any external) directories

mjp41:
Fixes #319


Theodus (approved):
LGTM

CMakeLists.txt
```diff
@@ -30,9 +30,18 @@ macro(clangformat_targets)
     message(WARNING "Not generating clangformat target, no clang-format tool found")
   else ()
     message(STATUS "Generating clangformat target using ${CLANG_FORMAT}")
-    file(GLOB_RECURSE ALL_SOURCE_FILES src/*.cc src/*.h src/*.hh)
+    find_package(Git)
+    execute_process(
+      COMMAND ${GIT_EXECUTABLE} ls-files *.cc *.h *.hh
```
rengolin, line 35: NIT: I didn't find any .hh file in the source...

CMakeLists.txt
```diff
@@ -30,9 +30,18 @@ macro(clangformat_targets)
...
+      WORKING_DIRECTORY "${CMAKE_CURRENT_SOURCE_DIR}/"
+      OUTPUT_VARIABLE ALL_SOURCE_FILES_STRING
+    )
+
+    string(REPLACE "\n" ";" ALL_SOURCE_FILES ${ALL_SOURCE_FILES_STRING})
```
rengolin, line 40: Why not `" "`? Isn't `;` going to stop at the first file?

rengolin (approved):

CMakeLists.txt
mjp41: We did, but not since we went public.  I can remove it.

Also, do we want clangformat to work on the 'td' files?

CMakeLists.txt
mjp41: Because if you use " ", then it outputs the " " escaped, so you get
```
   clangformat -i path1/file1\ path2/file2\ 
```
With the ";" cmake interprets this as a list, and it has special code for putting lists into the command line, that doesn't protect the " ".

Does this make sense? No! Does this work? surprisingly.  If you actually wanted the ";" to go to the command line, I have no clue how to do that.  But I stopped reading stack overflow at that point.

CMakeLists.txt
plietar: Last time I tried to run clangformat on the td files it worked very badly.

rengolin: Ignore td files. :)

rengolin (approved):

CMakeLists.txt
rengolin: You're clearly much braver than I am, sir.

PR #327, opened 2020-10-15
AST utils to use Ast instead of WeakAst

rengolin:
When creating this utility class, I didn't quite know what to expect.
But now, after enough commits, I can see it won't be more than just a
bunch of wrappers.

The ownership model was made "weak" to make sure we don't hold objects
for longer than they need, and don't create cycles in the reference
model. But since we hold nothing in the class itself, and all methods
are static, this will never happen anyway.

Moving to Ast shared pointers avoids all calls to `lock()` during the
interplay between weak_ptr and shared_ptr, and removes the need to
create `ptr` objects for multiple accesses. It also makes debugging the
code easier, as you can print the object with `*ast.get()`, while trying
to lock first won't work on most cases.

### DIFF
--- a/src/mlir/ast-utils.h
+++ b/src/mlir/ast-utils.h
@@ -21,16 +21,12 @@ namespace mlir::verona
    * operations. While the AST design is still in flux, we can keep this around,
    * but once we're set on its structure, this should move to src/ast instead.
    *
-   * Current Memory Model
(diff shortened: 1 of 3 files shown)


mjp41 (approved):
LGTM.  Seems purely mechanical.

Issue #321, opened 2020-10-14
De-sugar for loops into while loops


rengolin:
Currently, we lower `for` loops in MLIR in a different method than `while` loops, adding calls to `has_value`, `apply` and `next` directly. This is fragile, as it's assuming semantics without actually checking them.

With a new de-sugaring AST pass (#320), this would be a good fit for such a pass.

rengolin:
WIP: https://github.com/rengolin/verona/tree/parser

rengolin:
Some progress. I can now lower for loops into something that looks like a while loop close enough that the MLIR representation is identical. For this, I need #333 (which implement's `invoke` in MLIR). After #333, the only MLIR generator change is to remove the `parseForLoop`, which is awesome.

The local ast structure is the same but the `atom` and `blockexpr` nodes are not identical. In the original `while`, they have a number after them (ex. `atom/15`) and when I create a `node(ast, "atom")`, they don't. In the end, after all passes, the final ast is identical, so I'm not sure there is a big difference.

Missing bits:
 * merge #333 into master
 * hygienic name for the iterator
 * understand impact of the `atom`/`blockexpr` difference
 * create more patterns to improve robustness (both ast and mlir)
 * general code cleanup & extensive code review

rengolin:
Update on the numbering system in the ast. That's a feature specific to the peg library we use that exposes which of the choices in the grammar was taken to generate that node. So, `atom/15` means it had many choices and it picked the 15th one in the list. This is irrelevant for us, and when we create the atom by hand, it should have no number anyway, because there's only one choice.

Also, #333 was merged, so now on to implementing hygienic ids.

PR #329, opened 2020-10-19
Three changes to the ast parser: comments on ast.h, new pass.h, ast_dump

rengolin:
This PR adds three things:
1. Comments to all `ast.h` methods, to help clarify what they do and what are the inter-dependencies.
2. A `Pass` class wrapper for the function pointers of each pass to have a name, so that we can print an `after-tree` with a name associated.
3. An `ast_dump` method, for use in debuggers.

It's not possible to create tests yet due to the nature of our (hard-coded) test runner (in CMake).

### DIFF
--- a/src/ast/ast.h
+++ b/src/ast/ast.h
@@ -9,41 +9,68 @@ namespace ast
   struct SymbolScope;
   struct Annotation;
 
-  using Scope = std::shared_ptr<SymbolScope>;
+  /// Ast nodes
   using AstImpl = peg::AstBase<Annotation>;
   using Ast = std::shared_ptr<AstImpl>;
+  /// Weak reference to an Ast node
   using WeakAst = std::weak_ptr<AstImpl>;
+  /// IDs: (variable, function, etc) names
   using Ident = std::string;
+  /// Numeric tag from "name"_ operator
   using Tag = unsigned int;
 
+  /// Annotation wrapper for AstBase with Symbol scope
+  using Scope = std::shared_ptr<SymbolScope>;
   struct Annotation
   {
     Scope scope;
   };
 
+  /// Symbol scope (ID -> Ast)
   struct SymbolScope
   {
     std::map<Ident, WeakAst> sym;
   };
 
+  /// Create a new token (value node) with name and token value
+  /// using a previous token's source location, position and length.
   Ast token(const Ast& ast, const char* name, const std::string& token);
+  /// Create a new node with name and no children
+  /// using a previous token's source location, position and length.
   Ast node(const Ast& ast, const char* name);
+  /// Add a child to an existing node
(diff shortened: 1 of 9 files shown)


rengolin:
These commits help work on the ast and are part of #321 effort.

rengolin:
Will convert this to stop at a specific pass (by name). If quick, will also add a dump method that we can call from the debugger.

rengolin:
Re-implemented as a "stopAt" flag rather than a "dump-before-all", so that we can only see one ast and not the whole list of things.

Also added a `dump` method that only seems to work on GDB, not LLDB.

src/ast/module.cc
```diff
@@ -187,11 +188,15 @@ namespace
...
+  bool run_passes(
+    ModulePtr& m,
+    const std::string& stopAt,
+    const pass::Passes& passes,
+    err::Errors& err)
   {
     return dfs::post(
       m,
-      [](auto& m, auto& passes, auto& err) {
+      [&stopAt](auto& m, auto& passes, auto& err) {
```
sylvanc, line 199: I should really rewrite dfs::post to expect the lambda to capture more. It's better to do what you've done here (capture &stopAt) than to thread things through.

src/ast/pass.h
```diff
@@ -0,0 +1,34 @@
...
+
+#include "ast.h"
+#include "err.h"
+
+#include <vector>
+
+namespace pass
+{
+  /// A Pass object has a name and an operation, that is executed on AST nodes.
+  class Pass
```
sylvanc, line 13: Personally, I would just use a `std::tuple` here. Of course it is fine as it is!

src/ast/ast.h
```diff
@@ -69,4 +96,16 @@ namespace ast
...
+  extern "C"
+  {
+    /// This method is for debug only,
+    /// when needing to dump ast trees inside a debugger.
+    /// This is needed because `peg::ast_to_s` is templated.
+    /// Return 0 to make gdb happy.
+    /// FIXME: This does not work on LLDB:
+    ///   error: Couldn't lookup symbols
+    ///     ast::dump(std::shared_ptr<peg::AstBase<ast::Annotation> > const&)
+    size_t dump(const ::ast::Ast& ast);
```
sylvanc, line 109: As we talked about, perhaps `::ast_dump` will work in LLDB.

sylvanc (approved):

src/ast/ast.h
rengolin: It does, updated now.

src/ast/pass.h
rengolin: Personal preference. I think it's marginally easier to read the code (`pass.name` rather than `get<0>(pass)` and `pass(args)` rather than `get<1>(pass)(args)`).

PR #334, opened 2020-10-22
Backpressure blocker

Theodus:
This PR detects muted cowns lingering in the mute map of the last scheduler thread to shut down. A runtime test is also added for a scenario where the backpressure system requires tracking the `blocker`.

For a behaviour that requires the cowns `{a, b}` and `a` is acquired before `b`, if cown `b` cannot be fast-acquired then `b` will be tracked as the blocker for `a` until the behaviour is run. This allows the backpressure system to raise the priority of `b` in the event that `a` has it's priority raised after it has been acquired for a message on `b` and `b` may become muted before the behaviour runs and `a` is rescheduled.


src/rt/sched/cown.h
```diff
@@ -528,26 +530,38 @@ namespace verona::rt
...
         auto* next = body->cowns[body->index];
         Systematic::cout() << "MultiMessage " << m << ": fast requesting "
                            << next << ", index " << body->index << std::endl;
 
-        {
-          // Hold epoch in case priority needs to be raised after message is
-          // placed in queue. TODO: only hold if `high_priority` is true
-          Epoch e(alloc);
+        if (body->index > 0)
+          body->cowns[body->index - 1]->blocker = next;
```
mjp41, line 539: I think you need to check if 
```
body->cowns[body->index - 1]
```
has a `Priority::High`, and use that for `high_priority`.

mjp41 (changes requested):

src/rt/sched/cown.h
```diff
@@ -804,13 +821,18 @@ namespace verona::rt
...
+      {
+        auto sleeping = queue.wake();
+        UNUSED(sleeping);
+        assert(!sleeping);
+        schedule();
+      }
+      else if ((state == Priority::High) && (blocker != nullptr))
+      {
+        Systematic::cout() << "Unblock cown " << blocker << std::endl;
+        blocker->backpressure_transition(Priority::High);
```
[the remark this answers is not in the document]
mjp41: Also, as discussed on Teams.  I think the `blocker` needs protecting from deallocation here by holding an epoch.

src/rt/sched/cown.h
```diff
@@ -850,6 +850,9 @@ namespace verona::rt
     /// Return true if a sender to this cown should become low priority.
     inline bool triggers_muting()
     {
+      if (is_collected())
+        return false;
+
```
mjp41, line 855: If `is_collected()` is false, will that stay true at this point?

Theodus: It doesn't need to stay false in this execution since the next mute map scan would see that the mutor has been collected. I've moved the condition after the yield.

src/rt/sched/cown.h
```diff
@@ -115,6 +115,8 @@ namespace verona::rt
 
     std::atomic<Status> status{};
     std::atomic<Priority> bp_state = Priority::Normal;
+    // TODO: combine into `bp_state` field
+    Cown* blocker = nullptr;
```
mjp41, line 119: I think this should be `std::atomic` as it is a communication channel between threads.

src/rt/sched/cown.h
```diff
@@ -525,29 +527,66 @@ namespace verona::rt
...
+          // Double check the priority of the most recently acquired cown to
+          // prevent deadlock.
+          auto* cur = body->cowns[body->index - 1];
+          high_priority = high_priority ||
+            (cur->bp_state.load(std::memory_order_acquire) & PriorityMask::High)
+#ifdef USE_SYSTEMATIC_TESTING
+            || Systematic::coin(3)
+#endif
+            ;
+          yield();
```
mjp41, line 561: I am not happy this is correct.  I understand what you are trying to do, but I think the issue is that a thread can set a cown to high-priority and not see a `blocker`, but the owning scheduler thread does not see it is set to high-priority, so does not perform `backpressure_unblock(next, e);` on the subsequent cowns.

I am coming to the conclusion that compressing `bp_state` and `blocker` into a single machine word is pretty much the only solution. Or you need to add fences.

mjp41: I think something like
```
when () {
    for (lots of times) { 
        when (c_muted1) { 
            when (c_overloaded1) { 
                something slow 
            }
        }
    }
}

when () {
    for (lots of times) { 
        when (c_muted2) { 
            when (c_overloaded2) { 
                something slow 
            }
        }
    }
}

when (c_overloaded1, c_muted2)
{

}
```
might exhibit the problem.  `backpressure3` almost does this, but the pair of the muted cown and overloaded cown are too related.  The muting is caused by the overloading, which induces additional orderings.  The example above has independence between the muted and the overloaded cown.

Theodus: The priority and blocker have been merged into the same atomic variable. The test you have described has also been added to `func-backpressure`.

mjp41 (approved):
LGTM

PR #335, opened 2020-10-24
Update LLVM and use new ODS Type definitions

### LINKED ISSUE #331: New MLIR ODS type declaration
MLIR has a new [ODS type definition](https://mlir.llvm.org/docs/OpDefinitions/#type-definitions) that we may be able to use for our interned types. Though, to me, it still looks like moving the C++ boilerplate to the table-gen files (although less of it). 

It does allow use in the table-gen operation definitions, but we're getting mostly unknown types, the type inference is independent of the table-gen code and the complexity of type unions and intersections is too great for a clean ODS usage.

Overall, I'm not convinced it would help us. The current table-gen representation is clean enough as it is.

@plietar

plietar:
This cuts down on a lot of boilerplate around the definition of types, accessors and storage classes.

Because of complications due to its recursive structure, we can't use ODS with ClassType yet. We may be able to use ODS with a custom storage class and extra class declarations, but it leads to a messy Frankenstein definition, so it is easier to keep it all in C++.

Similarly, we can't use the parsing/printing code ODS generates, because our implementations need extra context about recursive classes (ie. the class_stack fields of TypePrinter and TypeParser).

Fixes #331

### DIFF
--- a/src/mlir/dialect/VeronaDialect.h
+++ b/src/mlir/dialect/VeronaDialect.h
@@ -7,4 +7,4 @@
 
 // This line must come after "normal" includes.
 // Having this comment prevents clang-format from re-ordering it.
-#include "dialect/VeronaOpsDialect.h.inc"
+#include "dialect/VeronaDialect.h.inc"
--- a/src/mlir/dialect/VeronaDialect.h
+++ b/src/mlir/dialect/VeronaDialect.h
@@ -3,11 +3,8 @@
 
 #pragma once
 
-#include "dialect/VeronaTypes.h"
 #include "mlir/IR/Dialect.h"
 
-namespace mlir::verona
-{
+// This line must come after "normal" includes.
+// Having this comment prevents clang-format from re-ordering it.
 #include "dialect/VeronaOpsDialect.h.inc"
-
-} // namespace mlir::verona
--- a/src/mlir/dialect/VeronaDialect.td
+++ b/src/mlir/dialect/VeronaDialect.td
@@ -27,18 +27,9 @@ class Verona_Op<string mnemonic, list<OpTrait> traits = []> :
     let verifier = [{ return ::verify(*this); }];
 }
 
+class Verona_TypeDef<string name> : TypeDef<Verona_Dialect, name> { }
+
 def Verona_Type : DialectType<Verona_Dialect,
     CPred<"isaVeronaType($_self)">, "Verona type">;
-
-def Verona_TypeAttr :
-    Attr<And<
-      [CPred<"$_self.isa<TypeAttr>()">,
-       CPred<"isaVeronaType($_self.cast<TypeAttr>().getValue())">]>>
-{
-  let storageType = [{ TypeAttr }];
-  let returnType = [{ Type }];
-  let valueType = NoneType;
-  let convertFromStorage = "$_self.getValue().cast<Type>()";
-}
 
 #endif // VERONA_DIALECT
--- a/src/mlir/dialect/VeronaDialect.td
+++ b/src/mlir/dialect/VeronaDialect.td
@@ -13,7 +13,7 @@ include "mlir/IR/OpBase.td"
 def Verona_Dialect : Dialect {
     let name = "verona";
     let summary = "A prototype Verona language MLIR dialect.";
-    let cppNamespace = "verona";
+    let cppNamespace = "::mlir::verona";
 }
 
 //===----------------------------------------------------------------------===//
--- a/src/mlir/dialect/VeronaTypes.h
+++ b/src/mlir/dialect/VeronaTypes.h
@@ -8,11 +8,21 @@
 
 namespace mlir::verona
 {
+  /// Describes properties of references, as part of their type.
+  /// This enum is embedded in a CapabilityType, as defined in ODS.
+  enum class Capability
+  {
+    Isolated,
+    Mutable,
+    Immutable,
+  };
+
   Type parseVeronaType(DialectAsmParser& parser);
   void printVeronaType(Type type, DialectAsmPrinter& os);
 
   /// Returns true if the type is one defined by the Verona dialect.
   bool isaVeronaType(Type type);
+
   /// Returns true if all types in the array are ones defined by the Verona
   /// dialect.
   bool areVeronaTypes(llvm::ArrayRef<Type> types);
@@ -37,90 +47,8 @@ namespace mlir::verona
 
   namespace detail
   {
-    struct MeetTypeStorage;
-    struct JoinTypeStorage;
-    struct IntegerTypeStorage;
-    struct FloatTypeStorage;
-    struct DescriptorTypeStorage;
-    struct CapabilityTypeStorage;
     struct ClassTypeStorage;
-    struct ViewpointTypeStorage;
   }
-
-  /**
-   * Meet types are intersections between types (A & B).
-   */
-  struct MeetType
-  : public Type::TypeBase<MeetType, Type, detail::MeetTypeStorage>
-  {
-    using Base::Base;
-    static MeetType
-    get(MLIRContext* ctx, llvm::ArrayRef<mlir::Type> elementTypes);
-    llvm::ArrayRef<mlir::Type> getElements() const;
-  };
-
-  /**
-   * Join types are unions between types (A | B).
-   */
-  struct JoinType
-  : public Type::TypeBase<JoinType, Type, detail::JoinTypeStorage>
-  {
-    using Base::Base;
-    static JoinType
-    get(MLIRContext* ctx, llvm::ArrayRef<mlir::Type> elementTypes);
-    llvm::ArrayRef<mlir::Type> getElements() const;
-  };
-
-  /**
-   * Unknown types are derived types from operations that cannot define the type
-   * at lowering stage, but will later be replaced by other types during type
-   * inference.
-   */
-  struct UnknownType : public Type::TypeBase<UnknownType, Type, TypeStorage>
-  {
-    using Base::Base;
-
-    static UnknownType get(MLIRContext* context);
-  };
-
-  /**
-   * A static class descriptor type, used for access to static members of the
-   * class, including fields and methods.
-   */
-  struct DescriptorType
-  : public Type::TypeBase<DescriptorType, Type, detail::DescriptorTypeStorage>
-  {
-    using Base::Base;
-
-    static DescriptorType get(MLIRContext* context, Type ref);
-    Type getDescribedType() const;
-  };
-
-  enum class Capability
-  {
-    Isolated,
-    Mutable,
-    Immutable,
-  };
-
-  /**
-   * Capability types represents properties of individual references eg.
-   * a `String & imm` and a `String & mut` could point to the same object.
-   *
-   * Isolated: An entry point to a new region. There can be more than one
-   * reference to the entrypoint, but only one of them can be Isolated ie.
-   * the others must be Mutable.
-   *
-   * Immutable: A stronger property than Read-Only. It guarantees that no
-   * mutable aliases to that object exist anywhere else.
-   */
-  struct CapabilityType
-  : public Type::TypeBase<CapabilityType, Type, detail::CapabilityTypeStorage>
-  {
-    using Base::Base;
-    static CapabilityType get(MLIRContext* ctx, Capability cap);
-    Capability getCapability() const;
-  };
 
   /**
    * A class is described both by its name and its list of fields. A class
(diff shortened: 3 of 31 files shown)


plietar:
As a side-effect of the LLVM update, I think we could now use a namespace that is not a descendant of `mlir`, eg. `verona::dialect` or `verona::compiler` (though that's what the old compiler uses). This has the benefit (or disadvantage, depending on how you look at it) of requiring uses of mlir types to be qualified everywhere eg. `mlir::Type`.

I'll leave that to another PR, if this is even something we want to do.

plietar:
@rengolin This needs an update of the prebuilt LLVM blob. I don't know how to do this (and probably wouldn't have permission to anyway). Would you mind taking care of it please?

I don't know what your process is for picking a commit version. I just use whatever what HEAD when I did the update, ie. https://github.com/llvm/llvm-project/commit/73811d32c72d0760c8c2066e4675dd6f1a7bbef7.

rengolin:
> @rengolin This needs an update of the prebuilt LLVM blob. I don't know how to do this (and probably wouldn't have permission to anyway). Would you mind taking care of it please?

I can do that, no worries.

> I don't know what your process is for picking a commit version. I just use whatever HEAD was when I did the update, ie. [llvm/llvm-project@73811d3](https://github.com/llvm/llvm-project/commit/73811d32c72d0760c8c2066e4675dd6f1a7bbef7).

That's not good because we don't test it, we rely on it passing buildbots. What I do is to use [my scripts](https://github.com/rengolin/verona/tree/zzz/bot-scan/utils/bot-scan) to scan the last 100 or so commits and pick the ones that passed most of the critical buildbots.

Then I trigger an LLVM Build of that commit hash and once the blobs are in the Azure storage, I checkout that hash on my local branch. Once you submit the PR, the change in the commit hash will trigger the CI build to pull from the right blob from the storage.

I then keep both previous and new blobs for a while in case people have old branches with the old LLVM, especially if they're doing things that are not related to LLVM, which means they don't need to rebase to master until right before the pull request.

src/mlir/dialect/VeronaDialect.h
```diff
@@ -3,11 +3,8 @@
 
 #pragma once
 
-#include "dialect/VeronaTypes.h"
 #include "mlir/IR/Dialect.h"
 
-namespace mlir::verona
-{
+// This line must come after "normal" includes.
```
rengolin, line 8: Keeping it inside the namespace as before would have prevented that, too. Is the removal of the namespace a required change from the new tree?

src/mlir/dialect/VeronaDialect.td
```diff
@@ -27,18 +27,9 @@ class Verona_Op<string mnemonic, list<OpTrait> traits = []> :
     let verifier = [{ return ::verify(*this); }];
 }
 
+class Verona_TypeDef<string name> : TypeDef<Verona_Dialect, name> { }
```
rengolin, line 30: Does this mean "any type defined in the verona dialect is a verona type"?

rengolin (commented):
Much cleaner implementation, nice!

You should pick a commit that has been tested and passed the buildbots we care about (the list is in the `zzz/bot-scan` branch). You should also be able to trigger the LLVM Builds by hand  in our DevOps, but you need to change the GOOD_HASH variable to the one you picked above.

If you can't do that, let me know and I can help you.

src/mlir/dialect/VeronaDialect.h
plietar: Sorry this is referenced in the commit message but not in the PR description.

Yes, all the namespace stuff is required by changes in MLIR. TableGen now wraps the generated code with `namespace` blocks, using whatever `cppNamespace` we define in the `.td`.

It’s possible we could leave `cppNamespace` empty, I’m not sure, but that would certainly not be the intended use case.

Cf. https://reviews.llvm.org/D86811

src/mlir/dialect/VeronaDialect.td
plietar: Sorry I don’t really understand the question. The switch to ODS does not affect the implementation of `isaVeronaType`, which is roughly how we define what a “Verona Type”.

To be honest I don’t really know what impact the dialect we pass as the first argument of `TypeDef` has. I’ll try and dig into that.

plietar:
I can't seem to get your scripts to work.

The whole `http://lab.llvm.org:8011/json` API has been giving me 404s for the past few days.
Not sure if that's a transient issue on LLVM's side, or if the API got removed.

rengolin:
Dang, the API has changed with the buildbot upgrade. :(

You can look at the [waterfall page](http://lab.llvm.org:8011/#/waterfall) and see if they're all green/yellow. Looking at them now, either the Windows clang or the Windows MSVC bots are red. There may be a new build that makes them yellow/green, but it could also be that they haven't failed yet.

The bots we care are Clang on Linux/Windows, MSVC on Windows, all on x86_64. If the AArch64 bots are green too, that's a bonus.

plietar:
I'm sorry, I am pretty confused about how to read the waterfall (where do I find the commit hash?), or even which bots I need to care about (is `clang-x64-windows-msvc` clang or msvc?). I've tried to script the new Buildbot API, but similarily I have no idea what I actually need to consider.

(This can wait Monday, don't worry)

rengolin:
I think [this commit](http://lab.llvm.org:8011/#/changes/2318) (235dfcf70ab) should be good enough.

What I did was to go through the latest changes, moved past ~2 days ago (to give time for slow builders) and picked a commit that doesn't introduce a new feature (which could break new builds) but rather fix a crash from a previous build. It lists all of its builds as green.

The assumption is that if that build had broken something else, the author would have fixed everything in them. If another bug had caused direct failure to that commit, it would have shown in its result status as red, which it didn't.

I know it's not optimal, but the JSON API has changed and I can't seem to easily fix. Instead of holding your patch until I fix it, I think we better ignore it this time around (and have a strategy if we can't fix it).

src/mlir/dialect/VeronaDialect.h
rengolin: Makes sense, thanks! Maybe we should list all the changes in each update and what made us do it, in the commit message.

src/mlir/dialect/VeronaDialect.td
rengolin: I was under the impression that the type attribute below (now removed) had some implicit connection with the type storage defined for each of our internal "Verona types", and that somehow the new typedef was replacing it.

Maybe they're two independent changes?

src/mlir/dialect/VeronaTypes.h
```diff
@@ -8,11 +8,19 @@
 
 namespace mlir::verona
 {
+  enum class Capability
```
rengolin, line 11: This deserves a comment making it explicit how and where it's used.

rengolin:
All failures seem to be because the module dump doesn't print two empty lines in the beginning any more (which is actually a good change). :)

Also, `ninja clangformat` now works as expected, if you have `clang-format-9` in the path.

plietar:
I'm getting a [build failure on Windows](https://dev.azure.com/ProjectVeronaCI/Project%20Verona/_build/results?buildId=1718&view=logs&j=49e8ed45-37ec-5c12-712f-f6a8b8f168e0&t=acbf99da-c3f7-5be9-e9fd-50b286298f18):
```
CMake Error at build/Release/mlir/install/lib/cmake/mlir/MLIRTargets.cmake:798 (message):
The imported target "mlir_async_runtime" references the file
         
    "D:/a/1/s/build/Release/mlir/install/lib/mlir_async_runtime.lib"
         
but this file does not exist.
```

`mlir_async_runtime` is a new target in MLIR.

Will investigate some more.

plietar:
Looking at the prebuilt archive, there is a `bin/mlir_async_runtime.dll` file but no `lib/mlir_async_runtime.lib`.
This is unlike other comparable targets eg. `mlir_runner_utils`, which have both the dll and the lib files.

The build log seems to suggest that the file was created though, so I don't know why it wasn't included in the archive.

rengolin:
Is it possible that the file was created but not installed? Perhaps it's missing from a CMake file and no one found it because most people use the build directory not the install one.

rengolin:
Looking at the [CMake file](https://github.com/llvm/llvm-project/blob/master/mlir/lib/ExecutionEngine/CMakeLists.txt#L104), I can see the option `EXCLUDE_FROM_LIBMLIR`, which seems to be the case for all libraries in the EE directory, including `mlir_runner_utils`.

This library was added recently ([36ce915ac54a, 22nd Oct 2020](https://github.com/llvm/llvm-project/commit/36ce915ac54a97d40c8e340afe7d46398d97b4ee)) which seems a reapply of a previously failed commit.

The reason given by the [revert](https://github.com/llvm/llvm-project/commit/4986d5eaff359081a867def1c6a2e1147dbb2ad6) is: "The build with SHARED_LIBRARY=ON is broken."

Perhaps we should pick a commit slightly before that one, or one much later (hoping whatever it was broken got fixed).

rengolin:
I started a [new build](https://dev.azure.com/ProjectVeronaCI/Project%20Verona/_build/results?buildId=1738&view=results) on [b2ac9681a70](https://github.com/llvm/llvm-project/commit/b2ac9681a70) which seems to have passed MLIR bots and others.

We can try the Verona build as soon as that's finished.

plietar:
I finally sorted out my Windows VM and got LLVM building on there, so I was able to do a bit of investigation.

In [AsyncRuntime.h](https://github.com/llvm/llvm-project/blob/58c586e701889250c08285193dc75d94a7ed302d/mlir/include/mlir/ExecutionEngine/AsyncRuntime.h#L19), the `MLIR_ASYNCRUNTIME_EXPORT` macro is set to `dllexport` only if `mlir_c_runner_utils_EXPORTS` is defined. However that is most certainly a copy/paste error and should have been `mlir_async_runtime_EXPORTS` instead. Since nothing gets exported, [the linker does not produce a `.lib` file](https://docs.microsoft.com/en-us/cpp/build/reference/implib-name-import-library?view=msvc-160).

With that simple change applied, the build process correctly produces a `lib\mlir_async_runtime.lib` file. I haven't tried building Verona against that, but it should work.

plietar:
https://reviews.llvm.org/D91196

rengolin:
https://reviews.llvm.org/rG6fd9e59e1b3ac15430eccd4011e063605f58039c

plietar:
Triggered a new LLVM build for https://github.com/llvm/llvm-project/commit/f6a326adef4fa2b7654f0d12d8c9be32aecbd2e3. It looked [green enough](http://lab.llvm.org:8011/#/changes/3380).

rengolin:
The new LLVM build fails on Windows with the message:
```
Generator
  Visual Studio 16 2019
could not find any instance of Visual Studio.
```

I am at a loss. the build worked on Windows on `master` and you have not changed anything on the Windows side in your branch. Why are these things so horrendously random? What did we do to upset the infrastructure gods? Why the gremlins only come out on Fridays?

rengolin:
Ok, I managed to get all tarballs in the blob storage. See if you can use that to build Verona with. Sorry it took so long, the gremlins had a field day.

plietar:
> I am at a loss. the build worked on Windows on master and you have not changed anything on the Windows side in your branch.

Could Azure have changed their VMs?

> Ok, I managed to get all tarballs in the blob storage. See if you can use that to build Verona with. Sorry it took so long, the gremlins had a field day.

Thanks! I'll give it a try now.

rengolin:
> Could Azure have changed their VMs?

Those are our VMs and I specifically create it with a `vs2019` image: `MicrosoftVisualStudio:visualstudio2019latest:vs-2019-ent-latest-ws2019:latest`.

We install CMake on every build, so all deps should be there. 

My Windows test environment is in the office and lockdown. :(

I'll see what I can do. Hopefully the new build will "just work" and we'll have another month to solve that problem.

src/mlir/dialect/VeronaTypes.h
```diff
@@ -155,8 +83,8 @@ namespace mlir::verona
    *   FieldsRef fields = { { "f", a } };
    *   a.setFields(a);
    *
-   * TODO: While recursive types can be constructed programmatically, they can
-   * neither be parsed nor printed yet.
+   * Because of these extra complications, this type cannot be generated using
```
rengolin, line 86: Nit: perhaps would be nice to very briefly list what those complications are.

rengolin (approved):
With a small nit on a comment, LGTM. Thanks!

plietar:
Woohoo, finally!

PR #338, opened 2020-11-02
Relaxed semantics for message send and deallocation

mjp41:
This PR is more of a RFC on altering the semantics of the language to simpify the type system design. It is partly a brain dump to flesh out the ideas.

### DIFF
--- a/docs/alternativesemantics.md
+++ b/docs/alternativesemantics.md
@@ -142,9 +142,13 @@ The definitions take a region parameter so that "freshness" can be suitably defi
    field_read(tau1 | tau2, f) = field_read(tau1, f) | field_read(tau2, f)
    field_read(tau1 & tau2, f) = field_read(tau1, f) & field_read(tau2, f)
    field_write(tau1 | tau2, f) = field_write(tau1, f) & field_write(tau2, f)
-   field_write(tau1 & tau2, f) = field_write(tau1, f) & field_write(tau2, f)
+   field_write(tau1 & tau2, f) = field_write(tau1, f) | field_write(tau2, f)   // Need to think about this.
+
+   field_write(tau1 & tau2, f) = field_write(tau1, f) & field_write(tau2, f)   // Need to think about this.
 ```
 ]
+
+
 
 ### Aliasing assignment generates a fresh region
 if it follows an `iso`. If it follows an `imm` then it is in the immutable space, otherwise it is in the same region. 
@@ -336,6 +340,7 @@ It also means when manipulating things of type `imm`, we would need dynamic test
 
 
 ## Interesting examples
+
 This example is trying to think up places where accessing inside regions would be interesting. 
 ```
 recv_packet(packet: Array[byte] & iso)
@@ -367,3 +372,16 @@ recv_packet(packet: Array[byte] & iso)
     }
 }
 ```
+
+So discussing this examples led to a a question from Renato:
+```
+// z: iso
+y = (x.fiso = z);
+// y: iso, z: mut 
+// or 
+// y: iso, z: no access
+```
+I think the first makes the most sense.  But the second might be easier to implement.
+
+The first is closer to the semantics of `::insert` in C++ that returns an iterator to the inserted element.  `mut` is kind of like an iterator of a region.
+
--- a/docs/alternativesemantics.md
+++ b/docs/alternativesemantics.md
@@ -2,7 +2,7 @@
 
 ## Abstract
 
-This document is a thought experiment in how runtime support can reduce the complexity of the type system.  The key insight is to delay certain operations to a point where we can trivially know that there is no aliasing.  Aliases into a region are difficult to track accurately at compile time, by relaxing the semantics of when a programmer can assume an event occurs allows for a simpler type system.
+This document is a thought experiment in how runtime support can reduce the complexity of the type system.  The key insight is to delay certain operations to a point where we can trivially know that there is no aliasing.  Aliases into a region are difficult to track accurately at compile time.  We propose relaxing the semantics of when a programmer can assume an event occurs, which allows for a simpler type system.
 
 ## Motivation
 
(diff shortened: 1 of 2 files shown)


mjp41:
@lukecheeseman, @sophiaIC @SeanTAllen  for some reason I couldn't tag you as reviewers.

docs/alternativesemantics.md
```diff
@@ -0,0 +1,202 @@
...
+]
+
+```
+field_R(f, tau_x) = tau_f
+fresh r
+-------------------------------------
+G |- z = x.f -| G, z: tau_x ->r tau_f
+```
+
+Extraction uses a fresh region
```
rengolin, line 141: Use ### to indent the sections.

docs/alternativesemantics.md
```diff
@@ -0,0 +1,202 @@
...
+field_r(f, tau_x) = tau_f_r
+exists r'. tau_y <: tau_x <~r' tau_f_w
+fresh r
+--------------------------------------
+G |- z = (x.f = y) -| G, z: tau_x <~r tau_f_r
+```
+Is y still accessible?  If it was an iso, then no.  But otherwise, yes.  So remove y, if we can prove it was not an iso?
+
+
+Method call?
```
rengolin, line 155: Use ### to indent the sections.

docs/alternativesemantics.md
```diff
@@ -0,0 +1,202 @@
...
+
+### Additional root sets
+
+Not tracking regions and subregions precisely means we may need to be more careful with codegen for the additional root set.  For instance, we can use have the same region multiple times, which means we use the region additional roots in a globally stack based way.  I.e.
+```
+  r1.add_root(a)
+  r2.add_root(b)
+  ...
+  r1.remove_root(a)
+  ...
```
rengolin, line 189: I'm suspecting that this area is where the problem lies. It'd be easier to see if you had some `code here` text and explain in the paragraph below that this is what breaks the stack discipline.

docs/alternativesemantics.md
```diff
@@ -0,0 +1,202 @@
...
+  r2.remove_root(b)
+```
+Is not allowed as if r1 and r2 are the same region, then we will not be following the stack discipline.
+
+**[This may already be the case with cown aliasing, where we don't know if two cowns are the same or different.]**
+
+### Performance
+
+This approach can cause more memory and less concurrency of messages.  This may be bad for performance.
```
rengolin, line 198: "cause higher memory usage"?

docs/alternativesemantics.md
```diff
@@ -0,0 +1,202 @@
...
+
+Using compiler optimisations and global analysis could mitigate the default performance by finding cases where messages can be sent promptly.
+
+This however leads to hard to predict performance for the programmer.  We can mitigate this by providing annotations the programmer provides to guarantee certain properties must be true for optimisation, but then we are effectively implementing the complex type system.  This may still be acceptable as the barrier to entry is lower, and the performance work will require a more specialist view.  We can also potentially forbid more in the perf setting, allowing programmers to trade off flexibity of implementation with performance.
```
rengolin, line 202: My worry with type annotations is that people can find multiple special cases that provide more performance and hard-code those in their programs (and evangelise other programmers to do the same), which will make it harder for us to improve the semantics later on, with hindsight.

At such an early stage, I'd pay the performance cost and keep all annotations "private" (ie. explicitly not part of the language, but available for the compiler) and try to clean them up before we start asking people to use Verona.

mjp41: Agreed, this is forward looking, but we want to have in mind we will be doing this.  Part of the reason for making the semantics allow for the messages send at any point after the references into the region have gone away. We need to leave space in the semantics for optimisation.

docs/alternativesemantics.md
```diff
@@ -0,0 +1,202 @@
...
+The definitions take a region parameter so that "freshness" can be suitably defined. 
+
+### Type Rules
+
+Aliasing assignment generates a fresh region
+if it follows an `iso`, if it follows an `imm` then it is in the immutable region, otherwise it is in the same region. 
+
+Read versus Write types for a field
+```
+   field_r(tau1 | tau2, f) = field_r(tau1, f) | field_r(tau2, f)
```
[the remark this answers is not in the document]
mjp41: Agreed

docs/alternativesemantics.md
mjp41: After a bit more thought, I'm not sure this will be a problem.  I think we just need to scope the accesses and not try to be too clever with liveness, and stick to a stack discipline then it should be fine.

docs/alternativesemantics.md
```diff
@@ -0,0 +1,202 @@
...
+```
+G(x) : tau_x
+G(y) : tau_y
+field_w(f, tau_x) = tau_f_w
+field_r(f, tau_x) = tau_f_r
+exists r'. tau_y <: tau_x <~r' tau_f_w
+fresh r
+--------------------------------------
+G |- z = (x.f = y) -| G, z: tau_x <~r tau_f_r
+```
```
mjp41, line 151: I don't think this is enforcing that `y` is in the same region as `x` correctly.

If we have no region restrictions on either `y` or `x`, then if subtyping is subset with the semantics above, then it will be allowed, but shouldn't be.

docs/alternativesemantics.md
```diff
@@ -0,0 +1,202 @@
+# Alternative semantics for Verona.
+
+## Abstract
+
+This document is a thought experiment in how runtime support can reduce the complexity of the type system.  The key insight is to delay certain operations to a point where we can trivially know that there is no aliasing.  Aliases into a region are difficult to track accurately at compile time, by relaxing the semantics of when a programmer can assume an event occurs allows for a simpler type system.
```
ronorton, line 5: > ...compile time, by relaxing... 
This sentence is a bit garbled . Maybe should be a new sentence here without 'by'?

plietar:
Here are some rough notes, trying to compare this formulation of the type
system with previous versions. This isn't particularly thorough or structured,
just a brain dump after today's discussion.

I call "Winter 2019" the type system the work I started in in my internship,
and kept using until around Spring 2019. "POPL20" is the type system based on
topological fact I wrote in May/June/July for the (unsubmitted) POPL20 paper.

A key commonality between all variants of the type system is that there is
"must" information (eg. x and y are in the same region) and "may" information
(eg. x and y may be in the same region tree). The "must" is necessary to
preserve external uniqueness, the "may" is necessary to perform the right
invalidations.

# Design points

## "may" vs "may-not"

The "may" can be interpreted as the negation of "may-not" (eg. x is definitely
not in a sub-region of y), and vice-versa. In fact, the POPL20 type system used
"may-not".

## Encoding

There are two places we can encode region information, embedded in the types vs
as a side structure (eg. a set of topological facts in Γ, or
MayAliasRegionTree). Embedding in types has the benefit of supporting generics
and disjunction more naturally. A side structure leads to a (possibly) simpler
format (eg. finite domain), and allows proofs by saturation.

In the Winter 2019 type system, everything was represented inside types.  In
POPL20, everything was in side structures (so called topological facts). In
Matt's system, "must" information is in types, "may" information is in
side-structures (ie. MayAliasRegionTree).

On the surface, this distinction feels arbitrary, but I find it very appealing,
and I can rationalize the divide.

Essentially, "must" information is public and symbolic. It gets mentioned in type
signatures ("public"), which means while the function that produces this
information may not be able to use the information, a caller, with more precise
type information, might.  Within the following function, we do not know the
precise relationship between a and b.  However, the caller may know that `X =
mut`, in which case it should be able to recover the fact that the argument and
the return type are in the same region.  This requires us to keep the
information "symbolic" wrt. type variables.

```
read[X](a: HasField[X] & mut): a->X {
  b = a.f
  return b
}
```

On the other hand, "may" is, for the most part, local to a function. If we
don't have enough information because of a generic, we have to take the
conservative approach and assume the worst and invalidate more things. "may"
information is *probably* not useful on method signatures, and therefore
doesn't need to be symbolic (more on that later though).

This distinction is actually present in the POPL20 system. Positive facts (ie.
must) are parametrized over types, whereas negative facts (ie. may not) are
not.

## Introducing "may"s

There are two options when it comes to introducing "may"/"may-not" information.

The first, which I'll call "derived may", is to have typing rules introduce
"must" information only, and have subtyping/fact deduction/... rules to derive
"may" from "must". For example, "x may be in a strict sub-region of y, unless y
is in the same or sub-region of x".

The second, which I'll call "explicit may", is to have typing rules directly
state what forms of aliasing they could introduce. For example, "x = (y.f = z)
may cause x and y to alias (for some definition of alias)".

I think the "derived may" approach is easier to make sound, especially taken
with the "may-not" perspective. If we forget cases, the system will simply be
less precise. In the "explicit may" approach, if we forget cases, we will be
possible aliasings and miss invalidations. On the other hand, the "explicit
may" approach is more, *well*, explicit. It does not require a parallel
inference system.

"derived may" has problems with proving whether that an argument must not
alias a locally defined variable:
```
m(x: mut) {
  y: iso = new
  z: mut = view y

  // How do you reach the conclusion that:
  // z may alias y
  // x may not alias y
}
```

With "explicit may", we need a way to express these on method signatures. It's
not clear 1) how expressive do we want this to be 2) do we need a whole surface
syntax for it 3) is it derived from the types, in a restrictive version of
"derived may" 4) is it instead derived from some "where" sugar

```
m1(x: mut): x->mut
m1(x: mut): mut
  where return in x
m1(x: mut): mut
  where return may-alias x
```

```
m2(x: mut, y: mut, z: mut): (x->mut | y->mut)

m2(x: mut, y: mut, z: mut): mut
  where return in x, y

m2(x: mut): mut
  where return may-alias x,
        return may-alias y
```

Both Winter 2019 and POPL20 use a "derived may". Matt's type system uses
"explicit may".

# Various thoughts

- "in a sub-region" is generally thought of as a "must" information. However,
  it's only purpose is to derive "may". In a full blown "explicit may" model,
  do we even need this anymore?

- Granularity of "may": Matt's system uses a single transitive, symmetric,
  reflexive relation for "may". POPL20 uses 3 (reduced to 2 without tidy
  shenanigans) "may-not" relations, which are non-transitive, non-symmetric and
  one which is non-reflexive. The negated perspective might make this comparison
  a bit too apples/oranges. Finer grained information presumably gives us more
  precision.

- Is either "may" or "may-not" better? I find "may-not" simpler to reason about
  in terms of soundness, due to the constructive nature, but I'm still
  unconvinced about the overall comparison.

docs/alternativesemantics.md
```diff
@@ -0,0 +1,387 @@
...
+   // Captures something that is still live due to sub-reference
+}
+when (b)
+{
+  // Does not capture anything.
+}
+```
+It is sound to dispatch the second before the first, if `a` and `b` are different cowns, but otherwise it is not.
+
+The runtime if any message has been delayed, and if so, all subsequent messages in the behaviour are delayed until the end of the behaviour.  This is the simplist solution, and would probably work well in practice.
```
rengolin, line 285: Could you have a "depends on" pointer from one message to another if it "may depend on" and it's the compiler/runtime's job to optimise that away by checking alias etc?

mjp41: That would be an approach but would probably be too expensive to maintain.

PR #341, opened 2020-11-05
Initial stab at documenting the internal ABI.

### LINKED ISSUE #272: Type representation, sizeof, structure layout
The MLIR representation is taking shape, but we's still using mostly opaque instructions and types, because the Verona types are not complete yet (in MLIR). However, to start lowering classes and members, we need to know how to represent them at a lower level than just an opaque MLIR dialect.

Right now, we don't seem to have any particular type defined. But in the [grammar](https://github.com/microsoft/verona/blob/master/src/ast/grammar.peg), we have some constants:
* Int (which will be either `U` or `S`, with bit-size from 8 to 128)
* Float (with bit-size from 8 to 128?)
* Octal/Hex, which I'm assuming will default to Int (but could also be float or pointer type)
* String (multi-byte characters? which? all?)

We'll also need to represent pointers/references, which will have additional capability/range tags in it, so I'm assuming it will be at least 128-bits (like CHERI).

There are also tuples and structures, which will bring ABI issues into light. We need to decide how we pack them, what is the optimal padding between elements and derived class storage, vtable layout and offset, etc. If we want to control packing, we'll need a special keyword for that, otherwise we must guarantee correct behaviour at all times. For example, if we don't pack anything, it would be interesting to have tuples and structs with the same layout, so we can instantiate one from the other with a simple `memcpy`.

Finally, there's type unions and intersections. Depending on what they mean if done with two different concrete types (like `int and float`, `int or float`), we'll need to create a storage for the largest value and pad the smallest so reads and writes are aligned, etc. Capabilities would probably not have a concrete representation in memory, other than pointers (that already count on that space being used), so this may not be an issue for numeric type representation.

It would be nice to write down the set of concrete decisions on type sizes, padding, alignment for all types we support in Verona, so that issues on the MLIR / LLVM layers can take that into account when implement lowering of load/store patterns.

@davidchisnall @sylvanc @mjp41 @Theodus @plietar

davidchisnall:
Fixes #272


docs/internal_abi.md
```diff
@@ -0,0 +1,321 @@
...
+
+### Pointers
+
+We anticipate three common representations for pointers on Verona target systems:
+
+ - 32-bit integers representing an address
+ - 64-bit integers representing an address
+ - CHERI capabilities granting access to an address range.
+
+All Verona objects are at least 8-byte aligned and so the low three bits of a pointer will always be zero.
```
rengolin, line 63: What do you mean by objects here? Instances of concrete classes?

Does it mean that you can't have pointers to native types (which are naturally aligned)?

docs/internal_abi.md
```diff
@@ -0,0 +1,321 @@
...
+Pointers to concrete types are simply pointers.
+Pointers to interface types are a pair of a pointer to the object followed by a pointer to the vtable for the interface.
+There are no additional requirements on alignment for interfaces: the pair of pointers is aligned on the same boundary as a single pointer.
+
+Union types
+-----------
+
+Variables or fields of union types are *discriminated unions*.
+The simplest representation would be to store a buffer of the largest type and an integer large enough to store a unique value for each type.
+For example, a `U8 | Float64 | Foo*` could be represented as a 64-bit buffer followed by a 2-bit integer.
```
rengolin, line 91: Suggestion: ...64-bit buffer [for the values] ... 2-bit integer [for the discriminator].

docs/internal_abi.md
```diff
@@ -0,0 +1,321 @@
...
+At an abstract level, the Verona internal ABI does exactly that, with two observations that allow denser packing:
+
+ - Singleton types (equivalent to Pony primitives [*note:* do we have a name for these yet]) have a single instance and so the body can be elided in some cases and the value stored entirely in the tag discriminator.
```
rengolin, line 94: "Pony primitives" hasn't been defined here, perhaps a description of "Singleton Types" on a section above this one?

If my interpretation of [this document](https://www.ponylang.io/media/cheatsheet/pony-cheat-sheet.pdf) is correct, a "primitive type" is like a class with only members, which isn't necessarily what I would call "singleton".

docs/internal_abi.md
```diff
@@ -0,0 +1,321 @@
...
+ - Pointers to object types (including external reference types) require as many bits as are required for an unambiguous pointer; however, all pointer types require only a single discriminator state because all pointees can be discriminated by inspecting their descriptor.
+   The number of bits required for a pointer varies depending on the pointer representation:
+   - 32-bit integer pointers require 29 bits of state.
+   - 64-bit integer pointers require either 45 or 52 bits, depending on the virtual address space size of the target.
+   - CHERI capabilities require 125 bits of state but provide a 'free' discriminator bit because the tag bit can be used to discriminate CHERI capabilities from non-pointer data.
+ - Value types are still being defined; however, it seems plausible that they will each require one descriptor state and as many payload bits as it has non-padding bits.
+   Making efficient use of padding bits as discriminators seems very difficult and so is likely to be left to a later ABI, if ever implemented.
+
+Once the payload sizes are all known, the algorithm for implementing the packing depends on the type of the largest payload.
```
rengolin, line 129: I worry about the performance differences of saying `U16 | F32` versus `U32 | F16`. It's not entirely clear to the programmer model that they will be encoded completely differently, which could have impact on memory and run time.

rengolin (commented):
Some initial comments.

docs/internal_abi.md
```diff
@@ -0,0 +1,321 @@
...
+### Booleans
+
+Boolean values are stored as bytes, with the bit pattern 0 or 1 in the platform's native endian.
+The compiler is responsible for ensuring that no bits other than the least-significant bit in a boolean are ever non-zero.
+
+### Integers
+
+Power-of-two-sized integers in Verona can be used for arithmetic and so are expected to use the representation that the target architecture uses for integer operations.
+The types `U8`, `U16`, `U32`, `U64`, and `U128 are all unsigned fixed-width integers with the platform's native endian.
+The types `S8`, `S16`, `S32`, `S64`, and `S128 are all signed fixed-width integers represented as twos complement numbers with the platform's native endian.
```
rengolin, line 39:
```suggestion
The types `U8`, `U16`, `U32`, `U64`, and `U128` are all unsigned fixed-width integers with the platform's native endian.
The types `S8`, `S16`, `S32`, `S64`, and `S128` are all signed fixed-width integers represented as twos complement numbers with the platform's native endian.
```

docs/internal_abi.md
```diff
@@ -0,0 +1,321 @@
...
+These types are always stored naturally aligned.
+For example, a `U64` is always stored on an 8-byte boundary.
+
+Other integer types, for example `U56` are *storage-only* types.
+These must be extended to a power-of-two sized type for arithmetic.
+When stored on the stack, these should also be sign- or zero-extended (for signed and unsigned types, respectively) to the next native integer type.
```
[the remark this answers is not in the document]
davidchisnall: Sorry, this is unclear.  The idea is that storage-only types are useful in fields, but on the stack you probably don't want actually save anything and the compiler is free to just store a `U52` as a `U64`.

docs/internal_abi.md
davidchisnall: I believe our current plan is that all pointers are pointers to objects (or sufficiently object-like things that they do have a Verona object header).

docs/internal_abi.md
davidchisnall: Yes, it's not quite the same.  The term 'primitive' is confusing in Pony.  They're used as tag values (also overloaded and tag means something else in Pony) where the type is used as a value, rather than as a meaningful holder of state.

docs/internal_abi.md
davidchisnall: Hopefully if people need to care, they are able to read the docs on how the encoding works.  Lazy people can assume that the encoding of both is the same as:
```c
struct
{
  union
  {
      float f;
      short i;
  };
  int discriminator:1;
};
```
Or
```c
struct
{
  union
  {
      fp16 f;
      int i;
  };
  int discriminator:1;
};
```
Anything else is an optimisation.

docs/internal_abi.md
rengolin: I honestly can't think of a good name. "Concrete interfaces", "Behaviour class" would all overload existing terms and be even more confusing. I think we can just say "classes without data" or "POBOFP" (plain old bag of function pointers:).

docs/internal_abi.md
rengolin: Actually, it wouldn't be bad to have at least one of those examples in the document.

What I was worried about is that we do such a good job at the optimisation side for some and not others, the "others" will look like we did a bad job. Like Java's Integer caches.

docs/internal_abi.md
```diff
@@ -0,0 +1,321 @@
...
+Power-of-two-sized integers in Verona can be used for arithmetic and so are expected to use the representation that the target architecture uses for integer operations.
+The types `U8`, `U16`, `U32`, `U64`, and `U128` are all unsigned fixed-width integers with the platform's native endian.
+The types `S8`, `S16`, `S32`, `S64`, and `S128` are all signed fixed-width integers represented as twos complement numbers with the platform's native endian.
+These types are always stored naturally aligned.
+For example, a `U64` is always stored on an 8-byte boundary.
+
+Other integer types, for example `U56` are *storage-only* types.
+These must be extended to a power-of-two sized type for arithmetic.
+When stored on the stack, these should also be sign- or zero-extended (for signed and unsigned types, respectively) to the next native integer type.
+The value of these types is primarily for use in structures or union types, where special layout rules apply.
```
rengolin, line 46:
```suggestion
The value of these types is primarily for use in structures or union types, where special layout rules apply. Their storage semantics are explained below.
When stored on the stack, however, these should also be sign- or zero-extended (for signed and unsigned types, respectively) to the next native integer type.
```

docs/internal_abi.md
```diff
@@ -0,0 +1,321 @@
...
+ - The architecture can load and store a single byte without read-modify-write operations
+ - The order of bits within different-sized values is consistent
+ - Signed arithmetic on integers uses twos complement
+
+This is a slightly tighter set of requirements than the C specification but is looser than the de-facto C standard as used in large deployed codebases.
+
+### Booleans
+
+Boolean values are stored as bytes, with the bit pattern 0 or 1 in the platform's native endian.
+The compiler is responsible for ensuring that no bits other than the least-significant bit in a boolean are ever non-zero.
```
rengolin, line 33:
```suggestion
The compiler is responsible for ensuring that no bits other than the least-significant bit in a boolean are ever non-zero.

The alignment of booleans is 1 byte.
```

docs/internal_abi.md
```diff
@@ -0,0 +1,323 @@
...
+### Example: `Float32 | U20 | U16`
+
+The size of `Float32` is 32 bits and requires one discriminator state.
+The size of `U20` is 20 bits and requires one discriminator state.
+The size of `U16` is 16 bits and requires one discriminator state.
+
+The first special case applies: the largest type is a floating-point type and the remaining types require 20 bits of payload and one of discriminator, which fits in the significand for a `Float32` signalling NaN.
+
+Placing the discriminator in the low bits would give one byte and two partial bytes for the payload.
+Placing the discriminator in the high bits would give two complete bytes and one nybble for the payload.
```
rengolin, line 205:
```suggestion
Placing the discriminator in the high bits would give two complete bytes and one nibble for the payload.
```

docs/internal_abi.md
```diff
@@ -0,0 +1,323 @@
...
+If the comparison result is equal, then a value of bit 20 is tested, with zero indicating a `U16` and one indicating a `U20`, either of which can be extracted by a bit shift (note: On Arm systems, either value can be extracted with a single `UBFX` instruction, on x86 systems that support the BMI1 instructions a BEXTR can do the same).
+
+### Example: `Float64 | T* | U*`
+
+The payload size of `Float64` is 64 bits and it requires one discriminator state.
+The payload size of `T*` and `U*` is each either 32, 48, or 128 bits, depending on the target architecture and they require one discriminator state between them.
+The lowering for this union therefore depends on the target.
+
+On 32-bit systems, the largest type is `Float64` and the remaining types all fit easily in the significand bits.
+There are no discriminator bits required but the 'as-if' rules triggers as if there were a single bit for determining which end of the space should be used and the remaining values are stored at the end of the space that is strongly aligned: bits 0-31.
```
rengolin, line 222: Shouldn't there be a discriminator for the difference between `U*` and `T*`?

docs/internal_abi.md
```diff
@@ -0,0 +1,323 @@
...
+The discriminator values are assigned starting from the singletons, so zero is used for `None` and one for `T*`.
+
+A pattern match expression on this type is compiled into a compare-and-branch-on-{non}-zero, which is a single instruction on most architectures.
+Accessing the `T*` on a CHERI system requires no further operations.
+On integer-pointer systems, it requires subtracting one.
+On any mainstream architecture other than RISC-V, this subtraction can be folded into the immediate offset on a load instruction and so requires no additional instructions.
+
+If the `None` value is used (rather than just being pattern matched on and ignored) then the compiler must be able to materialise the address (or capability) to the singleton.
+This is no worse than if we stored the address inline; however, because in that case the match would require a comparison against the address and so the address would need to be loaded before the pattern match, rather than conditionally in one path.
```
rengolin, line 243: This seems orthogonal to the type encoding, as you'd have to do that anyway, even if the descriptor was a whole new byte.

rengolin (commented):
A few more comments on the examples.

They make sense to me, but I worry that discerning between all these cases and having special rules for each particular case may make it harder to pass union types along.

For example, `a : A* | B*` has descriptors to differentiate `A` and `B` on some bits, and `c : C* | D*` happens to use the same bits, now you write a function that takes in `a | c` and the descriptor bits are now in conflict.

If you can solve that statically, you can make sure A, B, C, D never use the same bits, as they sometimes overlap, but now you're expanding your bit-space of all cases just because in one case they merge (which may not even be hot-code).

So, I can see a complex search of all possibilities and a common pathological case where one new thing breaks the performance of everything else, which could possibly be entirely unrelated code, and hard for programmers to understand why it's so much slower or even how to fix.

This is obviously all hand-wavy, I may not have understood completely all the edge cases.

Next, I'll read the objects section...

docs/internal_abi.md
```diff
@@ -0,0 +1,323 @@
...
+External references are structures with the same header as a Verona object followed by two pointers, one to the table that stores the external reference and one to the object.
+The rules for operations on this are defined in [`externalreference.h`](../src/rt/region/externalreference.h).
+External references are never embedded in other structures, they are always referenced by pointer and so fields or arrays of external references follow the same rules as any other kind of object pointer.
+
+Arrays
+------
+
+Arrays are objects (with the standard header) followed by a run of values of the same type.
+
+*Open question*: Should `Array[Bool]` be equivalent to `Array[U1]` or `Array[U8]` in storage?
```
rengolin, line 321: Similar to `std::vector<bool>`, I think this could be an optimisation. First we implement as `U8`, then we optimise if necessary.

docs/internal_abi.md
```diff
@@ -0,0 +1,323 @@
...
+
+*Open question*: Should descriptors for array elements be collected in the same way as for structures?
```
rengolin, line 323: This means an unbound number of descriptors, which can only be packed at runtime. Also, if you add them to the beginning (or the end) of arrays, you end up having to read two cache pages for elements far from the table.

You could pack them at intervals, for example, as they accumulate into the size of the element itself, so you have a simple algorithm without breaking the alignment and caching locality, but making random access slightly harder. Though, if the interval is regular, it could be simple arithmetic.

Something like: `obj | obj | obj | obj | desc | obj | ...`, both `obj` and `desc` with the same alighment.

rengolin (commented):
This is my final review at this stage. Same as the string document, it's probably as far as I can go with such a high level description. 

I imagine there are a number of edge cases on the descriptor magic that we're not considering, and I'd prefer if we took a more conservative approach for now and only optimise when we make sure (at compile time?) it cannot break at runtime. 

It'll be easier to relax later than retroactively fix a broken runtime.

A general comment about struct layout and all the clever bits: where do we store that information?

When the compiler is generating accessors, we can encode straight into the IR the mechanism for doing so, but we'll need to remember which structure has which optimised (or not) layout, and access them in the same order / style.

This could be a compiler map, but it may be fragile to do partial lowering if different parts of the compiler have passed in some areas and not others.

Perhaps the easiest thing to do is to create static accessors for each class, even if obvious, and then have all accesses to call them. Once we're happy with it, perhaps even after lowering to LLVM IR, we can just get them inlined by marking them `always_inline`.

davidchisnall:
> A few more comments on the examples.
> 
> They make sense to me, but I worry that discerning between all these cases and having special rules for each particular case may make it harder to pass union types along.

It shouldn't.  Once the layout is done, they're no different from any other type.
 
> For example, `a : A* | B*` has descriptors to differentiate `A` and `B` on some bits, and `c : C* | D*` happens to use the same bits, now you write a function that takes in `a | c` and the descriptor bits are now in conflict.

`A* | B*` is a plain pointer because you can use the descriptor in the pointee to work out which, but in the general case you shouldn't assume that `A | B` (where `A` and `B` are type expressions) has the same representation as `A` or `B`.  Passing an `A | B` to an argument of type `(A | B) | (C | D)` is no different from passing a `U8` to an argument of type `U8 | U16`: in both cases, you need to construct the new union type representation.

This is a good motivating example for wanting the lowering to the concrete representation to happen after reification though: If we reify the callee then we can eliminate all of the match expressions that look for `C` or `D` in the version that is always called with an `A | B`.

> 
> If you can solve that statically, you can make sure A, B, C, D never use the same bits, as they sometimes overlap, but now you're expanding your bit-space of all cases just because in one case they merge (which may not even be hot-code).

This is effectively what selector colouring does, but it's not likely to be feasible for union types.
 
> So, I can see a complex search of all possibilities and a common pathological case where one new thing breaks the performance of everything else, which could possibly be entirely unrelated code, and hard for programmers to understand why it's so much slower or even how to fix.

I don't think these problems appear if you don't guarantee that union types have the same representation as union types that define a superset of types.  You definitely can't do that in the degenerate case (`U64 | T`, for any `T`, will have a different representation to `U64`, for example), so I don't think it's worth trying to guarantee it anywhere else.

davidchisnall:
> I imagine there are a number of edge cases on the descriptor magic that we're not considering, and I'd prefer if we took a more conservative approach for now and only optimise when we make sure (at compile time?) it cannot break at runtime.
> 
> It'll be easier to relax later than retroactively fix a broken runtime.

I think it's important that we get this mostly right quite early because telling systems programmers 'you don't need to write this fragile and error-prone logic because your compiler does the right thing for you' is an easier sell than 'you can't write this, you have to use much less space-efficient structures'.  The technical debt that we'll accrue from hacks to work around not being able to rely on the compiler doing this is likely to pile up quickly.

> A general comment about struct layout and all the clever bits: where do we store that information?

In the compiler?  Wherever is most convenient.  Note that, because these rules are deterministic, we don't *need* to store it for correctness, we can always recompute it if necessary.

> When the compiler is generating accessors, we can encode straight into the IR the mechanism for doing so, but we'll need to remember which structure has which optimised (or not) layout, and access them in the same order / style.
> 
> This could be a compiler map, but it may be fragile to do partial lowering if different parts of the compiler have passed in some areas and not others.

The result of the layout will change only if the types change.  We shouldn't be doing any of this until we're very late in the MLIR pipeline, after reification, when types are as known as they will ever be.

> Perhaps the easiest thing to do is to create static accessors for each class, even if obvious, and then have all accesses to call them. Once we're happy with it, perhaps even after lowering to LLVM IR, we can just get them inlined by marking them `always_inline`.

In the higher-level MLIR, we'll want to do explicit Verona-specific field access operations.  Whether we lower these to calls to field accessor functions that are `always_inline` or whether we just insert the body of such a method doesn't matter too much to me: the generated code should be equivalent.

docs/internal_abi.md
davidchisnall: Wikipedia says both spellings are allowed, I prefer nybble because it's less ambiguous.

docs/internal_abi.md
davidchisnall: No, you can always do the equivalent of `((Verona::Object*)ptr)->descriptor` and compare that against the descriptor pointers for the matched types for any concrete type.  @sylvanc, I'm not 100% sure what you do when you have something like `Cls1 | Cls2 | Interface1` and want to match on `Interface1`?

docs/internal_abi.md
davidchisnall: I don't think I understand this comment.  The alternative representation is to use the address of a `None` singleton global as the descriptor and just match on the descriptor field in the pointee object, rather than special casing for 'is the pointer 0?'

docs/internal_abi.md
```diff
@@ -0,0 +1,323 @@
...
+-------
+
+Verona objects contain a header with metadata describing the object, followed by user-defined fields.
+
+### Object headers
+
+Object headers are two pointer-size words.
+These are described in detail in [src/rt/object/object.h](../src/rt/object/object.h).
+
+The first word contains the region metadata.
```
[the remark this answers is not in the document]
davidchisnall: The thing it points to is 8-byte aligned even on 32-bit systems so that we can still steal the three low bits.  It should be possible to use a pointer-sized value here, but I'm not sure.

docs/internal_abi.md
```diff
@@ -0,0 +1,323 @@
...
+Methods in the vtable use *selector colouring*.
+Each selector (method name and types) is assigned a separate vtable index.
+Pairs of selectors that do not appear in the same interface are then combined to use the same index.
+Selector colouring is not essential but can dramatically reduce the size of vtables.
+
+*Note:* If we can distinguish between sentinel objects and others at object-creation time, the region metadata field for non-sentinel objects could be a 32-bit value that is either an integer or an index into a table, rather than a pointer.
+This would be a noticeable saving on CHERI systems and a small saving on 64-bit systems, at the cost of an extra indirection for some memory-management operations.
+We could also use a 32-bit index into a descriptor table, as the Fuchsia C++ ABI does: Programs that have more than 2^32 distinct types arise only as a result of poor life choices.
+
+*Note:* The above choices are visible to the runtime but are part of the internal ABI and so it would be possible to have size and performance optimised variants available on a per-program basis.
```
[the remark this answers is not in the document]
davidchisnall: In theory, yes, though I don't know if we can automatically differentiate between objects that are touched only in cold code paths and ones that aren't.  We might be able to do this on a per-type basis with explicit programmer opt-in, though we might end up needing to steal one bit in the header to differentiate between the two.

docs/internal_abi.md
```diff
@@ -0,0 +1,323 @@
...
+
+*Note:* On AArch64 systems where the top-byte-ignore feature is available, using the top bits in the descriptor pointer instead of the bottom ones would improve efficiency of descriptor accesses.
+
+### Field layout
+
+In C, `struct` layout is a trade between storage density and accidental aliasing.
+If two fields are accessed concurrently, false sharing in the cache can cause significant performance bottlenecks.
+In Verona; however, no mutable objects are ever shared and so, aside from reference count manipulations on immutable objects and interactions between adjacently allocated objects in different regions, there are no negative interactions with modern caches.
+The Verona field layout algorithm is therefore free to optimise for space.
```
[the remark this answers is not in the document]
davidchisnall: Dynamically loadable code is an entirely separate discussion and is explicitly not planned for MVP, but I don't think this is actually a concern.  As long as the layout algorithm is the same in both cases, it will give the results.  This is not an optional compiler optimisation, this is the goal of the design of the layout algorithm.

docs/internal_abi.md
```diff
@@ -0,0 +1,323 @@
...
+
+All fields whose size is a power-of-two should be sorted by size and laid out largest to smallest.
+The end of each section is more strongly aligned than the next requires and so this does not require any padding.
+
+Next, any union types that have a power-of-two sized payload but a smaller discriminator should be split into payload and discriminator.
+The payloads should be inserted into the sections for the same sized power-of-two objects.
+The discriminators should be packed into a single larger value.
```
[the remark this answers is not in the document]
davidchisnall: Taking those in opposite order:

For (2), anything with descriptors embedded in the payload, the result should be a power-of-two type and so the first rules apply.

In think (1) is a much more interesting question and one that I hadn't fully considered.  It probably does make sense to float them backwards because then we can always do an aligned load + ubfx to get any field's descriptor.

docs/internal_abi.md
```diff
@@ -0,0 +1,323 @@
...
+
+*Rationale:* Extracting from a bitfield is no more expensive if the bitfield is densely packed and if a function is pattern matching on multiple fields then they may be able to reuse a register for the descriptor values.
+
+The remaining values should be stored in the order in which they are declared, between the power-of-two region and the descriptors.
+
+*Rationale:* We could do some clever bin packing here but systems programmers who are using odd-sized storage-only types are likely to be careful about arranging them usefully already.
+A future version can improve bin packing and not break anything.
```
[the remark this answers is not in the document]
davidchisnall: I think these rules implicitly align them to a power-of-two boundary, I'm not sure what we'd get from stronger alignment.  Eventually, I'd like to move to a better bin-packing algorithm but the proposal here is simple to implement and any code that works well with this approach will work at least as well with a better algorithm.

docs/internal_abi.md
```diff
@@ -0,0 +1,323 @@
...
+
+
+*Note:* Java implementations typically pack all pointers together such that the trace logic just needs to look at a range, rather than have data or code describing the complete layout of the object.
+This may be possible for Verona, depending on the constraints for value types as fields and on pointers to value types.
+If value types cannot be passed by reference then they do not need to be stored contiguously in fields, because they will always be either copied when passed down the stack or accessed from functions that are aware of the location of their components.
```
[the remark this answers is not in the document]
davidchisnall: Value types are still largely undefined, but I was imagining that if we had a value type that contained, for example, a `Foo*` and a `U64`, we could store the `Foo*` in the block of pointer fields in the enclosing object and the `U64` separately.  This means that the descriptor would just need to encode the number of pointer fields (and possibly the offset of the first pointer field, if we have things that need stronger-than-pointer alignment, such as 256-bit SIMD vectors) and let the tracing code scan objects without needing a custom `scan` function in the descriptor.  This avoids a function call per object when tracing a traced region.

docs/internal_abi.md
davidchisnall: Agreed for the initial version, it remains an open question longer term.

docs/internal_abi.md
davidchisnall: I'm assuming something along the lines you describe.  It may not be something that actually matters, but something like `Array[U64 | T*]` is annoying because you effectively have an array of 65-bit values that require 64-bit alignment.  Your choices in representation are either 16 bytes per element or trying to collect the descriptor bits.  Having one 64-bit word of descriptors per 64 array elements isn't too bad from the perspective of iteration (you'll just grab it once and then if you're matching in the loop body then you'll shift and branch on the low one) but it sucks for random access because now you need to load two cache lines to get a single value.

It might be better to give programmers predictable bad behaviour than unpredictable slightly better (for some value of better) behaviour.

docs/internal_abi.md
```diff
@@ -0,0 +1,323 @@
...
+
+Union types
+-----------
+
+Variables or fields of union types are *discriminated unions*.
+The simplest representation would be to store a buffer of the largest type and an integer large enough to store a unique value for each type.
+For example, a `U8 | Float64 | Foo*` could be represented as a 64-bit buffer for the values followed by a 2-bit integer for the discriminator.
+At an abstract level, the Verona internal ABI does exactly that, with two observations that allow denser packing:
+
+ - Singleton types (equivalent to Pony primitives [*note:* do we have a name for these yet]) have a single instance and so the body can be elided in some cases and the value stored entirely in the tag discriminator.
```
SeanTAllen, line 96: "do we have a number for these yet"

is that a question missing a `?` or is there some other typo/missing word?

docs/internal_abi.md
SeanTAllen: @rengolin if I understand what you mean by "members" then no "primitive type"s in Pony aren't like a class with only members.

primitive is a singleton. There are no fields. Only methods.  I think you misread the cheatsheet section.

docs/internal_abi.md
davidchisnall: Yes, this was meant to be a question for reviewers.  'Do we have a name for the thing that Pony calls primitives, ideally one that is less likely to confuse people?'

rengolin:
> in both cases, you need to construct the new union type representation.

Which means shifts and masks for every argument and potentially different logic for different return values.

```
foo() : A | B;
bar() : B | C;
baz() : C | A;
function(a) : A | B | C
{
  switch (a) {
    case 1: return foo();
    case 2: return bar();
    default: return baz();
  }
}
```

The compiler will have to generate generic conversion routines for every (potential) combination (dynamic matches left over after reification) and add to it everywhere the types are not guaranteed to be the same.

```
...
    case 2: return __B_or_C_2_A_or_B_or_C(bar());
...
```

> This is effectively what selector colouring does, but it's not likely to be feasible for union types.

Agree. Though the document wasn't explicit and it wasn't clear to me that wasn't what you were suggesting. Perhaps making explicit that we'll have to write conversion routines for all combinations would help that.

> I think it's important that we get this mostly right quite early because telling systems programmers 'you don't need to write this fragile and error-prone logic because your compiler does the right thing for you' is an easier sell than 'you can't write this, you have to use much less space-efficient structures'. The technical debt that we'll accrue from hacks to work around not being able to rely on the compiler doing this is likely to pile up quickly.

Good point.


> In the compiler? Wherever is most convenient. Note that, because these rules are deterministic, we don't _need_ to store it for correctness, we can always recompute it if necessary.

I think that, if we handle as always_inline functions with explicit names, we can cache that in the function table. Given they only depend on the types involved, it should be guaranteed that we can't generate a slightly different version with the same types.


> The result of the layout will change only if the types change. We shouldn't be doing any of this until we're very late in the MLIR pipeline, after reification, when types are as known as they will ever be.

Depending on how we create the dialect, we can do this as a replacement of `verona.cast`s (to get the type checks passing) or it could be the lowering of `verona.cast` instructions straight into LLVM IR.

> In the higher-level MLIR, we'll want to do explicit Verona-specific field access operations. Whether we lower these to calls to field accessor functions that are `always_inline` or whether we just insert the body of such a method doesn't matter too much to me: the generated code should be equivalent.

Precisely.

docs/internal_abi.md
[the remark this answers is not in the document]
rengolin: Right, I meant an older version of the compiler created `libfoo.so` and a newer version (with improved packing) created the executable `foo.exe`. But I guess this is just like any other ABI breaking change, people will have to recompile everything.

Means we'll have to track any change to this algorithm very closely to make sure there are no ABI-breaking scenarios.

We'll also need some ABI versioning in the objects to warn users that include older libraries into their code, but that's a separate discussion.

docs/internal_abi.md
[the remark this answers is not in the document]
rengolin: Ah, yes! Otherwise we'd create two implementations of the same conversion and that would be against the core design goal of being able to recompute, cache, etc.

The hot/cold part could be done using profile information or user pragmas, but also, this is subject for another discussion.

docs/internal_abi.md
rengolin: I never seen `nybble` before... Just thought it was a typo, doesn't matter much.

docs/internal_abi.md
rengolin: I can't remember what I meant by this and I also don't understand my own question... :S ignore me.

docs/internal_abi.md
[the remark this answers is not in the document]
rengolin: Perhaps also make it explicit in this document that this encoding is a core part of the ABI, and that not only it's deterministic on the types only, but it has strong ABI guarantees and that we'll change the ABI version if any of that changes.

docs/internal_abi.md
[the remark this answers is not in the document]
rengolin: Indeed, we are reserving the right to manipulate the object layout, which could be non-contiguous. Makes sense.

docs/internal_abi.md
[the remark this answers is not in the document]
davidchisnall: I think you are working under the assumption of a C linkage model.  This is something that we have explicitly discounted for Verona.  Our loadable code model is not going to be supported at all in v1 and will not look like a UNIX .so when it is implemented.

docs/internal_abi.md
```diff
@@ -0,0 +1,323 @@
...
+There are two special cases:
+
+If the largest type is a floating-point type and the payload of all other types plus the bits required for the discriminator will fit in the unused bits of the significand of a signalling NaN minus one then the top bit of the significand is reserved to discriminate between a real signalling NaN and another value and the remainder of the types are encoded in the spare bits, starting this algorithm again.
+If the remaining bits of the significand are not sufficient to encode the remaining states then this rule is ignored.
+
+If the largest type is a CHERI capability then the tag bit is used to differentiate pointer and non-pointer types.
+This leaves 128 bits to use for the remainder of the values.
+
+The size of the final object is the number of payload bits, plus the base-two logarithm of the number of discriminator states rounded up.
+The alignment of the result is the alignment of the most strongly aligned type.
```
nwf, line 141: This paragraph seems premature given the next, or perhaps the next is intended as an example?

docs/internal_abi.md
```diff
@@ -0,0 +1,323 @@
...
+
+If one of the types is a pointer, the object size is pointer sized, and the discriminator is three or fewer bits, then the low bits of the pointer are used for the discriminator.
```
nwf, line 143: What about unions of pointers with things larger than pointers?

docs/internal_abi.md
```diff
@@ -0,0 +1,323 @@
...
+We first try the special-case rule for the case that the largest value is a floating-point value.
+The remaining type (`U64`) requires 64 bits for payload and 0 bits for discriminator, and so does not fit into the significand part of a `Float64` signalling NaN.
+The first special case is therefore not used
+The second special case is not used because neither type is a pointer.
+
+We are now left requiring one bit of discriminator and 64 bits of payload.
+The payload size is a power of two, and so the address of the union type is the address of the start of a 64-bit payload.
+The 1-bit discriminator is stored afterwards, with a value of 0 representing a `Float64` and a value of `1` representing a `U64`.
+The field-packing algorithm is free to use the remaining 7 bits in the discriminator for other values.
+
```
nwf, line 195: Possibly worth adding, just for completeness, that, in this example, the alignment requirement is 8 bytes and so there are 7 bytes (and 7 bits) of padding.

docs/internal_abi.md
```diff
@@ -0,0 +1,323 @@
...
+ - For singleton types, there are two possible answers:
+   - Each type requires one state in the discriminator but no bits to encode the value.
+   - Each type requires one pointer's worth of state to encode the value but no bits in the discriminator.
+   We choose the former, to maximise our ability to use the CPU's branch-on-zero instructions.
+ - Each integer requires one state in the discriminator and as many bits as its width defines for the payload.
+   For example, in the union `U8 | U16`, each requires one state in the discriminator (giving a 1-bit discriminator to represent both state) and 8 and 16 bits of payload, respectively.
+ - Floating point types follow the same rules as integers, with special handling if a floating point type is the type with the largest payload in a union type, described below.
+ - Pointers to object types (including external reference types) require as many bits as are required for an unambiguous pointer; however, all pointer types require only a single discriminator state because all pointees can be discriminated by inspecting their descriptor.
```
nwf, line 123: I originally missed the implication of "all pointer types require only a single discriminator state"; perhaps "any number of pointer types in a union requires only a single discriminator state"?

docs/internal_abi.md
```diff
@@ -0,0 +1,323 @@
...
+The `T*` requires one pointer's worth of space for the payload (as above, this varies depending on the target) and one discriminator state.
+The `None` value requires one discriminator state and no payload.
+
+The largest value is a pointer and so on CHERI systems the tag bit is used as the discriminator, on other architectures the low bit of the pointer is used (because the pointer is the largest type and the discriminator fits in a three-bit value).
+The discriminator values are assigned starting from the singletons, so zero is used for `None` and one for `T*`.
+
+A pattern match expression on this type is compiled into a compare-and-branch-on-{non}-zero, which is a single instruction on most architectures.
+Accessing the `T*` on a CHERI system requires no further operations.
+On integer-pointer systems, it requires subtracting one.
+On any mainstream architecture other than RISC-V, this subtraction can be folded into the immediate offset on a load instruction and so requires no additional instructions.
```
nwf, line 240: Why isn't that true on RISC-V?  AFAICT loads and stores have a sign-extended 12 bit immediate offset; is that not sufficient?

docs/internal_abi.md
```diff
@@ -0,0 +1,323 @@
...
+Pointers to interface types
+---------------------------
+
+Pointers to concrete types are simply pointers.
+Pointers to interface types are a pair of a pointer to the object followed by a pointer to the vtable for the interface.
+There are no additional requirements on alignment for interfaces: the pair of pointers is aligned on the same boundary as a single pointer.
+
+Union types
+-----------
+
```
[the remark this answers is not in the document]
rengolin: Types are canonicalised, yes. `A | B` is the same as `B | A` (the document hints at some "default" ordering, not clear yet what). Same for combinations of unions and intersections, there will be a canonical form and all variations that represent the same type will be simplified to their canonical forms.

While unions (specifically) may not be flattened during lowering (IR representation: `!verona.meet<A, meet<B, C>>`) they should be so for binary representation and there can be no difference between any random order of the same types.

docs/internal_abi.md
davidchisnall: Good point.  I think this should state that we do this only when the rest of the union can be stored in the remaining pointer bits.

docs/internal_abi.md
davidchisnall: I don't want to talk about padding here, because padding is a thing that happens in aggregate types and there are more complex rules for that.

docs/internal_abi.md
davidchisnall: They might have fixed this after the v8 team yelled at them, but last time I looked the immediate field in the RISC-V load/store instructions was a multiple of the loaded value size, so you can't do a 3-byte offset on a 16-bit or larger load.  This means you can't fold the clearing of the low bits into the immediate, because it will be base + field offset - {1-7}.

This is what happens when you optimise your ISA based on the output of a prototype compiler on a small corpus of C code.

davidchisnall (commented):
Thanks.  Please can you add ```suggestion blocks for improvements to the wording / explanation and other fixes?

docs/internal_abi.md
```diff
@@ -0,0 +1,323 @@
...
+Any type that does not use the full width of a machine word is amenable to dense packing in discriminated unions.
+This is a sufficiently common pattern in systems code that Verona has chosen to bake it into the language.
+
+The rules for determining the correct encoding for a union type begins by identifying the number of payload bits and the number of states required for the type discriminator:
+
+ - For singleton types, there are two possible answers:
+   - Each type requires one state in the discriminator but no bits to encode the value.
+   - Each type requires one pointer's worth of state to encode the value but no bits in the discriminator.
+   We choose the former, to maximise our ability to use the CPU's branch-on-zero instructions.
+ - Each integer requires one state in the discriminator and as many bits as its width defines for the payload.
```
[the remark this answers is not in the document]
nwf:
```suggestion
 - Each integer requires one state in the discriminator and as many bits as its width defines for the payload.  (Integer types are assumed to range over all values permitted by their bit width, and so, unlike aligned pointers, must be externally discriminated.)
```

docs/internal_abi.md
nwf: It looks like it's been corrected.  Sail since reading the official ISA doc is an exercise in frustration.

https://github.com/rems-project/sail-riscv/blob/ae6cb1de092e9ea727e2318d76e9b88999bbee59/model/riscv_insts_base.sail#L317-L337

and

https://github.com/rems-project/sail-riscv/blob/f9f066e89a5c20f2769fbfb2b6a585af9b3f0561/model/riscv_addr_checks.sail#L49-L52

davidchisnall: Ah, great!  Odd that there's no mention of this change in the history bit of the RISC-V spec, given that it's a breaking change from the original.

docs/internal_abi.md
```diff
@@ -0,0 +1,323 @@
...
+
+### Integers
+
+Power-of-two-sized integers in Verona can be used for arithmetic and so are expected to use the representation that the target architecture uses for integer operations.
+The types `U8`, `U16`, `U32`, `U64`, and `U128` are all unsigned fixed-width integers with the platform's native endian.
+The types `S8`, `S16`, `S32`, `S64`, and `S128` are all signed fixed-width integers represented as twos complement numbers with the platform's native endian.
+These types are always stored naturally aligned.
+For example, a `U64` is always stored on an 8-byte boundary.
+
+Other integer types, for example `U56` are *storage-only* types.
```
rengolin, line 45: I think there's a consensus building that non-power-of-two integers be left out of the ABI, at least for v1.

I'm not sure what kind of document you want to write: A strict one with only agreed things, or a more open one, with ideas for the future.

If the former, this paragraph should probably be removed, for now. If the latter, I'd rewrite this as "There's an open discussion about allowing non-power-of-two integers, in which case the semantic is...".

docs/internal_abi.md
```diff
@@ -0,0 +1,323 @@
...
+These must be extended to a power-of-two sized type for arithmetic.
+The value of these types is primarily for use in structures or union types, where special layout rules apply. Their storage semantics are explained below.
+When stored on the stack, however, these should also be sign- or zero-extended (for signed and unsigned types, respectively) to the next native integer type.
+
+### Floating-point values
+
+The types `Float16`, `Float32`, `Float64` and `Float128` correspond to IEEE 754 binary16, binary32, binary64, and binary128 types, respectively.
+*Note:* A future version of the language may provide IEEE decimal types, binary256, bfloat32, or smaller floating point types but these are not necessary for MVP.
```
rengolin, line 53: What's `MVP`?

docs/internal_abi.md
rengolin: To me, personally, singleton is good enough for this purpose.

docs/internal_abi.md
```diff
@@ -0,0 +1,323 @@
...
+
+This form of encoding is one of the motivations for supporting storage-only types such as `U42`.
+Any type that does not use the full width of a machine word is amenable to dense packing in discriminated unions.
+This is a sufficiently common pattern in systems code that Verona has chosen to bake it into the language.
+
+The rules for determining the correct encoding for a union type begins by identifying the number of payload bits and the number of states required for the type discriminator:
+
+ - For singleton types, there are two possible answers:
+   - Each type requires one state in the discriminator but no bits to encode the value.
+   - Each type requires one pointer's worth of state to encode the value but no bits in the discriminator.
```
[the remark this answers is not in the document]
rengolin: But in the case of "all pointers", the payload is "pointer sized" (ex. 48-bits) so the discriminator's space is still unused. I don't see why we'd want to differentiate that.

docs/internal_abi.md
rengolin: Again, Verona's runtime use of lowbits may mean we won't do this. I think we should just look at the high (48+) bits for now.

docs/internal_abi.md
```diff
@@ -0,0 +1,323 @@
...
+
+If the largest type is a CHERI capability then the tag bit is used to differentiate pointer and non-pointer types.
+This leaves 128 bits to use for the remainder of the values.
+
+The size of the final object is the number of payload bits, plus the base-two logarithm of the number of discriminator states rounded up.
+The alignment of the result is the alignment of the most strongly aligned type.
+
+If one of the types is a pointer, the object size is pointer sized, and the discriminator is three or fewer bits, then the low bits of the pointer are used for the discriminator.
+Pointer values are stored with their bits in the natural positions and must have the low bits subtracted away on use.
+Other values are stored in the most significant bits.
```
rengolin, line 145: I assume you mean the payload's most significant bits, as the actual 12 top bits are all `1`s.

Ex, `U8` in [47, 40], etc?

docs/internal_abi.md
```diff
@@ -0,0 +1,323 @@
...
+
+*Rationale:* Most ISAs make subtracting a 3-bit immediate trivial (often something that can be folded into another instruction).
+Other values that fit in the payload space can be shifted down as required in a single-cycle instruction.
+
+If the payload is an integer number of bytes, the payload is stored first (independent of byte order) so that the payload can benefit from strong alignment, which avoids the payload accidentally spanning a cache line.
+In this case, all elements of the payload are stored starting from the beginning of the payload.
```
rengolin, line 151: I'm not sure what you mean by "stored first". Most or least significant bits? Assuming (for now) all types will be power of twos, this paragraph is unnecessary.

docs/internal_abi.md
```diff
@@ -0,0 +1,323 @@
...
+
+*Rationale:* The payload can be loaded with a single aligned load on any architecture and depending on the target byte order then needs to do either a single zero-extending shift right or a left shift followed by a zero-extending right shift.
+
+If storing the discriminator bits at either end of the space results in different numbers of complete bytes for the payload then the discriminator should stored at the end that results in the larger number of complete bytes for the payload.
+If there are no discriminator bits, this rule is applied as if there were a single bit.
+Payload values that are smaller than the payload space are then stored aligned to the opposite end of the value to the discriminator.
```
rengolin, line 157: I'm not sure what the example is here. This paragraph is confusing. Perhaps adding a simple example for each of these topics with rationales will help.

docs/internal_abi.md
```diff
@@ -0,0 +1,323 @@
...
+
+*Rationale:* This case is reached if we are packing a union type inside the significand bits of a signalling NaN and guarantees that we can load the payload in a single aligned load and need to zero bits at only one end to extract the payload.
+
+In other cases, the discriminator bits are stored in the low bits (as defined by the platform byte order) of the object and the payload in the high bits.
```
rengolin, line 161: Same comment as before, probably high bits.

docs/internal_abi.md
```diff
@@ -0,0 +1,323 @@
...
+
+*Rationale:* The discriminator is easy to mask off with a right shift and extract with an xor with an immediate operand for comparisons.
+
+The types identified in a discriminator are sorted lexically and discriminator values are assigned in order.
+Singletons are sorted before all other types.
+
+*Rationale:* The representation of `A | B` and `B | A` should be the same.
+It is useful to sort singletons before pointers so that the common case of `T* | NotAThing` is able to use null-pointer checks and branches, which are typically optimised aggressively in modern microarchitectures.
+
+{*Note:* I'm not sure lexical sorting will work because types imported from the same package by two other packages will have different names and so we may need some other stable ordering for types.}
```
rengolin, line 171: I believe all types will be normalised by their full scope, so even if you name them differently from different modules, they should, at the end, have the same canonical qualified name.

docs/internal_abi.md
```diff
@@ -0,0 +1,323 @@
...
+
+*Open question*: As described, this requires a pointer dereference do disambiguate pointer types.
+It would be possible to use any space in the low three bits that is not used to disambiguate non-pointer types to differentiate a small number of pointee types.
+This would save a load in pattern matching and might be useful, though given that the main reason for pattern matching is to then call a method on the concrete type, it's not clear that this is actually a win.
+Note that, in the CHERI case, the low bits are still available and so it may be useful to use them to differentiate between pointee types.
+
+These rules are somewhat complex in the abstract.
```
rengolin, line 178: What the document hasn't clarified for me, so far, is if those rules are static (ie. the same for any union type, however complex) or will be optimised to allow corner cases to be represented in the discriminator bits IFF the compiler can guarantee those union types won't intersect with the common case in code.

While I think that kind of optimisation is doable, I believe it makes the life of the compiler more complicated and by consequence helps introduce bugs in code generation that are hard to detect, debug and create tests for.

docs/internal_abi.md
```diff
@@ -0,0 +1,323 @@
...
+The size of `Float64` is 64 bits and it requires one discriminator state.
+
+We first try the special-case rule for the case that the largest value is a floating-point value.
+The remaining type (`U64`) requires 64 bits for payload and 0 bits for discriminator, and so does not fit into the significand part of a `Float64` signalling NaN.
+The first special case is therefore not used
+The second special case is not used because neither type is a pointer.
+
+We are now left requiring one bit of discriminator and 64 bits of payload.
+The payload size is a power of two, and so the address of the union type is the address of the start of a 64-bit payload.
+The 1-bit discriminator is stored afterwards, with a value of 0 representing a `Float64` and a value of `1` representing a `U64`.
```
rengolin, line 193: I believe this means each case will have a special encoding, from 0 to N-1 for N types, which is the point I was trying to make above. 

This "works" if we keep a map of every union in the compiler (or reconstruct the representation every time) and know how to check types and extract the payload. But when looking at the code generated, it'd be hard to know what's the shift and mask for each type, many of which could be merged with some other instruction by the optimising compiler.

I wish we could have a single representation, using the 4-bits in the high-bit-gap of the FP64 representation for all types that fit the payload (`{I,U}x{8,16,32}` + `{F16, F32}` + `ptr`) and do pointer and class separation in the same way we'll do `{I,U}x{Size,64,128}` and `F128`, which is, with an additional descriptor.

This means, I'd prefer to have a separate encoding for boxed primitives and boxed complex types, including singleton types.

I'm assuming that developers understand that usually `for (int in list) -> sum+=int` is faster than `for (obj in list) -> sum+=obj.field`, and that the additional run-time requirements of indirectly calling methods or accessing fields as well as the memory requirements for dynamic classes (perhaps excluding singletons) won't have a huge impact.

docs/internal_abi.md
davidchisnall: I initially wrote this as a straw-man for discussion.  There are a few places where non-power-of-two integers (or, really, bitfields) can be really useful.  The two that come to mind are:

 - Implementing a generic Integer type that stores the value inline if it fits or as a pointer if it doesn't.  
 - String representations that want to store as many characters as possible in a pointer or a pointer to the real data for larger strings.

docs/internal_abi.md
davidchisnall: Minimum Viable Product.  The pre-1.0 release that's complete enough for folks to actually try writing real code in Verona.

docs/internal_abi.md
```diff
@@ -0,0 +1,323 @@
...
+
+In the above example, a `U8` defines 8 bits but, if stored in a 64-bit buffer, leaves 56 unused.
+The `Float64` does not have any unused bits but if the 11 exponent bits and the signalling bit are all 1, the value is a signalling not-a-number (NaN) encoding and the remaining 52 bits are unused.
+The `Foo*`, because it points to a Verona object, has zero in the low three bits and, on a typical 64-bit system with a 48-bit address space has zeroes in the top 16 bits and so needs only 45 bits to encode all valid values (on a system with a 56-bit address space, it still requires only 52 bits to represent any userspace address and so can fit in the NaN space).
+
+It is therefore possible to fully encode this example in a single 64-bit word on any common 64-bit system.
+For example, in userspace on a little-endian 64-bit system with a 48-bit virtual address space that uses integers as pointers:
+
+ - If any of bits 50-62 inclusive is non-zero, then the entire value is a `Float64`.
+ - Otherwise, if the value of bit 0 is 1, then bits 8-15 inclusive contain a `U8` value.
```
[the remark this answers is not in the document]
davidchisnall: According to @mjp41, the runtime is not currently stealing any of the low bits from object pointers.  It's stealing all of the low bits from descriptor pointers (and a few other pointers to runtime data types), so those bits can be available for us to use if we want them.

docs/internal_abi.md
rengolin: Right, then perhaps write that explicitly, as I keep looking at it and thinking of minimum / maximum and the acronym completions make no sense. :)

docs/internal_abi.md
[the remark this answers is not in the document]
rengolin: Ah, I see. Ok, then ignore all the comments I made below on this topic.

docs/internal_abi.md
rengolin: Bitfields are an interesting case per se, but they can also be implemented in Verona over a larger type + shift and mask semantics, so they match as an `U32` for example.

The "generic integer" sounds to me like `(U8 | U16 | U32 | U64 | U128)` and therefore implemented by the standard ABI.

The string representation is an odd one, because only very short strings will be in this representation, so you get two very different performance characteristics with the same type. Not to mention if the string isn't plain ASCII, getting the actual byte representation might be non-trivial. I'd prefer to implement that in Verona...

docs/internal_abi.md
rengolin: Right, so beginning is least-significant. Perhaps use that language instead.

docs/internal_abi.md
rengolin: So, doing the same lowering for any (canonicalised) union type can mean two things:
1. Each case has its own representation, so `(U8 | U16)` can have a different bit-pattern to represent `U8` than in `(U8 | U32)`, but all `(U8 | U32)` will have the same bit-pattern for `U8` and `U32`.
2. Each type is represented in the same way, regardless of the union type, so `U8` would always be the same bit-pattern, no matter the union it's in.

(1) allows us to (consistently) represent a larger set of union types (including classes, interfaces, singletons, pointers), but also requires us to know how to read the discriminators from each possible union. This can be done with a cached version of the patterns on a map to avoid recomputation, but it can also lead to less clear code-gen, which is harder to debug.

(2) allows us to have very predictable code-gen, but at the expense that any non-primitive type will require more loads and branches than they could.

I don't like (2)'s run-time costs, but I also don't like (1)'s compile time costs. I guess we can't just pick one now and change later without some major refactoring on the implementation of the compiler as well as some builtin behaviour.

I'm not sure what is best, tbh, and I'm happy with either if people feel strongly about them.

docs/internal_abi.md
davidchisnall: If you are on a little-endian system.  If you are on a big-endian system, it is the most significant.

rengolin: Ah right, I see, makes sense.

docs/internal_abi.md
davidchisnall: Is there a GitHub syntax highlighting thing that makes it easy to write bit layouts?

docs/internal_abi.md
davidchisnall: I don't think the compile-time cost for calculating the layouts will be very high (and it can be memoised).  

My bias is towards different representations for each union type and explicit operations to transform, but I am also aware of the objection that we may end up with a lot of those that are not no-op casts and that can dominate performance.  I would like it if the compiler could:

 - Preserve these as opaque things for as long as possible.
 - Have an early and at-lowering-to-LLVM point for eliminating no-op casts.
 - Initially use a very simple model for these.

Once we have a reasonable corpus, we can extract the sets of types and casts from the compiler and I can spend some time with Z3 trying to tune the set of representations to minimise the number of conversions that require anything other than a no-op.

docs/internal_abi.md
rengolin: I see, each variation of the union with 1 or 2 types would require a change in representation, but with 2 types you (may) have to deconstruct first, then reconstruct. As the number of types in the union grows, and the number of different usages grow, this can become a bottleneck.

What I'm worried about is also the logic to map N to M conversions where both N and M are unbound. Having to deconstruct to the naive representation and back into another packed form would make the compiler a lot simpler, but would also be worst for run-time performance. 

Unless we come up with a generic algorithm (which I'm not sure yet we can), we'll end up catering for a few particular pain points and leaving the rest as naive conversions. That works, I guess, but it'll be a source of constant pain.

docs/internal_abi.md
```diff
@@ -0,0 +1,323 @@
...
+### Example: `Float64 | T* | U*`
+
+The payload size of `Float64` is 64 bits and it requires one discriminator state.
+The payload size of `T*` and `U*` is each either 32, 48, or 128 bits, depending on the target architecture and they require one discriminator state between them.
+The lowering for this union therefore depends on the target.
+
+On 32-bit systems, the largest type is `Float64` and the remaining types all fit easily in the significand bits.
+There are no discriminator bits required but the 'as-if' rules triggers as if there were a single bit for determining which end of the space should be used and the remaining values are stored at the end of the space that is strongly aligned: bits 0-31.
+
+On 64-bit systems, the same rules apply, packing the used bits of the pointer into bits 0-50.
```
rengolin, line 224: Why 0-50? Isn't it 0-47?

docs/internal_abi.md
```diff
@@ -0,0 +1,323 @@
...
+
+The largest value is a pointer and so on CHERI systems the tag bit is used as the discriminator, on other architectures the low bit of the pointer is used (because the pointer is the largest type and the discriminator fits in a three-bit value).
+The discriminator values are assigned starting from the singletons, so zero is used for `None` and one for `T*`.
+
+A pattern match expression on this type is compiled into a compare-and-branch-on-{non}-zero, which is a single instruction on most architectures.
+Accessing the `T*` on a CHERI system requires no further operations.
+On integer-pointer systems, it requires subtracting one.
+On any mainstream architecture other than RISC-V, this subtraction can be folded into the immediate offset on a load instruction and so requires no additional instructions.
+
+If the `None` value is used (rather than just being pattern matched on and ignored) then the compiler must be able to materialise the address (or capability) to the singleton.
```
rengolin, line 242: This address is where the vtable is, right?

docs/internal_abi.md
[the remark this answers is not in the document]
rengolin: > Make one of them zero and handle the others differently either as:

I think we'll have to make sure the one we pick is being used as a "comparison to zero shortcut" in the code to do this. I'd go with making them all pointers and optimising it later.

> For example, if I have a call site that uses an interface that 100 classes implement, without any run-time profiling information, if I assume that this is megamorphic then am I likely to be right?

IIUC, we won't have to guess by looking at the code itself, but at the AST after reachability analysis. If the code has an open world for interfaces and 1000s of classes implement that, but only two are ever used in my code, then the list of possibilities is actually 2, not 1000.

I believe the compiler will be able to considerably reduce the number of types in unions, creating versions with individual types, or even sub-sets of types (ex. `foo(Interface) -> foo(Class1 | Class2)`), which will make life easier on most cases.

I imagine more complex cases like `foo(I1 | I2)` and where other interfaces are passed as arguments (ex. `I3 : I2`) could prove more complex, but in that case, reverting to the naive implementation is possible and correct.

docs/internal_abi.md
rengolin: Tables?

docs/internal_abi.md
rengolin: This is one possible candidate for a dialect operation, then. That's the only sane way to keep semantics around without pattern-matching later, and we can perform the actual computation (and elision) at the "lower-to-LLVM-dialect" stage.

docs/internal_abi.md
```diff
@@ -0,0 +1,323 @@
...
+
+The first word contains the region metadata.
+This is a discriminated union where the low three bits describe how the remainder is used (for example, as a reference count for one immutable object in an SCC, a pointer to that object for other immutable objects in an SCC, and a pointer to the region metadata for sentinel objects).
+
+The second is a pointer to the descriptor for the object.
+The low bits of this are used by the runtime for memory management state and so must be zeroed before accesses.
+
+The object descriptor layout is still likely to change, the current layout is:
+
+64-bit integer size.
```
rengolin, line 263: I'm confused. Does "64-bit integer size" means there's a 64-bit padding before the trace pointer? Or that integers here are 64-bit, but actually all elements are pointers? Or that all pointers are encoded in a 64-bit integer by means of offsets into other known vtables?

docs/internal_abi.md
```diff
@@ -0,0 +1,323 @@
...
+Pointer to a trace function, used in traced regions to walk reachable objects.
+Pointer to a finalise function, called when an object is dead but before other objects in the same region are deleted (though they may have already been finalised.
+Pointer to a notify function, invoked when an object is notified of a noticeboard event.
+Pointer to a destructor, which is called after some objects in a region have been deallocated and may not follow any `mut` pointers.
+The remainder of the descriptor is the vtable.
```
rengolin, line 268: Are we having a vtable per type or is this a pointer to a global vtable?

docs/internal_abi.md
```diff
@@ -0,0 +1,323 @@
...
+
+Methods in the vtable use *selector colouring*.
+Each selector (method name and types) is assigned a separate vtable index.
+Pairs of selectors that do not appear in the same interface are then combined to use the same index.
+Selector colouring is not essential but can dramatically reduce the size of vtables.
+
+*Note:* If we can distinguish between sentinel objects and others at object-creation time, the region metadata field for non-sentinel objects could be a 32-bit value that is either an integer or an index into a table, rather than a pointer.
+This would be a noticeable saving on CHERI systems and a small saving on 64-bit systems, at the cost of an extra indirection for some memory-management operations.
+We could also use a 32-bit index into a descriptor table, as the Fuchsia C++ ABI does: Programs that have more than 2^32 distinct types arise only as a result of poor life choices.
```
rengolin, line 277: I have to strongly agree with you on this one.

rengolin (commented):
I've re-read to the end again and my comments are all that I can think of.

I think the main issue is the non-PO2 integers being considered as a harder consensus than I have seen in meetings, but otherwise, looks good.

rengolin:
Closing this in favour of #467, which is leaner in some areas and broader in others. It's also up-to-date with a stronger consensus within the team, after we discussed all the topics in this document.

PR #342, opened 2020-11-05
Fix false leak detection of muted cowns

Theodus:
This PR changes the condition to delay a muted cown. Previously, a cown would not be muted on a thread in the `Prescan`, `Scan`, or `AllInScan`. Now we only delay muting cowns that have the `SCANNED` epoch mark.

This PR also makes a change to rescan the mute map so long as a key in the map has been unmuted as a member of an invalid mute set. This is necessary to ensure that deadlock detection avoids false positives after an incremental scan.

### DIFF
--- a/src/rt/sched/schedulerthread.h
+++ b/src/rt/sched/schedulerthread.h
@@ -267,33 +267,43 @@ namespace verona::rt
      */
     void mute_map_scan(bool force = false)
     {
-      for (auto entry = mute_map.begin(); entry != mute_map.end(); ++entry)
+      while (true)
       {
-        auto* m = entry.key();
-        auto& mute_set = *entry.value();
-        if (force || !m->triggers_muting() || m->is_collected())
+        bool scan_again = false;
+        for (auto entry = mute_map.begin(); entry != mute_map.end(); ++entry)
         {
-          yield();
-          for (auto it = mute_set.begin(); it != mute_set.end(); ++it)
+          auto* m = entry.key();
+          auto& mute_set = *entry.value();
+          if (force || !m->triggers_muting() || m->is_collected())
           {
-            Systematic::cout()
-              << "Mute map remove cown " << it.key() << std::endl;
-            it.key()->backpressure_transition(Priority::Normal);
-            T::release(alloc, it.key());
-            mute_set.erase(it);
+            yield();
+            for (auto it = mute_set.begin(); it != mute_set.end(); ++it)
+            {
+              assert(m != it.key());
+              Systematic::cout()
+                << "Mute map remove cown " << it.key() << std::endl;
+              it.key()->backpressure_transition(Priority::Normal);
+              T::release(alloc, it.key());
+              scan_again =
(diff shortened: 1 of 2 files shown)


src/rt/sched/schedulerthread.h
```diff
@@ -223,9 +223,9 @@ namespace verona::rt
         assert(p != Priority::Low);
 
         if (
```
mjp41, line 225: Please add a comment explaining this condition.

src/rt/sched/schedulerthread.h
```diff
@@ -223,9 +223,9 @@ namespace verona::rt
...
-          (p & PriorityMask::High) || (state == ThreadState::PreScan) ||
-          (state == ThreadState::Scan) || (state == ThreadState::AllInScan))
-        { // Messages in this cown's queue must be scanned.
+          (p & PriorityMask::High) ||
```
mjp41, line 226: Please add a comment explaining this condition.

src/rt/sched/schedulerthread.h
```diff
@@ -267,33 +267,43 @@ namespace verona::rt
      */
     void mute_map_scan(bool force = false)
     {
-      for (auto entry = mute_map.begin(); entry != mute_map.end(); ++entry)
+      while (true)
```
mjp41, line 270: Why do you have to scan the mute map multiple times?  Could you add a comment to explain.I think it is because you might make something normal, which then requires removing from the mute map?

mjp41, line 270:
```C++
bool scan_again = true;
while (scan_again)
{
  scan_again = false;
  ...
}
```
might be slightly cleaner than the break at the end.

src/rt/sched/schedulerthread.h
```diff
@@ -267,33 +267,43 @@ namespace verona::rt
...
-          alloc->dealloc<sizeof(ObjectMap<T*>)>(&mute_set);
-        }
-        else if (mute_set.size() == 0)
-        {
-          m->weak_release(alloc);
-          mute_map.erase(entry);
-          mute_set.dealloc(alloc);
-          alloc->dealloc<sizeof(ObjectMap<T*>)>(&mute_set);
         }
+
```
mjp41, line 304: Is it the case that if `force==true` then everything has been emptied at this point, and we don't need to go round again?  I.e. is the loop only required in the non-force case?

mjp41 (approved):
LGTM, just add a few comments

src/rt/sched/schedulerthread.h
Theodus: yes, and the `scan_again` condition has been updated to skip the check for an unmuted key if `force` is true.

src/rt/sched/schedulerthread.h
Theodus: Comment added. Yes, when a cown is unmuted it may have previously muted cowns that had scheduled behaviours on it. So their entries should be removed as well.

Issue #344, opened 2020-11-10
Typed AST


plietar:
The compiler currently uses an "untyped" AST, which is essentially an S-expression. This makes working on it very painful and code hard to maintain, as there is no central definition of what the shape of the tree needs to be.

We should instead replace it with an explicit and typed AST, akin to [what the original prototype had](https://github.com/microsoft/verona/blob/4d6f471b28f5223aa5468e4918f9ea888ced3315/src/compiler/ast.h).

The plan for now is:
1. Define the typed AST
2. Add a translation pass from the untyped AST to the typed one
3. Change the AST->MLIR lowering to operate on the typed AST.
4. Incrementally shift existing passes from the untyped AST to the typed one. This needs to happen in reverse compared to the execution order.
5. Make the parser produce a typed AST. There are different ways we can achieve this, either by using peglib's semantic actions, or by writing the parser by hand, or we never do this step and keep the untyped->typed conversion "forever" (until the next time we rewrite the compiler).
6. Delete all untyped AST code.

There's a couple of open questions on the design of the AST itself:

- How are variants of expressions, types, class members, ..., represented?
    - Abstract class with concrete subclasses.
        - This is what the prototype's AST used.
        - Should we set up [LLVM-style RTTI](https://llvm.org/docs/HowToSetUpLLVMStyleRTTI.html)?
    - `typedef std::variant<FooExpr, BarExpr, ...> Expr;`
        - This is what the prototype's IR used.
- How are variants traversed, aka [the expression problem](https://en.wikipedia.org/wiki/Expression_problem):
    - Add virtual methods to each variant.
    - Use a visitor pattern
    - std::visit (assuming we chose `std::variant`)
    - LLVM TypeSwitch
    - dynamic_cast
    - Use a [cast + visitor pattern](https://github.com/microsoft/verona/blob/4d6f471b28f5223aa5468e4918f9ea888ced3315/src/compiler/visitor.h). This allows the visitor to be templated over additional arguments and a return type.
- How are source locations represented? Do all nodes carry a location, or only select ones?
- How are symbol tables and symbol references represented?
    - In particular, I think that after a dedicated resolution pass, all symbol references have direct pointers to the defining node. Ulterior passes can directly access that pointer and use it as a simple unique identifier, rather than reimplementing their own lookup tables.
- Are strings in the AST interned?
- Can we reuse any of the original prototype's AST, visitor infrastructure and name resolution pass?

cc @rengolin @sylvanc @mjp41

[...]

PR #345, opened 2020-11-12
Create an initial Typed AST

plietar:
This introduces an initial version of the typed AST, and a conversion from the untyped to the typed version.

The AST is still incomplete. ~~It is missing a couple of expressions (tuple, when, new at least)~~. Most importantly does not encode any of the types and signatures that appear in the program.

I would like to ~~add the last few expressions in this PR and then~~ merge the PR without the type hierachy and without using the AST anywhere yet.

See #344


rengolin:
Quick comment: don't try to fix issues in the generator now, make sure your patch is completely independent, otherwise it will be harder for both of us to work in parallel. The typed AST discussion is, for now, orthogonal to how we lower it to MLIR, and only once we all agree, that we'll create the layer that lowers it down.

Also, don't introduce *any* new feature while doing so (ex. yield, etc). If you want those supported before this patch, submit separate patches. Otherwise, just wait until this is finished. We don't have enough tests and we don't trust enough our current tests (because they don't lower to LLVM at all) to be able to do large changes and still hope nothing on the side was changed.

An approximate roadmap would be:
 * start with the AST design loose as an optional step off the last pass int he parser only.
   * add tests with the expected results to make sure we're not regressing on the agreed syntax
   * meanwhile, both me and sylvan can continue working on the MLIR and AST side
   * only start the next step when this is approved and merged
 * make the generator read the new AST
   * refactor `ast-utils.h` in order to reduce the changes to the generator to a minimum
   * doesn't mean it's perfect now, but allows us to compare results independently
   * only start the next step when this is approved and merged
 * Refactor the generator to improve use of the new AST features
   * this should still produce similar IR but there could be enough changes to cloud our view in previous steps
   * this is the time to change maps and control data structures in the generator

Each step needs to be completed because there will be a lot of back and forth on the design and we shouldn't have to refactor all steps on every little change. This leads to a lot of work being thrown away and it leads to bad decision making.

plietar:
> Quick comment: don't try to fix issues in the generator now, make sure your patch is completely independent, otherwise it will be harder for both of us to work in parallel. The typed AST discussion is, for now, orthogonal to how we lower it to MLIR, and only once we all agree, that we'll create the layer that lowers it down.

Sorry these were just some compiler warnings that were getting on my nerves. Will remove.

> Also, don't introduce any new feature while doing so (ex. yield, etc). If you want those supported before this patch, submit separate patches.

I don't think I'm adding any existing features. These already exist in the parser, untyped AST and relevant tests, so I'm including them in the typed AST. I of course won't implement them in the generator now (or anytime soon).

----

Your Step 1 is essentially this PR, right? Apart from the warnings I was fixing in `generator.cc`, I'm not changing any existing code.

I'm not sure it makes sense to change ast-utils to use the typed AST. Shoehorning the new AST into it will require more work than it will to adapt the generator to the new AST.

rengolin:
> Sorry these were just some compiler warnings that were getting on my nerves. Will remove.

They're good catches, and now I'm wondering why my build doesn't give those warnings. Are you using clang or gcc?

> I don't think I'm adding any existing features. These already exist in the parser, untyped AST and relevant tests, so I'm including them in the typed AST. I of course won't implement them in the generator now (or anytime soon).

Sorry, not features, I meant the new values in the `ast-utils.h` enum, which are unused. I want to make sure we only add there what we really use, when we use, so it's clear what they refer to (alongside implementation helps clarify their intent and semantics).

> Your Step 1 is essentially this PR, right? Apart from the warnings I was fixing in `generator.cc`, I'm not changing any existing code.

Yes, but without any change to MLIR infrastructure, including `ast-utils.h`, `generator.cc` and `verona-mlir.cc`. Just adding new files (source and tests) and updating CMake stuff to make it build.

> I'm not sure it makes sense to change ast-utils to use the typed AST. Shoehorning the new AST into it will require more work than it will to adapt the generator to the new AST.

I understand. The helper is a documentation on how we generate MLIR from the existing AST. A new AST won't need that, as hopefully it will have a syntax so obvious that all of those methods will be implemented in the AST structure itself. 

What I meant is "use the helper as a guide to design the new AST in a way that will require minimal changes to the MLIR lowering". So, in the simplest case (probably not going to happen), we could just replace `AST::` to `ast->` and all methods will be implemented by the nodes (ex. `ast->isLoop()` instead of `AST::isLoop(ast)`). Which leaves the changes to the generator only what really improves readability, simplicity and performance (ex. exposing `ArrayRef` to internal vectors instead of building new ones every time).

I'm really hoping that by the time you finish we'll happily delete `ast-utils.h` (which was my plan all along for that file and why it's so horrendous).

plietar:
> They're good catches, and now I'm wondering why my build doesn't give those warnings. Are you using clang or gcc?

Ah right, so LLVM's cmake was accidentally disabling all warnings. [This got fixed upstream a couple of months ago](https://github.com/llvm/llvm-project/commit/b36bdfe5ca0c2b863248f327b03d41516b38dc11), so since I've been working with an updated LLVM for #335, they started popping up on my machine.

plietar:
Split off some of the unrelated cleanups to #346 and #347.

plietar:
> Sorry, not features, I meant the new values in the ast-utils.h enum, which are unused.

They are used in the new `conversion.cc` file. Would you suggest I not use `ast-utils.h` at all to define the uptyped -> typed conversion then? At the moment I'm including it just for `NodeKind` and `isA`/`isAny`. I could certainly do without those.

(It's actually annoying to depend on `ast-utils.h` because that depends on `LLVMSupport` for `ArrayRef`/`StringRef`).

src/typed-ast/print.cc
```diff
@@ -0,0 +1,250 @@
...
+      print_compact(out, node);
+      return buf.size();
+    }
+  }
+
+  void print(std::ostream& out, const Node& node, size_t indent, size_t width)
+  {
+    // We first try to print the node in its compact form. If this takes more
+    // space than what is available, we fall back to the extended form.
+    size_t len = compact_length(node);
```
mjp41, line 111: I worry about the quadratic blow up here on printing large ASTs.  But fine for a first cut, and I guess we will only ever use this for testing.  It seems you are really calculating `compact_length` for each node, adding a memoization to that would make things much better from a complexity point of view.

src/typed-ast/expr.h
```diff
@@ -0,0 +1,637 @@
...
+    {
+      return E->getKind() == Kind::New;
+    }
+
+    // The implementation needs MemberDef to be complete, therefore it is moved
+    // to `print.cc`.
+    void print(NodePrinter& out) const override;
+  };
+
+  class LambdaExpr final : public Expr
```
mjp41, line 618: Does a lambda not have parameters?

src/typed-ast/expr.h
```diff
@@ -0,0 +1,637 @@
...
+      return E->getKind() == Kind::StringLiteral;
+    }
+
+    void print(NodePrinter& out) const override
+    {
+      out.begin("string-literal").field(value).finish();
+    }
+  };
+
+  class InterpolateExpr final : public Expr
```
mjp41, line 525: `InterpolateExpr`  pretty printing as `seq`?

src/typed-ast/print.cc
plietar: Yeah, I wonder how bad the blow up is. The idea was taken from [Pony's own pretty-printer](https://github.com/ponylang/ponyc/blob/eb6fbbb16600249e1d90cd3e6674ebe75fecc5b0/src/libponyc/ast/ast.c) which probably has the same issue. The pony one doesn't use the fake ostream trick to compute the length, but instead has an explicit computation, which is probably faster but more fragile too. The complexity would be the same in both cases though.

src/typed-ast/expr.h
plietar: Yes they do, but they are part of a `signature` node in the untyped AST, which I've ignored for now.

I can add TODOs for fields that should exist in the future (I've done that in `conversion.cc` already).

src/typed-ast/node.h
```diff
@@ -0,0 +1,52 @@
...
+#include <string>
+
+namespace verona::ast
+{
+  class NodePrinter;
+  typedef std::string Symbol;
+
+  /// This represents a location in the source.
+  /// TODO: reuse the code from `compiler/source_manager.h` for a more efficient
+  /// representation.
```
rengolin, line 15: It would be nice if this used MLIR-compatible source location, or even reused it to avoid creating infrastructure that exists (and we use).

src/typed-ast/ast.h
```diff
@@ -0,0 +1,160 @@
...
+/// automatically using `std::unique_ptr` on every edge of the tree. In the
+/// future, we should consider using an arena to allocate all the nodes, since
+/// we are unlikely to ever deallocate individual nodes.
+///
+/// [LLVM-style RTTI]: https://llvm.org/docs/HowToSetUpLLVMStyleRTTI.html
+namespace verona::ast
+{
+  /// This is the base class for entity members, from which all concrete nodes
+  /// derive.
+  class MemberDef : public Node
```
rengolin, line 35: What's behind the name `MemberDef`? Why not `NodeDef`? To me, `Member` means a class member (field, method).

src/typed-ast/print.cc
```diff
@@ -0,0 +1,250 @@
...
+      out << '"';
+      for (char c : s)
+      {
+        if (c == '\\' || c == '"')
+          out << '\\' << c;
+        else if (c == '\a')
+          out << "\\a";
+        else if (c == '\b')
+          out << "\\b";
+        else if (c == '\e')
```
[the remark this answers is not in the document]
plietar: Heh, that explains why my editor was highlighting it weirdly.

src/typed-ast/expr.h
plietar: Good spot. Thanks.

src/typed-ast/ast.h
```diff
@@ -0,0 +1,160 @@
...
+    void print(NodePrinter& out) const override
+    {
+      out.begin("field").field(getName()).optional_field(initializer).finish();
+    }
+  };
+
+  /// This represents a method definition.
+  class MethodDef : public MemberDef
+  {
+    ExprPtr body;
```
rengolin, line 124: What about arguments, types, wheres, etc?

src/typed-ast/expr.h
```diff
@@ -0,0 +1,637 @@
...
+    explicit Expr(Kind kind, SourceLocation location)
+    : Node(location), kind(kind)
+    {}
+    Kind kind;
+  };
+
+  using ExprPtr = std::unique_ptr<Expr>;
+
+  /// This represents a local variable declaration, e.g. `let x`.
+  /// TODO: unclear what this "evaluates" to.
```
rengolin, line 64: In the generator, this is only used to fetch the name for the symbol table and the type for overriding in an assignment.

ex:
```
let a : U32 = 10;
let b : U16 = a;
```

When assigning `b` we add a `verona.cast -> U16` and set that as the new symbol's SSA value.

src/typed-ast/expr.h
```diff
@@ -0,0 +1,637 @@
...
+  class LocalDeclExpr final : public Expr
+  {
+    Symbol name;
```
rengolin, line 68: Also need declared type.

src/typed-ast/expr.h
```diff
@@ -0,0 +1,637 @@
...
+    void print(NodePrinter& out) const override
+    {
+      out.begin("yield").field(value).finish();
+    }
+  };
+
+  /// This represents a call to an object, e.g. `E1(E2, E3)`.
+  ///
+  /// In the case of method calls, eg. `E.m()`, the receiver of the call is
+  /// `E.m`, a partial application of the method.
```
rengolin, line 336: Why do we have this to support method calls when we already have the `MethodCallExpr` node?

src/typed-ast/expr.h
```diff
@@ -0,0 +1,637 @@
...
+    {
+      out.begin("invoke").field(receiver).field(arguments).finish();
+    }
+  };
+
+  /// This represents a static method call, e.g. `T.m(E1, E2)`
+  class StaticCallExpr final : public Expr
+  {
+    Symbol name;
+    // Type receiver;
```
rengolin: Is this comment intentional?

src/typed-ast/expr.h
```diff
@@ -0,0 +1,637 @@
...
+    {
+      out.begin("call").field(name).field(receiver).field(arguments).finish();
+    }
+  };
+
+  /// This represents an integer literal, e.g. 345
+  class IntegerLiteral final : public Expr
+  {
+    // TODO: use something like APInt
+    uint64_t value;
```
rengolin, line 428: what about signed values? if we're going to have a cheap default for now, then `int64_t` would be more accomodating.

src/typed-ast/expr.h
```diff
@@ -0,0 +1,637 @@
...
+
+    void print(NodePrinter& out) const override
+    {
+      out.begin("bool-literal").field(value).finish();
+    }
+  };
+
+  class StringLiteral final : public Expr
+  {
+    std::string value;
```
rengolin, line 509: // TODO: Support unicode

src/typed-ast/expr.h
```diff
@@ -0,0 +1,637 @@
...
+      return E->getKind() == Kind::BooleanLiteral;
+    }
+
+    void print(NodePrinter& out) const override
+    {
+      out.begin("bool-literal").field(value).finish();
+    }
+  };
+
+  class StringLiteral final : public Expr
```
rengolin, line 505: missing doc comment

src/typed-ast/expr.h
rengolin, line 534: missing doc comment

src/typed-ast/expr.h
```diff
@@ -0,0 +1,637 @@
...
+
+    void print(NodePrinter& out) const override
+    {
+      out.begin("when").field(arguments).field(body).finish();
+    }
+  };
+
+  class NewExpr final : public Expr
+  {
+    std::vector<std::unique_ptr<MemberDef>> elements;
```
rengolin, line 614: Do you really mean `MemberDef` or `FieldDef`?

src/typed-ast/expr.h
rengolin: Yes please. I've been adding comments about them until I reached this comment. :)

Or better yet, add the fields directly even if you initialise them as null.

src/typed-ast/main.cc
```diff
@@ -0,0 +1,31 @@
...
+  pass::Passes passes = {{"sugar", sugar::build},
+                         {"sym", sym::build},
+                         {"ref", ref::build},
+                         {"prec", prec::build}};
+  err::Errors err;
+
+  auto opt = cli::parse(argc, argv);
+  auto m =
+    module::build(opt.grammar, opt.stopAt, passes, opt.filename, "verona", err);
+  auto typedModule = verona::ast::convertModule(m->ast);
```
rengolin, line 24: Are you planning to add a set of passes on top of your converted module?

So it becomes:
```
old ast -> sugar -> sym -> ref -> prec;
new ast = convert(prec) -> return;
```
then
```
old ast -> sugar -> sym -> ref;
new ast = convert(ref);
new ast -> new prec -> return;
```
then
```
old ast -> sugar -> sym;
new ast = convert(sym);
new ast -> new ref -> new prec -> return;
```
etc.

src/typed-ast/node.h
```diff
@@ -0,0 +1,52 @@
...
+    size_t line;
+    size_t column;
+
+    explicit SourceLocation(std::string file, size_t line, size_t column)
+    : file(file), line(line), column(column)
+    {}
+  };
+
+  /// Base class from which all AST nodes derive from.
+  class Node
```
rengolin, line 30: This file could actually move entirely to `ast.h`?

rengolin (commented):
Many comments inline, some overall comments here:

The `ast.h` / `node.h` division seems unnecessary and confusing. Perhaps `ast.h` could be only a catch-all header to include in the generator or it could merge what `node.h` has now.

I know you're only testing the printing, but many nodes have missing get methods. Perhaps the printer could use those methods, too.

Choosing which sub-node it it should also work with switch/case, in addition to `isa` and `cast`. The enums are split between `Expr::Kind` and `MemberDef::Kind` which makes that not possible (clash of values).

A lot of nodes are missing doc comments. Some are redundant, others are important. All need comments.

The conversion file looks nice, but that's only because there was a lot of cleanup from the other passes. Once you look at the original AST tree, you'll see that it will become hard to reason about the position and existence of nodes.

Perhaps we should always have a cleanup pass before converting to typed AST, even after we move all passes to the typed side. This could be a step in the migration:
```
old ast -> passes... -> cleanup -> conversion -> new ast -> new passes
```
and at the end:
```
old ast -> cleanup -> conversion -> new ast -> new passes
```

src/mlir/verona-mlir.cc
```diff
@@ -54,8 +54,8 @@ namespace
   // Output file
   cl::opt<std::string> outputFile("o", cl::init(""), cl::desc("Output file"));
 
-  // Grammar file is not optional
-  std::string grammarFile;
+  // Grammar file
+  cl::opt<std::string> grammarFile("g", cl::init(""), cl::desc("Grammar file"));
```
rengolin, line 58: This change it not relevant

src/mlir/ast-utils.h
```diff
@@ -99,6 +102,12 @@ namespace mlir::verona
       InitExpr = peg::str2tag("initexpr"), // = 3935567717
       Member = peg::str2tag("member"), // = 75189616
       Lookup = peg::str2tag("lookup"), // = 4099072514
+      InterpString = peg::str2tag("interp_string"), // TODO
```
rengolin, line 105: Please don't add anything to the MLIR side in this PR.

src/typed-ast/ast.h
plietar: It is a class member (well an entity member, which could be a class/module/interface). Why would it be `NodeDef`?

src/typed-ast/expr.h
plietar: Added a TODO.

src/typed-ast/expr.h
plietar: If we have `x.f(y)`, we don't know yet if this is a field lookup, followed by a call the resulting object, or if it is a call to the method `f`. `MethodCallExpr` is used for operator calls, which aren't allowed to be fields.

 I'm not happy with it either, but that's what the untyped AST looks like right now.

src/typed-ast/ast.h
plietar: Added TODO.

src/typed-ast/print.cc
plietar: Added a note explaining the risk, and pointed to memoization as a possible solution.

src/typed-ast/expr.h
plietar: Changed it to a clearer TODO.

src/typed-ast/expr.h
plietar: Good point. Will change.

src/typed-ast/expr.h
plietar: Done.

src/typed-ast/expr.h
plietar: Done.

src/typed-ast/expr.h
plietar: `MemberDef`. You are allowed to do, for example:

```ts
interface I {
  foo();
}

new I {
  foo() { ... }
}
```

Or at least that's how I understand the grammar.

plietar:
On the topic of file layout:
I wanted the "entrypoint" node, ie. EntityDef, to be in `ast.h`.
However I also wanted expressions to be in their own file.

Because `ast.h` depends on `expr.h`, and `expr.h` needs to include the `Node` class, I had to put the latter in a separate file.

I'm open to suggestions if you think there is a better option.

src/mlir/verona-mlir.cc
plietar: Yep. It will go away once I rebase onto #347.

src/typed-ast/expr.h
plietar: Done.

src/typed-ast/expr.h
plietar: Done.

src/mlir/verona-mlir.cc
plietar: Done.

src/typed-ast/expr.h
plietar: Well the question is, if `let x` is as good an expression as anything else, does `(let x).foo()` mean anything? What about `(let x = E).foo()`?

We probably want to introduce a verification pass that checks that `let x` does not appear in an rvalue context. Therefore the question "what does it evaluate to", in an rvalue context, is meaningless.

src/typed-ast/main.cc
plietar: I don't know yet. I don't have any immediate plans to, but it is bound to happen at some point.

src/mlir/ast-utils.h
plietar: See my top-level comment. I am using these in `conversion.cc`. I'm happy to not do that anymore, and use something else.

rengolin:
> They are used in the new `conversion.cc` file. Would you suggest I not use `ast-utils.h` at all to define the uptyped -> typed conversion then? At the moment I'm including it just for `NodeKind` and `isA`/`isAny`. I could certainly do without those.
> (It's actually annoying to depend on `ast-utils.h` because that depends on `LLVMSupport` for `ArrayRef`/`StringRef`).

Yes, just replicate the logic for the time being. We'll hopefully remove that file by the end of this effort.

Also, feel free to not include the actual number that `str2tag` evaluates to in the comments, I needed that before we had `ast_dump`. Now it's not necessary any more.

src/typed-ast/ast.h
rengolin: Hm, I see, "Entity member definition". The way I read it (and its usage in field access below) makes me think of "class member". I think if we consider modules, interfaces and classes as the same thing, than this is actually correct.

Perhaps a more elaborate comment on that before the declaration would help those uninitiated in Verona parlay. :)

src/typed-ast/main.cc
rengolin: Makes sense, one thing at a time.

src/typed-ast/expr.h
rengolin: Ah, right, I have to handle that in the `parseCall` too and they're virtually identical. Perhaps we could common them up here even if the old ast doesn't, if they're semantically equivalent for all passes after this one.

src/typed-ast/expr.h
rengolin: Hm, this looks weird to me, but so did a lot of Verona code before. Should be fine for now.

plietar:
> Yes, just replicate the logic for the time being. We'll hopefully remove that file by the end of this effort.

Done.

src/typed-ast/expr.h
plietar: They aren't equivalent though. If `f` is a field, then the semantics of InvokeExpr and MethodCallExpr are different. Except we won't know what it is until late in the pipeline.

rengolin: Hm, ok, so we'll have to update the generator, because I'm treating both the same. There weren't any example of calling a field (or a method FWIW), so I was a bit lost. It should be simple to do that distinction, though, either as a new dialect operation or a variation on the existing one. We'll worry about that when we have the new ast.

rengolin:
> On the topic of file layout:
> I wanted the "entrypoint" node, ie. EntityDef, to be in `ast.h`.
> However I also wanted expressions to be in their own file.
> 
> Because `ast.h` depends on `expr.h`, and `expr.h` needs to include the `Node` class, I had to put the latter in a separate file. Putting `Node` in `ast.h` would create a cycle.
> 
> I'm open to suggestions if you think there is a better option.

Right, my suggestion was uninformed, it just "feels" weird which is not much of an argument. We can worry about that if it becomes a problem.

plietar:
This is pretty much ready in my opinion. Let me know if there is anything you think needs changing.

rengolin (approved):
This looks great to me, thanks!

Perhaps @mjp41 and @sylvanc should have a final look (and approve) before merge, as this will hopefully replace our current ast.

mjp41 (approved):

PR #346, opened 2020-11-12
Fix some warnings in generator.cc

plietar:
LLVM was accidentally disabling all warnings, which let a couple of
them sneak into generator.cc

We can re-enable them by setting LLVM_ENABLE_WARNINGS to ON. This was
fixed upstream, so we can get rid of it next time we update.


plietar:
For reference, https://github.com/llvm/llvm-project/commit/b36bdfe5ca0c2b863248f327b03d41516b38dc11 this is the upstream fix.

mjp41 (approved):
LGTM

PR #347, opened 2020-11-12
Fix and tidy the testsuite CMake scripts.

plietar:
- ~~Add a -g option to verona-mlir. It was used by update-dump but not
  actually supported.~~
- **Don't pass the -g option to `verona-mlir`. That binary doesn't support it.**
- Correct the path for the IR output file in update-dump. It should be
  `mlir.txt`, not `out.mlir`.
- Use the `CheckStatus` helper in `ast-parse.cmake` and `mlir-parse.cmake`.
  This is equivalent to execute_process, but has builtin support for
  checking the exit code.
- Change the name of the `PARSER` and `VERONAM` variables to match the
  actual binary names. This is neater as we add more binaries.


testsuite/update-dump.cmake
```diff
@@ -12,18 +12,17 @@ foreach(TEST_FOLDER ${TEST_FOLDERS})
...
-        COMMAND ${PARSER} -a -g ${GRAMMAR} ${PARSE_TEST}
+        COMMAND ${VERONA_AST} -a -g ${GRAMMAR} ${PARSE_TEST}
         OUTPUT_FILE ${OUT_FILE})
     endforeach()
   elseif(${TEST_FOLDER} MATCHES ".*/mlir")
     file(GLOB MLIR_TESTS "${TEST_FOLDER}/mlir-parse/*.verona" "${TEST_FOLDER}/mlir-parse/*.mlir")
     foreach(MLIR_TEST ${MLIR_TESTS})
       get_filename_component(TEST_NAME ${MLIR_TEST} NAME_WE)
-      set(OUT_FILE ${TEST_FOLDER}/mlir-parse/${TEST_NAME}/out.mlir)
+      set(OUT_FILE ${TEST_FOLDER}/mlir-parse/${TEST_NAME}/mlir.txt)
```
mjp41, line 22: Using `out.mlir` will mean VS code will syntax highlight more nicely.  Would it be better to switch the other way?

src/mlir/verona-mlir.cc
```diff
@@ -97,7 +97,8 @@ namespace
 
     // Default grammar
     // FIXME: Move to llvm::sys::path, but LLVM's GetMainExecutable is horrible
-    grammarFile = path::directory(path::executable()).append("/grammar.peg");
+    if (grammarFile.empty())
```
mjp41, line 100: Do we ever change the grammar for tests?  If not, I am not sure why we need the tests to supply it?  It seems you are adding functionality here, where as we could just drop the argument from the tests?

mjp41 (approved):
Looks fine to me.  Couple of minor bits.

testsuite/update-dump.cmake
plietar: An issue with out.mlir is that the copyright check script will look at these files and fail. Also the check_dump script is configured to only look at .txt files.

These are both fixable issues, but naming it mlir.txt sounds like the simplest option.

mjp41: They sounds good reasons.  So please ignore.

rengolin: Ah, I changed this on the test runner and forgot to update the dump. This change is correct, even though I agree with @mjp41 's points. :)

rengolin (commented):
Please don't change code and CMake files at the same time unless it's strictly necessary.

src/mlir/verona-mlir.cc
rengolin: This change is irrelevant here and makes no difference in the CMake files or tests. Leave this for another PR.

rengolin (changes requested):

src/mlir/verona-mlir.cc
plietar: I disagree. The update-dump script was broken because it was passing the `-g` flag to `verona-mlir`. I can fix this by removing it from update-dump or by adding the flag to the source, but either approach seems “relevant” to me.

mjp41: I would prefer to remove from `update-dump`.

plietar: @mjp42 I was mostly trying to make the way tests are run uniform across different binaries. `verona-ast` is called with the grammar file, so I figured so should `verona-mlir`. I’m happy to just remove it from the cmake script instead.

To be honest I’m not really sure why we allow the path to the grammar file to be configurable (and in my opinion it should even be embedded in the binary).

rengolin: The comment says: "Grammar file is not optional" and you're changing it to optional. This is change in behaviour that does not belong to a CMake change.

If the update script is using the program in the wrong way, we need to fix the update script, not the program.

If we want to change the behaviour to make grammar files optional, that's another discussion.

plietar: I really feel like this is splitting hairs. Saying commits that touch cmake scripts aren’t allowed to change implementation is such an arbitrary rule.

Additionally this line of code is in a function called `cmdLineDefaults`, along with a comment saying this is the “default grammar”. That makes it sound like it was intended to be a flag all along, but was just forgotten. “The update script is using the program in the wrong way” is nothing but the other side of the “the program doesn’t accept options it is supposed to” coin.

Whatever. I’ll change the cmake script and undo the C++ change.

mjp41:
I wonder if we should make the test do
* Check for no changes in `testsuite` that are not in `git`.
* Run update-dump
* Check for no changes in `testsuite` again.
Then 
1. we only have one utility.  
2. We are testing it in CI.  `update-dump` is not tested by CI, hence why some of this issues exist.
Might be a bit of a pain, but depending on `git` for checking if things have changed is reasonable. We already expect `git` on every platform, so not making portability worse.

plietar:
I think that makes local workflow more annoying. It means I can only run tests if I’ve committed my changes.

I do agree however that the update-dump script should be tested, and that the duplication between that script and the ones that actually run tests is annoying.

rengolin:
It doesn't have to be a local workflow process, and could be done as an extra job in the CI.

I think it would be good to have that because it would have caught all these issues (and others we fixed earlier) straight away.

mjp41:
The script could create a temporary commit for any changes in that directory, and then undo after testing, but that worries me for things going wrong.

Having a different process for local and CI worries me as that then means things don't get tested.

plietar:
I've undone the change in `verona-mlir`. Is it okay if I merge this PR, and we can design a bigger refactor later?

rengolin (approved):
Thanks! Please squash before merge, to avoid touching files that didn't change.

PR #350, opened 2020-11-14
Replace load metric with cown RC

Theodus:
This PR removes the `status` bits from the cown and instead tracks load via the reference count of the cown. This allows for the runtime to manage a growing messages queue on a cown that has been acquired and therefore cannot compute its load until it is rescheduled. The bit used for tracking the existence of a message token in the cown's queue has been moved into the `bp_state` field.


Theodus:
Blocked on https://github.com/microsoft/verona/issues/355

Theodus:
Closing in favor of #390

PR #353, opened 2020-11-17
Move from alloca/load/store to block arguments

### LINKED ISSUE #322: Implement BasicBlock arguments
Earlier [prototype](https://github.com/rengolin/verona/commits/zzz/freevar) need to be extended to allow basic block arguments instead of `alloca/load/store` chains.

It could be done in the parser (since it's done in the AST), or it could be done in the generator (since it's only used for the initial MLIR lowering).

Any further transformation of the IR should handle arguments and reachability as usual.

rengolin:
This is a big change in two areas: conditionals and loops. 

Essentially, it moves away from `alloca` for all variables to use SSA values and track them via `writes minus defs` in the AST and use those as arguments to the basic blocks. This allowed us to remove a lot of code that was generating loads and stores, checking for allocas and makes the IR much leaner and simpler to process.

It has two main shortcomings:
1. Because of the flat analysis (on the AST), it may add additional arguments to basic blocks that don't need it, especially in loops' back-edges. This can later be removed with a pass that looks at back-edges and clean them up, or possibly a more generic pass that builds on the existing MLIR block argument pass.
2. To avoid complexity in the basic type check (on `verify`), all block arguments have type `unknown`. A type check pass can populate those with the right level of abstraction, considering the types of branch operands and arguments that reach each other, and make sure all types are covariant, contra-variant and whatever is needed in each case.

The current consensus is that there will be a stage, possibly right after type inference, where there could be no `unknown` types at all in the IR, and no new ones can be created. From then on, *all* type validation should pass after every transformation.

Both of these shortcomings have solutions that shouldn't explode in complexity, so for the time being, it should be fine.

Fixes #322.

### DIFF
--- a/src/mlir/ast-utils.h
+++ b/src/mlir/ast-utils.h
@@ -386,7 +386,7 @@ namespace mlir::verona
       return getTokenValue(findNode(ast, NodeKind::Localref));
     }
 
-    /// Return true if node is a variable definition
+    /// Get the assignment variable name (new local or localref)
     static llvm::StringRef getLocalName(::ast::Ast ast)
     {
       // Local variables can be new 'local' or existing 'localref'
@@ -398,13 +398,15 @@ namespace mlir::verona
       return getTokenValue(findNode(ast, NodeKind::Local));
     }
 
-    /// Return true if node is an ID (func, var, type names)
+    /// Get the token value of the ID (func, var, type names, localref, etc.)
     static llvm::StringRef getID(::ast::Ast ast)
     {
       if (isAny(ast, {NodeKind::Call, NodeKind::StaticCall}))
         return getTokenValue(findNode(ast, NodeKind::Function));
       if (isMember(ast))
         return getTokenValue(findNode(ast, NodeKind::Lookup));
+      if (isLet(ast))
+        return getTokenValue(findNode(ast, NodeKind::Local));
       if (isQualType(ast))
         return getTokenValue(findNode(ast, NodeKind::Typeref));
       if (isInvoke(ast))
--- a/src/mlir/ast-utils.h
+++ b/src/mlir/ast-utils.h
@@ -357,6 +357,16 @@ namespace mlir::verona
       nodes.insert(nodes.end(), ast->nodes.begin(), ast->nodes.end());
     }
 
+    /// Return the sequence of a block ast, in case we don't want to
+    /// push to the variable scope (when it needs to exist for longer).
+    /// If not a block, just return the ast.
+    static ::ast::Ast skipBlock(::ast::Ast ast)
+    {
+      if (isBlock(ast))
+        return findNode(ast, NodeKind::Seq);
+      return ast;
+    }
+
     // ================================================= Value Helpers
     /// Get the string value of a token
     static llvm::StringRef getTokenValue(::ast::Ast ast)
@@ -662,7 +672,7 @@ namespace mlir::verona
     }
 
     // ================================================= Condition Helpers
-    /// Return the condition from an if statement
+    /// Return true if the 'if' node has an else block
     static bool hasElse(::ast::Ast ast)
     {
       // Else nodes always exist inside `if` nodes, but if there was no `else`
@@ -671,7 +681,7 @@ namespace mlir::verona
         hasSubs(findNode(ast, NodeKind::Else));
     }
 
-    /// Return the block from an if statement
+    /// Return the condition from an if statement
     static ::ast::Ast getCond(::ast::Ast ast)
     {
       // These are the nodes that could have conditions as subnodes
@@ -681,7 +691,7 @@ namespace mlir::verona
       return findNode(cond, NodeKind::Seq);
     }
 
-    /// Return true if the 'if' node has an else block
+    /// Return the if block from an if statement
     static ::ast::Ast getIfBlock(::ast::Ast ast)
     {
       assert(isIf(ast) && "Bad node");
--- a/src/mlir/ast-utils.h
+++ b/src/mlir/ast-utils.h
@@ -91,7 +91,6 @@ namespace mlir::verona
       If = peg::str2tag("if"), // = 3567
       Else = peg::str2tag("else"), // = 3742303
       While = peg::str2tag("while"), // = 140358143
-      For = peg::str2tag("for"), // = 112155
       Continue = peg::str2tag("continue"), // = 2929012833
       Break = peg::str2tag("break"), // = 117842911
       New = peg::str2tag("new"), // = 120796
@@ -293,22 +292,10 @@ namespace mlir::verona
       return isA(ast, NodeKind::Else);
     }
 
-    /// Return true if node is a while loop
-    static bool isWhile(::ast::Ast ast)
-    {
-      return isA(ast, NodeKind::While);
-    }
-
-    /// Return true if node is a for loop
-    static bool isFor(::ast::Ast ast)
-    {
-      return isA(ast, NodeKind::For);
-    }
-
     /// Return true if node is any kind of loop
     static bool isLoop(::ast::Ast ast)
     {
-      return isWhile(ast) || isFor(ast);
+      return isA(ast, NodeKind::While);
     }
 
     /// Return true if node is a loop continue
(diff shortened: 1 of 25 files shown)


plietar:
```ts
(x: U64): U64 {
  while (true) {
    x = x.m();
  }
  return x;
}
```

Is giving me the following IR:

```mlir
  func @apply(%arg0: !verona.class<"U64">) -> !verona.class<"U64"> attributes {class = !verona.class<"$module">} {
    %0 = "verona.cast"(%arg0) : (!verona.class<"U64">) -> !verona.unknown
    br ^bb1(%0 : !verona.unknown)

  ^bb1(%1: !verona.unknown):  // 2 preds: ^bb0, ^bb2
    %2 = "verona.constant(true)"() : () -> !verona.class<"bool">
    %3 = "verona.cast"(%2) : (!verona.class<"bool">) -> i1
    cond_br %3, ^bb2(%1 : !verona.unknown), ^bb3(%1 : !verona.unknown)

  ^bb2(%4: !verona.unknown):  // pred: ^bb1
    %5 = verona.call "m"[%4 : !verona.unknown] ( : ) : !verona.unknown

    /// ----- This looks incorrect -----
    br ^bb1(%4 : !verona.unknown)

  ^bb3(%6: !verona.unknown):  // pred: ^bb1
    %7 = "verona.cast"(%6) : (!verona.unknown) -> !verona.class<"U64">
    return %7 : !verona.class<"U64">
  }
```

In `^bb2`, the terminator `br ^bb1(%4)` should be passing `%5`, not `%4`. Otherwise the loop with be executing with the same value everytime.

plietar:
In an assignment `let y = x`, the generator re-uses the same SSA variable for x and y. I think this is going to cause issues, or at least confusion, when it comes to linearity:
```ts
// x: iso
let y = x; // This should move x into y
use(x); // This should not be allowed
```
versus
```ts
// x: iso
let y = x; // This should move x into y
use(y); // This should be allowed
```

However in your implementation, since `x` and `y` are assigned to the same SSA value, the two `use` calls are indistinguishable from each other. This somewhat works in this particular case, where you treat x and y as synonymous, and either `use` call is allowed, but it breaks down for more complicated cases:
```
use(a: iso, b: iso) { .... }

let x = ...;
let y0 = ...;
if true {
   y1 = x;
}
y2 = φ(y0, y1)
use(x, y2);
```

This program must not be allowed, since it duplicates the iso value. I'm not sure how do we ban it without making copies effectful, ie. `y1 = x` is represented in the IR and actually consumes x.

rengolin:
> This is not just an efficiency problem: A large segment of our region analysis will be about requring certain properties about live variables. For instance consider `x` to be an `iso`, and `use(x)` actually consumed it. You shouldn't be allowed to use x again (ie. it is not live after the `use(x)`), but the SSA program does.

This kind of lifetime analysis isn't supported yet at all. All variables are (wrongly) assumed to live through their lexical contexts. This is not a problem with this commit, but with the current state of the IR at the moment.

But lifetime is orthogonal to block arguments.

If that `x` is the only change in the `if`, we still need the PHI node because if the branch is taken, `x` is used and dead, if not taken, `x` can still be used and is alive. We must be able to encode that.

Due to a requirement in the MLIR structure, the basic block MUST have the same number (and type) of operands and arguments, regardless of the lifetime of variables. So, if `x` dies there and we still must pass it to the exit block, we need to pass _something_.

A simple solution would be to update the value of the symbol in the table to some dead value (that can't be used and means mead), and then the branch operation lowering is still exactly the same: take the value of the symbol from the table, use as operand.

rengolin:
> In `^bb2`, the terminator `br ^bb1(%4)` should be passing `%5`, not `%4`. Otherwise the loop would be executing with the same value everytime.

That is a bug, one that I thought I had fixed. I'll have a look and add your examples to the list of tests. Thanks!

rengolin:
> This program must not be allowed, since it duplicates the iso value. I'm not sure how do we ban it without making copies effectful, ie. `y1 = x` is represented in the IR and actually consumes x.

The first program looks fine to me, that's what assignments normally mean. If we mean something different, than we need special operations for that.

The second program looks like it's just wrong and we can infer that statically and emit an error.

In all your comments, you seem to be implying that the IR must express perfect semantics from step 1 and there can be no misunderstanding or ambiguity at any level. But we have unknown types and a lot of operations that we don't even know are possible.

Some errors will be caught by the AST, others by the MLIR generation, others by the MLIR lowering and others only by LLVM lowering and codegen. That's why we propagate line information all the way down, so that we can always report good error messages to users.

This is just one step in that long line of transformations that each will discover new errors, warnings as well as avenues for optimisation. As long as we eventually catch the error, and lifetime issues are related but mostly orthogonal to how we propagate values through the IR (load/store, PHI nodes, block arguments), it should be fine to have an incomplete picture at this high level.

plietar:
> The first program looks fine to me, that's what assignments normally mean.

I don’t know think “normal” means anything.
In Rust, [an equivalent program](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=12b26ed93e33f0c9be5b0dcb01080f2d) will fail to compile with an error along the lines "use of moved value: `x`". It is pretty "normal" for languages to treat assignemnts as consuming the right-hand side, even if it does not seem strictly necessary.

What I’m suggesting is that “y = x” needs to be lowered as “%y = verona.copy %x”. If we don’t do this, we will have lost information that we cannot recover later.

> The second program looks like it's just wrong and we can infer that statically and emit an error.

It is wrong, but we will have a very hard time detecting that if we don’t encode copies.

> In all your comments, you seem to be implying that the IR must express perfect semantics from step 1 and there can be no misunderstanding or ambiguity at any level. But we have unknown types and a lot of operations that we don't even know are possible.

Maybe I wasn’t clear. I’m not saying at all that these invalid programs be caught during this step. What I’m saying is the IR needs to encode enough information for later passes to do so. With the transform as it is in the PR, this is not the case

sylvanc:
@plietar is, in my opinion, correct here.

Assignment in Verona is region-aware. That means two things:

1. Assignment must return a value, and that value must be the _previous_ value of the left-hand side. This is because the previous value of the left-hand side may be an `iso` or an `imm`, which means there is a discharge obligation. If an `iso` result of an assignment is not used, is must be dropped. If an `imm` result of an assignment is not used, it's refcount must be decremented.
2. Assignment creates a new binding for the right-hand side. This may be in the form of a new alias (if the left-hand side is local variable) or a new path (if the left hand side is a path such as `y.f = x`). This means that if the right-hand side is an `iso`, the right-hand side _must not continue to exist_. It must be consumed. Otherwise there would be more than one entry-point to the region, which is not allowed.

This is not a special operation. This is the fundamental semantics of assignment in a region-aware language.

If we write:

```ts
let y = x;
use(y);
use(x);
```

We must have an IR that allows use to determine that this program is incorrect. We are doing type inference and type checking on the IR, so all of this information must be available (that is, we must be able to distinguish `x` from `y`). If we can't do that, we will have to do type inference and type checking on the AST.

sylvanc:
Looks like @plietar and I wrote basically the same thing :) Bike-shedding: I would prefer `%y = verona.assign %x`, as `copy` seems to indicate some deeper operation to me, but... that's just bike-shedding.

rengolin:
> Maybe I wasn’t clear. I’m not saying at all that these invalid programs be caught during this step. What I’m saying is the IR needs to encode enough information for later passes to do so. With the transform as it is in the PR, this is not the case

Right, and this is about adding new operations (like `verona.copy`, which should be done in a different PR) and not about how do we propagate values across basic blocks.

rengolin:
> @plietar is, in my opinion, correct here.

I did not disagree with him at all.

My usage of `normal` and `special` was about `standard` and `Verona` dialects. There is no equivalent operation in the standard dialects that encode assignment like Verona, thus we need a new dialect operation.

My point was only that this isn't supported at all in the current version of the IR because we haven't got there yet. When we do, we'll still have to track lifetime of variables (through all branches of the CFG) and we'll still have to send those as basic block arguments, so that issue is largely orthogonal to this PR.

We may have to change some things in this implementation in the light of the new constraints, but we can't avoid basic block arguments and building on top of a simplified implementation is better than trying to implement two (or more) complex changes at the same time.

rengolin:
@plietar, thanks for the review, your check found a serious problem with the lowering that was lingering for a while. Basically, I was considering every `block` as a new scope, but on conditions and loops I had to create the scope manually to update the block arguments from the right scope. Turns out I was popping the updated scope too early and the values we reverted to the previous scope.

Reviewing giant IR diffs isn't easy and I missed that. Now I think I got it right, please have a second check. I've also added a test like your example in `conditional.verona`. 

Thanks!

plietar:
> Right, and this is about adding new operations (like verona.copy, which should be done in a different PR)

The operation already exists in the dialect.

rengolin:
Abandoning this PR as we'll move to a different implementation.

PR #356, opened 2020-11-19
Fix a typo in the tidy-protect test comments.

plietar:
The comment was incorrectly stating that o2 is live, while
simultaneously stating that it is unreachable. This was a simple typo:
o2 is indeed unreachable, whereas o5 is live. The program itself and its
assertions are correct.


mjp41 (approved):
LGTM

PR #357, opened 2020-11-23
Document the IR.

plietar:


src/compiler/ir/README.md
```diff
@@ -0,0 +1,162 @@
...
+
+## `EndScopeStmt`
+
+This statement is inserted whenever variables go out of scope. This may not be
+the end of a BasicBlock, if a block expression is used for example.
+
+This is where a value's "release" code should be executed. This could be
+`rt::Region::release`, `rt::Cown::release` or `rt::Immutable::release`, for
+owned, cown and immutable values, respectively. Unowned values do not need to
+be released.
```
rengolin, line 124: This is just related to lexical blocks, not liveness analysis, right?

src/compiler/ir/README.md
```diff
@@ -0,0 +1,162 @@
...
+
+```
+0 <- new A
+1 <- new B
+overwrite(0)
+// ...
+end-scope(1)
+```
+
+The local `x` is originally assigned the SSA variable `0`, and later `1`. After
```
rengolin, line 141: Why isn't this just `end-scope(0)`? If the variable `x` was updated in the symbol table to point to `1`, then `0` is dangling and `end-scope` would just free it.

src/compiler/ir/README.md
```diff
@@ -0,0 +1,162 @@
...
+## Match Terminator
+
+# Builtins
+
+A number of language features don't need special syntax, but are instead
+exposed as builtin methods, defined in `stdlib/builtin.verona`. This includes
+arithmetic operations and some region operations such as freeze and trace.
+
+The language features they represent isn't even expressible in the IR. Instead,
+`codegen/builtins.cc` provides hardcoded lowerings into bytecode.
```
rengolin, line 161: IIUC, we'll still need to carry those as calls to functions, ex. `U32.add(x, y)` in MLIR, and then have a special lowering to LLVM that picks the codegen alternative and either call the intrinsic or lower to a native operation, right?

rengolin (commented):
Thanks for this document, it really helps me understand the low level semantics of the language, many of which I was still missing.

I have added a few inline comments that are most for my own education. Some are related to MLIR lowering, but others may be ambiguity on the document and could lead to adding a bit more context.

But otherwise, looking great, thanks!

src/compiler/ir/README.md
```diff
@@ -0,0 +1,162 @@
...
+occasionally refers to "phi nodes", the IR uses basic block arguments rather
+than phi nodes.
+
+The IR only represents the body of methods. The "shape" of the program, eg. its
+classes, fields, method names and signatures, ... are all stored in the AST
+instead. For each `Method` node in the AST, a corresponding `FnAnalysis` is
+created, which, among others, contains the `MethodIR`.
+
+A `MethodIR` is composed of one or more `FunctionIR`. This distinction is used
+to encode the bodies of `when` clauses that appear within the method. The first
```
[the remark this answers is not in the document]
plietar: Lambdas aren't implemented yet. I'm still undecided how they should be represented.

In theory both lambdas and when clauses lower to an anonymous class. However `when` clauses are "easier" to implement because the program never receives a reference to that object, therefore we don't need to give when clauses any type, which made it very easy to implement. Lambdas will need more thinking.

src/compiler/ir/README.md
```diff
@@ -0,0 +1,162 @@
...
+
+This will correspond to either a `rt::RegionTrace::create` or
+`rt::RegionTrace::alloc` call, dependending on whether a new region needs to be
+allocated.
+
+The `parent` specified in the statement could be any object in the target
+region; it may not be a reference to the region's entrypoint. On the other
+hand, the runtime's `RegionTrace::alloc` method expects to be given the
+region's entrypoint. The compiler therefore needs to do some work to find the
+entrypoint given the specified parent.
```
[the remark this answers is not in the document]
plietar: When allocating an object you always need the region, because there's a bunch of metadata the runtime needs to update.

What the VM does is that every object has a pointer to its region. Whenever we want to get the region of a reference, we look up this pointer. Because regions can be merged, and we don't want to update every object to point to the new region, [this is actually a union-find](https://github.com/microsoft/verona/blob/322dc03dfbfcbf34a2de3d47ae337cdfe07fa88d/src/interpreter/object.cc#L50-L61), which is very similar to your tree idea.

However this wastes 8 bytes on every object, so it's very probably not what we want to do in the real backend, and instead have "shadow" references for every reference (and hope copy elision removes most of these). We could still do the extra pointer in the object as an initial version.

I think @mjp41 was planning on writing a more detailed section about this.

src/compiler/ir/README.md
```diff
@@ -0,0 +1,162 @@
...
+In order to dispatch the call, in the general case, the selector should be used
+to index into the object's vtable, to retrieve the address of the target.
+Optimizations may enable us to dispatch the call statically, using for example,
+the type of the receiver.
+
+The method being called could be static or not (ie. does it accept a `self`
+parameter, not to be confused with the dispatch strategy). In either case, the
+receiver of the call is always passed as the first argument to the method.  In
+the case of a static method call, the receiver may not be an actual object, but
+only a pointer to a descriptor, as created by a `StaticTypeStmt`.
```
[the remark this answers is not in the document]
plietar: Yeah, because I pointed you towards this IR when adding static calls to the MLIR dialect 😀

src/compiler/ir/README.md
```diff
@@ -0,0 +1,162 @@
...
+The closure's body will accept as many parameters as there are cowns and
+captures specified. When invoking the body, each cown argument should be
+transformed into a reference to its contents, before being passed as an
+argument. The captures should be passed as is.
+
+## `ReadFieldStmt`
+
+Read a field from an object. The field is identified by its name. The general
+case is similar to method calls: the field's selector is used to look up the
+field's offset in the object's vtable.
```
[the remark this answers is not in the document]
plietar: Two parts of the same vtable, or two vtables per class, depends on how you look at it 😁 

I don't know yet whether at the surface syntax we want to use the same namespace for methods and fields, but I think having separate vtables allows us to have field entries that are 32-bit (4GB objects ought to be enough for anyone), whereas the method vtable entries need to be full pointers.

On the other hand, we could also lower field access into a pair of accesor methods, and not have to deal with this at all (this has a bunch of other advantages too, and I think we'll have to do it anyway).

src/compiler/ir/README.md
```diff
@@ -0,0 +1,162 @@
...
+Otherwise, the reference count should be incremented.
+
+## `IntegerLiteralStmt`, `StringLiteralStmt`, `UnitStmt`
+
+These create values from a literal. The support for strings is very special
+cased for demonstration purposes.
+
+## `MatchBindStmt`
+
+This is an unchecked cast, used at the top of the target of a match statement.
```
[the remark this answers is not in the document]
plietar: Sorry I was gonna write more about matches, haven't gotten to it yet.

In short, your example lowers to
```
BB0:
  x : Type1 | Type2 | Type3
  match x {
    Type1 => BB1
    Type2 => BB2
    Type3 => BB3
  }

BB1:
  a: Type1 = bind x, Type1
  ...

BB2:
  b: Type2 = bind x, Type2
  ...


BB3:
  c: Type3 = bind x, Type3
  ...
```

x only ever has type `Type1 | Type2 | Type3` (in general an SSA variable only ever has one type).
The match terminator will dispatch to the correct branch. It only affects control flow, not the type of variables nor does it introduce any variable. Every branch uses match bind to get the right type.

Doing a "match bind" with the wrong type is UB, but the compiler will never produce that.

In theory, match statements are required to be exahustive (ie. at least one pattern must match), but that check isn't implemented yet, which means the compiler could generate a `match` terminator where none of branches are taken. The bytecode backend introduces an abort instruction if all the patterns fail. We probably want to do the same in the MLIR backend until the typechecker is fixed to check this.

src/compiler/ir/README.md
```diff
@@ -0,0 +1,162 @@
...
+
+## `ViewStmt`
+
+If the argument to the statement is an owned value, the statement produces an
+unowned reference to the same object. Otherwise this is equivalent to a copy.
```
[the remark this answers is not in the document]
plietar: Yeah, I meant equivalent to CopyStmt, which I now realize isn't explained anywhere.

src/compiler/ir/README.md
plietar: Correct. Liveness isn't encoded in the IR, although it can be accessed in `analysis->liveness` if you need it.

src/compiler/ir/README.md
plietar: Meh, `end-scope` and `overwrite` basically have the same semantics, but we want different error messages out of them.

src/compiler/ir/README.md
plietar: So in my opinion, `x + y` is always lowered as a method call (`x.add(y)`), all the way through to LLVM. We also generate an add method which is lowered using an intrinsic / native operation. Inlining is responsible for replacing the method call with the native operation, not the lowering.

If we find out that inlining is too unreliable / extra work, we can do the lowering to native ops at a later point. But I don't see the point in doing that yet.

src/compiler/ir/README.md
```diff
@@ -0,0 +1,162 @@
...
+
+The IR only represents the body of methods. The "shape" of the program, eg. its
+classes, fields, method names and signatures, ... are all stored in the AST
+instead. For each `Method` node in the AST, a corresponding `FnAnalysis` is
+created, which, among others, contains the `MethodIR`.
+
+A `MethodIR` is composed of one or more `FunctionIR`. This distinction is used
+to encode the bodies of `when` clauses that appear within the method. The first
+`FunctionIR` corresponds to the method body itself, whereas subsequent blocks
+correspond to when block bodies.
```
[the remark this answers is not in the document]
plietar: Will do.

src/compiler/ir/README.md
```diff
@@ -0,0 +1,162 @@
...
+In both cases however, these are the polymorphic types. If a method or class is
+generic, these types may mention type variables. The non-generic types will
+only be known after reachabilitiy and reification.
+
+# Statements
+## `NewStmt`
+
+Allocates a new object of a given class. It may optionally be given a `parent`
+which designates the region in which the object should be allocated. If no
+parent is specified, the object is allocated in a new region.
```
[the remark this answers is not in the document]
plietar: It depends whether we consider this an optimisation or an inference. In the former case, programs would be correct without that pass, and I'm happy for this to happen in MLIR (it would be a "merge elision" pass). If it's an inference pass, programs can't type check without that pass, so we.

The truth is we'll need both. Region inference is necessary to get certain programs to compile, but "merge elision" is more powerful (it can take advantage of inlining and other generic optimizations, for example).

src/compiler/ir/README.md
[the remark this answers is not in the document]
plietar: There's other ways we can represent this btw, and I'm considering taking a more MLIR-y approach, taking advantage of block parameters:

```
BB00:
  match x {
    Type1: BB1
    Type2: BB2
    Type3: BB3
  }

BB1(a: Type1): ...
BB2(b: Type2): ...
BB3(c: Type3): ...
```

This removes the implicit invariant that exists between the match terminator and the "match bind" statement.

The existing implementation matches Rust's IR ([SwitchInt terminator](https://doc.rust-lang.org/nightly/nightly-rustc/rustc_middle/mir/terminator/enum.TerminatorKind.html#variant.SwitchInt) and [the downcast operation](https://doc.rust-lang.org/nightly/nightly-rustc/rustc_middle/mir/enum.ProjectionElem.html#variant.Downcast)), the basic block parameter style [matches Swift's](https://github.com/apple/swift/blob/main/docs/SIL.rst#switch-enum).

src/compiler/ir/README.md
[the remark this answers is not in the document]
plietar: Btw I'm thinking we should rename `FunctionIR` since I find the naming confusing (method and function are used interchangebly almost everywhere else).

What about `IRBody`?

src/compiler/ir/README.md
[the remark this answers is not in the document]
rengolin: Right, this is how I thought the high-level would be like. 

Part of the question was about the failure recovery of `match` if there is no match, and you answered that: it won't.

The other part was the low level implementation of how we `cast`, but that can be a runtime implementation that doesn't need to be visible to (ie. lowered by) the compiler.

src/compiler/ir/README.md
rengolin: Fair enough.

src/compiler/ir/README.md
rengolin: I think Sylvan said this was an issue with Pony, that the LLVM inliner was unreliable and that's why they lowered to intrinsic/op directly into LLVM. 

One thing that could be a problem is if we lower calls with more than just `call` (like lowered vtable access). In that case, `x.add(y)` would actually be a sequence of instructions which culminates in a dynamic call (call on the offset of the vtable), and LLVM would have no idea what function is actually being called.

The simple case of a static call (like `U32.add(x,y)`) would be converted into a call to a mangled function, available in the library and lowered in the same LLVM IR, and then the inliner could spot that and inline. But we'd still need to convert from `x.add(y)` to `TypeOfX.add(x, y)` before lowering to LLVM.

Makes sense?

src/compiler/ir/README.md
```diff
@@ -101,7 +101,122 @@ This could be refined by considering, which classes could reach `i`.  For instan
...
+
+
+Aside:  We do not have to worry about allocating on a null region
+variable. The front-end will perform sufficient work such that
+```
+  x = new C in y
+```
+will guarantee that `y` is not an `iso` by adding coercions to
+`mut` such that all accesses in a function will determine they 
+are part of the same region
```
plietar, line 213: This isn’t actually true today. That code, with `y: iso` will pass the typechecker. Of course we can change that in the future.

src/compiler/ir/README.md
```diff
@@ -101,7 +101,122 @@ This could be refined by considering, which classes could reach `i`.  For instan
...
+```
+y = x.f
+``` 
+we need translate this to
+```
+y = // Read field described below.
+if (y->rt::Object::get_class() == RegionMD::ISO)
+  y_region = y
+else 
+  y_region = x_region
```
plietar, line 127: This needs x_region to not be null as well, just like the allocation stuff further down.

Why did you go with the region is null, rather than tagging the low bits as you’d mention previously?

mjp41: I like the idea of using the front-end to separate out allocation on an `iso` reference from `mut`, and then this seemed cleaner.

We can put the alternative in.

src/compiler/ir/README.md
mjp41: Yeah, sorry that was more how I was thinking it should be, not how it is.

rengolin:
Just noticed, this PR has an LLVM update. Probably a left-over of some other branch, but we need to clear that before merge.

src/compiler/ir/README.md
```diff
@@ -0,0 +1,522 @@
...
+`end-scope` effect should only release the value was not taken.
+
+```
+BB0:
+  %x = ...
+  %y = ... // %y holds an owned value
+  if %x then goto BB1 else goto BB2
+
+BB1:
+  call foo(%y) // Consuming use of %y, its value is moved
```
rengolin, line 377: I would have thought the `end-scope(%y)` would be here, right after use, not after the PHI node in the tail block. This would get rid of the conundrum, no?

src/compiler/ir/README.md
```diff
@@ -0,0 +1,522 @@
...
+  goto BB2
+
+BB2:
+  // %y has been moved conditionally, we need to release it only if the
+  // branch was not taken.
+  end-scope(%y) 
+```
```
rengolin, line 384: In the first example, you state: "if a variable is passed as basic block argument, it may not be used again".

In this example, you don't pass `%y` to BB1 but you do use it. 

Does that mean you have both dominance use and block arguments and you only use arguments in some special condition?

src/compiler/ir/README.md
plietar: `end-scope` represents the point in the surface language where a variable goes out of scope, ie. the block in which that variable was declared in closed.

In this particular example, nothing happens between the call and the end-scope. However a more complex example:

```pony
{
  var y = ...;
  if * {
    foo(y); // y is conditionally moved
  }
  bar();
  // y goes out of scope
}
```

the `end-scope` cannot be simply moved into the conditional block.

src/compiler/ir/README.md
plietar: > Does that mean you have both dominance use and block arguments

Yes, but you cannot have both at once. What I meant by "if a variable is passed as basic block argument, it may not be used again" is this:

```
BB1:
  // define %x, %y
  goto BB2(%y)

BB2(%z):
  // %x may be used, since BB1 dominates BB2
  // %y may not be used since there has already been a destructive use on the path, ie. the BB2(%y) line.

  // %x and %z must have a terminating use at some point
  end-scope(%x, %z)
```

> you only use arguments in some special condition?
I don't understand what this means.

src/compiler/ir/README.md
plietar: The issue is that this will not always be a statically dispatched call. If you consider the slightly simpler `-x` expression, it could be that `x` has type `Neg`, where `Neg` is defined as :
```
interface Neg {
  neg(self: imm): Neg & imm
}
```

Now the expression `-x` (which is really `x.neg()`) cannot be dispatched to a specific method but instead does the usual vtable lookup and dynamic dispatch stuff. But I32 has this `neg` method, therefore there must actually exist a vtable for `I32`, which has a pointer to the body of the method. That body cannot be implemented in Verona code and must be some compiler magic eg. in the bytecode backend this is the `builtins.cc` file that directly emits bytecode instructions that have no equivalent in the IR.

Since the code path for treating arithmetic operations as functions, and lowering of the function bodies, must exist for this purpose, we can always use it and 1) hope LLVM inlines it in the concrete arithmetic type case 2) do something better ourselves at a later point if this is a problem.

src/compiler/ir/README.md
rengolin: I see. You use basic block arguments as basically killing the variable elsewhere. This is not enforced by MLIR (or LLVM IR), but it should be fine because in the 2IR I will assume variable use and scope have already been checked and are safe.

Passing an argument to a block already updates the symbol table for that binding, so no additional work should be needed.

src/compiler/ir/README.md
rengolin: Right, I consider this to be a language definition problem. 

If we guarantee native type arithmetic will always be lowered to machine instructions on registers, we need to build interface/library/lowering in a way that allows us to achieve this without complex hacks in the compiler.

Relying on the compiler matching functions by name on "known" exceptions can lead to horrible and poorly maintained code. 

Perhaps we could add a `native` keyword that means "this is implemented outside of the language" or something like this, so that at least we know what must and what must not be done in that way and develop the library around that concept.

plietar: Yes that’s what the `builtin.verona` file already does. A number of methods in there are tagged with a `builtin` keyword.

docs/internal/1IR.md
```diff
@@ -0,0 +1,350 @@
...
+The local `x` is originally assigned the SSA variable `0`, and later `1`. After
+the assignement, the `overwrite(0)` statement encodes the fact that the old
+value needs to be released.
+
+Note that an SSA value that has been overwritten is never passed to an
+"end-scope" statement. Only `x`'s new value, `1`, goes out of scope.
+
+## Terminators
+
+### Branch Terminator
```
rengolin, line 338: Are these the last topics to cover? If so, we should finish this soon, so we can start working on the code to read this.

rengolin:
Hey, I know these are incomplete, but can we just merge this with a TODO on the missing parts and fill them up later?

There's no point in keeping a documentation PR waiting for more content, when we can add it on following PRs.

rengolin (approved):
I have rebased and added a TODO on the terminators, the only part that was missing, I think.

PR #360, opened 2020-12-03
Support matching with an interface type.

plietar:
For every interface, codegen emits the list of subtypes. During
matching, rather than requiring strict equality of descriptors, we
search through the subtypes of the match type looking for the
descriptor of the object being matched on.

### DIFF
--- a/src/compiler/codegen/descriptor.cc
+++ b/src/compiler/codegen/descriptor.cc
@@ -46,7 +46,7 @@ namespace verona::compiler
           break;
 
         case Entity::Interface:
-          emit_interface_descriptor(entity);
+          emit_interface_descriptor(entity, info);
           break;
 
           EXHAUSTIVE_SWITCH;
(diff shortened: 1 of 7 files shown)


mjp41 (approved):
I think this is fine for making progress at the moment.  Long term we should integrate selector colouring and interface matching, so this does not require a search.  

My current thoughts are to store an entry in the type for `is_an_I` and then propagate on to all the types that can reach the match site:
```
   /// x : A | B | C | D
  match x {
      var _: A => Builtin.print("A\n"),
      var _: I => Builtin.print("I\n"),
      var _: B => Builtin.print("B\n"),
    }
```
This would generate the `is_an_I` on `A`, `B`, `C` and `D`.  With the `Any` type in your example this would mean all possible types would need the `is_an_I` method (without global analysis).

If we combine this with allowing constant values in the descriptor table (not just methods)
```
  class A
  {
     constant U64 class_field = 5;
  }
```
Then the `is_an_I` could then be a class constant. If we then look at allowing those to be bit sized, then we might make this fairly efficient.

plietar:
Yeah I agree we can do better. This was just a quick fix to unblock Luke.

If you are allowed to use match in order to perform a downcast, eg. `Any -> I`, then in the presence of such a match `is_an_A` would have to appear in every class too.

The advantage of encoding the actual list of subtypes (we could alternatively do supertypes too, it just happens that subtypes were already computed) is that we only consume space when the subtyping relation holds.

PR #361, opened 2020-12-03
NFC: Refactoring header includes.

mjp41:
This is reducing the dependence on compiler/ast.h as ground work
for switching to the new parser.


plietar (approved):
I'd rather `typevisitor.h` be called `type_visitor.h`. Otherwise LGTM.

PR #362, opened 2020-12-03
Widen the set of types that may be used in pattern matching

plietar:
This adds support for matching on primitives (cown, string, U64), capabilities, and union/intersection types.

This is done by:
- Turning String into a primitive, rather than being hardcoded in the compiler
- Add a `MatchCapability` opcode to the VM. The existing `Match` opcode is renamed to `MatchDescriptor`.
- Use a `TypeVisitor` to explore the match pattern, and using boolean operators to decompose complex types.


mjp41 (approved):
LGTM

PR #363, opened 2020-12-07
Add initial Verona web page.

davidchisnall:
This should be automatically rendered by GitHub Pages at:
https://microsoft.github.io/verona

You can test it locally by installing the github-pages gem and running
`jekyll serve` in the docs directory.

### DIFF
--- a/docs/_layouts/default.html
+++ b/docs/_layouts/default.html
@@ -1,0 +1,64 @@
+{% comment %}Copied from the slate them to add a navbar{% endcomment %}
+<!DOCTYPE html>
+<html lang="{{ site.lang | default: "en-US" }}">
+
+  <head>
+    <meta charset='utf-8'>
+    <meta http-equiv="X-UA-Compatible" content="IE=edge">
+    <meta name="viewport" content="width=device-width,maximum-scale=2">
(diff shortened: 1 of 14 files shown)


docs/_layouts/default.html
```diff
@@ -0,0 +1,64 @@
...
+    <div id="footer_wrap" class="outer">
+      <footer class="inner">
+        {% if site.github.is_project_page %}
+        <p class="copyright">{{ site.title | default: site.github.repository_name }} maintained by <a href="{{ site.github.owner_url }}">{{ site.github.owner_name }}</a></p>
+        {% endif %}
+        <p>Published with <a href="https://pages.github.com">GitHub Pages</a></p>
+      </footer>
+    </div>
+
+    {% if site.google_analytics %}
```
rengolin, line 53: Will we ever use this?

rengolin (commented):
We probably also want to change the readme file to point to this page, as well as remove the internal links it has to the same pages.

mjp41 (approved):
Looks good as far as I can tell.

docs/_layouts/default.html
davidchisnall: No, I just wanted to minimise diffs with the upstream version of this file.

PR #364, opened 2020-12-07
Add a better DSL for emitting instructions. 

plietar:
The emit method's signature is deduced from the same opcode spec which
drives instruction decoding. This ensures instructions are written in
a format which is consistent with.

~~This PR depends on #362 because they conflict a little, so it was easier for me to develop them this way rather than fix the conflict later. The last commit (c93b02a), is the only that is relevant to this particular PR.~~


mjp41 (approved):
Looks like a nice tidying.

PR #367, opened 2020-12-08
Documenting new low-level IR

rengolin:
This is the document that describes what the first version if the 2IR will look like, which is very similar to what we had before.

There will be two other documents, one that describes the optimisations we want to do and the last one that will describe how we lower to Std/LLVM dialect.

None of them will be complete and accurate, just ball park estimations to make sure there is consensus on the high-level ideas. the details should be hashed on the implementation PRs.

This is still in draft and is based on #357


rengolin:
This is a simpler view of the document:
https://github.com/rengolin/verona/blob/new-mlir/src/mlir/lowering.md

rengolin:
Ok, so I've gone through the 1IR document and tried to map 1:1 the concepts exposed there. I have not updated it to match the new "IR Invariants" section, but the statement that it's irrelevant to lowering perhaps means I don't need to. I'll read that section and make sure what I wrote matches the expectation, but I think this document could already start to be reviewed for its own content.

A warning though: I am making lots of assumptions on semantics! I tried to be as explicit as possible as to what it will be lowered to, so people should look at those and see if it matches their understanding of the semantics.

However, good part of it was borrowed from the existing MLIR dialect, so at least the bulk has been reviewed alread.

rengolin:
New passes document:
https://github.com/rengolin/verona/blob/new-mlir/src/mlir/passes.md

rengolin:
New LLVM document:
https://github.com/rengolin/verona/blob/new-mlir/src/mlir/llvm.md

src/mlir/llvm.md
```diff
@@ -0,0 +1,178 @@
...
+### when
+
+`when` statements have already been converted to a sequence of creating the
+cown and calling the schedule runtime, which later on (see above) will be
+converted to standard calls, so again, nothing special here.
+
+### field_read / field_write
+
+All types at this stage are complete and concrete, which means reading/writing
+from/to fields in LLVM are just direct access to struct fields (ie. via GEP).
```
plietar, line 119: We allow reading and writing fields in a union or interface type, which means we may not know the offset of the field. In this case, we need to look up the offset in the vtable (similar to a method call).

src/mlir/llvm.md
```diff
@@ -0,0 +1,178 @@
...
+
+### numeric literals
+
+All numeric literals, including compiler generated ones, can be represented
+directly in LLVM. However, in MLIR, they have been lowered as an `imm` anonymous
+class, wich access control (ex. reference count) and thus calls to the runtime
+library or lowered directly as code.
```
plietar, line 126: I'm not sure I understand this section. Numeric literals are always just an llvm constant. No need for an anonymous class or reference counting or anything.

src/mlir/llvm.md
```diff
@@ -0,0 +1,178 @@
...
+
+However, struct fields can only be valid LLVM types, so we need to convert all
+Verona types into LLVM types and appropriate functions to handle the semantics.
+
+For example:
+ * Class types are just LLVM struct types, recursively. To avoid declaration
+   order issues, a type can be forward-declared as `opaque`, and then fully
+   defined later.
+ * Pointer/reference types need to be pointer types in LLVM, to some other
+   struct types. There may be some extra handling of the pointers (for instance
```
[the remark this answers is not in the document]
rengolin: `void*` in LLVM is really messy. We need to keep the layout in an array and then lower the access with ABI-correct GEPs (right offset / size / alignment) and it's not trivial. I'd only use this if I had no other choice. Given we do have information on all concrete classes, and GEPs work well on them, I don't think it's a good idea to just flatten everything into a `void*`.

We may have to do that for the cases we don't know (or haven't specialised / versioned for specific concrete types), but only for those cases.

src/mlir/llvm.md
rengolin: In theory, union/interface field access constructs will not survive the MLIR pass, just like method calls, and will be replaced by vtable offset + load/store, and those lower directly. Whatever's left is on concrete types.

src/mlir/llvm.md
rengolin: Sorry, this is completely wrong. It's based on my previous interpretation that was wrong and I changed recently but somehow my brain didn't commit. I'll change the wording, thanks!

rengolin: updated the text, hope it makes more sense now.

src/mlir/llvm.md
plietar: I guess it doesn't really matter much, but I'm not sure I like treating operations on non-concrete types in a different place than operations on concrete types.

If we have a pass that lowers fields reads into a vtable load + field load, then we should also lower field read on concrete types. We just replace the first step of it by a constant value.

src/mlir/lowering.md
```diff
@@ -0,0 +1,517 @@
...
+
+### drop
+
+Dropping (deallocating) regions need to know that there are no more active users
+of that region (ex. reference counting) and also need to find which region it is.
+If we're dropping an `iso`, it's straight forward, if it's a `mut`, we simply
+follow the annotation to find the `iso`.
+
+The actual deallocation will be a call to the runtime library if the region is
+gone (ie. equals `nullptr`). This is a simple check that can be lowered by the
```
plietar, line 241: I don't understand what you mean by `gone`.

src/mlir/lowering.md
```diff
@@ -0,0 +1,517 @@
...
+Mutable references must belong to a region and must have access to that region,
+ie. they need to belong to the owning `iso` of the region. For this, mutable
+references must point to the `iso` references. This can be represented as
+MLIR annotation, referring the SSA value (ex. `#region<%22>` where `%22` must be
+an `iso` reference of the same region).
+
+### drop
+
+Dropping (deallocating) regions need to know that there are no more active users
+of that region (ex. reference counting) and also need to find which region it is.
```
plietar, line 236: Regions are deallocated when the `iso` goes out of scope / is overwritten. We don't do any kind of smart trick to find "active users" of it.

src/mlir/lowering.md
```diff
@@ -0,0 +1,517 @@
...
+The operation has two semantics plus the merge:
+```
+  // new object in a new region
+  %0 = verona.new_region @Class [initialiser]
+
+  // new object in an existing region
+  %1 = verona.new_object @Class [initialiser] in @Other
+
+  // Merge two regions
+  verona.merge %0, %1
```
plietar, line 267: `merge` returns a mut reference to the object `%1` referenced (`%1` would have been an `iso`).

src/mlir/lowering.md
```diff
@@ -0,0 +1,517 @@
...
+
+Example:
+```
+  // 10 + a : U32
+  %const = verona.constant "10" : U32
+  %0 = verona.new_object @U32 [ %const ]
+  %1 = verona.call "+" [ %0 ] ( %a )
+```
+
+### StringLiteralStmt
```
plietar, line 388: Can we just ignore strings for now? It's a huge can of worms. I wrote the current implementation just for the sake of demonstration purposes, so I wouldn't build anything on top of it for now.

I suspect 80% of the functionality of strings will be managed in the standard library and the rest will almost entirely be in the frontend. Anything that's left in the backend will likely be trivial and actually be more about array literals than anything else.

src/mlir/lowering.md
```diff
@@ -0,0 +1,517 @@
...
+
+At the 2IR level, there is no need to represent this statement as an operation
+and the only implementation detail is to add the previous value to the new
+symbol.
+
+### EndScopeStmt / OverwriteStmt
+
+When variables go out of scope, either lexical or type related (ex. move
+semantics). We need to call `release` on owned region types (region, iso, imm)
+but not on unowned types (mut).
```
plietar, line 491: It's more complicated than types. It's about *values*. If you have a variable of type `iso | mut | imm`, there are 4 possible ways to deal with it (`{Cown,Immutable,Region}::release`, or nothing), and you cannot pick the correct one until runtime.

Also "owned region types (region, iso, imm)" is confusing. `imm` is not a region.

src/mlir/passes.md
```diff
@@ -0,0 +1,207 @@
...
+## Virtual table colouring
+
+Concrete classes implementing the same interface will have similar (if not
+identical) method layout. Concrete implementation of generic functions can
+proliferate concrete types, increasing the number of similar/identical type
+representations.
+
+If each type has its own vtable, we're spending a large amount of read-only
+memory to store them all. This can also hurt caching, if objects are used in hot
+code but are too far apart from each other in the read-only section.
```
plietar, line 102: I have no data to back this up, but I seriously think this is negligeable and is certainly a problem that doesn't need solving now.

src/mlir/passes.md
```diff
@@ -0,0 +1,207 @@
...
+This minimises the size and complexity of the vtables and improve runtime
+performance avoiding unnecessary indirect calls.
+
+### IR state afterwards
+
+At this stage, all static calls should have moved to a direct `std.call` to
+named (and possibly mangled) names that exist as functions in the IR, with the
+appropriate lexical context (module name, etc).
+
+## Virtual table colouring
```
plietar, line 93: For the time being, you can stick with a very simple "each name gets a different offset" and not bother do anything smart. It'll be inefficient, but easy to implement and easy to improve later.

src/mlir/passes.md
```diff
@@ -0,0 +1,207 @@
...
+
+If each type has its own vtable, we're spending a large amount of read-only
+memory to store them all. This can also hurt caching, if objects are used in hot
+code but are too far apart from each other in the read-only section.
+
+In order to make the vtables more compact, we join all vtables into a single
+block, where the type points to a position in that table and the offset of each
+method is added to that position. Methods with the same name always have the
+same offset. But methods that aren't shared across multiple types can reuse
+offsets.
```
plietar, line 108: "Methods with the same name always have the same offset. But methods that aren't shared across multiple types can reuse offsets." this is true regardless of whether we use the single compact vtable trick or not.

src/mlir/passes.md
```diff
@@ -0,0 +1,207 @@
...
+    a: 0
+    b: 1
+    c: 2
+    d: 0
+    e: 1
+    f: 0
+
+  Layout:
+  +-----+
+  |  a  | <- A, B
```
plietar, line 127: This doesn't work. A and B need to have different values for entry `a`.

src/mlir/lowering.md
```diff
@@ -0,0 +1,517 @@
...
+  %utf8 = verona.new_object @UTF8String [%0, %name_str, %1, %age_str, %2]
+```
+
+In the case above, if all variables are already compile-time constant strings
+at the time of lowering, the compiler has the ability to fuse the strings,
+creating a simple string (or at least less splits).
+
+### UnitStmt
+
+No idea.
```
plietar, line 434: In the long term `unit` won't be a special concept in the IR and will be replaced with something from the standard library.

plietar (commented):
I think it may be more useful to describe the IRs themselves, rather than the lowering process. Given a target IR, the process of lowering to it is mostly obvious. However by describing the lowering process instead, it is still really unclear what the target looks like.

In many ways, I still have a hard time understanding what the abstraction level of the 2IR is supposed to be.

src/mlir/llvm.md
rengolin: > If we have a pass that lowers fields reads into a vtable load + field load, then we should also lower field read on concrete types. We just replace the first step of it by a constant value.

That's what it does, which means by the end, it's already a static call that have been lowered to a `std.call`.

rengolin:
> I think it may be more useful to describe the IRs themselves, rather than the lowering process. Just giving the list of operations and their signatures would help, without having to delve too much into details.

This is intentional. I don't want to fix what the IR will look like but I want to make sure what we expect from the 1IR and its semantics. Ultimately, the 2IR will depend heavily on what the 1IR will be.

> Given a target IR, the process of lowering to it is mostly obvious. However by describing the lowering process instead, it is still really unclear what the target looks like.

What the IR will "look like" is almost entirely irrelevant. What really matters is what kind of information we'll have in it, and that's defined by what the 1IR looks like and what transformations we do with it.

My aim with this document is to make clear that I don't want to destroy any information in the first lowering phase. I may lower custom operations into code but only when the semantics is identical and there's no value in keeping it (ex. we won't use in optimisations).

> In many ways, I still have a hard time understanding what the abstraction level of the 2IR is supposed to be.

Me too, and that's also intentional.

The three documents I wrote so far are brain dumps of what I expect that we'll do. They're in no way a contract of what has to be done and more importantly, they shouldn't be fixing form over function.

I'll reply to your inline questions in separate.

src/mlir/lowering.md
rengolin: Isn't this just a runtime `match`? I'm expecting that all runtime unions will need runtime `match` statements. It should be fine for either 1IR or 2IR to add the calls depending on which level we're lowering.

Fixed the "owned types" confusion.

src/mlir/passes.md
```diff
@@ -0,0 +1,207 @@
+# MLIR passes
+
+Long term, we want the following passes to run in MLIR:
+ * Type inference & check (removing unknown types)
+ * Free variable & liveness analysis (sanity check, phi nodes, clean-up calls)
+ * Reification & type-based code ellision (instantiating all generics)
+ * Region based alias analysis (disjoint regions never alias)
```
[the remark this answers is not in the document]
rengolin: None of the current MLIR passes are necessary for correctness. That's the main difference between the two levels for now.

I think we're all clear that there are correctness and performance passes, but I can add a paragraph on that here.

src/mlir/passes.md
rengolin: this is a document on the intention and listing the things we'll need to do, not a plan to do it all now.

src/mlir/passes.md
rengolin: No one is saying we'll solve it now.

src/mlir/passes.md
rengolin: sigh, you're right. It's hard to create an example demonstrating the point.

src/mlir/lowering.md
```diff
@@ -0,0 +1,517 @@
...
+## Basic Blocks
+
+So far, there is only two constructs that create new basic blocks: conditionals
+and loops. Both will end with a tail block that will be the main block of the
+subsequent operations.
+
+The main issue when lowering basic blocks is figuring out their arguments. For
+this we need free variable and liveness analysis, which will have already been
+done by the 1IR passes. That information is in the `FnAnalysis` structure for
+now and it should be trivial to use it for arguments without having to backtrack.
```
plietar, line 182: The 1IR is already in SSA form and has all the basic block arguments you need. There's no need to worry about that anymore.

plietar:
Then I don't understand what these documents are trying to acheive. It is either documenting 1-1 abstraction mappings (Verona calls will become LLVM calls) and describing passes that won't be relevant for another 6-12 months (selector colouring, region based alias analysis).

Meanwhile, it omits any description of lowering high level abstractions into lower ones:
- At what stage will increfs be introduced? How do you handle `mut | imm` variables, which may or may not need an incref.
- At what stage will boxing and unboxing occur? If a U64 (ie. a machine word) is written to an `Any` (the empty interface) variable, we need to box the value and give it a descriptor. Now if a class' method returns a U64, but it used to fulfill an interface that returns an `Any`, we need to generate glue code that wraps the method.
- At what stage will additional region arguments be added to methods? Given a method `m(x: mut)`, in LLVM this needs to have two parameters, one for the pointer to `x` and one for the pointer to its region.
- At what stage will "region propagation" happen? On a field read, we need to dyamically compute the region of the value we just read based on a number of factors. Matt has described a possible version of this in pseudo code in #357.

These steps could all happen on the 1IR->2IR lowering, as a 2IR pass or on the 2IR->LLVM dialect lowering.

rengolin:
> Then I don't understand what these documents are trying to acheive. It is either documenting 1-1 abstraction mappings (Verona calls will become LLVM calls) or describing passes that won't be relevant for another 6-12 months (selector colouring, region based alias analysis).

They are meant to be a guide for implementation at all levels and time-frames. Not a TODO list per se, but a list of objectives and the semantics required to achieve them. I have suffered in the past for the lack of such documents and I want to make sure neither me nor anyone else have to go through that again.

> Meanwhile, it omits any description of lowering high level abstractions into lower ones:

One of the goals of these documents was precisely to get feedback like the list below. Thanks! :D

> * At what stage will increfs be introduced? How do you handle `mut | imm` variables, which may or may not need an incref.

Good question. I'm expecting this to be a runtime implementation that is coupled with "constructors" (new) and "destructors" (end-scope). We'll lower `new` as a call to the right `allocate` and end-scope should already have been inserted by the 1IR after liveness analysis.

What checks if the count is still positive could be either inside the runtime or as a lowered sequence after every end-scope.

Would that work?

> * At what stage will boxing and unboxing occur? If a U64 (ie. a machine word) is written to an `Any` (the empty interface) variable, we need to box the value and give it a descriptor. Now if a class' method returns a U64, but it used to fulfill an interface that returns an `Any`, we need to generate glue code that wraps the method.

I'd say type logic should have been done by the 1IR as per my view of the current division. The 2IR will receive types that `validate` in MLIR, which means you can't assign a `U64` into a `Any` variable type without a cast or an assignment operator.

As previously, my aim is to add casts to whatever doesn't match, assuming that the 1IR will have checked it already. I don't want to introduce auto-boxing in the 2IR because that would be adding automatic casts that I'm not sure it's safe, because the 2IR doesn't have a sophisticated type check/inference.

> * At what stage will additional region arguments be added to methods? Given a method `m(x: mut)`, in LLVM this needs to have two parameters, one for the pointer to `x` and one for the pointer to its region.

IIUC, the pointer to a `mut` region is a runtime call (as described by @mjp41), so this can be just an additional call, either before the method (which will change the number of arguments) or before every use (which could be sub-optimal, but still valid).

> * At what stage will "region propagation" happen? On a field read, we need to dyamically compute the region of the value we just read based on a number of factors. Matt has described a possible version of this in pseudo code in #357.

This will be a pass in 2IR that I forgot to add. It needs to run before "region alias analysis" as we want to have all regions fixed by then. I'll add it to the document. Thanks!

> These steps could all happen on the 1IR->2IR lowering, as a 2IR pass or on the 2IR->LLVM dialect lowering.

Right, and these are each of the three documents: `lowering.md`, `pass.md` and `llvm.md`. I don't want to commit to the order and separation now, but we need to have a ball park separation, and writing a document and discussing it with others makes it much easier and clearer to all what the expectations are.

rengolin:
> > * At what stage will "region propagation" happen? On a field read, we need to dyamically compute the region of the value we just read based on a number of factors. Matt has described a possible version of this in pseudo code in #357.
> 
> This will be a pass in 2IR that I forgot to add. It needs to run before "region alias analysis" as we want to have all regions fixed by then. I'll add it to the document. Thanks!

Wait, I thought that what Matt described in #357 was going to happen on the 1IR (thus being in that document), and I'd get the regions already tracked and merged. 

The pass I wanted to add before the region analysis is region merging (for objects that are not declared in the same region but could have been).

plietar:
> Good question. I'm expecting this to be a runtime implementation that is coupled with "constructors" (new) and "destructors" (end-scope). We'll lower new as a call to the right allocate and end-scope should already have been inserted by the 1IR after liveness analysis.

What I mean is that a number of operations in the 1IR implicitely need reference count manipulation, not just `new` and `end-scope`. Copy needs to incref, method calls need to incref (maybe, we may optimize the calling convention at a later point), writing to a field needs to add the RC value to the target region's remembered set.

When all these operations get lowered to their `std`/`llvm` equivalent, the increfs need to be inserted in order to preserve the semantics. If we have a hybrid translation, where static calls are `std.call` but dynnamic calls are `verona.call`, does this mean that refcounts have been inserted for the former, but not the later?

> What checks if the count is still positive could be either inside the runtime or as a lowered sequence after every end-scope.

This is already inside the runtime, as part of `Immutable::release`. Even reading the refcount is a little bit involved and not something we should move out of the runtime, due to the way immutables are represented.

> I'd say type logic should have been done by the 1IR as per my view of the current division. The 2IR will receive types that validate in MLIR, which means you can't assign a U64 into a Any variable type without a cast or an assignment operator.

Right, the 1IR does not do that. The IR is built straight from the AST before we have types. While the type inference adds those types to the IR, it does not insert new operations.

For reference, Pony does this step directly when lowering the LLVM, which is a bit messy but works.

> As previously, my aim is to add casts to whatever doesn't match, assuming that the 1IR will have checked it already.

The 1IR will have checked that there exists a coercion from source type to target type anytime a variable is used, although it does not insert such an operation into the IR. I guess these can be considered "casts", but they are not trivial ones, the layout of the input and output is different. Adding a bunch of `verona.coerce` operations during the 1IR->2IR would work, and then have a pass that lowers these to various operation sequences based on the input and output types.

There is still the problem of generating the "glue" code around methods. Effectively if you have:
```
interface I {
  m(): Any
}
class C {
  m(): U64 { return 5; }
}
```

then this needs be transformed into

```
interface I {
  m_any(): Any
}
class C {
  m_u64(): U64 { return 5; }
  m_any(): Any { verona.coerce(m_u64()); }
}
```

because if someone call method `m` on a value of type `I`, they won't know whether they need to coerce the result, hence the callee needs to do that. Essentially you can think of this as multiple calling conventions for the same function.

> IIUC, the pointer to a mut region is a runtime call (as described by @mjp41), so this can be just an additional call, either before the method (which will change the number of arguments) or before every use (which could be sub-optimal, but still valid).

No, if you only have a `mut` reference to an object, there is no cheap way the runtime could find the reference to the region. However many operations do need a reference to the region. This is the whole point of "region tracking": for every operation that could produce a `mut` value we need to emit code that computes that value's region. When a `mut` value is passed into a function, we must also pass its region to the function. When a `mut` value is returned from a function, we must also return its region (there's optimizations we can do on this, but this is what happens in the general case). The 1IR does none of this (the existing VM adds a pointer to each object to point to the regio, but that is overhead we don't want).

> Wait, I thought that what Matt described in #357 was going to happen on the 1IR (thus being in that document), and I'd get the regions already tracked.

Yeah that document was a mix of describing the 1IR and "concepts Renato should know", which is maybe not the best structure. I might split the file up into separate things. The 1IR does nothing for region tracking.

rengolin:
> What I mean is that a number of operations in the 1IR implicitely need reference count manipulation, not just `new` and `end-scope`. Copy needs to incref, method calls need to incref (maybe, we may optimize the calling convention at a later point), writing to a field needs to add the RC value to the target region's remembered set.

Right, `new` was an example of "constructor". I don't want to have to special case things, so if we have rules for passing arguments we need to make that explicit (in a document or in a previous pass). So, shallow copies are another type of constructor, and method calls will call them for each argument where meaningful.

What about decrefs? We could try to guess which values need decrementing at each function exit, but this is already a result of the liveness analysis that will be done in 1IR, so shouldn't all calls will be placed already?

> When all these operations get lowered to their `std`/`llvm` equivalent, the increfs need to be inserted in order to preserve the semantics. If we have a hybrid translation, where static calls are `std.call` but dynnamic calls are `verona.call`, does this mean that refcounts have been inserted for the former, but not the latter?

Good question. We need to be consistent if we don't want to introduce special rules (and with it, special bugs). But this will also depend on what the runtime ctors/dtors will do and what we'll need to lower as a call to other calls.

> This is already inside the runtime, as part of `Immutable::release`. Even reading the refcount is a little bit involved and not something we should move out of the runtime, due to the way immutables are represented.

Perfect.

> Right, the 1IR does not do that. The IR is built straight from the AST before we have types. While the type inference adds those types to the IR, it does not insert new operations.
> 
> For reference, Pony does this step directly when lowering the LLVM, which is a bit messy but works.

I think this is far too late. Even at the 2IR level, I don't know which types I have to add simple casts (ie nops) and which I have to insert constructors (coerce), and how they behave. 

Perhaps this could even be done at the AST level. If the type is unknown, it could add a "unknown-type-ctor" call that will later move into the type's own ctor (in 1IR, after type inference), and then I can just call that function in the 2IR. Or it could be done at the 1IR level, after type inference, as we'll have multiple data structures with the relationship between the types etc.

> The 1IR will have checked that there exists a coercion from source type to target type anytime a variable is used, although it does not insert such an operation into the IR. I guess these can be considered "casts", but they are not trivial ones, the layout of the input and output is different. Adding a bunch of `verona.coerce` operations during the 1IR->2IR would work, and then have a pass that lowers these to various operation sequences based on the input and output types.

Right, when I say `cast` I do mean `noop`. We don't have enough information in the 2IR to know when we add coerce instead of casts.

If we make into a similar logic than C++ (which is not a good idea), we could look up a function in the type that "construct" from the input type and apply that, but this would be like implicit casts, which we don't want.

> because if someone call method `m` on a value of type `I`, they won't know whether they need to coerce the result, hence the callee needs to do that. Essentially you can think of this as multiple calling conventions for the same function.

Exactly. But this has to have been done at a higher level, either AST or 1IR for the same reason as above: we don't have enough type information at this level to guess correctly. Also, the 1IR will have the infrastructure to create new functions from old ones (reification) and is the easiest place to insert such a pass.

> No, if you only have a `mut` reference to an object, there is no cheap way the runtime could find the reference to the region.

Ack.

> The 1IR does none of this (the existing VM adds a pointer to each object to point to the regio, but that is overhead we don't want).

I see. It should be reasonably simple to add new arguments to each function that receives or return `mut` values. We just need to agree on the calling convention (thinking about spilling arguments to stack):
```
  func(a : mut, b : mut, c : mut);
  // option a, best if mut and region are usually accessed with the same frequency
  func(a : mut, ar : region, b : mut, br : region, c : mut, cr : region);
  // option b, best if regions are rarely accessed
  func(a : mut, b : mut, c : mut, ar : region, br : region, cr : region);
```

MLIR accepts multiple return values, but LLVM doesn't, so we'd do:
```
  func(...) : mut;
  // option a, best if mut and region are usually accessed with the same frequency
  func(region, ...) -> mut;
  // option b, best if regions are rarely accessed
  func(..., region) -> mut;
```

rengolin:
Folks, I think the documents are as ready as they can be. This is not meant to be a TODO list, nor an accurate reference. Just an high-level idea on how to proceed. So, I'm trying not to focus too much on the details and more on the overall architecture. In that sense, I think it's good enough for now.

I'd like to merge this so I can start working on the actual code.

PR #368, opened 2020-12-08
Use a LinkedObjectStack to collect objects.

plietar:
This produces the exact same behaviour, only it makes use of an
existing higher abstraction.


mjp41 (approved):
LGTM

PR #369, opened 2020-12-08
Remove extraneous 'template' keyword.

plietar:
This causes a compiler error when attempting to use the forall method.
I think the keyword was a leftover from a previous version, where the
apply function was passed as a template parameter.


mjp41 (approved):
LGTM

PR #370, opened 2020-12-08
Remove the `mark_pending`/`unmark_pending` methods.

plietar:
They aren't used anywhere. `mark_pending` was identical to
`set_pending` anyway, which does get used. In practice,
`unmark_pending` is actually done with the `set_atomic`/`set_scc`
methods.


mjp41 (approved):
LGTM

PR #372, opened 2020-12-09
Improvements to CI and build.

mjp41:
Make testing part of top-level build to simplify
use of some scripts such as update-dump.  Also, gives
nicer experience for running individual tests.


mjp41:
FYI.  Successfully passed a nightly build, but on the commit without updated instructions. 
https://dev.azure.com/ProjectVeronaCI/Project%20Verona/_build/results?buildId=1854&view=results

devops/ci.yml
```diff
@@ -90,7 +90,7 @@ jobs:
       set -eo pipefail
       export ASAN_SYMBOLIZER_PATH=/usr/lib/llvm-6.0/bin/llvm-symbolizer
       export ASAN_OPTIONS="alloc_dealloc_mismatch=0 symbolize=1"
-      ninja check
+      ctest -C $(BuildType) -j $(nproc) --output-on-failure --timeout 400 --interactive-debug-mode 0
```
rengolin, line 93: Will this require other users to call the same convoluted line, or does `ninja check` still mean the same thing and you're just making it more configurable in the build?

mjp41: `ninja check` uses the default setting for `ctest`. This does not contain `timeout`s or `output-on-failure`.  I'll do a little bit more, so that `check` has better defaults.

rengolin (approved):
LGTM, thanks!

plietar:
This breaks the workflow of passing `-DVERONA_LLVM_LOCATION` to point to a manually built LLVM.

Given how much trouble ExternalProject and the recursive invocation has caused, could we instead revert to a "simpler" setup:

- If `-DVERONA_LLVM_LOCATION` exists, use that
- Otherwise, use `CMAKE_BINARY_DIR/llvm`
   - If this directory contains stuff, use it
   - Otherwise don't include targets for `src/` (or build the LLVM-free parts of the compiler) and instead expose a `ninja download-llvm` target. Print a clear warning message that the user should run this command if they want the llvm backend.

This adds a minor hurdle for the first build (having to run `cmake; ninja download-llvm; cmake; ninja`, we might be able to do the second cmake implicitely), but it will save us on a lot of complexity and discrepancy between different workflows.

I don't mind putting in the work for this, if there are no objections.

PR #373, opened 2020-12-10
[Website] Add people to the web site

davidchisnall:
Adds a people page (currently missing most of the Imperial folks) and automatically link authors in blog posts to their URLs in the people thingy.  Two separate commits in this PR, which might be easier to read independently (the second builds on the first).


mjp41:
Does the order in `docs/_data/people.yml` affect layout?  If so, this should either be alphabetical, or carefully chosen not to offend people.

davidchisnall:
It does, I think we can sort it alphabetically.  The order is the same as on the MSR web site currently :-)

davidchisnall:
Done in the last update.

mjp41:
What do you need from @plietar and @lukecheeseman to add them to the list?

Also, it would be good to circle back to the MSR web team, so they can get updates from here.

davidchisnall:
The just need entries in the .yml file.  We can add names now, but we probably want photo and web page URL at the same time.  

The MSR web site is very difficult to add external people to.

davidchisnall:
Added Paul and Luke, missing photos.

mjp41 (approved):
LGTM
- Is it alphabetically by first name, or surname?
- Could you explicitly say listed alphabetically?

davidchisnall:
It's alphabetically by first name (which is why Juliana is second).  Eventually I'll figure out how to sort by surname...

davidchisnall:
@mjp41, I've made it sort by surname.  Let me know if you'd like some squashing here or if it's just a rebase.

docs/_includes/person.html
```diff
@@ -0,0 +1,36 @@
+<h2>{{ include.title }}</h2>
+{% comment %}
+We want to sort people by surname.
```
plietar, line 3: Would it be easier to have a "last_name" entry in the yaml and concatenate them here, or have a "sort_key" entry that doesn't even get displayed?

davidchisnall: I thought about this and I'm not sure which route is correct.  I see two options:

 - Add a sort key and also a unique identifier (github username, maybe?) that we can use to refer to people from blog posts.
 - Do the current set of hacks to sort things.

If we go with the second one, it works for now and there's some fairly mechanical transformation needed to get to the other approach.  I'm starting to think it is better to do that now, rather than have to update it later.

davidchisnall: I've added a sort key.  This simplifies everything and seems nicer.

davidchisnall:
@mjp41, sorting is now by surname (or other sort descriptor for folks that don't have European style names), so I'm happy for this to be merged now.

docs/_data/people.yml
```diff
@@ -0,0 +1,50 @@
...
+- name: Paul Liétar
+  affiliation: Imperial College London
+  sort-key: Liétar
+- name: Luke Cheeseman
+  affiliation: Imperial College London
+  sort-key: Cheeseman
+- name: Albert Mingkun Yang
+  url: http://www.it.uu.se/katalog/albya111
+  photo: https://www.microsoft.com/en-us/research/wp-content/uploads/2020/04/Albert-Mingkun-Yang-e1586347975790-360x360.jpg
+  sort-key: Cheeseman
```
plietar, line 44: Yang?

davidchisnall: Copy and paste fail :-(

mjp41:
I think this can be added merged now?

davidchisnall:
I thought it had been, so yes.

PR #374, opened 2020-12-11
Simplify CMake setup.

plietar:
Just trying out what the CI says.


plietar:
Not going down this route anymore.

PR #376, opened 2020-12-17
Fix the use of RT_TESTS in CI.

plietar:
With `RT_TESTS=Off`, we were not adding any `rt_tests` target, but check was still depending on it.
This is what is currently breaking CI in #364.


mjp41 (approved):
LGTM

PR #377, opened 2020-12-18
Refactor CMake

mjp41:
* Makes a top-level approach of supplying your own installation
* Makes the previous recursive approach now just submodules.
* Applies some of Paul Lietar's refactorings
* Fix LLVM submodule build


src/CMakeLists.txt
```diff
@@ -1,5 +1,34 @@
...
+
+if (VERONA_CI_BUILD)
+  # Specify policy to go into child projects.
+  set (CMAKE_POLICY_DEFAULT_CMP0077 NEW)
+  set (SNMALLOC_CI_BUILD ON)
+endif ()
+
+include(../cmake/enable-asserts.cmake)
+
+add_subdirectory(../external/CLI11 ./external/CLI11 EXCLUDE_FROM_ALL)
```
plietar, line 19: Wasn’t there a thing about snmalloc has to live in `src/rt` because we can’t import stuff from parent directories?

If this works here, could we use the same trick to move snmalloc to the top level external?

(at a later point, I’m obviously not suggesting we do this now)

mjp41: I was thinking exactly the same.  I was going to do that as a separate PR.

CMakeLists.txt
```diff
@@ -113,128 +87,70 @@ if (NOT DEFINED VERONA_LLVM_LOCATION)
...
+set (VERONA_EXTRA_CMAKE_ARGS)
+
+# Point Verona at the MLIR
+list (APPEND VERONA_EXTRA_CMAKE_ARGS
+  -DVERONA_LLVM_LOCATION=${MLIR_INSTALL}/install
+)
+
+# Define project level defaults
+list (APPEND VERONA_EXTRA_CMAKE_ARGS
+  -DCMAKE_CXX_STANDARD=17)
```
plietar, line 107: Should this just be define from within `src/CMakeLists.txt`?

CMakeLists.txt
```diff
@@ -113,128 +87,70 @@ if (NOT DEFINED VERONA_LLVM_LOCATION)
...
-        CMAKE_C_FLAGS_RELWITHDEBINFO
-        CMAKE_C_FLAGS_MINSIZEREL)
-      string (REGEX REPLACE "(^| )[/-]D *NDEBUG($| )" " "
-        "${flags_var_to_scrub}" "${${flags_var_to_scrub}}")
-    endforeach()
+set (VERONA_EXTRA_CMAKE_ARGS)
+
+# Point Verona at the MLIR
+list (APPEND VERONA_EXTRA_CMAKE_ARGS
+  -DVERONA_LLVM_LOCATION=${MLIR_INSTALL}/install
```
plietar, line 102: The `/install` suffix will cause problems with an external LLVM, since it won’t have that layout.

CMakeLists.txt
```diff
@@ -113,128 +87,70 @@ if (NOT DEFINED VERONA_LLVM_LOCATION)
...
+add_custom_target(check DEPENDS verona)
+add_custom_command(TARGET check 
+  COMMAND ${CMAKE_CTEST_COMMAND} -C $<CONFIG> -j ${N} --output-on-failure --timeout 400 --interactive-debug-mode 0
+  USES_TERMINAL
+)
+
+# Main target does not test runtime, build subdirectory for that
+# and add to check target.
+if (RT_TESTS)
+  add_subdirectory(src/rt)
```
plietar, line 155: I know this was already the case before, and it probably works fine, but I find it a bit surprising that RT tests are built by the outer cmake when everything else is built by the inner one.

src/CMakeLists.txt
plietar: Great

CMakeLists.txt
mjp41: The aim is to have the tests not to be part of an external project.  There are a few reasons for this.  

  1. ExternalProject with tests on MSbuild doesn't work quite right,  I was not able to build a subproject without also running its tests. This hides that fact.  
  2. It also means we have a top level ctest target, and can call it directly on particular failing tests.  Rather than having to go diving through the layers of external project to call an individual test.

CMakeLists.txt
mjp41: Done

CMakeLists.txt
mjp41: Fixed, but not nicely.

mjp41: Nicer fix now

CMakeLists.txt
```diff
@@ -11,89 +13,90 @@ option_top (VERONA_DOWNLOAD_LLVM "Download cached version of LLVM" ON)
...
+endif ()
 
+if (VERONA_DOWNLOAD_LLVM AND DEFINED VERONA_EXTERNAL_LLVM_DIR)
+  message (WARNING "Ignoring VERONA_EXTERNAL_LLVM_DIR as using downloaded LLVM.")
+endif ()
+
+##########################################################
+#  Configure MLIR installation
+##########################################################
+if (DEFINED VERONA_EXTERNAL_LLVM_DIR)
```
plietar, line 83: Rather than relying on `VERONA_EXTERNAL_LLVM_DIR` being undefined, I think it would be better to have, in the option group at the top,

```cmake
set (VERONA_EXTERNAL_LLVM_DIR "" CACHE PATH "Path to an existing LLVM installation.")
```

so the option is documented. This defaults to the empty string, which is false-y in a if block.

The option should also be added to `VERONA_DONT_PASS` to avoid getting a "unused variable" warning.

CMakeLists.txt
```diff
@@ -114,127 +116,57 @@ if (NOT DEFINED VERONA_LLVM_LOCATION)
...
-  # Adds a target check that runs the tests.
-  add_custom_target(check
-    COMMAND ${CMAKE_CTEST_COMMAND} -C $<CONFIG> -j ${N} --output-on-failure --timeout 400 --interactive-debug-mode 0
-    USES_TERMINAL
-  )
-
-  # If RT_TESTS are enabled, we need to make sure they are
-  # up-to-date before we run tests.
-  if (RT_TESTS)
-    add_dependencies(check rt_tests)
```
plietar, line 179: This line is missing in the new version I think.

plietar (commented):
Mostly cosmetic comments. Overall it builds and tests fine for me, including passing the inner build weird flags required by my setup (`CMAKE_OSX_DEPLOYMENT_TARGET`).

Thanks!

CMakeLists.txt
plietar: Same applies to `LLVM_EXTRA_CMAKE_ARGS` I guess.

Given that `option` is pretty much a shorthand for `set(XXX Default CACHE BOOL "Description")`, I think we could replace `option_top` with something like:
```cmake
macro (set_cache_top VAR DEFAULT TYPE DEFAULT)
  set (${VAR} ${DEFAULT} CACHE ${TYPE} ${DEFAULT})
  list (APPEND VERONA_DONT_PASS ${VAR})
endmacro()
```

CMakeLists.txt
```diff
@@ -11,89 +13,90 @@ option_top (VERONA_DOWNLOAD_LLVM "Download cached version of LLVM" ON)
...
 
-  set (MLIR_INSTALL ${CMAKE_BINARY_DIR}/$<CONFIG>/mlir)
+set(DEFAULT_BUILD_TYPE "Release")
+if (NOT CMAKE_BUILD_TYPE)
+  message(STATUS "Setting build type to '${DEFAULT_BUILD_TYPE}' as none was specified.")
+  set(CMAKE_BUILD_TYPE "${DEFAULT_BUILD_TYPE}" CACHE STRING "Choose the type of build." FORCE)
+endif()
+
+if (VERONA_DOWNLOAD_LLVM AND DEFINED LLVM_EXTRA_CMAKE_ARGS)
+  message (WARNING "Ignoring LLVM_EXTRA_CMAKE_ARGS as using downloaded LLVM.")
```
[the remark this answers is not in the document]
plietar: `LLVM_EXTRA_CMAKE_ARGS` and `VERONA_EXTERNAL_LLVM_DIR` are also mutually exclusive I guess.

CMakeLists.txt
mjp41: Good catch got dropped in a merge conflict when I rebased.

CMakeLists.txt
mjp41: I have done approximately what you ask.  I don't really understand the CMake variable/cache model, so if you could take a final look that would be great.

plietar: Sorry that wasn't clear, what I meant was that `option_top` is subsumed by `set_cache_top`, ie.
```cmake
option_top (VERONA_DOWNLOAD_LLVM "..." ON)
```
Can be replaced by
```cmake
set_cache_top(VERONA_DOWNLOAD_LLVM "..." BOOL ON)
```

then we only need one of the two macros.

plietar (approved):

CMakeLists.txt
mjp41: I've refactored slightly, but kept two macros, it make it easier to switch from `option` to `option_top`, if we need to move things around a bit.

PR #378, opened 2021-01-12
Count in-flight behaviours with cown RC

### LINKED ISSUE #355: Cown reference counting for in-flight behaviours
The backpressure system requires a metric for a cown's "load" that approximates how many messages are in its queue. Currently, this is only tracked on a cown as it runs by counting the amount of messages processed between tokens. This is insufficient when messages start piling up on a cown's queue while it is acquired for an incomplete multi-message. One solution with fairly low-overhead would be to use the cown's reference count to determine when a cown is overloaded. This approach may also reduce how aggressively muting propagates from muted cowns, since we would be able to tell if the muted cown is actually being overloaded instead of assuming that it is.

For this to work, a cown's refcount would need to be incremented when a behaviour is scheduled on it and decremented when the behaviour has completed. The runtime currently keeps a single reference count on scheduled cowns.

Theodus:
Whenever a behaviour is scheduled to run on a set of cowns, the reference count for each cown in the set is incremented. This reference count is then dropped once a fast acquire is attempted, if the cown is already schdeuled. This PR modifies this by keeping the reference count until the scheduled behaviour is complete so that the reference count for each cown is greater than or equal to the count of messages in its queue. This new referece count replaces the "load" metric used to trigger backpressure.

Closes #355


src/rt/sched/cown.h
```diff
@@ -953,54 +917,48 @@ namespace verona::rt
...
       if (curr == nullptr)
       {
         Systematic::cout() << "Reached message token on cown " << this
                            << std::endl;
-        assert(stat.has_token());
-        stat.set_has_token(false);
-        status.store(stat, std::memory_order_release);
+        assert(has_token);
+        has_token = false;
+        if (overloaded())
```
mjp41, line 943: Could the token message track the RC last time we sent the token message, so cowns with high in degree can not be considered overloaded, i.e. if the RC is not growing, then it don't consider it overloaded?

mjp41: Also, does it make sense to scan the message queue for its length when you observe a high RC?  I.e. most of the time, we know it is not a problem due to the RC, and only scan if it is a problem.

src/rt/sched/cown.h
```diff
@@ -1038,16 +996,9 @@ namespace verona::rt
...
-      const auto stat = status.load(std::memory_order_acquire);
       assert(priority() != Priority::Low);
 
-      // The batch limit is between 100 and 251, depending on the load.
-      const auto batch_limit = (size_t)100 | ((size_t)stat.total_load() >> 3);
-
-      Systematic::cout() << "Cown " << this << " load: " << stat.total_load()
-                         << std::endl;
-
+      static constexpr size_t batch_limit = 100;
```
mjp41, line 1001: Unused?

src/rt/sched/cown.h
```diff
@@ -113,8 +123,8 @@ namespace verona::rt
      **/
     std::atomic<size_t> weak_count = 1;
 
-    std::atomic<Status> status{};
     std::atomic<uintptr_t> bp_state{(Cown*)nullptr | Priority::Normal};
+    bool has_token = false;
```
mjp41, line 127: Can you leave a comment to pack this into something else?  Seems pretty wasteful on space.

src/rt/sched/cown.h
Theodus: It's used in the do-while condition down at line 1103.

src/rt/sched/cown.h
```diff
@@ -839,12 +811,9 @@ namespace verona::rt
         if (prev == state)
           return prev;
 
-      } while (
-#ifdef USE_SYSTEMATIC_TESTING
-        Systematic::coin(9) ||
-#endif
-        !bp_state.compare_exchange_weak(
-          bp, blocker | state, std::memory_order_acq_rel));
+      } while (coin(9) ||
```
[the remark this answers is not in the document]
Theodus: When the coin returns true, simulating a spurious failure, the exchange should not occur.

src/rt/sched/cown.h
Theodus: I'll look into this later today.

src/rt/sched/cown.h
```diff
@@ -25,6 +25,16 @@ namespace verona::rt
 #endif
   }
 
+  static bool coin(size_t range_bits = 1)
```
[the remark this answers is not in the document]
Theodus: Comment added.

src/rt/sched/cown.h
Theodus: Comment added.

src/rt/sched/cown.h
Theodus: Checking `next->overloaded()` in `fast_send` means that we can't check the actual queue length when needed. So instead, it should correct for the in-degree using a value updated whenever a cown finds its token message. The in-degree correction could be compressed into 8 bits using something like the snmalloc sizeclass, but that still raises the problem of having to make space for 8 bits. Does that make sense, or do you have something in mind that takes less space to correct for the in-degree?

src/rt/sched/cown.h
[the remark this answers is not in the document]
mjp41: Good point.  That makes sense. Perhaps add a comment to stop me breaking it in the future.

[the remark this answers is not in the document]
Theodus: Comment added.

mjp41 (commented):
Sorry, I am not sure how this comment got missed with my last review.

PR #379, opened 2021-01-15
Update sandbox code to new snmalloc and fix Linux.

davidchisnall:
This includes a few cleanups:

 - This code no longer depends on libc using snmalloc and so doesn't use
   the late-binding external pagemap support.
 - Linux pipes are not bidirectional, so pipes have been replaced with
   `AF_LOCAL` sockets where necessary.
 - All updates to the pagemap or requests for large allocations now go
   via a (trivial!) RPC mechanism to the host process.  This should also
   help define the set of upcalls required for other sandbox
   implementations.
 - The base address and length of the shared region are passed to the
   child via an environment variable, rather than in arguments.  This
   allows the child to map the shared memory region on the first call to
   any allocation function, rather than in `main` and so we no longer
   need to do the dance of replacing the thread allocator.

Note: Although the Linux version builds and passes a smoke test, it does
not actually sandbox the child process and so a compromise in the child
process retains ambient authority.  The Linux version does not yet
correctly align the shared memory region and so may have problems with
the pagemap with larger allocations.


davidchisnall:
This fixes building on FreeBSD and Linux, at least to the point where the smoke test can run.  It is not yet in a state where it's ready for moving out of experiments.  Still to do:

 - [ ] Clean up the structure of the code a bit (we still have a `libsandbox.cc` that is `#include`d in another file as well as being compiled separately, for example.  A lot more things should be pulled into separate files).
 - [ ] Extract all of the `#ifdef __linux__` and `#ifdef __FreeBSD__` into a sandbox PAL.  It's a little bit clunky at the moment, it will get much worse when we have more than two platforms.
 - [ ] Separate the C++ function export stuff into a a test harness.
 - [ ] Make the alloc outside the sandbox validate free lists and things.
 - [ ] Refactor the bootstrap code to use the snmalloc slow path code (this requires tweaking ).
 - [ ] Add privilege-elevation callbacks for opening files / sockets based on some policy.
 - [ ] Add a useful interface for the Verona runtime to use.

I think this is currently in a state where we could at least try to build it in CI, if not actually run it.

rengolin:
There's an empty file inside docs that got committed by accident, I think.

experiments/process_sandbox/host_service_calls.h
```diff
@@ -0,0 +1,97 @@
...
+  enum HostServiceCallID : uintptr_t
+  {
+    /**
+     * Push a large allocation to the stack.  The first argument is the address
+     * of the slab, the second the large sizeclass.  The return value is unused.
+     */
+    MemoryProviderPushLargeStack,
+    /**
+     * Pop a large allocation from the stack.  The first argument is the large
+     * sizeclass, the second is unused.  The return value is the address of
```
rengolin, line 28:
```suggestion
     * sizeclass, the second is unused.  The return value is the address that was
```

experiments/process_sandbox/host_service_calls.h
```diff
@@ -0,0 +1,97 @@
...
+namespace sandbox
+{
+  /**
+   * The ID of the method that's being proxied.
+   */
+  enum HostServiceCallID : uintptr_t
+  {
+    /**
+     * Push a large allocation to the stack.  The first argument is the address
+     * of the slab, the second the large sizeclass.  The return value is unused.
```
rengolin, line 23: the address of the slab was returned by reserve below, right? Perhaps add that to the comment.

experiments/process_sandbox/sandbox.hh
```diff
@@ -46,6 +46,51 @@ namespace sandbox
   struct SharedMemoryRegion;
   struct SharedPagemapAdaptor;
   struct MemoryProviderBumpPointerState;
+
+  /**
+   * A PAL that cannot be used to allocate memory.  This is used for sandboxes,
```
rengolin, line 51: It's hard to find what `PAL` means and this code is distant enough from snmalloc that it could warrant the expansion of the TLA.

experiments/process_sandbox/library_runner.cc
```diff
@@ -335,73 +485,123 @@ namespace sandbox
...
+    void* addr = nullptr;
+    size_t length = 0;
+    // Find the correct environment variables.  Note that libc is not fully
+    // initialised when this is called and so we have to be very careful about
+    // the libc function that we call.  We use the `environ` variable directly,
+    // rather than `getenv`, which may allocate memory.
+    for (char** e = environ; *e != nullptr; e++)
+    {
+      char* ev = *e;
+      const char ev_name[] = "SANDBOX_LOCATION=";
```
rengolin, line 513: Who sets this? Shouldn't the error message be a little more descriptive if the user is expected to know this?

I know this is an experiment, but without reading the github description of the PR, this would have been lost to me.

We can't expect people to git-blame to find the PR number and read the thread.

rengolin (commented):
Some quick comments inline.

The RPC socket should work well, but the code is a bit scattered. It would be nice if that was an isolated code (its own class, used by the Sandbox library, for example). I agree with your TODO list on this PR, especially the last point.

We should be looking at each functionality here and separating it as a library (or library looking class), so that we can more easily compose and refactor once things start to get serious.

The more entangled this code is, the harder it will be to move it to the Verona runtime library.

davidchisnall:
@rengolin, I think I've addressed all of your comments.  The `sandbox_pal` branch has my WIP changes to provide cleaner platform abstractions.  The RPC mechanism will end up factored out because we want something different for the coprocess version.

rengolin (approved):
LGTM, thanks!

rengolin:
@davidchisnall, anything holding up the merge?

davidchisnall:
No, this can probably be merged.  I'm working in the sandbox_pal branch, which builds on top of this and wasn't sure if it was worth merging this without the other changes.

rengolin:
It's probably easier to merge this than be constantly rebasing on the PR branch to make sure you didn't mess up the tested part. Or, you can change this to a draft and work on it as a whole, then request another review when you're done.

davidchisnall:
Closing, I'll raise a new PR that's rebased for the snmalloc pulls and includes the newer sandbox_pal code.

PR #382, opened 2021-01-29
Refactoring code around test harness

mjp41:
Small changes to make the test harness write to a log file in dump mode, and tidy up some of the error message production.

This is a precursor to larger test harness refactoring.


src/interpreter/value.cc
```diff
@@ -97,8 +98,8 @@ namespace verona::interpreter
   {
     if (tag != UNINIT)
     {
-      std::cerr << "Dropped an initialized Value" << std::endl;
-      abort();
+      // TODO: Should this be an internal error?
```
plietar, line 101: I think there is no way for bad bytecode to trigger this, unless it’s a VM bug, so internal error seems fine to me.

There’s actually an explicit check in the return opcode to make sure all registers in the frame have already been cleared, in which case it calls `fatal`.

src/compiler/error.h
```diff
@@ -0,0 +1,45 @@
+#pragma once
```
plietar, line 1: Since this is used by the VM, should this live somewhere else? eg. `src/ds/`

plietar (approved):

src/compiler/error.h
```diff
@@ -0,0 +1,45 @@
+#pragma once
+
+#include <iostream>
+#include <type_traits>
+#include <fmt/ostream.h>
+
+
+class InternalError
```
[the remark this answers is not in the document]
mjp41: Yeah, was trying to minimise the diff, can consolidate.

src/compiler/error.h
mjp41: Good point

src/compiler/error.h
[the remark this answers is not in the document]
mjp41: Did this in the end as MSVC gets upset about non-returning destructors.
https://docs.microsoft.com/en-us/cpp/error-messages/compiler-warnings/compiler-warning-level-1-c4722?view=msvc-160

PR #384, opened 2021-02-02
Add copyright/license check to CMake

rengolin:
This helps us running `ninja copyright` to check locally without having to rely on the CI or running it by hand. Should work on anything that has `bash` including Windows.

I tested both checks by changing/removing the headers from some files and the output (and exit status) are as expected.

### DIFF
--- a/devops/ci.yml
+++ b/devops/ci.yml
@@ -204,12 +204,6 @@ jobs:
 
   - script: |
       set -eo pipefail
-      git submodule init
-      git submodule update --depth 1 --recursive
-    displayName: 'Update submodules'
-
-  - script: |
-      set -eo pipefail
       wget -q -O - https://apt.llvm.org/llvm-snapshot.gpg.key | sudo apt-key add -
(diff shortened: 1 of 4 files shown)


devops/ci.yml
```diff
@@ -232,9 +226,6 @@ jobs:
 
   - script: |
       set -eo pipefail
-      files=`git ls-files -- '*.cpp' '*.cc' '*.h' '*.hh' '*.verona' '*.td' '*.mlir' | xargs`
-      grep -L "Copyright Microsoft and Project Verona Contributors."  $files | tee header_missing
-      [ ! -s header_missing ]
-      grep -L "SPDX-License-Identifier: MIT" $files | tee header_missing
-      [ ! -s header_missing ]
+      make copyright
+    workingDirectory: build
```
mjp41, line 230: Don't you need to check for a diff?
```
git diff --ignore-submodules=dirty --exit-code $(Build.SourceVersion)
```

mjp41: Sorry, I see build just checks it.

mjp41 (approved):
LGTM

PR #385, opened 2021-02-03
Move LLVM to dd2dac2fd07

rengolin:
The new commit fixes a problem on MSVC builds and doesn't seem to break anything upstream either, with the specific commit passing on many buildbots. Appropriate changes to the MLIR code to match included.

Additional changes:
* We're not building debug versions of LLVM anymore. We don't use them, they were huge and had random build failures on Windows. It's better to build a custom debug version for what we need locally when we do need than having multiple cached builds on storage that we never use.
* Moved LLVM's default branch from "master" to "main"
* Building Clang in addition to MLIR. The FFI code will depend on that, so we need the libraries in there.

There is still one build failure (Windows) because of a pegmatite library export issue that @mjp41 has identified. Once that's fixed, we can merge this.

### DIFF
--- a/devops/llvm.yml
+++ b/devops/llvm.yml
@@ -28,7 +28,7 @@ jobs:
 
   - script: |
       set -eo pipefail
-      git checkout master
+      git checkout main
       git pull
       git checkout $(GOOD_HASH)
     displayName: 'Move tree to known good LLVM commit for Verona'
@@ -96,7 +96,7 @@ jobs:
     displayName: 'Install CMake'
 
   - script: |
-      git checkout master
+      git checkout main
       git pull
       git checkout $(GOOD_HASH)
     displayName: 'Prepare tree'
@@ -151,7 +151,7 @@ jobs:
 
   - script: |
       set -eo pipefail
-      git checkout master
+      git checkout main
       git pull
       git checkout $(GOOD_HASH)
     displayName: 'Move tree to known good LLVM commit for Verona'
--- a/devops/llvm.yml
+++ b/devops/llvm.yml
@@ -20,13 +20,6 @@ jobs:
         BuildType: Release
         BuildName: release
         Sanitizer:
-      Clang Debug+ASAN:
-        CC: clang
-        CXX: clang++
-        CXXFLAGS: -stdlib=libstdc++
-        BuildType: Debug
-        BuildName: debug
-        Sanitizer: Address
   variables:
   - name: PKG_NAME
     value: verona-llvm-install-x86_64-linux-$(BuildName)-$(GOOD_HASH)
@@ -80,9 +73,6 @@ jobs:
   timeoutInMinutes: 120
   strategy:
     matrix:
-      Debug:
-        BuildType: Debug
-        BuildName: debug
       Release:
         BuildType: Release
         BuildName: release
@@ -152,9 +142,6 @@ jobs:
       Release:
         BuildType: Release
         BuildName: release
-      Debug:
-        BuildType: Debug
-        BuildName: debug
   variables:
   - name: PKG_NAME
     value: verona-llvm-install-x86_64-darwin-$(BuildName)-$(GOOD_HASH)
--- a/devops/llvm.yml
+++ b/devops/llvm.yml
@@ -45,7 +45,7 @@ jobs:
     displayName: 'CMake'
     inputs:
       cmakeArgs: |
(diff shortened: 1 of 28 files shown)


devops/llvm.yml
```diff
@@ -20,13 +20,13 @@ jobs:
...
         BuildName: release
         Sanitizer:
-      Clang Debug+ASAN:
-        CC: clang
-        CXX: clang++
-        CXXFLAGS: -stdlib=libstdc++
-        BuildType: Debug
-        BuildName: debug
-        Sanitizer: Address
+#      Clang Debug+ASAN:
```
davidchisnall, line 23: Please can you add a comment explaining why this is disabled and what the condition for reenabling it would be?

devops/llvm.yml
```diff
@@ -80,12 +80,12 @@ jobs:
   timeoutInMinutes: 120
   strategy:
     matrix:
-      Debug:
-        BuildType: Debug
-        BuildName: debug
       Release:
         BuildType: Release
         BuildName: release
+#      Debug:
```
davidchisnall, line 86: As above.

devops/llvm.yml
```diff
@@ -152,9 +153,9 @@ jobs:
       Release:
         BuildType: Release
         BuildName: release
-      Debug:
-        BuildType: Debug
-        BuildName: debug
+#      Debug:
```
davidchisnall, line 156: Also here (alternatively, delete these rather than commenting them out).

src/mlir/generator.cc
```diff
@@ -811,6 +811,8 @@ namespace mlir::verona
     // Create function
     auto funcTy = builder.getFunctionType(types, retTy);
     auto func = mlir::FuncOp::create(loc, name, funcTy);
+    // FIXME: This needs to be set correctly later
```
davidchisnall, line 814: What does 'correctly' mean here?  I think everything in Verona will have internal / private linkage except for the main entry point.

davidchisnall (approved):
Please add some comments for the commented-out-but-left-in things or remove them.  Otherwise, it looks fine.

devops/llvm.yml
rengolin: I thought I should keep it in case we want to enable it later, but I don't think we will need it.

src/mlir/generator.cc
rengolin: That's what I meant by "correctly": confirming that internal is the correct visibility here. I'm assuming the main entry point isn't any user function, but the entry point of the executable, right?

mjp41:
I've put a new Pegmatite commit on here.  This isn't the updated master yet, I wanted to test through Verona as well, before that PR is completed.

rengolin:
@mjp41 do you want to test the pegmatite update and submit a separate PR or is it ok to keep it here?

In any case, let me know when you finish your testing so we can merge this PR.

mjp41:
I am happy.  This should all work now.

mjp41:
>We're not building debug versions of LLVM anymore. We don't use them, they were huge and had random build failures on Windows. It's better to build a custom debug version for what we need locally when we do need than having multiple cached builds on storage that we never use.

So this means we can no longer build a local windows build with the default Debug settings.  As this needs to download the Debug build.  So currently the only way to build a local windows build is to build the submodule directly, which is not ideal.

rengolin:
> So this means we can no longer build a local windows build with the default Debug settings. As this needs to download the Debug build. So currently the only way to build a local windows build is to build the submodule directly, which is not ideal.

Not ideal, I agree. Windows doesn't allow release libraries to build with debug programs, so it's hard to work around that.

However, since we keep LLVM versions for at least a month, this would be a monthly cost at best. I do this on Linux myself (build, install, tar, keep the archive somewhere safe). This is still not ideal, but limits usage to people in the project, rather than every Windows build out there.

I had a really hard time reproducing the CI Windows environment on my machines. The LLVM build Windows VM is substantially different from the Verona build ones, and I can't have both on the same box. Some of the configuration on the Microsoft hosted VMs (tools, paths) I have no idea how to do.

When I was working on this PR, I kept getting flaky failures on the Windows debug builds that had no relation with the changes I was making. Perhaps it was something to do with building clang. Perhaps we are doing something silly (either during CI execution - bad yaml files - or during the machine creation - wrong VM instance)?

If we can fix the CI debug builds on Windows and make sure they're stable, then we can just re-enable the debug builds. As I said, worst case this is a monthly cost, and we don't keep many copies of the tar balls anyway in the blob storage.

PR #389, opened 2021-02-05
allow oftype on any expression, add try_rule for non-packrat retry

sylvanc:


mjp41 (approved):
LGTM.  I like the oftype node being everywhere.  That seems really good.

I wonder if we should change the syntax of lambda as the parsing seems to be challenging, which makes me worry that it will also be hard to read.

PR #390, opened 2021-02-07
Replace mute map with message to mutor

Theodus:
Each scheduler thread would previously manage a map to track which cowns had been muted. The keys in this map were the "mutor", which is the high priority cown that received a message from the set of cowns that were muted. The mute map managed by each scheduler thread has been replaced by a message to the mutor containing the set of cowns that have been muted. This mechanism is simpler than the mute map, removes the need for frequent mute map scans, and it should fix possible sources of deadlock from the mute maps.

### DIFF
--- a/src/rt/sched/cown.h
+++ b/src/rt/sched/cown.h
@@ -995,6 +995,11 @@ namespace verona::rt
       if (mutor == nullptr)
         return false;
 
+      // The array of senders is reused for the unmute message. Since fewer than
+      // the original count of cowns may be muted, a null terminator may be
+      // added before the end of the allocation to mark the end of the muted
+      // set.
+
       Scheduler::local()->mutor = nullptr;
       size_t muting_count = 0;
       for (size_t i = 0; i < count; i++)
--- a/src/rt/sched/cown.h
+++ b/src/rt/sched/cown.h
@@ -1058,8 +1058,7 @@ namespace verona::rt
       if (muting_count < count)
         senders[muting_count] = nullptr;
 
-      auto* body = MultiMessage::make_body(alloc, count, senders, nullptr);
-      auto* msg = MultiMessage::make_message(alloc, body, epoch);
+      auto* msg = unmute_msg(alloc, count, senders, epoch);
       bool needs_scheduling = mutor->try_fast_send(msg);
       if (needs_scheduling)
         mutor->schedule();
@@ -1336,10 +1335,24 @@ namespace verona::rt
       alloc->dealloc<sizeof(MultiMessage)>(stub);
     }
 
+    /**
+     * Creates a `MultiMessage` that is never sent or processed.
+     */
     static MultiMessage* stub_msg(Alloc* alloc)
     {
-      // This is not a real message it is never sent or processed.
       return MultiMessage::make_message(alloc, nullptr, EpochMark::EPOCH_NONE);
+    }
+
+    /**
+     * Creates a `MultiMessage` with no behaviour. The given array of cowns may
+     * be null terminated, but the count must always be count of pointers that
+     * indicates the size of the allocation.
+     */
+    static MultiMessage*
+    unmute_msg(Alloc* alloc, size_t count, Cown** cowns, EpochMark epoch)
+    {
+      auto* body = MultiMessage::make_body(alloc, count, cowns, nullptr);
+      return MultiMessage::make_message(alloc, body, epoch);
     }
   };
 
--- a/src/rt/sched/cown.h
+++ b/src/rt/sched/cown.h
@@ -39,6 +39,8 @@ namespace verona::rt
 #endif
   }
 
+  static Behaviour unmute_behaviour{Behaviour::Descriptor::empty()};
+
   /**
    * A cown, or concurrent owner, encapsulates a set of resources that may be
    * accessed by a single (scheduler) thread at a time. A cown can only be in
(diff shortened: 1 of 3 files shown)


src/rt/sched/cown.h
```diff
@@ -967,16 +963,103 @@ namespace verona::rt
...
 
+    inline bool check_unmute_message(Alloc* alloc, MessageBody* msg)
+    {
+      if (msg->behaviour != nullptr)
+        return false;
+
+      for (size_t i = 0; i < msg->count; i++)
+      {
+        auto* cown = msg->cowns[i];
+        if (cown == nullptr)
```
rengolin, line 974: why would a cown be null here? would it be first? last? anything in between? multiple?

src/rt/sched/cown.h
```diff
@@ -967,16 +963,103 @@ namespace verona::rt
...
+        const bool high_priority = bp.high_priority();
+        yield();
+        assert(bp.priority() != Priority::Low);
+
+        // The cown may only be muted if its priority is normal and its epoch
+        // mark is not `SCANNED`. Muting a scanned cown may result in the leak
+        // detector collecting the cown while it is muted.
+        if (high_priority || (cown->get_epoch_mark() == EpochMark::SCANNED))
+        {
+          cown->schedule();
```
rengolin, line 1013: what if the list has multiple high priority? do you schedule all of them at the same time?

rengolin (commented):
No blocking comments, just trying to understand the code.

src/rt/sched/cown.h
Theodus: Any of the cowns with a high priority will be scheduled, yes.

src/rt/sched/cown.h
Theodus: The array of cowns from the MultiMessage that caused their muting is reused in the unmute message. If fewer than the original count of cowns were muted, then the array is null terminated. I'll add a comment in `apply_backpressure` that mentions this.

src/rt/sched/schedulerthread.h
```diff
@@ -99,9 +99,9 @@ namespace verona::rt
     size_t total_cowns = 0;
     std::atomic<size_t> free_cowns = 0;
 
-    ObjectMap<std::pair<T*, ObjectMap<T*>*>> mute_map;
     typename T::MessageBody* message_body = nullptr;
     T* mutor = nullptr;
```
mjp41, line 103: Similarly adding a comment on `mutor` would be good, while your here.

src/rt/sched/cown.h
```diff
@@ -902,7 +898,7 @@ namespace verona::rt
         {
           assert(Scheduler::local()->mutor == nullptr);
           Scheduler::local()->mutor = receiver;
-          receiver->weak_acquire();
+          Cown::acquire(receiver);
```
mjp41, line 901: Out of interest why has this changed from weak to strong.

src/rt/sched/cown.h
```diff
@@ -967,16 +963,108 @@ namespace verona::rt
...
+      {
+        alloc->dealloc(senders, count * sizeof(Cown*));
+        Cown::release(alloc, mutor);
+        return true;
+      }
+
+      if (muting_count < count)
+        senders[muting_count] = nullptr;
+
+      auto* body = MultiMessage::make_body(alloc, count, senders, nullptr);
```
mjp41, line 1061: You're using the `nullptr` behaviour here for the mute message?  I think it might be nice to introduce an actual named behaviour, and detect it later.  It makes the code more descriptive, and less likely to hide errors, where a nullptr behaviour ends up not crashing the system.

src/rt/sched/cown.h
```diff
@@ -967,16 +963,108 @@ namespace verona::rt
       return false;
     }
 
+    inline bool check_unmute_message(Alloc* alloc, MessageBody* msg)
+    {
+      if (msg->behaviour != nullptr)
```
mjp41, line 968: The Behaviour `nullptr` means `mute message`.

src/rt/sched/cown.h
```diff
@@ -923,7 +919,7 @@ namespace verona::rt
     /// Update priority based on the occurrence of a token message, or replace
     /// the token if it is not in the queue. Return true if the current message
     /// is a token.
-    inline bool check_message_token(Alloc* alloc, MessageBody* curr)
+    inline bool check_token_message(Alloc* alloc, MessageBody* curr)
```
mjp41, line 922: If a message without a body, the token message?

src/rt/sched/cown.h
```diff
@@ -1231,6 +1318,8 @@ namespace verona::rt
       }
 
       yield();
+      // Systematic::cout() << "Collect cown " << this << " rc: " << debug_rc()
```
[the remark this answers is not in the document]
Theodus: Fixed, thanks

src/rt/sched/schedulerthread.h
```diff
@@ -99,9 +99,9 @@ namespace verona::rt
...
+    ObjectMap<T*> mute_set;
```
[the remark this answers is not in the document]
Theodus: Comments added

src/rt/sched/cown.h
Theodus: Yes, it's a stub message

src/rt/sched/cown.h
Theodus: yes

src/rt/sched/cown.h
Theodus: We may need to send it an unmute message, so we need to keep it alive.

src/rt/sched/cown.h
Theodus: Done.

PR #391, opened 2021-02-12
TestRunner prototype

rengolin:
This is just an idea of what could replace the CMake runner in Python. I haven't done much around it before we're sure this is the path we want to take.

Does have:
* Multiple run lines with arbitrary command lines
* Substitution of `%s` with the file name
* Compares results with golden file

Doesn't have (yet):
- [x] A way to recursively inspect directories for new tests
- [x] Support for PIPE (command | OutputCheck)
- [x] CMake driver via ctest (I don't know how to do that but sounds like a really good idea)
- [ ] Create an update-dump script to update the golden files
- [ ] Migrate existng tests to the new framework

If this is what people want, I can continue working on the script to implement the remaining features and port our testing to this.


testsuite/golden/test.foo.out.2
```diff
@@ -0,0 +1,5 @@
+// RUN: grep foo %s
```
rengolin, line 1: This line will run and fail the diff because the golden file is wrong.

testsuite/golden/test.foo.out.2
```diff
@@ -0,0 +1,5 @@
...
+# RUN: cat %s
```
rengolin, line 2: This line will run and pass, as the golden file is good.

testsuite/golden/test.foo.out.2
```diff
@@ -0,0 +1,5 @@
...
+# RUN: bash %s
```
rengolin, line 3: This will fail on execution as `test.foo` is not a `bash` file.

rengolin (commented):
This is the output:
```
$ ./run_test.py test.foo 
/home/rengolin/devel/msrc/verona/testsuite/test.foo test 1 FAIL ['grep', 'foo', '/home/rengolin/devel/msrc/verona/testsuite/test.foo']
- /-/ RUN: grep foo %s-
?  -                  -
+ // RUN: grep foo %s
/home/rengolin/devel/msrc/verona/testsuite/test.foo test 2 PASS
/home/rengolin/devel/msrc/verona/testsuite/test.foo test 3 FAIL ['bash', '/home/rengolin/devel/msrc/verona/testsuite/test.foo']
/home/rengolin/devel/msrc/verona/testsuite/test.foo: line 1: //: Is a directory
/home/rengolin/devel/msrc/verona/testsuite/test.foo: line 4: unexpected EOF while looking for matching `''
/home/rengolin/devel/msrc/verona/testsuite/test.foo: line 6: syntax error: unexpected end of file
Invalid path to golden file for comparison /home/rengolin/devel/msrc/verona/testsuite/golden/test.foo.out.3
```

rengolin:
I have now moved the MLIR tests to the new script and all tests pass. The CMake logic is extremely simple: listing directories, adding tests that call the python script on the files, and that's it.

If everyone is happy with it, I can move all remaining tests to this scheme and remove all previous CMake kludge.

davidchisnall:
I'd really like us to avoid adding another Python dependency, we only agreed to take the OutputCheck one on the understanding that it would go away once we moved to FileCheck.

Looking at these files, it appears that every single one of these is running a single command and most of them are running the same command.  Is there a reason that we need to parse that from a `RUN` line in the file and can't just add them to a test category (manually, via a directory per test type, or via a pattern in the file name) based on that?

rengolin:
> I'd really like us to avoid adding another Python dependency, we only agreed to take the OutputCheck one on the understanding that it would go away once we moved to FileCheck.

I really see no problem in depending on Python at all. It's just for tests, not build, not execution. And almost all developers already have it installed, regardless of the platform.

Any logic programmed in Bash or CMake is way safer in Python, not least because it's an actual programming language (unlike CMake or bash), but also because it works similar across different platforms and has no quoting problems (unlike bash).


> Looking at these files, it appears that every single one of these is running a single command and most of them are running the same command. Is there a reason that we need to parse that from a `RUN` line in the file and can't just add them to a test category (manually, via a directory per test type, or via a pattern in the file name) based on that?

This is a restriction of the old testing framework that I carried away "as is" and is not a requirement of the new one. That was precisely why I needed to change it: I need to have multiple and more complex RUN lines. But I also don't want to break any other test while at it.

So all previous tests are simple with the same RUN line, and any new test can do whatever they want. We can (should) also improve the old tests to encompass more coverage by improving the existing RUN lines and adding more of them. But this is not a job for a simple migration PR.

rengolin:
After out call, we decided to go with a different approach: reuse the current CMake strategy but passing a config file to the binary instead of a source file. The config file can have any arbitrary configurations that the tool understands, which can be as simple as dumping the tokens inside `argv`, or as complex as YAML/JSON files, if needed.

This can create a scenario where we have multiple config files for a single source file, creating the equivalent of multiple RUN lines, and since the golden file will be per configuration, the comparison is also equivalent to this proposal. I'll keep this branch around in case we want to go back to it, but right now, I'll drop the PR for the different approach.

PR #393, opened 2021-02-16
Support for IO

marioskogias:
I rebased the code I started writing before christmas for IO. I implemented the token IO idea we talked about. I am creating the PR to re-initiate the discussion on how to expose IO and not to be merged in this way. That's the code I used to run for benchmarking. Some of this is necessary for the integration for DPDK.

### DIFF
--- a/src/rt/pal/pal_linux.h
+++ b/src/rt/pal/pal_linux.h
@@ -1,0 +1,105 @@
+#pragma once
+
+#if defined(__linux__)
+
+#  include <fcntl.h>
+#  include <netinet/in.h>
+#  include <netinet/ip.h>
+#  include <netinet/tcp.h>
+#  include <sys/epoll.h>
+#  include <sys/socket.h>
+
+#  define BACKLOG 8192
+
+namespace verona::rt
+{
+  class LinuxTCPSocket
+  {
+  public:
+    static void make_nonblocking(int sock)
+    {
+      int flags;
+
+      flags = fcntl(sock, F_GETFL, 0);
+      assert(flags >= 0);
+      flags = fcntl(sock, F_SETFL, flags | O_NONBLOCK);
+      assert(flags >= 0);
+    }
+
+    static int socket_config(int fd)
+    {
+      int one = 1;
+
+      make_nonblocking(fd);
+      if (setsockopt(fd, IPPROTO_TCP, TCP_NODELAY, (void*)&one, sizeof(one)))
+      {
+        perror("setsockopt(TCP_NODELAY)");
+        return -1;
+      }
+
+      return 0;
+    }
+
+    static int server_accept(int fd)
+    {
+      return accept(fd, nullptr, nullptr);
+    }
+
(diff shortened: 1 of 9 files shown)


rengolin:
You can convert this PR to a "draft" so it's clear it's an RFC, not a merge request. It also avoids accidentally merging to master.

src/rt/pal/pal_linux.h
```diff
@@ -0,0 +1,141 @@
...
+      if (!sock)
+      {
+        perror("socket");
+        return -1;
+      }
+
+      make_nonblocking(sock);
+
+      one = 1;
+      if (setsockopt(sock, SOL_SOCKET, SO_REUSEPORT, (void*)&one, sizeof(one)))
```
rengolin, line 78: You could have an additional wrapper for this, like:
```c
static int socket_option(int sock, int level, int option_name)
{
    int unused = 1;
    return setsockopt(sock, level, name, (void*)&unused, sizeof(unused));
}
```

src/rt/sched/cown.h
```diff
@@ -359,12 +370,18 @@ namespace verona::rt
       yield();
     }
 
+    void would_block_in_io()
```
rengolin, line 373: nit: name doesn't imply it's setting things. Perhaps something like `set_block_io` or something?

src/rt/sched/cown.h
```diff
@@ -719,6 +736,28 @@ namespace verona::rt
       return true;
     }
 
+    void check_io()
+    {
+      Cown* ready_cowns[MAX_EVENTS];
```
rengolin, line 741: Is it possible to have more ready cowns than MAX_EVENTS?

src/rt/pal/pal_linux.h
```diff
@@ -0,0 +1,141 @@
...
+
+      ev.events = EPOLLIN;
+      ev.data.ptr = (void*)cookie;
+      ret = epoll_ctl(efd, EPOLL_CTL_ADD, fd, &ev);
+      assert(!ret);
+
+      return 0;
+    }
+
+    static int check_network_io(int efd, void** ptrs)
```
rengolin, line 128: Why not `struct epoll_event*` instead of `void*`?

src/rt/sched/schedulerthread.h
```diff
@@ -285,7 +301,8 @@ namespace verona::rt
 
           // If we can't steal, we are done.
           if (cown == nullptr)
-            break;
+            continue;
```
rengolin, line 304: I don't understand this change

rengolin (commented):
@marioskogias thanks for the patch!

I understand this is work in progress, but would you mind trying to describe what you're trying to achieve with this change on the PR description? It'll make reviewing the code much easier.

Also, if you could add some short comments on the several new functions you created, even if simple, but especially if it does something fancy, so that we know what to expect for the arguments and results.

I have added some in-line comments, mostly to understand the code, not that there's something wrong with it.

PR #395, opened 2021-02-17
Move experiments/clang_interface to src/ffi and make it more generic

rengolin:
This PR takes the experiment on the clang interface and makes it more generic, easier to programmatically do things. It doesn't, however, add any new logic or change the internal logic of the AST handlers. In a way, this is mostly a refactoring, so we can work towards making use of the interface in the foreign function interface's sandbox remote procedure calls.

I've separated the internal logic into multiple classes, some helpers and created an example on how to use them in the testing program, `verona-ffi`. This test is still very brittle and can't do much other than running the (newly created) tests, but it gives us direction on where we want to go next.

Big things missing:
* Template specialisation still doesn't appear on the final file, need to implement them as functions maybe in the CU.
* Function creation still needs direct AST handling, exposing internal state from `CXXInterface`. It would be nice to make use of builders to allow users to create arbitrary AST nodes from expected semantics.
* Startup boilerplate is still very messy (diags, source managers, compiler creation, command line options, etc). I hope with time and more features we'll be able to factor more and more of those into nice private functions.

Tests are composed of a configuration file that has a command line to use with the headers on the same directory. This allows us to create multiple tests from the same header with different options. Tests were added with existing CMake infrastructure.

The code is a mess, but this isn't meant to be in production, but a starting point for different people to collaborate on the definition and implementation of the FFI.


src/ffi/CXXInterface.h
```diff
@@ -0,0 +1,645 @@
...
+        clang::StorageClass::SC_None,
+        nullptr);
+      func->setParams({arg});
+      return arg;
+    }
+
+    /**
+     * Create integer constant literal
+     *
+     * TODO: Can we have a generic literal creator or do we need one each?
```
davidchisnall, line 596: I think the only place we should be using this is for template specialisations, so combining them is probably not helpful.  The caller will be post-reification in Verona and the consumer of the result needs to know that it's an IntegerLiteral.

src/ffi/CXXInterface.h
```diff
@@ -0,0 +1,645 @@
...
+      llvm::APInt num{len, val};
+      auto* lit = clang::IntegerLiteral::Create(
+        *ast, num, getQualType(CXXType::getInt()), clang::SourceLocation{});
+      return lit;
+    }
+
+    /**
+     * Create a return instruction
+     *
+     * TODO: Can we have a generic instruction creator or do we need one each?
```
davidchisnall, line 610: We need very few things here.  The point of this function and the related ones are to generate functions that hide all of the C++ calling convention logic, like:
```c++
// Everything in this is a primitive type or pointer, with trivial alignment and layout requirements:
struct vector_int_pop_argframe
{
  int ret;
  std::vector<int> *v;
};
void call_std_vector_int_pop(vector_pop_argframe *args) {
  args->ret = args->v->pop_back();
}
```

We'll then generate a wrapper, so if we do something like this in verona:

```verona
try
{
  var i = someSandbox.vec.pop_back();
}
catch (oops: SandboxExploded)
{
}
```

On the Verona side, we'll generate a struct containing the vtable index of the exported function, a pointer to the vector, and space for the result and pass it into the sandbox.  On the sandbox side, we'll generate an LLVM function roughly equivalent to:

```c
void __verona_sandbox_dispatcher(int fn, void *argframe) {
  switch(fn)
  {
    case {whatever the vtable index for std::vector<int>::pop_back is}:
       call_std_vector_int_pop(argrame);
       break;
    ....
  }
}
```

Eventually, the dispatcher function will expose an entire sequence of things in a `try` block as a single vtable entry, so we avoid multiple domain-transition round-trips but in the MVP any function that we call (including virtual functions) will be exposed this way (so we don't need to understand C++ vtables.  If `pop_back` were `virtual` then we'd generate exactly the same code.  This means that the only thing that we actually need are:

 - Call a function / method.
 - Reference fields in a structure.
 - Perform any necessary casts.

This is *far* simpler than a generic instruction builder.  The most complex part is likely to be the cast logic.

Note that this return function actually does the wrong thing: we will want it to store the return in a struct field and return nothing.

src/ffi/Timer.h
```diff
@@ -0,0 +1,57 @@
...
+      std::atomic_signal_fence(std::memory_order::memory_order_seq_cst);
+      clock_gettime(clock, &end);
+      std::atomic_signal_fence(std::memory_order::memory_order_seq_cst);
+      auto interval_from_timespec = [](timespec t) {
+        return seconds{t.tv_sec} + nanoseconds{t.tv_nsec};
+      };
+      auto elapsed =
+        interval_from_timespec(end) - interval_from_timespec(start);
+
+      // TODO: Avoid printfs on headers
```
davidchisnall, line 49: Not urgent, but all of this should probably be using `std::chrono` things and the existing flight recorder logic for logging (this is a copy-and-pasted version of a quick hack I wrote many years ago).

src/ffi/verona-ffi.cc
```diff
@@ -0,0 +1,257 @@
...
+#include <iomanip>
+
+using namespace std;
+using namespace verona::ffi;
+using namespace clang;
+namespace cl = llvm::cl;
+
+namespace
+{
+  // Test function (TODO: make this more generic)
```
davidchisnall, line 16: I'd suggest the opposite.  The amount of logic required to pull out a handful of configuration things for testing here is significant, why not just make these individual unit test binaries?  We'll want some (many!) integration tests later that generate LLVM IR from Verona programs that use C++ types, I'm not convinced that it's worth building a generic test program here.

src/CMakeLists.txt
```diff
@@ -54,3 +54,4 @@ add_subdirectory(compiler)
 add_subdirectory(interpreter)
 add_subdirectory(stdlib)
 add_subdirectory(mlir)
+add_subdirectory(ffi)
```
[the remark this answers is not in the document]
rengolin: Fair enough, what do you suggest?

src/ffi/CMakeLists.txt
```diff
@@ -0,0 +1,45 @@
...
+    clangLex
+    clangParse
+    clangRewrite
+    clangRewriteFrontend
+    clangSema
+    clangSerialization
+    clangTooling
+    clangCodeGen)
+
+install(TARGETS verona-ffi RUNTIME DESTINATION .)
```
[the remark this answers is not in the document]
rengolin: All programs that we need to run as part of the CI are "installed". This is not a reflection of what the final install directory will be. My plan is to move the tests to `veronac` as soon as we're able and remove this binary, well before anything is to be considered "production".

src/ffi/Timer.h
rengolin: Honestly, this is here mostly as legacy, as I'm not using it at all. I'd be happy to remove now. If we need something similar later, we can always rewrite it with `std::chrono` (as I have done a few times already and is trivial).

src/ffi/verona-ffi.cc
rengolin: Right now, this is a hack in either direction and is just here as a "user"/"driver" of the internal API. This is neither a generic program nor a test file. Given this won't be used externally, there's little gain from having a truly generic "program", but not a more generic API.

I want to create proper unit-tests for these, but tests that express how we'll use the API from the sandbox side, which will need a more polished interface than we currently have. For example, I want to stop exposing the internal fields of `CXXInterface` and `Compiler` to the user and provide builders and helpers.

This will not only help writing tests that clearly mark the semantics and usability we need, but also simplify the sandbox code when creating the functions and querying the types it needs.

Right now, though, this is just a hack to show the API works in some way.

src/ffi/CXXInterface.h
```diff
@@ -0,0 +1,645 @@
...
+    /**
+     * Creates new AST consumers to add the AST back into the interface.
+     *
+     * Each traversal consumes the AST, so we need this to add them back
+     * for the next operation on the same AST. This is the way to expose
+     * the interface pointer so that we can update it again when needed.
+     *
+     * Use:
+     *  CompilerInstance->setASTConsumer(factory.newASTConsumer());
+     *  CompilerInstance->setASTContext(ast);
```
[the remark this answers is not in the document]
rengolin: That part in clang is a bit of a mess and I might have not understood it correctly. Feel free to suggest a better comment.

src/CMakeLists.txt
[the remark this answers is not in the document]
davidchisnall: Interop?  Foreign?

src/ffi/Timer.h
davidchisnall: If it's not used anywhere, removing it is probably a good idea.  It would be quite nice to have a generic thing like this: I found it very useful for seeing which bits of the clang logic were taking the most time.  It's not difficult to rewrite in the future if required.

src/ffi/CXXInterface.h
rengolin: So, IIUC, we'll need:
* Create structure types (either for Verona types mapped to C++ or internal compiler ones)
* Read / Write to their fields (this includes reading a pointer, which can be to a function)
* Direct / Indirect calls
* Simple control flow (switch / case)

To be clear, I didn't mean "a generic builder" like LLVM's `IRBuilder` (even though I have said _just that_ on one of the meetings), just some class separated from the `CXXInterface` that knows how to lower all the instructions we need (which is a small subset of all instructions possible).

src/CMakeLists.txt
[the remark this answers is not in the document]
rengolin: I'll use "interop". We can change it later if anyone has a better idea.

src/ffi/Timer.h
rengolin: File removed

rengolin:
@davidchisnall I have addresses all your comments (mostly on comments) and have fixed some MSVC issues building Clang.

It should be fine for an initial prototype, so we can start working on the RPC between the sandbox and the Interop.

davidchisnall (approved):
This will need more work as we move it into the rest of the compiler, but it's fine for now.

PR #397, opened 2021-02-22
Use a service principal that has access to AnyBuild to authenticate in CI

brphelps:
I added a service principal via variable group to the AnyBuild build for LLVM -- this service principal is something we'd be comfortable handing off, but have just temporarily stored in a keyvault that is under our subscription. 

These updates should use that principals service credentials to consume anybuild.


dmgonch (approved):
:shipit:

rengolin (approved):

PR #401, opened 2021-02-26
Interoperability Layer Document

rengolin:
This is a document describing the interoperability layer, from parsing external language, through integrating it with Verona code, to safely running them in a sandbox environment.

The syntax is not the core subject, so don't pay too much attention to that. The most important thing in this document is to make sure we all agree with the overall implementation details.

What I wrote is my interpretation of the multiple discussions we had so far, so make sure the text means what you think it means, to clear up any misunderstanding from my part.


docs/internal/interop.md
```diff
@@ -0,0 +1,351 @@
...
+and feels like Verona code. We want to abstract away the entirety of the sandbox
+functionality, the foreign function dispatch and serialisations, the execution
+models, etc.
+
+The syntax is still largely undefined, but this is an example of using the C++
+`std::vector` class inside Verona as pseudo-Verona code:
+
+```c
+// include/import the c++ code as a Verona module
+type StdVec = "c++::vector";
```
davidchisnall, line 19: Is this the latest version of Verona import syntax?

We should never write code that names C++ types as anything other than Verona types that are nested inside the class (package) the encapsulates a specific C++ library.  Something like:

```verona
import foreign_lib = some.package.foreign;
```

That will live in our filesystem as `some/package/foreign`, which will contain a `verona_interop.json` (or some better name), which will include everything required to generate verona types (C++ header, library, compilation flags).  If that header includes `std::vector` then `foreign_lib::std::vector` is a valid Verona type name and we can instantiate it (though its constructor, as exposed to Verona, is defined only on sandboxed regions).

docs/internal/interop.md
```diff
@@ -0,0 +1,351 @@
...
+
+// Create a new region (we know it's foreign, so it's a sandbox)
+reg = new StdVec;
```
davidchisnall, line 22: We still need to specify the kind of sandbox, because we're likely to have multiple sandboxing mechanisms available.

docs/internal/interop.md
```diff
@@ -0,0 +1,351 @@
...
+
+Each execution model (child process, web assembly, etc) will need explicit
+support for deploying the functionality at run time, hopefully not individually
+bound to the source language (N-to-M).
+
+But the overall architecture will be the same.
+
+### Parser
+
+1. Upon finding a file include that has foreign origin, the appropriate foreign
```
davidchisnall, line 65: Please can we avoid talking about 'file include's?  It seems like a minor nit, but it's actually important:

In Verona, we do not include other source files, we import modules.  Modules are classes that may contain other classes, methods, interfaces, and so on.  A Verona module is defined by a set of `.verona` files.  A foreign module is defined by some instructions that give the compiler a way of constructing a set of Verona types and generating FFI stubs.

For C/C++, this involves an include file.  For .NET, it might involve parsing the metadata structures in the assemblies.

This is important in the structure of the compiler, because whenever we encounter an import for a module for the first time in a compilation, we need to go and parse the module code anyway.

docs/internal/interop.md
```diff
@@ -0,0 +1,351 @@
...
+   language parser (clang for C++) will be asked to parse the file and generate
+   its own AST.
+2. A Verona module will be created, with all foreign builtin types.
```
davidchisnall, line 68: We may want to surface this as a generic module that is imported by the specific library.  For example, a `verona.interop.c` that includes things like `float` and so on.  We may also want to make this come from the sandbox instance and make the modules generic on the sandbox instance (for example, a wasm sandbox `long` and a LP64 process-based sandbox `long` are not the same type).

docs/internal/interop.md
```diff
@@ -0,0 +1,351 @@
...
+4. A dispatcher will be created to call those functions with serialised
+   arguments by an index that is associated when they're created.
+5. A bi-directional RPC layer is created between the Verona code and the sandbox,
+   where calls to external functions go one way, and calls to allocation, system
+   calls, etc. go the other way.
+6. The appropriate object files are created and linked with the main executable.
+
+Not all functions will be representable in Verona, not all structure layouts will
+be easily known at the Verona level (special hardware ABI), so errors may still
+occur here.
```
davidchisnall, line 103: Can you give an example of the kind of thing here that might be a problem?  I am expecting that we'll only ever pass primitives / pointers through the translation layer.  If there's a struct with a funky layout, we'll handle it in the same way as any other struct: Generate RPC calls for reading / writing the fields and then seeing if we can inline them in the caller safely later (i.e. if, after lowering them to LLVM IR, we can insert checks for load and store instructions and if there are no calls).

docs/internal/interop.md
```diff
@@ -0,0 +1,351 @@
...
+
+The compiler is responsible for emitting errors that inform the programmer what
+type/function cannot be represented and potentially why. The programmer is
+responsible for implementing a wrapper function that abstracts away the
+complexity of the language in a public function that Verona can call.
+
+Language support will be always evolving. This is expected to be a balance
+between "best effort" and "appropriate support".
```
davidchisnall, line 111: In particular, this should be test driven: We should document what works and we should only implement things needed to get examples to work.  Once we can do MLIR things in Verona, we can probably declare success.

docs/internal/interop.md
```diff
@@ -0,0 +1,351 @@
...
+
+### Run time
+
+1. When the sandbox is initialised, its allocator sets up a new isolated heap
+   that is used for all allocations in the sandbox (including Verona foreign
+   objects that belong to that region).
+2. When the foreign code makes calls to allocate memory or run potentially unsafe
```
davidchisnall, line 118: This is a bit misleading.  snmalloc runs inside the sandbox as well as outside, it *may* (depending on the sandboxing model) need an upcall to allocate superslabs.

docs/internal/interop.md
```diff
@@ -0,0 +1,351 @@
...
+   operations (file descriptors, system calls, etc.), the RPC layer will check
```
davidchisnall, line 119: Also a bit over generalised.  Creating new file descriptors is unsafe, most FD operations are safe.  A lot of system calls are allowed directly in process-based sandboxing because they don't affect anything other than explicitly delegated file descriptors or the sandboxed process' state.

docs/internal/interop.md
```diff
@@ -0,0 +1,351 @@
...
+   them and forward if explicitly marked as an upcall or if it's a direct
+   sandbox functionality (webasm, CHERI), otherwise fail and close the sandbox.
+3. When Verona code makes calls to the foreign library, the compiler-generated
+   serialisation code will transform the arguments and return value into a
```
davidchisnall, line 123: I'd prefer marshalling to serialisation here.  We're not really serialising, we're just putting the values in a shared buffer.

docs/internal/interop.md
```diff
@@ -0,0 +1,351 @@
...
+that function to the dispatcher, giving it its own (auto-incremented) index.
+
+Additional code needs to be generated to serialise the Verona arguments and
+return values into a buffer, so that the dispatcher can be generic, only taking
+an index and a memory buffer.
+
+For C++, further code will be created (via AST construction and LLVM code
+generation) if the types involved are parametrised. All template arguments are
+taken into consideration and a specialisation is defined and generated in a
+wrapper layer.
```
davidchisnall, line 150: Probably worth noting that, at least for v1, we don't allow Verona generic parameters to be passed to template instantiations.  This means that all possible template instantiations that a Verona generic may use must exist *before* reification, which means we have the same guarantees for Verona generics that use C++ templates as we do for other Verona generics (i.e. they either type check and work for all instantiations or they don't type check).

We may relax that later, but I believe we should probably do that only once C++ Concepts have landed and then we can map concepts into Verona interface types and keep the same guarantees.

docs/internal/interop.md
```diff
@@ -0,0 +1,351 @@
...
+
+## Child process execution environment
+
+Our initial implementation is to separate the sandbox execution by creating a
+child process. This is not strictly speaking a sandbox, but gives us an easy
+way to test the rest of the framework without much additional work. Our aim is
+to use WebAssembly as our default sandbox environment, but that will take time
```
davidchisnall, line 157: Is it?  Wasm is a distribution format, not a sandboxing environment.

docs/internal/interop.md
```diff
@@ -0,0 +1,351 @@
...
+generation stage above, and the buffer is the contents of the arguments and a
+place for the return value.
+
+Argument decomposition will be done by the sandbox code that will call the actual
+function with the actual arguments and, if there is a return value, set it on
+the right place of the buffer and return.
+
+### Memory Management
+
+To avoid the sandbox from requesting memory in the wrong places (accidentally
```
davidchisnall, line 176: This is... not a very close approximation of what's going on.

For integer-pointer environments (processes, SFI) we carve out a chunk of the virtual address space for the shared heap.  We use some explicit calls to transfer ownership of large regions within this (superslabs) between the Verona code and the sandbox and then both sides have their own snmalloc `Alloc` instances that are able to allocate memory.  We don't need an RPC per malloc, that would be prohibitively expensive.  We started by making malloc across security contexts cheap and then built a lot of the other mechanisms on top of this.

docs/internal/interop.md
```diff
@@ -0,0 +1,351 @@
...
+running the C++ snippet in the beginning of this document. We'll go line by line
+and describe what the compiler and the generated code will do for each of them.
+
+### Compile time
+
+```c
+type StdVec = "c++::vector";
+```
+
+* Call the C++ driver (clang wrapper) to parse the include file (and all its
```
davidchisnall, line 203: There are earlier steps:

 1. Find the module that the import directive corresponds to.
 2. Read the module description file and discover that it's a C++ module.

docs/internal/interop.md
```diff
@@ -0,0 +1,351 @@
...
+  includes) and keep as a C++ AST.
+* Create a Verona class called `StdVec` with C++ built-in types defined as Verona
```
davidchisnall, line 205: First, create a Verona class for the foreign module.  Then, lazily, construct the types within it.

docs/internal/interop.md
```diff
@@ -0,0 +1,351 @@
...
+  types (ex. `int` as `S32`, `size_t` as `U64`).
+* Expose an interface for the parser to query symbols (types, objects and
+  functions).
+
+```c
+reg = new StdVec;
+```
+
+* The parser recognises the region as a C++ sandbox.
```
davidchisnall, line 214: The parser recognises the type as coming from a specific module.  It queries that module for types within that module as they are used.

docs/internal/interop.md
```diff
@@ -0,0 +1,351 @@
...
+* Lower code to create a region with a sandbox allocator (RT calls).
+* Lower code to create the sandbox memory area (RT allocator calls).
```
davidchisnall, line 216: The current interfaces at the sandbox layer take the sandbox maximum heap size and the library to the loaded.  The former of these should come from the sandbox type, the latter is implicit from the module.

docs/internal/interop.md
```diff
@@ -0,0 +1,351 @@
...
+* Create a wrapper object for function dispatch and general sandbox utilities.
+* The region is bound to the variable `reg`.
+
+```c
+obj = new std::vector[int] in reg
+```
+
+* The parser recognises the region and tries to create a new object in the
+  sandbox region by using the C++ driver.
```
davidchisnall, line 225: This might be what happens post-optimisation, but the initial code will do a cross-domain call for every in-sandbox operation.  This will generate an in-sandbox function that looks like this:

```c++
void {random non-colliding name}(struct args *args)
{
    *args->ret = new std::vector<int>();
}
```

and generate a call on the Verona side to invoke that wrapper.  If, after some optimisation on the generated code in the sandbox side, we discover that this is just `malloc` and store constants into the newly `malloc`'d memory, we can hoist it into the non-sandboxed code, but that's an optimisation.

We don't involve the allocator from the Verona side unless we do that optimisation (after MVP).

docs/internal/interop.md
```diff
@@ -0,0 +1,351 @@
...
+* Assemble the foreign call signature (namespace, template args):
+  `std::vector<int> std::vector<int>::vector<int>()`
+* Instantiate the signature as a template specialisation in the AST, using
+  the query interface to validate the code and get an actual implementation.
+* If available, create the Verona equivalent inside the `StdVec` class and
+  Lower call to that function. Otherwise emit an error and stop processing.
+* The Verona signature would be something like
+  `create() : StdVec::std::vector[S32]`.
```
davidchisnall, line 233: I don't think it would be `S32`, it would be something like `ProcessSandbox::int` or `WAsmSandbox::int`.

docs/internal/interop.md
```diff
@@ -0,0 +1,351 @@
...
+* On the recently created `StdVec` Verona function, lower Verona code to call the
+  dispatcher with the index and the buffer and to extract the return value.
+* Bind the result of the Verona call to the variable `obj`.
+
+
+```c
+obj.push_back(42)
+```
+
+* The parser recognises the object is in a sanbdox region.
```
davidchisnall, line 246: The parser doesn't know anything.  The type inference pass should be sufficient to tell us that this is a foreign type in a specific kind of sandboxed region.  We don't need region analysis.  The main thing we get from region analysis is that we disallow a load of unsafe things.

docs/internal/interop.md
```diff
@@ -0,0 +1,351 @@
...
+* Assemble the foreign call signature (namespace, template args):
+  `void std::vector<int>::push_back(int)`
+* Same as above, creates a new Verona function in the `StdVec` class that will
+  serialise arguments and return value and call the dispatcher.
+* The Verona signature would be something like
+  `StdVec::std::vector[S32]::push_back(S32)`.
```
davidchisnall, line 252: As above, we don't want to use `S32` because that binds the Verona code to a specific sandbox ABI, which has all sorts of problems.

davidchisnall (changes requested):
I've left some comments.  Some style things:

 - Please can you use one line per sentence?  It's hard to review (and especially hard to provide suggestions or review diffs) when line breaks are randomly in the middle of sentences and text reflows in one sentence because you've edited an earlier one.
 - Please use numbered lists for sequences of actions, bulleted list for unsequenced things.

docs/internal/interop.md
rengolin: Oh, that's not what I meant. I meant it will call the constructor and place the object in the same sandbox region because the line specifies `in reg`. I'll make that clear.

docs/internal/interop.md
rengolin: I just mention `S32` to give an idea what kind of Verona type it will map to. Of course it won't always be a signed 32-bit integer.

I don't think we can have a single `ProcessSandbox::int` for all targets as it depends on the target architecture and the foreign language standard rules.

docs/internal/interop.md
rengolin: Each signature would be specific to the program/foreign language/hardware architecture. The same program and the same library compiled on a different platform could have a completely different type.

Calling it `ProcessSandbox::int` or using `S32`, for a specific instance of a program, is irrelevant if they both have the same sign, size, alignment and semantics.

It could be misleading having a single `ProcessSandbox::int` type name with multiple sizes, just as it is in C++, but that will have to adapt to multiple different foreign languages.

docs/internal/interop.md
rengolin: I think there will always be things we don't implement. I think success will be different depending on what you mean. v1 success is making it work at all with something non-trivial. Language success is not having to rely on ugly wrappers (like Java's JNI) for every little trip outside of the Verona's confines.

docs/internal/interop.md
davidchisnall: I think the key thing here is that there is no special syntax for 'this is a foreign class' at the import site.  There is just syntax (whatever it may be) for importing a module.  From the world outside of that module, it looks just like a Verona module, though the nested classes inside it have constraints on the kind of region that they can be created inside (i.e. only sandboxed regions, possibly only a specific kind of sandbox region, depending on how the code is provided to Verona).  We have an explicit design decision to not add any custom syntax for foreign code.

docs/internal/interop.md
davidchisnall: Not yet, but it will be the same syntax that we use for creating objects in regions of specific kinds.  I think @sylvanc had some straw-man syntax, once the compiler can handle the difference between traced and bump-allocated regions then we will use the same syntax.  The key point is that, from the perspective of this code, creating a new process-based sandboxed region is the same syntactic construct as creating a new region of any other explicit kind.

docs/internal/interop.md
davidchisnall: This sentence to me implies things that will be correctly parsed but where we generate incorrect code because we don't have enough information to implement it correctly.  I can't think of anything where I'd expect that to happen because we either support a feature or we don't.  I agree that we won't support everything, but I don't think we should be enabling any C/C++ language construct until we are happy that it either works or reports an error.

docs/internal/interop.md
davidchisnall: e.g. is the accepted English (well, technically, Latin) abbreviation.  I think the point that you're trying to make is that the sandbox will make cross-domain calls back into the Verona environment for anything that requires privileges that are not explicitly delegated to the sandbox.

docs/internal/interop.md
davidchisnall: Possibly `ProcessSandbox::C::int`, but this is part of the sandbox-environment ABI, which is not necessarily the same for all instantiations of a foreign library.  We should be able (eventually) on the same system to instantiate:

 - An SFI sandbox, possibly via WebAssembly
 - A process-based sandbox
 - A CHERI CheriABI sandbox
 - A CHERI A64-ABI sandbox

On Morello, we'd end up with:

 - `WAsmSandbox::C::intptr_t` is a 32-bit integer, which we can losslessly round-trip to and from a `S32`.
 - `ProcessSandbox::C:intptr_t` is a 64-bit integer, which we can losslessly round-trip to and from `S64`.
 - `CheriABISandbox::C::intptr_t` is a capability (and not something we currently have a Verona type for - something we need only for C interop).
 - `CheriLegacySandbox::C::intptr_t` is a 64-bit integer, which we can losslessly round-trip to and from `S64`.

Ideally, we don't want these to be properties of the foreign module, because the same module should be a generic class that takes that is instantiated with one of any of these sandboxing models (or some per-module subset if we don't have a version of the library for all of them).

docs/internal/interop.md
davidchisnall: `S32` is the same size on all platforms.  `ProcessSandbox::int` requires an explicit cast to and from *any* Verona native type.  This is necessary to be able to write portable the Verona code that handles these (I've made this mistake before, I don't want to do it again).

rengolin (commented):
Github review doesn't seem to be able to join conversations that are still relevant with code that has been updated, so it doesn't show the "outdated" comments in the review (files changed"), but it still shows in the "conversation" panel.

I wish Github would let us do a review (ie. answer multiple questions at once) in the conversation panel. This looks to me like a simple and extremely obvious change.

But alas, here are the answers to the ones I could change, and I'll change the others one by one, so people will get a bit of spam. Also, some of them will have changes to the code, others won't, and unfortunately, there's no telling Github to behave, so more comments can be dropped from the "review".

docs/internal/interop.md
rengolin: Ah, when I said "error may occur" I meant "report an error". I'll clarify.

docs/internal/interop.md
rengolin: I agree with everything, I just think this is too low level to be in this document. I'll just change `S32` to `Sandbox::int` and mention that it's sandbox specific.

docs/internal/interop.md
rengolin: Changed both to `Sandbox::int` and `Sandbox::size_t`.

rengolin: I think you're misunderstanding my point. To be explicit, I didn't mean to force the size here at a source level. This would be madness.

I meant in a particular platform (OS, arch, ABI) the size is `S32` so a binary of that platform, `Sandbox::int` will end up being with an identical representation as `S32`, so my example had that to mean what the actual size would be in this fictitious platform. This is only referring to this particular "already compiled" representation, and not the source code.

rengolin:
Converting to draft as I'll rewrite the document (almost entirely) based on the feedback here and meetings. I have identified some serious miscommunication issues with the text that need clarifying.

Once we're happy with the content, I'll move it out of draft and we can review the text.

rengolin:
@davidchisnall, this is the first iteration of the complete rewrite. 

The main changes are:
* Technical details in line with feedback here and in meetings.
* Changed the syntax of the example to be closer to what we'll actually have.
* Separated into four stages: parser, type checker, code gen, run time, and explained in detail each stage.
* C++ case re-written.
* Line-by-line example re-arranged to put both compile-time and run-time together, for each line.

A lot of the text and details have changed, so it probably needs a whole new read. Sorry about that.

Also, I proactively apologise for textual issues, typos, broken prose, etc. Feel free to correct them all.

rengolin:
@sylvanc @mjp41

rengolin:
Readable version: https://github.com/rengolin/verona/blob/docs/docs/internal/interop.md

davidchisnall (approved):

PR #402, opened 2021-03-01
Snmalloc updates

mjp41:


rengolin (approved):
Surprisingly simple. I think this makes it easier to know what the submodules are upfront.

src/rt/CMakeLists.txt
```diff
@@ -2,7 +2,7 @@ cmake_minimum_required(VERSION 3.8)
 project(verona C CXX)
 
 set(SNMALLOC_ONLY_HEADER_LIBRARY ON)
-add_subdirectory(external/snmalloc EXCLUDE_FROM_ALL)
+add_subdirectory(../../external/snmalloc ./external/snmalloc EXCLUDE_FROM_ALL)
```
plietar, line 5: Could this line be moved to the top-level `src/CMakeLists.txt`, with the other add_subdirectory?

mjp41: Not, if we want to be able to build the runtime as a self contained thing.  Currently, we can point cmake at src/rt, and it will correctly build all of the runtime, and not any of the other infrastructure.  We could move that functionality.  I am keen to keep the runtime fairly clear of the rest of the code for potential porting.

plietar: Fair enough

PR #405, opened 2021-03-08
Initial Support for IO

Theodus:
These changes add initial support for IO in the Verona runtime, based on #393. This only includes support for TCP sockets for now since the programmer interface currently makes assumplions that are not safely enforced.

Unresolved issues:
- Scheduler threads should probably block for some time when they are waiting for IO to make progress. I'm not sure how we should make the timeout though.
- An assumption is made that an IO cown that would block during a behaviour will have exactly one behaviour in its queue when the running behaviour ends. We should discuss how to prevent the user from scheduling multiple behaviours on a cown (possibly resembling [Pony's notify pattern](https://patterns.ponylang.io/code-sharing/notifier.html) or Sean's [lori](https://github.com/seantallen-org/lori)).
- This does not link the backpressure system to the outside word yet (via TCP backpressure).
- There are two boolean fields added to the Cown object that will need to be removed or stashed as bits in other fields.


Theodus:
I'll fix the CI errors later this week

src/rt/pal/pal_linux.h
```diff
@@ -0,0 +1,398 @@
+#pragma once
+
```
davidchisnall, line 2: This file is interleaving a lot of Linux-specific and generic-POSIX code, which will make bringing up the next platform painful.  The `sandbox_pal` branch includes [platform abstractions that hide the minimal bits that we need to paper over the differences between platforms](https://github.com/microsoft/verona/tree/sandbox_pal/experiments/process_sandbox/platform).  In particular, take a look at the [poller_epoll.h](https://github.com/microsoft/verona/blob/sandbox_pal/experiments/process_sandbox/platform/poller_epoll.h) and [poller_kqueue.h](https://github.com/microsoft/verona/blob/sandbox_pal/experiments/process_sandbox/platform/poller_kqueue.h) files.  These are very close to what you want and could easily be made more general.

src/rt/pal/pal_linux.h
```diff
@@ -0,0 +1,398 @@
...
+  };
+
+  static void make_nonblocking(int sock)
+  {
+    int flags;
+    flags = fcntl(sock, F_GETFL, 0);
+    assert(flags >= 0);
+    flags = fcntl(sock, F_SETFL, flags | O_NONBLOCK);
+    assert(flags >= 0);
+  }
```
davidchisnall, line 201: This is generic POSIX code.

src/rt/pal/pal_linux.h
```diff
@@ -0,0 +1,398 @@
...
+      int res =
+        setsockopt(sock, SOL_SOCKET, SO_REUSEADDR, &opt_val, sizeof(opt_val));
+      if (res == -1)
+        return -1;
+
+      return sock;
+    }
+  };
+}
+#endif
```
davidchisnall, line 398: This also looks like generic POSIX code.

PR #407, opened 2021-03-12
Refactor CXXInterface and tests

rengolin:
Splitting parts of `CXXInterface` into `CXXQuery` and `CXXBuilder`. Query only finds types and functions that exist in the foreign AST while the builder can create new AST representations, including type specialisation and functions.

This allows us to grow the three pieces independently:
 * Core interface: Knows how to build the compile units, parse files, etc.
 * Query interface: Knows all about AST types and functions.
 * Builder interface: Know how to build new AST nodes.

The part that lowers AST to LLVM IR is not yet in this design and will be done later. We don't need to query anything at that stage, and we'll trust that the types and functions created in the AST are correct and interface with the Verona AST in the right way.

Also refactor the tests, extract all functionality into those classes and simplify output to be able to test more than just return values.


davidchisnall (approved):
Looks fine for now.  I believe this is mostly just a cleanup of my code from experiments.  I expect that it will need further work once you start integrating it with the Verona compiler properly.

rengolin:
For now I'm just representing the API in the test binary. This is what the compiler will use to query types, fields and build functions.

The API is still in constant and radical motion, and a lot of code shuffling will still occur, at least until the compiler actually starts using it and we're happy with the interface.

Only then we'll have compiler tests with Verona and C++ files building the right IR.

Issue #394, opened 2021-02-17
Work required on process-based sandbox


davidchisnall:
This is a tracking issue for the various bits needed to turn the process-based sandbox code into something useable by Verona (extracted from #379):

 - [x] Clean up the structure of the code a bit (we still have a `libsandbox.cc` that is `#include`d in another file as well as being compiled separately, for example.  A lot more things should be pulled into separate files).
 - [x] Extract all of the `#ifdef __linux__` and `#ifdef __FreeBSD__` into a sandbox PAL.  It's a little bit clunky at the moment, it will get much worse when we have more than two platforms.
 - [x] Replace the condition variable / mutex code with something more robust.
 - [x] Separate the C++ function export stuff into a a test harness.
 - [x] Make the alloc outside the sandbox validate all pointers that it follows within the sandbox are within the sandbox.
 - [x] Refactor the bootstrap code to use the snmalloc slow path code (this requires tweaking ).
 - [x] Add privilege-elevation callbacks for opening files / sockets based on some policy.
 - [ ] Add a useful interface for the Verona runtime to use.
 - [x] Add a Linux sandboxing policy implementation.
 - [x] Write unit tests.

I think this is currently in a state where we could at least try to build it in CI, if not actually run it.

PR #408, opened 2021-03-19
Update sandbox code, clean up, and rebase.

davidchisnall:
This includes a few cleanups:

 - This code no longer depends on libc using snmalloc and so doesn't use
   the late-binding external pagemap support.
 - Linux pipes are not bidirectional, so pipes have been replaced with
   `AF_LOCAL` sockets where necessary.
 - All updates to the pagemap or requests for large allocations now go
   via a (trivial!) RPC mechanism to the host process.  This should also
   help define the set of upcalls required for other sandbox
   implementations.
 - The base address and length of the shared region are passed to the
   child via an environment variable, rather than in arguments.  This
   allows the child to map the shared memory region on the first call to
   any allocation function, rather than in `main` and so we no longer
   need to do the dance of replacing the thread allocator.
 - The bootstrapping process uses the snmalloc slow-path bootstrapping
   mechanism.
 -

There is now a platform-abstraction layer for the process sandbox code.
Currently the platform abstraction includes:

 - The type for an OS resource handle and a wrapper that provides
   lifetime management.  Implemented for UNIX file descriptors.
 - A poll loop that waits for one or more handles.  Implemented
   for `epoll` and `kqueue`.
 - Anonymous shared memory objects that are mapped naturally aligned.
   The shared memory objects are implemented with generic POSIX
   `shm_open`, `shm_open` with the FreeBSD `SHM_ANON` extension, and
   Linux's `memfd_create`.  The mapping is done via either a portable
   POSIX `mmap` of a large region to find a smaller aligned region
   inside and then a `MAP_FIXED` mapping over this, or with the
   non-standard (though portable to many POSIX systems) `MAP_ALIGNED()`
   flag.
 - A socket-pair abstraction for sending messages and handles over an
   inter-process channel.
 - A one-bit semaphore that is safe to use between mutually distrusting
   processes and can live in a shared memory region.  This is
   implemented with `futex` on Linux and `_umtx_op` on FreeBSD.
 - Child-process management using `vfork` or `pdfork`.
 - Sandbox abstractions using Capsicum and seccomp-bpf that aim to
   prevent the child process from accessing the global namespace at all,
   unless the parent explicitly delegates handles to it.
 - Wrappers around the information delivered in signals for sandbox
   policy violations that allow the system call number and arguments to
   be extracted and the return value re-injected.

The process-creation code is still very POSIX-specific.  It's not clear
how much code we are going to be able to share between POSIX and
non-POSIX systems and so for now I am focusing on factoring out the
variations between POSIX systems, rather than building a completely
portable abstraction layer.

On Windows, for example, we won't inherit handles, we'll use
`DuplicateHandle` to add handles to the child process and so the
socket-pair abstraction will need to carry the new handle value in-band.
We also won't be able to completely replace `malloc` and friends and
have a shared heap and it's not clear whether the Windows
process-sandboxing mechanisms allow us to approximate a Capsicum model.

The C++ API is now separated out.  This may be something that we expose
as a reusable component, but for now it's used only for testing.  The
C++ wrapper code will not be used for Verona, the generic code now only
contains things that Verona will need and is easier for veronac to plug
into.

The ExportedLibrary class is now entirely static (we only have one per
exported library) and the ExportedLibraryPrivate class is gone and all
of its logic moved into an anonymous namespace in the library runner.

Sandboxes now just need to provide init and dispatcher functions.  The
init function doesn't need to do anything, the dispatcher takes an int
containing the vtable index and a pointer containing the argument frame.
It is entirely up to consumers of the sandbox library how they lay
things out in the argument frame for any given vtable.  The C++ code
provides an example, Verona will want to do something different.
The library runner also provides an upcall function that allows
callbacks that are registered on the sandboxed library object to be
invoked.

The seccomp-bpf sandbox is still quite rough and has several FIXMEs.
The seccomp-bpf system allows userspace to provide eBPF filters that
define the system calls that can be authorised.  This is not ideal for
sandboxing for several reasons, most notably:

 - The eBPF cannot read userspace pointers and so cannot make decisions
   based on the path.  This is not possible to add safely with the
   current structure of seccomp-bpf because the kernel does not
   deep-copy arguments before the filter runs and so would be subject to
   TOCTOU vulnerabilities (e.g. read the path to `open`, approve it, the
   sandboxed process then modifies it in a different thread to one the
   policy would not have allowed).  This composes poorly with the recent
   trend in Linux towards system calls designed for future extensibility
   that take more complex data structures as arguments.
 - It can't handle dynamic properties and doesn't have any way of
   attaching metadata to file descriptors that can be queried from the
   eBPF, so you cannot easily restrict operations on individual file
   descriptors (for example, allowing `ioctl` on a file descriptor for a
   specific device node but no others).

This means that, for even basic usage, we need to provide some upcall
mechanism if we want to be able to do simple things like open a file.
The glibc run-time linker does not allow us to either provide file
descriptors or to interpose on its system calls, so we also need to be
able to trap and emulate these calls.

This implementation configures the seccomp-bpf filter to trap on certain
system calls (for example, `open`).  We then catch the signal that's
delivered, copy the arguments out of the trap frame, perform an upcall
to the parent process, and inject the result into the trap frame and
advance the program counter past the `syscall` instruction.  The same
mechanism is used for Capsicum, though Capsicum allows a lot more
things.

To support the emulation of `open`, this commit adds an initial
abstraction for an exported file tree.  This is currently used in the
parent for Capsicum as well, though eventually it can reside inside the
sandbox in the Capsicum implementation.

Various things were in `assert` statements that we want in release
builds and so trying a release build gave a lot of unused-variable
errors.  There is now a `sandbox::invariant` template function (and some
helper macros for using it) that allows invariants to be checked easily
in debug/release/any build configurations and to print an informative
message if the invariants are violated.  The format strings are checked
at compile time with libfmt >= 7.


davidchisnall:
#394 is tracking the work still to do to promote this from experiments.  This PR does not include all of that support, but it's close enough that it is probably ready for review.  This is a rebased and squahsed version of the sandbox_pal branch, so the commit history there may give more detail on specific things.

devops/ci.yml
```diff
@@ -94,6 +103,72 @@ jobs:
...
+  pool:
+    vmImage: 'ubuntu-18.04'
+  timeoutInMinutes: 120
+  strategy:
+    matrix:
+      GCC Release:
+        CC: gcc
+        CXX: g++
+        BuildType: Release
+      GCC Debug:
```
rengolin, line 119: We don't need all of these variations for CI, only one of them. Better move the rest to `nightly`.

rengolin (commented):
Two quick things: Still have `docs/website/empty` in the change tree. Some CI should move to nightly. I'll review this in due time, it's huge! :)

devops/ci.yml
davidchisnall: I've removed the GCC versions because GCC is too broken.  The compile + test for the others takes <1 minute each and is run only when the code changes, so I don't see any advantage in moving it to nightly.

rengolin: The main problem is that it starts in a new VM, so it uses more resources. We currently have a limit of 10 concurrent VMs and the CI uses 4, so we can do 2.5 builds in parallel (multiple changes, cancellation of previous builds, etc). If the number of VMs passes 5, we can't even do 2 at a time.

What we left in CI is the minimal configs to catch most errors, so one per type (gcc, clang, win, mac). Everything else goes into the nightly which can take as long as it takes.

Also, the GCC build will have to pass once we move this into `src` if we want to keep GCC support at all. Ignoring it now may make it a lot harder to enable it back later, as we've seen in the runtime last time we enabled it.

experiments/process_sandbox/cxxapi/cxxsandbox.h
```diff
@@ -0,0 +1,304 @@
...
+     * The return value of this function.  If the return value is void, uses
+     * char as an unused placeholder.
+     */
+    std::conditional_t<std::is_void_v<Ret>, char, Ret> ret;
+    /**
+     * The arguments to the function.
+     */
+    std::tuple<std::remove_reference_t<Args>...> args;
+    // FIXME: This strips references, so we copy arguments, but should probably
+    // turn references into pointers.
```
rengolin, line 37: IIUC, the compiler will check if the pointer being passed belongs to the same cown, and if (by construction) we create all Verona objects in an area of memory that the sandbox process can see (ex. its heap), then passing pointers would always be to a visible area on both sides, correct?

experiments/process_sandbox/cxxapi/cxxsandbox.h
```diff
@@ -0,0 +1,304 @@
...
+      {
+        export_function(type_encoding);
+      }
+      functions.emplace_back(new SandboxExportedFunction(fn));
+    }
+  };
+
+  extern "C" void sandbox_call(int idx, void* args)
+  {
+    ExportedLibrary::call(idx, args);
```
rengolin, line 252: It's unclear to me what `ExportedLibrary` is.

Being a static class means all users of `sandbox_call` will use the same exported library interface.

I expected that this code would be the base for each actual library (a Verona module in the source path) and each separate module would have their own dispatchers.

Do you expect the Verona compiler to generate this for every module at compile time? If not, how are we going to separate each module's implementation? Namespaces?

rengolin (commented):
Some initial comments, viewed up to `cxxsandbox.h`.

experiments/process_sandbox/filetree.h
```diff
@@ -0,0 +1,319 @@
...
+#include <variant>
+
+namespace sandbox
+{
+  /**
+   * Class encapsulating a filesystem exported to a sandbox.  This is a very
+   * lightweight simulation of a real filesystem.  This provides a directory
+   * structure where each leaf node is either a directory handle to a real
+   * directory or a file handle to a file (or something equivalent, such as
+   * an anonymous shared memory object).
```
rengolin, line 37: Is this a temporary solution to provide some abstraction to ease IPC mechanisms or an interim implementation of a more permanent solution for exposing the real file system?

experiments/process_sandbox/filetree.h
```diff
@@ -0,0 +1,319 @@
...
+   *
+   * Note: The directories in this structure that are not represented by
+   * handles cannot be opened or otherwise inspected in the sandbox.  If the
+   * tree stores `/tmp/example/somefile`, where only `somefile` is
+   * represented by a real file descriptor, the sandbox will not see `/tmp`
+   * or `/tmp/example`.  This restriction may be lifted in the future if it
+   * causes significant problems.
```
rengolin, line 44: Is there a reason for this, other than implementation complexity? Or is this restricted to avoid the sandbox from iterating the file system like `cat ../../var/log/security` without a proper permission management?

experiments/process_sandbox/filetree.h
```diff
@@ -0,0 +1,319 @@
...
+      /**
+       * A pointer to the child directory.  This could probably be a unique
+       * pointer as it's currently used, but `std::variant` can't contain
+       * references and so we'd have to use bare pointers during traversal.
+       */
+      using DirPtr = std::shared_ptr<ExportedDirectory>;
+
+      /**
+       * The contents of this virtual directory.  This maps from file names to
+       * another instance of this class or a file / directory handle.
```
rengolin, line 101: So, `DirPtr` is an object that I can do file lookup, while `DirHandle` / `FileHandle` are handlers to the actual dir/file objects in the file system?

This seems like a weird system to query file systems. Why not a simple dir/file handlers and a collection of methods (in a static class or something) that can traverse the actual file system?

Maybe this is a `view` into the filesystem that doesn't allow the sandbox to see everything, but only traverse certain directories and open certain files?

experiments/process_sandbox/filetree.h
```diff
@@ -0,0 +1,319 @@
...
+       * Add a directory tree represented by a directory handle.  Takes
+       * ownership of the handle.
+       */
+      void add_directory(const std::string& name, platform::Handle&& h)
+      {
+        directory[name] = DirHandle(std::move(h));
+      }
+
+      /**
+       * Add a file, represented by a directory handle.  Takes ownership of the
```
rengolin, line 181: "directory handle"?

experiments/process_sandbox/filetree.h
```diff
@@ -0,0 +1,319 @@
...
+    DirPtr root = std::make_shared<ExportedDirectory>();
+
+    /**
+     * Add a handle at a specific page.  This is a helper method for the public
+     * interfaces that add a file or directory handle.  The `add` parameter
+     * contains a callable object that takes the base file name and the
+     * directory pointer as arguments and adds it.
+     *
+     * Returns true on success, false on failure.  This can fail if one of the
+     * path components is a directory handle.
```
rengolin, line 207: So `DirHandle` can only be leaf directories? And if you add a sub-dir there, you need to convert to a `DirPtr`?

experiments/process_sandbox/host_service_calls.h
```diff
@@ -0,0 +1,105 @@
...
+ * complex. After sandboxes are made reentrant, the same upcall mechanism used
+ * for calling back into Verona can be used to replace this.
+ */
+
+namespace sandbox
+{
+  /**
+   * The ID of the method that's being proxied.
+   */
+  enum HostServiceCallID : uintptr_t
```
rengolin, line 22: Is this `uintprt_t` for alignment purposes?

rengolin (commented):
Some more comments. I'm slowly going through it, please bear with me.

experiments/process_sandbox/library_runner.cc
```diff
@@ -239,184 +219,531 @@ namespace sandbox
   void ProxyPageMap::set_large_size(void* p, size_t size)
   {
     size_t size_bits = bits::next_pow2_bits(size);
-    if ((p >= shared_memory_start) && (p < shared_memory_end))
+    SANDBOX_DEBUG_INVARIANT(
+      (p >= shared_memory_start) && (p < shared_memory_end),
```
rengolin, line 223: Shouldn't this be `p+size < shared_memory_end`? Same question below.

experiments/process_sandbox/library_runner.cc
```diff
@@ -239,184 +219,531 @@ namespace sandbox
...
+   * abstraction similar to OpenStep's modal runloop. Each recursion depth in
+   * an upcall has its own runloop that handles recursive invocations from the
+   * parent in response to the upcall.
+   */
+  void runloop(int upcall_depth = 0)
+  {
+    int new_depth;
+    do
+    {
+      do
```
rengolin, line 301: It's unclear why you need this busy loop and how `should_exit` can be set meanwhile. Perhaps a comment would help.

experiments/process_sandbox/library_runner.cc
```diff
@@ -239,184 +219,531 @@ namespace sandbox
...
+      shared->msg_buffer = nullptr;
+      try
+      {
+        if ((buf != nullptr) && (sandbox_invoke != nullptr))
+          sandbox_invoke(idx, buf);
+      }
+      catch (...)
+      {
+        // FIXME: Report error in some useful way.
+        printf("Exception!\n");
```
rengolin, line 322: print and continue? This sounds dangerous... I get there's a big orange FIXME up there, but perhaps the default behaviour would be to close the sandbox and cleanup everything before sending some type of `SandboxException` to the Verona code.

experiments/process_sandbox/library_runner.cc
```diff
@@ -239,184 +219,531 @@ namespace sandbox
...
+   */
+  void (*sandbox_invoke)(int, void*);
+
+  /**
+   * The run loop.  Takes the public interface of this library (effectively,
+   * the library's vtable) as an argument.  Exits when the upcall depth changes
+   * after executing the helper function.  This provides a nested runloop
+   * abstraction similar to OpenStep's modal runloop. Each recursion depth in
+   * an upcall has its own runloop that handles recursive invocations from the
+   * parent in response to the upcall.
```
rengolin, line 294: Is this about upcalls that need other upcalls to complete? So, the runloop triggers an upcall, which calls another runloop with `++upcall_depth` as an argument, and when that's done, it comes back to the current runloop, etc.?

experiments/process_sandbox/library_runner.cc
```diff
@@ -239,184 +219,531 @@ namespace sandbox
...
+    // Find the correct environment variables.  Note that libc is not fully
+    // initialised when this is called and so we have to be very careful about
+    // the libc function that we call.  We use the `environ` variable directly,
+    // rather than `getenv`, which may allocate memory.
+    //
+    // The parent process provides the shared memory object in the file
+    // descriptor with the number given by `SharedMemRegion` and the location
+    // where it should be mapped in an environment variable.  The child has to
+    // map this as the first step in bootstrapping (before most of libc
+    // initialises itself) to get a working heap.
```
rengolin, line 351: It would be good to have an example of the syntax this accepts for the address and size.

experiments/process_sandbox/library_runner.cc
```diff
@@ -239,184 +219,531 @@ namespace sandbox
...
+    if (ptr == MAP_FAILED)
+    {
+      err(1, "Mapping shared heap failed");
+    }
+
+    shared = reinterpret_cast<SharedMemoryRegion*>(ptr);
+    // Splice the pagemap page inherited from the parent into the pagemap.
+    void* pagemap_chunk = GlobalPagemap::pagemap().page_for_address(
+      reinterpret_cast<uintptr_t>(ptr));
+    munmap(pagemap_chunk, 4096);
```
rengolin, line 398: Magic number 4096. Would be nice to know why you're separating this chunk of the shared memory region.

experiments/process_sandbox/library_runner.cc
```diff
@@ -239,184 +219,531 @@ namespace sandbox
...
+   */
+  std::pair<uintptr_t, Handle>
+  upcall(sandbox::UpcallKind k, void* buffer, size_t size, int fd)
+  {
+    Handle out_fd(fd);
+    UpcallRequest req{k, size, reinterpret_cast<uintptr_t>(buffer)};
+    upcallSocket.send(&req, sizeof(req), out_fd);
+    out_fd.take();
+    int depth = ++shared->token.upcall_depth;
+    (void)depth;
```
rengolin, line 444: left over statement?

experiments/process_sandbox/library_runner.cc
```diff
@@ -239,184 +219,531 @@ namespace sandbox
...
+    runloop(depth);
+    Handle in_fd;
+    UpcallResponse response;
+    upcallSocket.receive(&response, sizeof(response), in_fd);
+    return {response.response, std::move(in_fd)};
+  }
+
+  /**
+   * Perform an upcall, of the specified kind, passing `data`.  The `data`
+   * argument must point to the shared heap.
```
rengolin, line 456: Shouldn't this function check the pointer is inside the shared heap? Instead of relying on the callers to do that? This is a publicly visible function, no?

experiments/process_sandbox/library_runner.cc
```diff
@@ -239,184 +219,531 @@ namespace sandbox
...
+  auto ret = upcall(static_cast<sandbox::UpcallKind>(idx), data, size, fd);
+  int result = static_cast<int>(ret.first);
+  if (ret.second.is_valid())
+  {
+    result = ret.second.take();
   }
+  return result;
+}
+
+int main()
```
rengolin, line 681: I'm guessing this would be a per-region "main" function, which will setup the sandbox and start the busy loop for the region and then tear down once the region is out of scope. The main function should only be called once the behaviour is scheduled, on the processor/core that the cown is executing. I'm assuming that apart from IPC handling, the core will be either executing parent or sandbox code, but not both at the same time (ie. no threading inside the sandbox).

The compiler should probably bundle all functionality of each region into a single object file, but we need to give them unique identifiers as to avoid clashing the static vtable in `ExternalLibrary` accesses when having multiple sandbox regions on the same binary (unless we want to share a single vtable, in which case we need to make sure two sandbox codes don't compile at the same time).

experiments/process_sandbox/platform/poller_epoll.h
```diff
@@ -0,0 +1,94 @@
...
+        {
+          ret = epoll_wait(ep, &event, 1, -1);
+        } while ((ret == -1) && (errno == EINTR));
+        if (ret == -1)
+        {
+          return false;
+        }
+        if (ret == 0)
+        {
+          sleep(1);
```
rengolin, line 74: Why sleep?

experiments/process_sandbox/platform/sandbox.h
```diff
@@ -0,0 +1,40 @@
+// Copyright Microsoft and Project Verona Contributors.
+// SPDX-License-Identifier: MIT
+#pragma once
+
+/**
+ * This file contains the definition of a per-platform sandboxing policy.
+ * Because this interface currently has only one implementation, it is almost
+ * certainly the wrong abstraction and will change when others (e.g.
+ * seccomp-bpf) are added.
```
rengolin, line 9: outdated comment. Also, add the nice interface comments other common headers have.

experiments/process_sandbox/platform/sandbox.h
```diff
@@ -0,0 +1,40 @@
...
+#include "sandbox_seccomp-bpf.h"
+
+namespace sandbox
+{
+  namespace platform
+  {
+    struct SandboxNoOp
+    {
+      template<typename T, typename U>
+      void restrict_file_descriptors(const T&, const U&)
```
rengolin, line 22: No comments whatsoever?

experiments/process_sandbox/tests/sandboxlib-upcall-basic.cc
```diff
@@ -0,0 +1,19 @@
+// Copyright Microsoft and Project Verona Contributors.
+// SPDX-License-Identifier: MIT
+
+#include "cxxapi/cxxsandbox.h"
+#include "sandbox.hh"
+#include "shared.h"
+
+int test(int idx)
+{
+  int v = 12;
```
rengolin, line 10: Ah, here's why 12!! Ok, the fact that this is in another file confuses people. Perhaps a comment?

experiments/process_sandbox/platform/syscall_context_freebsd.h
```diff
@@ -0,0 +1,153 @@
...
+#  include <sys/syscall.h>
+#  include <ucontext.h>
+
+namespace sandbox::platform
+{
+#  ifdef __x86_64__
+  /**
+   * FreeBSD system call information.
+   *
+   * Note: This depends on FreeBSD D29185 landing.
```
rengolin, line 13: Perhaps leaving a URL instead of a bug number would help people unfamiliar with FreeBSD to reach it more easily.

How soon is that likely to land, and how soon is that likely to be part of an official release?

experiments/process_sandbox/platform/syscall_context.h
```diff
@@ -0,0 +1,83 @@
...
+ * int get_syscall_number()
+ * ```
+ *
+ * Returns the number of the system call delivered in this frame.
+ * ```c++
+ * bool is_sandbox_policy_violation()
+ * ```
+ *
+ * Returns whether this signal was delivered as the result of a sandbox policy
+ * violation.
```
rengolin, line 66: Why is this a comment describing the API and not a virtual/base class that others should implement?

This makes it likely comments on a separate file won't be updated when the implementation is updated.

It also allows for changes in one (often tested) implementation (like Linux) to deviate from another (rarely tested) implementation (like FreeBSD) with time.

experiments/process_sandbox/platform/syscall_context_freebsd.h
```diff
@@ -0,0 +1,153 @@
...
+   */
+  class SyscallFrameFreeBSDX8664
```
rengolin, line 15: Nit: I'd call this and the Linux classes `...X64` instead of `X8646`.

rengolin (commented):
Final review, phew!

experiments/process_sandbox/tests/onebitsem-basic.cc
```diff
@@ -0,0 +1,44 @@
...
+
+// Timeout.  If you are debugging this test, increase this so that it doesn't
+// fail in the background while you're inspecting a breakpoint.
+constexpr int timeout_seconds = 5;
+
+template<typename Sem>
+void test_sem()
+{
+  Sem sem;
+  std::atomic<bool> passed;
```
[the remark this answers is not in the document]
davidchisnall: Weird.  I had the same bug in the two-process version of this test, but this one didn't cause the test to fail.

experiments/process_sandbox/tests/sandbox-upcall-basic.cc
```diff
@@ -0,0 +1,48 @@
...
+   * The library that defines the functions exposed by this sandbox.
+   */
+  SandboxedLibrary lib = {SANDBOX_LIBRARY};
+  decltype(make_sandboxed_function<int(int)>(lib)) call_upcall =
+    make_sandboxed_function<int(int)>(lib);
+};
+
+UpcallHandlerBase::Result upcall(SandboxedLibrary&, int val)
+{
+  SANDBOX_INVARIANT(val == 12, "Upcall argument is {}, expected 12", val);
```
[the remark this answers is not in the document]
davidchisnall: I like 12.  It is a good number.  There are many like it, but this one is mine.

experiments/process_sandbox/tests/sandbox-zlib.cc
```diff
@@ -0,0 +1,140 @@
...
+  {
+    return ::deflateEnd(strm);
+  }
+};
+
+template<typename ZLib>
+void test(ZLib& sandbox, const char* file, std::vector<char>& result)
+{
+  int fd = open(file, O_RDONLY);
+  if (fd >= 0)
```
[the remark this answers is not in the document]
davidchisnall: Changed to an invariant.

experiments/process_sandbox/README.md
```diff
@@ -1,37 +1,341 @@
...
+ - Process-related state, safe to use only if the process ID refers to this process, blocked otherwise.
+ - Unsafe to use in general but possibly safe in some cases and so emulated with a callback where possible (e.g. `openat` which can be transformed into a callback that may or may not open the file for you, depending on user policy).
+ - Unsafe to use in any case (e.g. `settimeofday`), includes intrinsically privileged operations that you should not try to use in a sandbox.
+
+### System-call interposition
+
+In the common case, POSIX software invokes system calls via libc wrappers.
+ELF linkage allows us to preempt those symbols by providing implementations in `library_runner` that will be use in preference to the ones in libc.
+This makes it possible to intercept calls that would normally result in a call that the operating system would block and issue a callback instead.
+Some software; however, issues system calls directly.
```
rengolin, line 277:
```suggestion
Some software, however, issue system calls directly.
```

rengolin (commented):
Review on the document only:
* Overall, very good document. Detailed and complete.
* Particularly like the sections that make inefficiencies explicit by saying what should be done better later, and which tools and technologies we can use, and what is OS dependent.
* Great to know why some solutions (ex. POSIX `select`) weren't used.
* The code layout is also very useful.

I have some nits and corrections on the text, but otherwise, looks great. Thanks!

experiments/process_sandbox/library_runner.cc
rengolin: That's not what I meant. By "all functionality of each region" I meant each invocation of a module inside a behaviour. So, if you use the same module in two different sandboxes, they'd be two different object files that get loaded when the behaviour is scheduled. Which is what you said.

My point here, however, was about the name of the function `main`.

IIUC, this is the main file per sandbox, that will set the heap, load the sandboxed library and start the busy loop, per sandbox, waiting for requests from the child process. I'm thinking if this will be part of the Verona program, we may have multiple "sandbox parent processes" running, so this `main` function would be something we call from the behaviour once it starts running (to communicate with the sandbox child process). Is that correct?

If not, then would we always fork a new parent (ie. a `library_runner`) process, per behaviour?

experiments/process_sandbox/cxxapi/cxxsandbox.h
rengolin: When you say "a single sandbox_call ... from the library", do you mean "one dispatcher per foreign module, so all sandbox regions using the same module will re-use the same dispatcher"?

I think this is the same issue I'm having with the other comment on the `main` function, and I think here you have answered that question: no, this isn't how we'll do in the compiler. But the scope of the dispatcher is still confusing to me.

I guess, as an experiment, this makes sense and works, so I don't want to hold on to this. We can discuss how we'll implement in the compiler at a later time, no worries.

experiments/process_sandbox/filetree.h
rengolin: You removed the `*` comment on your change request. But yeah, that looks good, thanks!

experiments/process_sandbox/platform/sandbox.h
rengolin: Well, something that tells you this is for testing only? I don't know, this is not important.

experiments/process_sandbox/platform/shm_posix.h
```diff
@@ -0,0 +1,349 @@
...
+
+      /**
+       * Generic POSIX shared memory object.  This uses the POSIX `shm_open`
+       * interface to create a new shared memory object and then makes it
+       * anonymous by unlinking it once it has been opened.  POSIX specifies
+       * that the shared memory object will remain in existence until after it
+       * has been unlinked and all file descriptors to it have been closed.
+       *
+       * There is a small window in between creating and unlinking the file
+       * where it can leak.
```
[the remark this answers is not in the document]
rengolin: Perhaps update the comment, then. This line scared me. :)

experiments/process_sandbox/platform/syscall_context_freebsd.h
rengolin: Good question, don't know. My dyslexia doesn't let me know that this is about `x86_64` unless I stare at it for a while. Maybe it's not a problem for others. Not an important point, though. Feel free to ignore.

experiments/process_sandbox/README.md
```diff
@@ -1,37 +1,341 @@
...
+ - Directly modify any memory owned by the parent.
+ - Cause the parent to modify any non-sandbox memory except as explicitly allowed via safe callbacks.
+ - Cause the parent to stop executing (denial of service) in any thread, except for the time that the parent allows for a call into the child to execute.
+ - Access any global namespace (filesystem, network, and so on) except as explicitly permitted by the parent.
+
+### Known issues
+
+The current implementation assumes that the snmalloc `Alloc` instance at the boundary of the sandbox validates all pointers that it follows within the sandbox.
+This is a trivial check to add (a simple range check), but is not currently implemented.
+As such, an attacker can corrupt allocator metadata in such a way that causes snmalloc to
```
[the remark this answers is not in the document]
davidchisnall:
```suggestion
As such, an attacker can corrupt allocator metadata in such a way that causes snmalloc in the parent to corrupt arbitrary memory.
```

experiments/process_sandbox/library_runner.cc
davidchisnall: I think I'm still a bit confused by your question.  In Verona, code is arranged in modules, which are classes that contain nested classes and act as a unit of encapsulation / namespacing.  A module that is implemented using foreign code is defined by a header file that describes its interface and library file (.so, .a, .dll + .lib, wasm assembly, whatever) that defines its implementation.  When you instantiate it in Verona, at the abstract-machine level, you get an `iso` object that has fields for all of the global symbols defined in that interface, which you can then use to call functions / create objects / whatever.  With the process-based sandbox, each one of those has an instance of the `SandboxedLibrary` class to define the runtime's interface to it, a shared memory region for the region's memory allocation (its heap, on the C++ side) and a running instance of the `library_runner` program, which loads the library and waits for calls.

The function here is called `main` because that is the name of the entry point for C++ programs.  The `library_runner` program is a C++ program that implements the in-sandbox run loop for a single Verona sandboxed region that is isolated using process-based sandboxing.

We will run one child process per process-sandboxed region that is live in the Verona program.  Every one of those child processes will be running the `library_runner` program and will load precisely one .so file generated by the Verona compiler (which, in turn, will link one or more .a or .so files that provide the C/C++/assembly/Rust/Intercal/whatever code that we are exposing to Verona).

rengolin (approved):
The remaining nits are comment/naming only and not important, since we're going to re-write most of it when we move that into the compiler. As a prototype, this looks good to me already.

rengolin:
FYI, all new commits since I approved look good to me still. Feel free to merge. Thanks!

PR #409, opened 2021-03-21
Programmable Control Flow

sylvanc:
This change produces a much simpler AST that uses `select` and `lambda` as its fundamental pieces. There is no distinction once the AST is built between prefix, infix, or dot notation, and the built in control flow is `try`, `match`, and `when`. Control flow is built by using non-local returns, which are encoded with `try/throw`.


docs/type-lists.md
```diff
@@ -0,0 +1,73 @@
...
+`A...` is a tuple of zero or more types. `(A, B...)` is a tuple of one or more types. That is, the type list is flattened into the tuple. `A...->B` is a function type with zero or more arguments. `(A, B...)->C` is a function type with one or more arguments.
+
+`A...-->B` is a curried function type (note the use of `-->` instead of `->`). That is, the type list is flattened into the arrow. If `A...` is bound to `(C, D)`, you would get `C->D->B`.
+
+Parameters are tuples that may contain type lists. Calling a function does a destructuring bind of the tuple argument. Results are tuples that may contain type lists. Using the result also does a destructuring bind.
+
+```ts
+papply[A, B..., C, R: B->C](f: R~>(A, B...)->C, x: R~>A): R
+{
+  { y: B... => f(x, y) }
```
nwf, line 14: May be important to indicate that tuple flattening also happens at the value level as well (i.e., this is not `(x, (y))`)?  And that this flattening is type-directed, based on the fact that `y : B...`.  If instead `z : (B...)` or `z : T` and `T ~ (B...)`, then `(x, z)` and `(x, (z))` are identical?

docs/unified-call-syntax.md
```diff
@@ -0,0 +1,192 @@
...
+
+// This also calls Foo::bar(4).
+Foo::bar(4);
+```
+
+## Unqualified Reference
+
+An unqualified function name `identifier` (as distinct from a qualified name such as `ident1::ident2`) can be used. If there is a function in the current scope (either defined in the current class or made available without qualification with a `using` directive), this will resolve as if it were a fully-qualified reference.
+
+If there is no static resolution to a function, this will instead resolve as if it were a dynamic unbound selection on the first argument. It's a compile-time error to try to use an undefined function with no arguments.
```
nwf, line 77: "unqualified", not "undefined"?

docs/unified-call-syntax.md
```diff
@@ -0,0 +1,192 @@
...
+
+```ts
+// The function `foo` must be in scope.
+let f = foo;
+
+// If `foo` is in scope, this calls `foo(x)`. If `foo` is not in scope, this
+// calls `x::foo(x)`.
```
nwf, line 84: "this behaves as `x.foo()`, that is, it calls `x::foo(x)`"?

docs/unified-call-syntax.md
```diff
@@ -0,0 +1,192 @@
...
+// These are equivalent
+f a b c;
+f(a, b) c;
+f a (b, c);
+a f (b, c)
+(a, b) f c;
+a f b c;
+
+// If the second argument is a tuple, nest it.
+f(a, (b, c));
```
nwf, line 127: This isn't `a f (b,c)` by the above rules (because that's `f(a,b,c)`)... but is it `a f ((b,c))`?

The examples so far don't, I think, distinguish between syntactic and type-based operation.  That is, is `let bc = (b,c) in a f bc` the same as `f(a,b,c)` or `f(a,(b,c))`?

docs/unified-call-syntax.md
```diff
@@ -0,0 +1,192 @@
...
+  create(args: Args): Window { ... }
+}
+
+let w = Window Args.width(1000).title("My Application");
+
+```
+
+## Variable Length Arguments
+
+There is no language support for variable length arguments. Instead, use an iterator.
```
nwf, line 178: Perhaps something like "(Recall that type lists, despite their apparent variable length, elaborate into concrete tuples of particular lengths)"?

docs/type-lists.md
```diff
@@ -0,0 +1,73 @@
+# Type Lists
+
+`T...` means a type list called `T...`. In a generic, if it is given a constraint, the constraint applies to each type in the list, not the list itself. So `[T...: U]` means a type list of any length such that each element is a subtype of `U`. The type `()` is an empty type list. Unlike a singleton type, this is not a subtype of the empty interface.
```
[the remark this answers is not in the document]
sylvanc: Yes, thanks!

docs/type-lists.md
```diff
@@ -0,0 +1,73 @@
...
+
+`A...` is a tuple of zero or more types. `(A, B...)` is a tuple of one or more types. That is, the type list is flattened into the tuple. `A...->B` is a function type with zero or more arguments. `(A, B...)->C` is a function type with one or more arguments.
+
+`A...-->B` is a curried function type (note the use of `-->` instead of `->`). That is, the type list is flattened into the arrow. If `A...` is bound to `(C, D)`, you would get `C->D->B`.
+
+Parameters are tuples that may contain type lists. Calling a function does a destructuring bind of the tuple argument. Results are tuples that may contain type lists. Using the result also does a destructuring bind.
+
+```ts
+papply[A, B..., C, R: B->C](f: R~>(A, B...)->C, x: R~>A): R
```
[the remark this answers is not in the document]
sylvanc: Nice catch! It's the latter, so this needs fixing.

docs/type-lists.md
sylvanc: Thanks, I've lifted you language!

docs/type-lists.md
```diff
@@ -0,0 +1,73 @@
+# Type Lists
+
+`T...` means a type list called `T...`. In a generic, if it is given a constraint, the constraint applies to each type in the list, not the list itself. So `[T...: U]` means a type list of any length such that each element is a subtype of `U`. The type `()` is an empty type list. Unlike a singleton type, this is not a subtype of the empty interface.
+
+`A...` is a tuple of zero or more types. `(A, B...)` is a tuple of one or more types. That is, the type list is flattened into the tuple. `A...->B` is a function type with zero or more arguments. `(A, B...)->C` is a function type with one or more arguments.
```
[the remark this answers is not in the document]
sylvanc: Yes. And if `B...` is bound to a tuple of length 1, then `(A, B...)` and `(A, (B...))` are identical.

docs/type-lists.md
```diff
@@ -0,0 +1,73 @@
...
+
+`A...-->B` is a curried function type (note the use of `-->` instead of `->`). That is, the type list is flattened into the arrow. If `A...` is bound to `(C, D)`, you would get `C->D->B`.
+
+Parameters are tuples that may contain type lists. Calling a function does a destructuring bind of the tuple argument. Results are tuples that may contain type lists. Using the result also does a destructuring bind.
```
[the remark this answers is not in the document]
sylvanc: Thanks, I lifted your language here too!

sylvanc:
Thanks @nwf ! I've incorporated your changes in the type lists doc, and rewritten the UCS doc completely. If you have a chance, would you mind reading the UCS doc again?

docs/programmable-control-flow.md
```diff
@@ -0,0 +1,145 @@
...
+```
+
+## Cases
+
+A case can can backtrack by throwing `NoMatch`.
+Labelled `NoMatch` gets multi-level backtrack.
+A match that doesn't match anything throws `NoMatch`.
+`match` can't be written in the source language because we want to detect exhaustive matches.
+
+> TODO: destructure in patterns
```
nwf, line 74: FWIW, the sort of analogy in Haskell is View Patterns, which don't project a `Bool` like `==` here, but project an arbitrary value which is then subjected to the usual pattern handling (I think including additional views, though I'm not sure why you'd do that).  So you'll see a lot of `view :: ... -> Maybe ...` views that get used like `foo (view x -> Just z) = ... z ...`.  (see https://gitlab.haskell.org/ghc/ghc/-/wikis/view-patterns)

docs/programmable-control-flow.md
```diff
@@ -0,0 +1,145 @@
+# Programmable Control Flow
+
+New language features:
+
+* Lambda parameters can be values, which turns the lambda into a case.
```
nwf, line 5: What's the rule for deciding whether a parameter is a value or a variable binder?  That is, when is `{ x => ... }` testing equality against `x` and when is it introducing `x`?

docs/type-lists.md
```diff
@@ -0,0 +1,75 @@
...
+
+`A...` is a tuple of zero or more types. `(A, B...)` is a tuple of one or more types. That is, the type list is flattened into the tuple. `A...->B` is a function type with zero or more arguments. `(A, B...)->C` is a function type with one or more arguments.
+
+`A...-->B` is a curried function type (note the use of `-->` instead of `->`). That is, the type list is flattened into the arrow. If `A...` is bound to `(C, D)`, you would get `C->D->B`.
+
+Function parameters and results may have type annotations containing type lists. Calling a function does a destructuring bind of the tuple argument. Using the result also does a destructuring bind.
+
+Tuple flattening also happens at the value level. This flattening is type-directed: given `x: A...`, uses of `x` are expanded to a sequence of values within the enclosing tuple. For example, given `x: A, y: B...`, if `B...` is bound to a tuple of length 2, `(x, y)` is a tuple of length 3. If instead `B...` is bound to a tuple of length 1, then `(x, y)` and `(x, (y))` are identical.
```
nwf, line 11: Perhaps also add to the "given `x: A, y: B...`, if `B...` is bound to a tuple of length 2," case that `(x, (y))` is not `(x, y)` but rather a pair whose second component is a pair?

docs/type-lists.md
```diff
@@ -0,0 +1,75 @@
...
+{
+  { x: A, y: B... => f x y }
+}
+
+curry_all[A, B..., C, R: A->B...-->C](f: R~>(A, B...)->C): R
+{
+  { x: A =>
+    match f
+    {
+      { g: R~>A->C => papply(g, x) }
```
nwf, line 34:
```suggestion
      { g: R~>(A->C) => papply(g, x) }
```

docs/type-lists.md
```diff
@@ -0,0 +1,75 @@
...
+
+Tuple flattening also happens at the value level. This flattening is type-directed: given `x: A...`, uses of `x` are expanded to a sequence of values within the enclosing tuple. For example, given `x: A, y: B...`, if `B...` is bound to a tuple of length 2, `(x, y)` is a tuple of length 3. If instead `B...` is bound to a tuple of length 1, then `(x, y)` and `(x, (y))` are identical.
+
+```ts
+papply[A, B..., C, R: B->C](f: R~>((A, B...)->C), x: R~>A): R
+{
+  { y: B... => f(x, y) }
+}
+
+curry_one[A, B..., C, R: A->B...->C](f: R~>(A, B...)->C): R
```
nwf, line 19:
```suggestion
curry_one[A, B..., C, R: A->B...->C](f: R~>((A, B...)->C)): R
```

docs/type-lists.md
```diff
@@ -0,0 +1,75 @@
...
+{
+  { x: A => papply(f, x) }
+}
+
+uncurry[A, B..., C, R: (A, B...)->C](f: R~>(A->B...-->C)): R
+{
+  { x: A, y: B... => f x y }
+}
+
+curry_all[A, B..., C, R: A->B...-->C](f: R~>(A, B...)->C): R
```
nwf, line 29:
```suggestion
curry_all[A, B..., C, R: A->B...-->C](f: R~>((A, B...)->C)): R
```

docs/type-lists.md
```diff
@@ -0,0 +1,75 @@
...
+}
+```
+
+```ts
+length[T...](x: T...): USize
+{
+  match x
+  {
+    { () => 0 }
+    { _, tl => 1 + length(tl) }
```
nwf, line 47: It probably deserves mention somewhere that despite the existence of type lists and their implicit tuple flattening behavior, that this flattening happens only on the right, so that there's no ambiguity about how many things get cleaved off by that `_`.  (Similarly, it might be good to say what happens---even if that's a failure to compile---if `x : A..., y : B...` in `(x, y)`.)

docs/programmable-control-flow.md
sylvanc: I've added a note here. The rule is: If it can be a parameter, it is. Otherwise, it is interpreted as a pattern.

docs/type-lists.md
sylvanc: Good call.

docs/programmable-control-flow.md
sylvanc: I'm definitely not following this. I'll bug you for a walkthrough.

PR #410, opened 2021-03-25
Type lists

sylvanc:
This adds type lists of the form `T...` to the parser. It also adds some uses of type lists to the stdlib.

### DIFF
--- a/src/stdlib/controlflow/bind.verona
+++ b/src/stdlib/controlflow/bind.verona
@@ -3,24 +3,12 @@
 
 // `~` implements explicit partial application.
 // f ~x rewrites to { y => f(x, y) }
-~[T, U, V, R: U->V](f: R~>((T, U)->V), x: R~>T): R
+~[T, U..., V, R: U...->V](f: R~>((T, U...)->V), x: R~>T): R
 {
-  { y => f(x, y) }
+  { y: U... => f(x, y) }
 }
 
-|>[T, U](x: T, y: T->U): U
+|>[T..., U](x: T..., y: T...->U): U
 {
   y x
 }
-
-// TODO: when type lists work, use these better implementations.
-
-// ~[T, U..., V, R: U...->V](f: R~>((T, U...)->V), x: R~>T): R
-// {
-//   { y: U... => f(x, y) }
-// }
-
-// |>[T..., U](x: T..., y: T...->U): U
-// {
-//   y x
-// }
--- a/src/stdlib/numbers/numbers.verona
+++ b/src/stdlib/numbers/numbers.verona
@@ -1,3 +1,5 @@
+// Copyright Microsoft and Project Verona Contributors.
+// SPDX-License-Identifier: MIT
 class I8 {}
 class I16 {}
 class I32 {}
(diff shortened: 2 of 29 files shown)


src/stdlib/sequence/array.verona
```diff
@@ -1,25 +1,55 @@
...
+  {
+    // TODO: builtin, append x to self
+  }
+
+  pushn[S: Array[T] & mut, U...: S~>T](self: S, x: U...)
+  {
+    match x
+    {
+      { () => }
+      { hd, tl => self.push(hd); self.pushn(tl) }
```
rengolin, line 44: Nit: I know this is clear form context, but it still takes me a few seconds to match `hd` with `head` and `tl` with `tail`. If neither are reserved words, I'd just use them directly.

src/stdlib/numbers/numbers.verona
```diff
@@ -0,0 +1,18 @@
+// Copyright Microsoft and Project Verona Contributors.
+// SPDX-License-Identifier: MIT
+class I8 {}
```
rengolin, line 3: So, numeric types are markers for code generation only, with storage size and signal defined by their names, location and duration defined by their region, right?

src/stdlib/controlflow/bind.verona
```diff
@@ -3,24 +3,12 @@
...
 // f ~x rewrites to { y => f(x, y) }
-~[T, U, V, R: U->V](f: R~>((T, U)->V), x: R~>T): R
+~[T, U..., V, R: U...->V](f: R~>((T, U...)->V), x: R~>T): R
 {
-  { y => f(x, y) }
+  { y: U... => f(x, y) }
 }
 
-|>[T, U](x: T, y: T->U): U
+|>[T..., U](x: T..., y: T...->U): U
```
rengolin, line 11: Does this means that `x` (being a type list) can still partially apply?

Ex:
```
f(a: A, b: B, c: C);
var p: C;
let partialC = f ~p; // creates partialC(a: A, b: B);
// here, does |>(x: T..., y: T...->U): U "partially applies"?
```

rengolin (commented):
Sorry for the unrelated questions, it's just it's the first time I'm seeing a concrete implementation of some concepts that were floating in my head.

Also, non-binding questions, they're mostly to clear my ignorance, not to challenge the implementation. :)

src/stdlib/numbers/numbers.verona
sylvanc: Yeah exactly. The representation of a numeric type is _not_ a pointer, but a concrete machine word. What we'll do is make sure that numbers (and bools) are always `imm`, and then the type system "just works".

src/stdlib/controlflow/bind.verona
sylvanc: In your example:

```
f: (A, B, C)->None
```

so you can't do `f ~p` because `p: C`. You could have `p: A`, for example:

```
var p: A = ...;
let partialF = f ~p; // partialF: (B, C)->None
```

Then, you could use `|>` like this:

```
var b: B = ...;
var c: C = ...;

(b, c) |> partialF;
// this gives partialF(b, c), which will call f(p, b, c)
```

rengolin: Right, I was unsure about which way partial application goes (my first example was with `a` but I got confused).

Ok, that looks like what I was expecting, thanks!

mjp41 (approved):

PR #412, opened 2021-04-01
Lowering calls, specialised methods to LLVM IR

rengolin:
This is another interim PR that allows us to construct functions that will call the specialised methods and make sure those methods will be lowered in LLVM IR. The AST and IR tests are unfortunately not possible, as both have platform specific output as well as random behaviour (AST prints pointers addresses).

This is not our final destination yet, but a good step forward.


rengolin:
Merging as interim progress.

PR #414, opened 2021-04-07
Add override to Kind / SymbolTable methods in AST

rengolin:
These were generating warnings in recent clang.


rengolin:
@sylvanc, ping. This is really simple and should be trivial to rebase your work on top.

davidchisnall (approved):
The new code looks correct, I can't see how the old code would work.  `override` implies `virtual`, without that you'll only ever get the `Kind` corresponding to the static type, so you can't use `kind` for any kind of down-casting and calling it never gives you information that you didn't have without calling it.

PR #415, opened 2021-04-08
Removed the dependencies of the runtime on std::endl outside of systematic testing

ihaller:
Systematic testing uses std::endl for printing, but it is otherwise not used.
Added Systematic::endl to avoid pulling in link-time dependencies towards <ostream> elements when not using systematic testing.
This allows using the Verona runtime in environments that don't support <ostream>.

### DIFF
--- a/src/rt/sched/cown.h
+++ b/src/rt/sched/cown.h
@@ -220,14 +220,14 @@ namespace verona::rt
 
     static void acquire(Object* o)
     {
-      Systematic::cout() << "Cown " << o << " acquire" << std::endl;
+      Systematic::cout() << "Cown " << o << " acquire" << Systematic::endl;
       assert(o->debug_is_cown());
       o->incref();
     }
 
     static void release(Alloc* alloc, Cown* o)
     {
-      Systematic::cout() << "Cown " << o << " release" << std::endl;
+      Systematic::cout() << "Cown " << o << " release" << Systematic::endl;
       assert(o->debug_is_cown());
       Cown* a = ((Cown*)o);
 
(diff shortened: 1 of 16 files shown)


src/rt/test/systematic.h
```diff
@@ -362,6 +364,7 @@ namespace Systematic
     }
 #endif
 
+#ifdef USE_FLIGHT_RECORDER
```
mjp41, line 367: If you push the `ifdef` inside the body, you will have a smaller diff.

src/rt/sched/cown.h
```diff
@@ -744,8 +751,10 @@ namespace verona::rt
     static void schedule(size_t count, Cown** cowns, Args&&... args)
     {
       static_assert(std::is_base_of_v<Behaviour, Be>);
+#ifdef USE_SYSTEMATIC_TESTING
```
Theodus, line 754: What is the purpose of this log message being excluded outside of systematic testing?

Theodus (commented):
LGTM, just one question

src/rt/sched/cown.h
ihaller: At some point I tried to remove the dependency on typeid, but it turned out that it is not important to do so. Thank you very much for pointing it it, removed it.

src/rt/test/systematic.h
ihaller: Changed it to "if constexpr" and moved it around the output to keep the diff smaller.

src/rt/test/harness.h
```diff
@@ -51,17 +51,17 @@ class SystematicTestHarness
 
     if (seed_upper < seed_lower)
     {
-      std::cout << "Seed_upper " << seed_upper << " seed_lower " << seed_lower
-                << std::endl;
+      Systematic::cout() << "Seed_upper " << seed_upper << " seed_lower "
```
mjp41, line 54: I am not sure I agree with pushing the command line messages only into the systematic logging.  These are really about how to run the application.

ihaller: You are right. I reverted all changes to harness.h as it does not involve the runtime just specific tests.

mjp41 (approved):

PR #417, opened 2021-04-08
Update AnyBuild pipeline

almiliMSFT:
- use yaml parameters instead of variables for the values settable at queue time
- remove the "validate agents" step (and instead instruct the AnyBuild process to dynamically fetch available agents)
- changed the "validate all remoted" step to instead print out some AnyBuild stats

### DIFF
--- a/devops/llvm-anybuild.yml
+++ b/devops/llvm-anybuild.yml
@@ -7,9 +7,26 @@ parameters:
   - name: BuildDebug
     type: boolean
     default: true
+  # extra command-line args to pass to AnyBuild
   - name: ExtraAnyBuildArgs
     type: string
-    default: ' '
+    default: '--Epoch 1'
+  # LLVM commit hash to build
+  - name: GOOD_HASH
+    type: string
+    default: '73811d3'
+  # args to pass to cmake when building
+  - name: CmakeBuildToolArgs
+    type: string
+    default: 'install'
+  # number of parallel cmake jobs
+  - name: CmakeJobs
+    type: number
+    default: 32
+  # whether to publish installed LLVM binaries as a package
+  - name: PublishPackage
+    type: boolean
+    default: false
 
 resources:
   repositories:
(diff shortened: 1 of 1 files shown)


devops/llvm-anybuild.yml
```diff
@@ -7,9 +7,26 @@ parameters:
...
-    default: ' '
+    default: '--Epoch 1'
+  # LLVM commit hash to build
+  - name: GOOD_HASH
+    type: string
+    default: '73811d3'
+  # args to pass to cmake when building
+  - name: CmakeBuildToolArgs
+    type: string
+    default: 'install'
```
rengolin, line 21: Having a default here means if some process overwrites this, the job won't install anymore unless the overwriting options have install in them. Maybe this could be empty and the usage below have a compulsory install in addition to any new options?

devops/llvm-anybuild.yml
```diff
@@ -7,9 +7,26 @@ parameters:
...
+  # number of parallel cmake jobs
+  - name: CmakeJobs
+    type: number
+    default: 32
```
rengolin, line 25: Just a note, but if you're using Ninja you don't need this option. Ninja works on Linux, Mac and Windows.

Alternatively, you could also get the number of cores with a script and use that instead.

devops/llvm-anybuild.yml
almiliMSFT: My idea was exactly to be able to overwrite this when doing test runs, so that I can avoid doing the `install` step.

That's how it's already set up in the current pipeline; the difference is that in the current pipeline it's a variable that is still settable at queue time.

But I understand your concern and I can change it to make sure that `install` is performed.

devops/llvm-anybuild.yml
almiliMSFT: Here the idea is to use many more jobs than the number of cores available, because all those compilation processes will be executed remotely.  So this number really depends on how big the AnyBuild cluster is.

devops/llvm-anybuild.yml
rengolin: Ah, I see. No, if the idea is to not install, than this should be fine. I just wanted to make sure this was the semantics. Thanks!

devops/llvm-anybuild.yml
rengolin: True, I forgot about that. Looks good, thanks!

rengolin (approved):

PR #418, opened 2021-04-11
Pre-refactoring for adding an abstraction layer

mjp41:
We are planning to abstract the notion of thread pool from the code, so that platform specifics can be overridden.  This PR contains some minor changes on the way to making that change. Each commit should be self contained.  The most important change is enabling systematic testing to run until thread completion, as this makes the notion of thread with and without systematic testing more uniform.


Theodus:
It looks like the last thread is being deleted before calling `thread_finished`

src/rt/sched/threadpool.h
```diff
@@ -238,6 +250,31 @@ namespace verona::rt
...
+
+      // Confirm at least one other thread is running,
+      auto curr = me;
+      while (curr->systematic_state != SystematicState::Active)
+      {
+        curr = curr->next;
+        if (curr == me)
+        {
+          // No threads left
+          running_thread = nullptr;
```
Theodus, line 271:
```suggestion
          sched.running_thread = nullptr;
```

mjp41: It's always the thing you rebase in at the last minute you get wrong ;-)

mjp41:
> It looks like the last thread is being deleted before calling `thread_finished`

So I added some logging to try to understand your comment:
```
    4     Teardown beginning
    4     Teardown: all threads stopped
    4     cv_notify_all() for teardown
  2       Unpausing
  2       Begin teardown (phase 1)
  2       End teardown (phase 1)
 1        Unpausing
 1        Begin teardown (phase 1)
 1        End teardown (phase 1)
   3      Unpausing
   3      Begin teardown (phase 1)
   3      End teardown (phase 1)
    4     Teardown: all threads beginning teardown
    4     Begin teardown (phase 1)
    4     End teardown (phase 1)
    4     Begin teardown (phase 2)
    4     End teardown (phase 2)
    4     Thread finished.
   3      Begin teardown (phase 2)
   3      End teardown (phase 2)
   3      Thread finished.
  2       Begin teardown (phase 2)
  2       End teardown (phase 2)
  2       Thread finished.
 1        Begin teardown (phase 2)
 1        End teardown (phase 2)
 1        Thread finished.
 1        Last thread finished.
0         All threads stopped
```

Do you mean thread `0`?  As that one does not previously participate in systematic testing.  Perhaps it should, and that would make testing external inputs easier.  But that is a larger change than I was aiming for at this point.

mjp41:
@theodus, are you referring to the ASAN errors? I wasn't testing with ASAN locally.

mjp41:
@theodus I have separated the blocking call to join from thread deallocation.  This fixes the ASAN crash from calling `thread_finished` traversing the `next` pointer of a deallocated thread structure.

Theodus (approved):
This looks good to me now.

PR #420, opened 2021-04-13
Make the scrambler global

Theodus:
The systematic scrambler was previously declared as thread_local, which caused issues when accessing the hash map from multiple threads. This makes the scrambler global and assumes that it will only be used via a const ref to produce permutations after setup.


mjp41:
The diff has changed a lot of lines, could you fix that please.

Theodus:
> The diff has changed a lot of lines, could you fix that please.

Fixed

mjp41 (approved):
LGTM

PR #421, opened 2021-04-14
Make the PRNG global

Theodus:
The systematic PRNG was previously declared as thread_local, and only one of the PRNGs would be seeded on setup. This makes the PRNG global and uses a spinlock to allow multiple threads to receive the next random number.


mjp41 (approved):
LGTM

PR #422, opened 2021-04-16
Systematic testing refactor

mjp41:
This PR primarily makes the systematic testing work when you are firing in work from outside the runtime.  The external sources are not systematically tested. But this gives some coverage. There are bunch of other minor fixes that were added to aid in debugging this code.


src/rt/CMakeLists.txt
```diff
@@ -332,4 +332,22 @@ if(NOT DEFINED VERONA_RT_ONLY_HEADER_LIBRARY)
...
+      add_test(${TESTNAME} func-sys-runtimepause --cores ${CORES} --seed ${SEEDLOWER} --seed_upper ${SEEDUPPER}  --log-all)
+    endforeach()
+  endforeach()
+
+  foreach(CORES 1 2 3 4)
+    foreach(SEED RANGE 1 2)
+      MATH(EXPR SEEDLOWER "${SEED} * 10")
+      MATH(EXPR SEEDUPPER "(${SEED} * 10) + 9")
+      SET (TESTNAME "runtime/func-con-runtimepause_${CORES}_${SEEDLOWER}")
+      add_test(${TESTNAME} func-con-runtimepause --cores ${CORES} --seed ${SEEDLOWER} --seed_upper ${SEEDUPPER})
```
Theodus, line 350: Should this be missing the `--log-all` option when the other added tests include it?

src/rt/test/harness.h
```diff
@@ -38,6 +38,12 @@ class SystematicTestHarness
 
   SystematicTestHarness(int argc, char** argv) : opt(argc, argv)
   {
+    for (int i = 0; i < argc; i++)
```
Theodus, line 41: If we're going to print the arguments here, then we should remove the [similar printouts](https://github.com/microsoft/verona/blob/1290ba708e9cace2df6b0735377f0e90f07de4d1/src/rt/test/perf/backpressure1/backpressure1.cc#L159) from some of the tests (like the backpressure ones).

src/rt/test/func/runtimepause/runtimepause.cc
```diff
@@ -28,53 +29,72 @@ struct M : public VBehaviour<M>
...
+      for (size_t i = 1; i <= pauses_; i++)
+      {
+        auto pause_time = std::chrono::milliseconds(dist(rng));
+        std::this_thread::sleep_for(pause_time);
+        Systematic::cout() << "Scheduling Message" << Systematic::endl;
+        Cown::schedule<M>(a, a, i, i == pauses_);
+      }
+
+      // auto pause_time = std::chrono::nanoseconds(dist(rng));
+      // std::this_thread::sleep_for(pause_time);
```
Theodus, line 73: Should these lines be deleted?

src/rt/test/func/runtimepause/runtimepause.cc
```diff
@@ -28,53 +29,72 @@ struct M : public VBehaviour<M>
...
-  });
+  StartExternalThread(size_t pauses) : pauses(pauses) {}
 
-  sched.run();
-  thr.join();
-  snmalloc::current_alloc_pool()->debug_check_empty();
+  void f()
+  {
+    auto a = new A;
+    auto pauses_ = pauses;
```
Theodus, line 58: Unless I'm mistaken, this new variable `pauses_` isn't necessary since a lambda capture of `pauses` will copy the value.

src/rt/sched/threadpool.h
```diff
@@ -120,14 +136,35 @@ namespace verona::rt
...
     {
+      // Must be called from inside a scheduler thread,
+      // A message can be LIFO scheduled to call this.
+      // Note, if this is not called from a scheduler thread, then the following
+      // can happen
+      //   1. All scheduler threads decide to pause the runtime but stop just
+      //   before
+      //      waiting on the condition variable.
+      //   2. This code runs and calls notify, but no threads see this
+      //   3. All the scheduler threads go call wait on condition variables
```
Theodus, line 147:
```suggestion
      // can happen:
      //   1. All scheduler threads decide to pause the runtime but stop just
      //   before waiting on the condition variable.
      //   2. This code runs and calls notify, but no threads see this.
      //   3. All the scheduler threads go call wait on condition variables.
```

src/rt/CMakeLists.txt
mjp41: The `--log-all` was for debugging, and shouldn't have been committed.

src/rt/test/harness.h
mjp41: That particular example doesn't use the harness.  It probably should be rewritten like the runtimepause example, but that can be a different PR after this one.

src/rt/test/func/runtimepause/runtimepause.cc
mjp41: `pauses` is a field of the messages, so can't appear in the capture list.  I could capture, `this` but it has the wrong lifetime.

src/rt/test/func/runtimepause/runtimepause.cc
mjp41: Yes. Sorry

src/rt/sched/threadpool.h
mjp41: I rebased the diff into the commit.

PR #423, opened 2021-04-20
Factor out thread creation

mjp41:
This commit takes the thread creation and joining code. and all the
affinity code and puts it in one place.


src/rt/pal/threadpoolbuilder.h
```diff
@@ -0,0 +1,89 @@
...
+    {
+      assert(index == thread_count);
+      for (size_t i = 0; i < thread_count; i++)
+      {
+        threads[i].join();
+      }
+
+      for (size_t i = 0; i < thread_count; i++)
+      {
+        threads[i].~thread();
```
Theodus, line 81: Is there a reason why we're not using std::vector<std::thread> to avoid the uses of explicit new, delete, and destruction?

Theodus (approved):
LGTM. Just one nit

src/rt/pal/threadpoolbuilder.h
mjp41: I was trying to minimise the C++ runtime dependence.  So when there is no C++ rt , then we could rewrite this with other calls to create the threads.

I wasn't sure std::vector would be guaranteed to be header only.  I guess it has to be due to the template?

PR #424, opened 2021-04-20
Re-implement MLIR generator over new AST

rengolin:
This PR replaces the old generator, using a custom MLIR dialect, with a new one, using the new AST (which will have type inference and will bring a canonical representation), and using only existing MLIR dialects (for now).

It can only lower a few constructs and has a very limited subset of the features we need, but it can generate standard MLIR, LLVM IR and even compile to machine code or execute using `lli`. This is a fundamentally different approach.

The idea is to have something simple that can be thoroughly tested, all the way to execution, and slowly adding support for the compiler features we need, with all steps through the pipeline (for example, MLIR passes, like the mangling one here).

When we want to add a new dialect operation, we have to make sure we either convert it to an existing dialect sequence (via partial lowering) or to LLVM dialect (through full lowering and conversion overrides). This guarantees that every decision we make at the MLIR level is sound all the way down to the execution semantics.

This PR, however, falls short on delivering most of that. This is just the first step in getting that working and a preview of what we'll have for code generation.

As an example, the `main.verona` test added produces this (sanitised) LLVM IR:
```
define i64 @bar() {
  ret i64 21
}
define i64 @foo(i64 %0) {
  %2 = add i64 %0, 21
  ret i64 %2
}
define i64 @main() {
  ret i64 42
}
```
via the command line:
```
verona-mlir main.verona -o - -out llvm -O3
```
Passing that through `lli` yields the result:
```
$ verona-mlir main.verona -o - -out llvm -O3 | lli
$ echo $?
42
```


src/mlir/driver.cc
```diff
@@ -106,4 +105,73 @@ namespace mlir::verona
...
+    llvm::LLVMContext llvmContext;
+    auto llvm = mlir::translateModuleToLLVMIR(module.get(), llvmContext);
+    if (!llvm)
+      return runtimeError("Failed to lower to LLVM IR");
+
+    // Optimise if requested
+    if (optLevel)
+    {
+      // Initialize LLVM targets.
+      llvm::InitializeNativeTarget();
```
davidchisnall, line 139: It would be good to thread a target triple through this early so we don't build something without cross-compilation support and have to add it later.  I presume the Verona lowering thing will eventually be parameterised on some aspects of the target, but most of the Verona MLIR will be target-independent.

src/mlir/design.md
```diff
@@ -0,0 +1,158 @@
+# Verona MLIR Generator Design Choices
+
+The MLIR generator sits in between the AST and LLVM IR, and is meant to ease
```
davidchisnall, line 3: Please don't line break in the middle of sentences, it makes reviewing diffs difficult.

src/mlir/design.md
```diff
@@ -0,0 +1,158 @@
...
+lowering while still maintaining some high level concepts to allow us to optimise
+the code without having to scan the whole IR on every pass.
+
+The LLVM pipeline intermixes analysis passes with transformation passes and has
+a way of tracking which transformation pass invalidates which analysis pass to
+avoid re-running complex analysis passes. However, more often than not, the
+transformation passes end up invalidating them anyway and a new run is needed.
+
+In Verona, we'll be bundling all modules, including the standard library and all
```
davidchisnall, line 12: I think there are two things being conflated here:

 - The compilation model requiring whole-program reachability analysis for reification.
 - A design choice to emit everything in a single MLIR module.

Long term, I'd hope that we'd have an internal compiler structure that is closer to thin-LTO: we need to walk the entire call graph for reachability, but a lot of the rest of the pipeline is amenable to parallelisation.  I'd like us to eventually end up with a load of different LLVM IR modules and module summary analyses that we can invalidate and regenerate independently when the corresponding Verona code changes and run the codegen pipeline on in parallel (for incremental debugging builds, doing less cross-module analysis so that we can keep .o files for individual modules around if we haven't changed any of the things that they depend on).

src/mlir/design.md
```diff
@@ -0,0 +1,158 @@
...
+## Generation Simplicity
+
+The new AST is much simpler than the previous. The new parser uses a lean typed
+AST and converts all language into those nodes, which becomes much simpler to
+lower to an IR. The parser also runs the type checks and inference, so the AST
+we get at this stage is already complete, and checked for syntactic and semantic
+errors.
+
+Lowering errors can still occur, if the semantics would be correct in some cases
+but not in this specific case, for example, when using it with a different
```
davidchisnall, line 38: It would be good to have an example of this, I don't know what it means.

src/mlir/design.md
```diff
@@ -0,0 +1,158 @@
...
+runtime functionality.
+
+The MLIR dialect that we'll develop at this stage will be solely for representing
+the lowering concepts in order to simplify optimisations. In theory, the AST
+could have been directly lowered to LLVM IR, but the LLVM pipeline wouldn't be
+able, for example, to optimise based on type and region information.
```
davidchisnall, line 44: I think the other key bit of motivation here is keeping some things abstract across early optimisations.  For example:

 - Object / closure layout and representation
 - Memory management mechanism (e.g. ref counting)

src/mlir/design.md
```diff
@@ -0,0 +1,158 @@
...
+    module @Foo {
+      func @foo() {
+        ...
+      }
+    }
+  }
+}
+```
+
+This structure, however, does not convert naturally to LLVM dialect. So we have
```
davidchisnall, line 71: I think the natural way of lowering this would be to a single LLVM module + ThinLTO module summary.  We could then run a ThinLTO-like CodeGen pipeline where each thread compiles one Verona module in parallel but also has (read-only) access to all of the others for the purpose of inlining.  If we're doing some early inlining in MLIR (especially lambda inlining for programmable control flow) then this should let us scale the compiler up to multiple cores fairly cleanly.

The mangling is still required to avoid symbol conflicts but all of the symbols are marked with private visibility or equivalent and so are erased at final link time.

src/mlir/design.md
```diff
@@ -0,0 +1,158 @@
...
+For now, every function argument and local declaration (`let`, `var`) are
+allocated in the stack and its value stored in the memory space. This simplifies
+code generation (all variable symbols point to an address), but it doesn't help
+with value symbols and are really slow.
+
+The LLVM pipeline, even at O1, recognises most of the argument, temporary and
+local allocas as bogus and optimises them away, so this isn't a critical issue
+from a performance point of view, but we will probably need to update the symbol
+table soon to account for more types and states of the variables we hold (ex.
+`let` vs `var`).
```
davidchisnall, line 109: Is there an MLIR equivalent of mem2reg / SROA?  This representation feels like it may make some MLIR-level optimisations harder but it might be fine.

src/mlir/design.md
```diff
@@ -0,0 +1,158 @@
...
+
+## Runtime Calls
+
+The Verona runtime is where the magic happens. Allocating regions, choosing the
+memory management strategy, scheduling behaviours, dispatching foreign language
+code, etc. The AST, however, has only the semantics of the language (with hints
+to the runtime behaviour), so we need to translate it at the MLIR lowering level.
+
+For example, heap allocation, which LLVM uses `malloc` calls, Verona uses
```
davidchisnall, line 118:
```suggestion
For example, heap allocation, for which C uses `malloc` calls, Verona uses
```
Maybe?  I'm not really sure what this is trying to say.

src/mlir/design.md
```diff
@@ -0,0 +1,158 @@
...
+`snmalloc`, which has a finer control over what we can do with the regions. In
+the end, however, it will be the same: calls to the runtime library in IR.
```
davidchisnall, line 120: Though if we're linking the runtime as IR, the snmalloc calls will actually be inlined and specialised for the correct size class, which is fun.

src/mlir/design.md
```diff
@@ -0,0 +1,158 @@
...
+
+The generator knows what runtime calls are appropriate for each AST construct,
+insert those calls in the right places and makes sure to declare the methods
+so that linking with the runtime library works.
```
davidchisnall, line 124: It would be good to discuss how the linking happens.  I presume the runtime will be compiled as IR and then made available as one of the inputs on a ThinLTO-like pipeline?

src/mlir/design.md
```diff
@@ -0,0 +1,158 @@
...
+
+## Arithmetic
+
+Arithmetic is generated by the generator via pattern matching on the types and
+"method" names. Adding two numbers (`a + b`) is the same as calling the method
```
davidchisnall, line 129:
```suggestion
selectors. Adding two numbers (`a + b`) is the same as calling the method
```

And this is a great example of why line breaks that are not on sentence breaks make life harder when reviewing text files.

src/mlir/design.md
```diff
@@ -0,0 +1,158 @@
...
+`+` on `a` over `b` (`a.+(b)`), but these methods don't exist (how would we
+implement addition if not with a `+`?).
+
+So the generator has to find those that weren't declared and replace with
+arithmetic if the types are compatible (int, float).
```
davidchisnall, line 134: Smalltalk handled this with ['primitive' methods](http://wiki.squeak.org/squeak/5716): The standard library `#new` method on `Object`, for example, invokes the `#basicNew' method on the class.  It would be good to discuss why we're not going down this path.

src/mlir/design.md
```diff
@@ -0,0 +1,158 @@
...
+
+Users are still allowed to use `+` as a method name, in which case the generator
+will recognise and call that method instead of transforming into arithmetic, even
+if the types are compatible. For example:
```
davidchisnall, line 138: Please can you describe what happens when you cast an integer to an interface type that defines a `+` method?  We'll need to have a `+` method that can be called via the vtable from somewhere, is this in the standard library?  In which case, we end up with code that looks something like this:

```verona
class U32
{
  static +(a: U32, B: U32) { a + b; }
}
```

Where the method appears to be infinitely recursive, but the compiler will generate the special version for the version in the body.  Or does the compiler synthesise these methods on demand?
}

davidchisnall (commented):
Comments mostly on the design doc (thanks for adding that!), I'll try to get to the code later.

src/mlir/driver.cc
rengolin: Adding target triple to this would be fairly trivial at any point, at least until we start doing target-specific things on our end (foreign code, struct layout, etc). Right now I'm focusing on target agnostic parts on purpose, to get the overall structure of the compiler in shape.

When we add target-specific things, the triple will be an obvious and trivial change, and I would prefer to defer it to when we do that, to make sure we're adding all the right hooks at the right places. If I add it now, I won't test and could be as bogus as not having it.

I'll reply to your review on the document in separate, as a single review.

src/mlir/design.md
rengolin: There's also an infrastructure issue. Emitting LLLVM IR form multiple MLIR modules isn't supported (we'd need multiple LLVM modules). So, if we want to keep the MLIR in modules all the way to LTO, we'll have to create multiple LLVM modules and handle them on our side when we link all LLVM IR modules together. This isn't a trivial infrastructure.

So, at the moment, the design decision is to simplify the infrastructure of some things, while we get other things done. Right now, I need to get something from Verona into LLVM IR and this is the simplest route. Simple enough, that adding functionality to it is easier than if we go down a route we "think is what we want" and then we discover it wasn't, and have to refactor everything from one complicated infrastructure into another completely different complicated infrastructure.

src/mlir/design.md
rengolin: I have no idea either, but I'm leaving it open to errors that can only be caught in MLIR. I'll rewrite it to be more vague.

src/mlir/design.md
rengolin: added those examples as well.

src/mlir/design.md
rengolin: You're thinking way ahead of this example. This is the first lowering into MLIR. This representation allows us to use the symbol table of the modules to refer to functions more easily, as well to have independent modules in which to run passes in parallel (ex. `isolatedFromAbove`).

We have a pass that transforms this into a mangled linear module, which can be run at any time during the pipeline. We can even transform that into another form before the mangled sausage that needs to lower to LLVM, or we can later not even have a sausage to begin with.

The analysis/optimisation passes should pass on the representation that is most efficient, and the final representation should also allow us to do what we want wrt LTO. But all this is well after what we have now, and what we have now is so simple that won't get in the way of what we need later.

src/mlir/design.md
rengolin: Not that I'm aware of, but the plan is to not need one.

At this early stage, it's a lot simpler to make everything load/store. Later on, when we start looking at some patterns, we can easily add values, too, and replace existing load/store patterns with them.

In the end, but still pretty soon, most patterns will be values with only actual stack variables that need their addresses taken will remain as load/store. I'd rather implement that when I have a concrete example I can test than now.

Same reason as above: move from simple to complex implementation is a lot easier than complex wrong to complex right.

src/mlir/design.md
rengolin: I actually meant LLVM IR. I'll update.

src/mlir/design.md
rengolin: I'd honestly rather leave that for later once we start implementing the first calls to the runtime. Also, this is not the compiler yet, and I imagine we'll have another module that will drive the parser, mlir and other libraries, including possibly a linking library (that will also look at foreign code, etc).

The fact that we can execute code at this early stage does not reflect on *how* we'll execute code in the final product.

src/mlir/design.md
rengolin: added

src/mlir/design.md
rengolin: I honestly don't know enough about Smalltalk (or language design) to start this discussion, nor I think this is the right document to do so. This document is about the MLIR lowering, not the language design.

The AST doesn't have a `primitive` attribute and, IIUC, we don't want one to exist, so we can create functions with the same names on different types without resorting to detecting what's primitive and what's not.

src/mlir/design.md
rengolin: All of that will have already been resolved by the parser and the AST will be unambiguous. What happens in the parser and if that's even legal is a matter for the parser to discuss, not the MLIR lowering of the AST.

src/mlir/design.md
davidchisnall: It would be good to capture this in the document, specifically the split between things that are:

- Intrinsic to the language (e.g. whole-program reachability for reification)
- Shortcuts to make the implementation proceed faster (single MLIR module)
- Long-term goals that we should try to avoid making harder but which won't be in the first implementation (multiple MLIR / LLVM IR modules).

src/mlir/design.md
davidchisnall: Primitive methods are not really part of the language in Smalltalk, they're a property of the implementation.  The basic idea is that the implementation reserves a few selectors as things that are not subject to dynamic dispatch and for which implementations can't be provided.  Applying this to Verona, we'd pick a prefix like `$primitive` and have a `$primitiveAdd` selector that would let you do `x $primitiveAdd y` where `x` and `y` are numeric types.  You'd then implement `+` in the standard library as a method that just calls `$primitiveAdd` on the receiver and the argument.  The `+` method would then appear in vtables (if necessary for any numeric types cast to interfaces), and would in every other way behave just like a normal method (because it is).  The `$primitiveAdd` method would be lowered here.  Implementing a `$primitive{X}` method would be a parse error.  There's then a clear contract between the compiler and the standard library that a set of `$primitive` methods must exist on certain types.

src/mlir/design.md
davidchisnall: I don't think I understand how this can be implemented with the lowering that you've described, so it would be good to state this in the document.

davidchisnall:
I actually asked for single line per sentence, not per paragraph.  That makes it easy to read line diffs (and write them in the review) because they never change any other sentences.  This is the style I use in LaTeX and Markdown for everything and I've not found anything that makes it easier to work with large amounts of text.

rengolin:
> I actually asked for single line per sentence, not per paragraph. That makes it easy to read line diffs (and write them in the review) because they never change any other sentences. This is the style I use in LaTeX and Markdown for everything and I've not found anything that makes it easier to work with large amounts of text.

This makes it hard to read, at least for me. Paragraphs are important to separate context, and it's not a given that changing one sentence won't change the surrounding sentences.

src/mlir/design.md
rengolin: I think you're looking for a larger document, one that spans across the whole compiler. This document is restricted to the `mlir` directory because it's only concerned with the MLIR lowering, from the AST. For example, reification and type checks will have already been done by the AST, so it should not be in this document.

The MLIR module discussion is an open problem. I don't really know if it's better or worse to do it that way. I'm trying different things, and those things change with time because MLIR is also changing fast. Currently, the symbol table doesn't seem to work like I expected, so it's quite possible that we'll have to go with single module and link the Verona modules with attributes. But it's too early to tell.

It seems to me like you're expecting a document that describes the whole process, from beginning to end, implemented in a way that is reasonably deterministic and based on well tested methods and infrastructures. We have none of that here.

src/mlir/design.md
rengolin: I think I understood it like you describe, in a way that's what we're doing, but there is no `primitiveAdd` in the AST and I believe this is intentional (@sylvanc?).

Numeric types in the AST are `class U32 {}` types, which in the AST are still classes that just happen to have that name. I am not sure if we want users to override numeric behaviour by creating their own `U32` classes instead of `using "numeric"`, but the current AST would allow that (we have tests that do that). In those cases, `+` could do whatever and it wouldn't be a `primitiveAdd` because it would have declared a '+' function in that context and we would have found it before trying arithmetic.

I believe Pony has a late binding of arithmetic at the LLVM IR lowering stage, once it knows there isn't anything else that overrides the behaviour (I'm not sure about the Pony semantics here). In Verona, I'm doing it a lot earlier, but also checking to see we don't have anything else defined (which should be fine for now).

Specifically for the MLIR lowering, being `+` with a late match or being `primitiveAdd` upfront is irrelevant. So, if we want to discuss the merits of each implementation, we need to look at the parser, not the MLIR lowering.

src/mlir/design.md
rengolin: Sorry, I misunderstood your question. You are right, we can't implement that in any reasonable (non recursive) way. Or at least I can't see a way that that would work as "expected".

However, this isn't any different than any infinitely recursive function that you can create, if you change the name from `+` to anything else. So, in a way, the programmer gets what they want, given the syntax and semantics of the language.

What could work is something like:
```
class U32
{
  static +(a: U32, b: U32) { @llvm.math.funky_add(a, b); }
}
```
which is meaningful but very low level to allow programmers to do that, so it would be probably done by the standard included modules.

A way to make sure users don't break that, for example, would be to mandate the inclusion of the standard modules for every project, so you don't need to add "using numeric", to your programs. This way, programmers wouldn't be able to create their own `U32` classes, because it would be a re-definition of an existing one, working around the problem entirely.

I'm not saying this is a good idea, I'm just saying this is a high-level problem that is best fixed above the MLIR generator level.

davidchisnall (commented):
Partial review, other things keep interrupting me.  Looks good overall, pedantic nitpicking comments inline.  None of them are particularly important.

src/mlir/generator.h
```diff
@@ -40,58 +37,74 @@ namespace mlir::verona
...
   {
-    using ValuesTy = llvm::SmallVector<mlir::Value, 4>;
+    using ResultTy = std::variant<Value, ModuleOp, FuncOp>;
+    using ValuesTy = llvm::SmallVector<ResultTy, 1>;
     ValuesTy values;
 
   public:
     /// Default constructor, builds an empty return value.
     ReturnValue() {}
     /// Single value constructor.
```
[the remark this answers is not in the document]
rengolin: Ah, yes, you're right! I'll update the comment.

src/mlir/generator.h
```diff
@@ -40,58 +37,74 @@ namespace mlir::verona
...
+    ReturnValue(ResultRange& range)
     {
       values.insert(values.begin(), range.begin(), range.end());
     }
-    /// Assignment operator for mlir::Value.
-    ReturnValue& operator=(mlir::Value& value)
+    /// Assignment operator for Value.
+    ReturnValue& operator=(Value& value)
     {
       values.push_back(value);
```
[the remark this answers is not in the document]
rengolin: Yes! Yes it does!

src/mlir/generator.h
```diff
@@ -40,58 +37,74 @@ namespace mlir::verona
...
     {
       assert(values.size() > 0 && "Access to empty return value");
       assert(values.size() == 1 && "Direct access to multiple values");
-      return values[0];
+      return std::get<T>(values[0]);
     }
 
-    /// Returns true if this return value has one or more values.
-    bool hasValues() const
+    /// Access a specific value held, error if none or not enough.
```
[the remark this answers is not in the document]
rengolin: I'll change the comment to "asserts". I don't want to create a full error handling now because I want it to be part of a diagnostics infrastructure and not make the same mistake we did last time, to bake in the error handling too soon.

src/mlir/generator.h
```diff
@@ -145,126 +132,102 @@ namespace mlir::verona
     /// Nested reference for head/exit blocks in loops.
     BasicBlockTableT loopTable;
 
-    /// Unknown types, will be defined during type inference.
-    mlir::Type unkTy;
-    /// MLIR boolean type (int1).
-    mlir::Type boolTy;
+    /// A list of types
+    using Types = llvm::SmallVector<Type, 1>;
```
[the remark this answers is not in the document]
rengolin: I removed this alias in a newer tree... It was used in the previous generator but I realised how useless it was. I'll merge the new tree later that will remove this.

src/mlir/generator.h
```diff
@@ -145,126 +132,102 @@ namespace mlir::verona
...
-
-    /// Wrapper for opaque operators before we use actual Verona dialect.
-    mlir::Value genOperation(
-      mlir::Location loc,
-      llvm::StringRef name,
-      llvm::ArrayRef<mlir::Value> ops,
-      mlir::Type retTy);
+      llvm::ArrayRef<Type> types,
+      llvm::ArrayRef<Type> retTy);
+    /// Generates a call to a static/dynamic function
```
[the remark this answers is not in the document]
rengolin: We do, because the selector for static calls is a type, while the selector for dynamic calls is an object. If there are no selectors, then we consider the current module as the selector and just make a static call to a method on the same module.

The comment isn't describing what the function currently does, but what it's supposed to do. I'll change to reflect reality and add a FIXME.

rengolin:
Converted to Draft because the idea of using modules for class representation [doesn't work](https://llvm.discourse.group/t/calling-functions-on-different-modules/3317). 

You can reference nested symbols from attributes, but not from the call operation. Creating a special call operation that allows it would create a dependency between the modules and kill the main reason I wanted them in the first place: to be isolated.

I'll re-implement as a flat module, with mangled function names, and annotate the classes/modules with attributes and types, which is what we were doing before.

rengolin:
Ok, @sylvanc & @davidchisnall, I think the current branch is good to go.

I have changed how the IR is represented and it made it a lot easier to call free functions and static methods. I've also added LLVM's `StructType`s and field access for good measure.

The `main.verona` uses all of those features, produces super clean LLVM IR and runs successfully, returning the correct value.

The code has **a lot** of `FIXME`s, but that's fine, because the current structure has held together after a number of changes, including flattening the main module, which means that it's reasonable enough to be a good seed.

For now we don't have any runtime calls, so no heap allocation, no regions, no capabilities. Meaning the Verona files are wrong and wouldn't type-check.

Once we have type inference in the AST we can start to implement them. Probably will need to change the `verona-mlir` binary to only `verify` with a flag, like `verona-parser`, so that the current tests still "pass" while we implement all the missing features to make them pass for real, on a valid Verona file.

To be clear, this isn't meant as complete work, but work-in-progress in tandem with the parser work, so we can progress in parallel.

davidchisnall (approved):
I might have time to get to this tomorrow, but probably not enough time to do a complete review.  As a WIP, I think it's fine, so I'm happy to continue with post-merge review.

src/mlir/generator.h
```diff
@@ -145,126 +138,102 @@ namespace mlir::verona
...
-    llvm::Expected<ReturnValue>
-    parseFieldWrite(const ::ast::Ast& ast, mlir::Value value);
-
-    // ============================================================== Generators
-    // These methods build complex MLIR constructs from parameters either
-    // acquired from the AST or built by the compiler as to mimic the AST.
-
-    /// Generate a Verona type from parsing its name, caching in the typeTable.
-    mlir::Type generateType(llvm::StringRef name);
+    llvm::Expected<ReturnValue> parseNode(Ast ast);
```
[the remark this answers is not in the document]
rengolin: Our AST is **so** much simpler than Clang's that it isn't hard nor confusing to create a simple scan.

But I agree with you, if this gets more complicated, at least each `parseX` method can be converted into the body of a visitor pattern and it should be simple enough to convert this into a visitor pattern.

src/mlir/generator.h
```diff
@@ -145,126 +138,102 @@ namespace mlir::verona
...
+    /// return the cast operations that make them the same. If types are
+    /// incompatible, assert.
+    std::pair<Value, Value> typePromotion(Value lhs, Value rhs);
 
-    // ================================================================= Parsers
-    // These methods parse the AST into MLIR constructs, then either return the
-    // expected MLIR value or call the generators (see below) to do that for
-    // them.
+    // =================================================================
+    // Parsers These methods parse the AST into MLIR constructs, then either
```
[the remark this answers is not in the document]
rengolin: Heh, clang format has broken the line above into this one.

rengolin:
Thanks David!

PR #426, opened 2021-04-20
Add pollers per scheduler thread

marioskogias:
Every scheduler thread has a map of poller cowns. The scheduler thread notifies those cowns every time it consumes the token cown.


src/rt/sched/schedulerthread.h
```diff
@@ -187,6 +201,14 @@ namespace verona::rt
             << "Should steal for fairness!" << Systematic::endl;
           should_steal_for_fairness = true;
         }
+
+        // Poll for IO periodically
```
mjp41, line 205: Should the polling occur after the executing the messages.  The messages may be removing the pollers.  Moreover, is there a semantic guarantee that a poller will run for one last time after it has been removed, so it will see everything up to the point where it is removed.

src/rt/sched/schedulerthread.h
```diff
@@ -105,6 +105,10 @@ namespace verona::rt
     /// cleared before scheduler sleep, or in some stages of the LD protocol.
     ObjectMap<T*> mute_set;
 
+    // The set of poller cowns. These cowns are notified periodically to handle
+    // IO.
+    ObjectMap<Object*> pollers;
```
mjp41, line 110: Does this keep the pollers live?  Is there a strong reference on the poller?

If the context does not keep a reference to a poller can it be collected.  I am concerned it can.

src/rt/test/func/pollers/pollers.cc
```diff
@@ -0,0 +1,48 @@
...
+
+  void notified(Object* o)
+  {
+    TimerPoller* p = (TimerPoller*)o;
+    if ((std::difftime(std::time(nullptr), p->start) > 1) && (p->active))
+    {
+      auto& sched = Scheduler::get();
+      sched.poller_remove(p);
+      p->active = false;
+      Cown::release(ThreadAlloc::get(), p);
```
mjp41, line 22: This example keeps the poller live, and then performs its own deallocation.  I think you need to add some systematic testing runs to the harness for this code.  I am concerned about the ordering of the `poller_remove` and the `release`.  

Can the following happen
  1. poller_remove is send to a token cown.
  2. p is released, and deallocated
  3. token cown performs notification of a deallocated cown
  4. poller_remove is executed by token cown.

mjp41 (changes requested):
I think we need more concurrent testing coverage for this.  Call me on Teams and I can explain how to get some coverage from systematic testing.  Basically, you need to add something like:
https://github.com/microsoft/verona/blob/204d11a96348bb010569b137ba07b6956ccb236f/src/rt/CMakeLists.txt#L342-L349
But for your test.

src/rt/sched/schedulerthread.h
```diff
@@ -168,6 +173,15 @@ namespace verona::rt
         stats.unpause();
     }
 
+    void notify_pollers()
+    {
+      for (auto* it : pollers)
+      {
+        T* cown = (T*)it;
+        cown->mark_notify();
+      }
```
Theodus, line 182:
```suggestion
      for (auto* cown : pollers)
        cown->mark_notify();
```

src/rt/sched/cown.h
```diff
@@ -1363,6 +1370,11 @@ namespace verona::rt
         MultiMessage::make_body(alloc, count, cowns, &unmute_behaviour);
       return MultiMessage::make_message(alloc, body, epoch);
     }
+
+    void send_remove_poller_msg(Cown* c)
+    {
+      schedule_lambda([=]() { Scheduler::local()->poller_remove(c); });
```
Theodus, line 1376: Shouldn't the behaviour be scheduled on the token cown (`this`)?
```suggestion
      schedule_lambda(this, [=]() { Scheduler::local()->poller_remove(c); });
```

docs/internal/io-proposal.md
```diff
@@ -0,0 +1,28 @@
...
+
+Below we plot two latency (99th percentile) vs throughput curves for a synthetic service time server. The server busy loops for a configurable amount of time that is in the request. In this case, 90% of the requests spin for 55 us and 10% of the requests spin for 505 us for an average service time of 100us. The experiment runs on a 8-core machine, thus the max expected throughput is 80kRPS. We consider two cases with biased and unbiased connections. In the unbiased case connections are chosen in a uniformly random manner, while in the biased case there are some connections that are slightly more popular than others. Requests arrive at the server in an open loop manner with a Poisson inter-arrival distribution.
+
+![unbiased bimosal](unbiased-bimodal.png)
+![biased bimosal](biased-bimodal.png)
+
+First, we observe that we need multiple poller owner cowns. Second, we observe that the fixed design behaves better than the floating one in the unbiased case and vice versa in the biased one. Unlike that, the proposed `notify-rr` design is the best peforming in terms of tail latency for both cases.
```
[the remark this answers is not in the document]
marioskogias: Round Robin. I'll update that.

PR #427, opened 2021-04-22
[llvm-anybuild] Update to match the LLVM CI pipeline

almiliMSFT:
- update GOOD_HASH to `9f33943ee01`
- update cmake to build clang + mlir
- make 'create package' deterministic by specifying a canned `mtime`
- update pool definition to demand an agent tailored for the requested build type (Release vs Debug)
- delete the produced tarball after publishing it

### DIFF
--- a/devops/llvm-anybuild.yml
+++ b/devops/llvm-anybuild.yml
@@ -14,7 +14,7 @@ parameters:
   # LLVM commit hash to build
   - name: GOOD_HASH
     type: string
-    default: 9f33943ee015
+    default: 9f33943ee01
   # args to pass to cmake when building
   - name: CmakeBuildToolArgs
     type: string
--- a/devops/llvm-anybuild.yml
+++ b/devops/llvm-anybuild.yml
@@ -95,6 +95,10 @@ jobs:
 
   - script: |
       set -euo pipefail
+      if [[ -z $(GOOD_HASH) ]]; then
+        echo "[ERROR] Variable 'GOOD_HASH' not set in Azure DevOps pipeline"
+        exit 1
+      fi
       git checkout $(GOOD_HASH)
     workingDirectory: $(LlvmSourceDir)
     displayName: 'Move tree to known good LLVM commit for Verona'
--- a/devops/llvm-anybuild.yml
+++ b/devops/llvm-anybuild.yml
@@ -11,10 +11,6 @@ parameters:
   - name: ExtraAnyBuildArgs
     type: string
     default: '--Epoch 1 --WhyCacheMiss --WhyCacheMissOptions CacheDataStoreKey=LLVMVerona'
-  # LLVM commit hash to build
-  - name: GOOD_HASH
-    type: string
-    default: 9f33943ee01
   # args to pass to cmake when building
   - name: CmakeBuildToolArgs
     type: string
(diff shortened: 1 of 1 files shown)


devops/llvm-anybuild.yml
```diff
@@ -10,19 +10,19 @@ parameters:
   # extra command-line args to pass to AnyBuild
   - name: ExtraAnyBuildArgs
     type: string
-    default: '--Epoch 1'
+    default: '--Epoch 1 --WhyCacheMiss --WhyCacheMissOptions CacheDataStoreKey=LLVMVerona'
   # LLVM commit hash to build
   - name: GOOD_HASH
     type: string
-    default: '73811d3'
+    default: 9f33943ee015
```
rengolin, line 17: The CMake script cuts at 11 digits (can't remember why), so this should be `9f33943ee01` instead.

devops/llvm-anybuild.yml
```diff
@@ -218,12 +224,16 @@ jobs:
   
   - script: |
       set -euo pipefail
-      rm -f $(PKG_NAME).tar.gz
-      tar zcf $(PKG_NAME).tar.gz build/install
-      md5sum $(PKG_NAME).tar.gz | awk '{print $1}' > $(PKG_NAME).tar.gz.md5
+      rm -f $(PKG_NAME).tar
+      tar --sort=name --owner=root:0 --group=root:0 --mtime='UTC 2021-01-01' -cf $(PKG_NAME).tar build/install
```
rengolin, line 228: I think the problem was the zip part that you had before, which is what creates the time stamp. We're not zipping anymore because it doesn't shrink enough to be worth the extra minutes, so we don't have that problem.

If you want to zip, there seems to be two [alternatives](https://serverfault.com/questions/110208/different-md5sums-for-same-tar-contents): use an external gzip or pass `-n` to the internal one.

devops/llvm-anybuild.yml
almiliMSFT: updated

devops/llvm-anybuild.yml
rengolin: Oh well, we are zipping! :astonished: 

I remember looking at it and thinking I should fix this, but apparently, I didn't. 

Honestly, this hasn't been a problem at all for us, so I think we shouldn't change the time on the file just for that.

Just upload the file and its MD5 and let's see how it works.

devops/llvm-anybuild.yml
```diff
@@ -10,19 +10,19 @@ parameters:
   # extra command-line args to pass to AnyBuild
   - name: ExtraAnyBuildArgs
     type: string
-    default: '--Epoch 1'
+    default: '--Epoch 1 --WhyCacheMiss --WhyCacheMissOptions CacheDataStoreKey=LLVMVerona'
   # LLVM commit hash to build
   - name: GOOD_HASH
     type: string
-    default: '73811d3'
+    default: 9f33943ee01
```
rengolin, line 17: On our side, we don't have a default value on the yaml file, and leave it as a variable in the build. This is meant to be a conscious change and error out if the value is not set, rather than build an older version by mistake.

Recurrent builds can still use the variable value, if driven by DevOps, so it shouldn't change stress testing either.

almiliMSFT: changed to a variable and added a check that it is set in DevOps

devops/llvm-anybuild.yml
almiliMSFT: removed --mtime, --owner, and --group

rengolin (approved):

PR #429, opened 2021-04-27
Simplify systematic testing

mjp41:
This introduces two changes to systematic testing.
*  Using guard functions to build condition variables without an
explicit notion of waiting.
* Use a step count so we perform fewer operations on the fast path.


Theodus (approved):
Looks good, just some suggestions

PR #433, opened 2021-05-04
Improve MLIR generator design document

rengolin:
Add missing topics, plus two TODO ones: exception and match, which aren't implemented yet and need further work to have a better idea how we'll do them.

There are some questions, too, to clarify before the document lands.

I want to make sure all those assumptions are valid for the _interim_ implementation, as well as making it easier to convert to the full implementation later.

Many of those assumptions will change with the new iteration of the parser (type inference, etc), so I just need to make sure I'm not going away from the final goal, but towards it.


rengolin:
Readable version: https://github.com/rengolin/verona/blob/mlir-gen/src/mlir/design.md

rengolin:
A lot is changing in this area, so I'll bundle this document with the other changes and drop this PR.

PR #435, opened 2021-05-13
Factor out thread pausing code. 

mjp41:
The thread pausing code provides a way to tell the OS that a thread is
not currently busy. This refactoring pulls that code out into a separate
class. This enables it to be replaced on platforms that do not implement
C++ condition variables.

As part of this refactoring, the systematic testing library was also
separated.


rengolin:
Once the CI passes, would be nice to run the nightly over this to see if the Windows build problem is fixed.

src/rt/sched/cown.h
```diff
@@ -21,7 +21,7 @@ namespace verona::rt
   static void yield()
   {
 #ifdef USE_SYSTEMATIC_TESTING
```
ihaller, line 23: Is the ifdef still needed?

src/rt/sched/threadsync.h
```diff
@@ -0,0 +1,101 @@
...
+
+      /**
+       * Pause this thread
+       *
+       * If this is the last thread, then something external must call
+       * unpause_all to restart the paused threads.
+       */
+      void pause()
+      {
+        Systematic::cout() << "Pausing" << Systematic::endl;
```
ihaller, line 52: The purpose of this debug message is unclear to me. My understanding is that when systematic testing is enabled, this class is not activated.

mjp41: The `flight_recorder` also logs these messages.  This should really be renamed.  The `flight_recorder` is for debugging crashes that can't be reproduced in systematic testing.

mjp41:
@theodus I think I have finally traced down the bugs in this.  I believe it is correct now.  There is one CI failure for Mac OS X, but the machine was extremely slow compared to other runs, so I think something might have been happening on that machine.  I will run several more Nightly builds to see if the problem reoccurs.

src/rt/sched/cown.h
```diff
@@ -1383,14 +1381,39 @@ namespace verona::rt
...
+    std::stringstream ss;
+    bool short_id = external_id <= 26;
+    size_t spaces = short_id ? 9 : 8;
+    size_t offset = (external_id - 1) % spaces;
+    if (offset != 0)
+      ss << std::setw(spaces - offset) << " ";
+    if (short_id)
+      ss << (char)('a' + (external_id - 1));
+    else
+      ss << 'e' << (external_id - 26);
```
Theodus, line 1414: Should this start with `E` instead of `e` to distinguish from the result when `external_id` is 5?

src/rt/sched/cown.h
```diff
@@ -1383,14 +1381,39 @@ namespace verona::rt
...
+    ss << std::setw(offset) << " ";
+    return ss.str();
 #endif
     return 0;
```
Theodus, line 1418:
```suggestion
#else
    return "";
#endif
```

src/rt/sched/cown.h
```diff
@@ -1383,14 +1381,39 @@ namespace verona::rt
 
 namespace Systematic
 {
-  inline size_t get_systematic_id()
+  inline string get_systematic_id()
```
Theodus, line 1384:
```suggestion
  inline std::string get_systematic_id()
```

src/rt/sched/threadsyncsystematic.h
```diff
@@ -0,0 +1,335 @@
...
+        // This is an external wake up
+        start = first_thread;
+      }
+
+      // Skip to a first choice for selecting.
+      while (i > 0)
+      {
+        start = start->next;
+        i--;
+      }
```
Theodus, line 82:
```suggestion
      for (; i > 0; i--)
        start = start->next;
```

src/rt/sched/threadsyncsystematic.h
```diff
@@ -0,0 +1,335 @@
...
+        if (result == start)
+        {
+          // The following note is incase anyone tries to add an assertion about
+          // the conditions that could be true at this point.
+          //
+          // Note, this could have zero external event sources, with the runtime
+          // waking up from pausing.  The external thread that dropped the event
+          // source, can will still be `unpausing` the runtime.  So it appears
+          // all threads are going to sleep, but the external thread will
+          // continue to wake them. Until `runtime_pausing` is unset.
```
Theodus, line 98:
```suggestion
          // The following note is for anyone wanting to add an assertion about
          // the conditions that could be true at this point.
          //
          // Note, this could have zero external event sources, with the runtime
          // waking up from pausing.  The external thread that dropped the event
          // source, can still be `unpausing` the runtime.  So it appears all
          // threads are going to sleep, but the external thread will continue to
          // wake them until `runtime_pausing` is unset.

```

src/rt/sched/threadsyncsystematic.h
```diff
@@ -0,0 +1,335 @@
...
+    void wait_for_my_turn_inner(std::unique_lock<std::mutex>& lock, T* me)
+    {
+      assert(lock.mutex() == &m_sys);
+      Systematic::cout() << "Waiting for turn" << Systematic::endl;
+      while (running_thread != me)
+        me->cv.wait(lock);
+      assert(me->systematic_state == SystematicState::Active);
+    }
+
+    /// lock, must be holding the m_sys mutex.
```
Theodus, line 122:
```suggestion
    /// Must hold the systematic testing lock to call this.
```

src/rt/sched/threadsyncsystematic.h
```diff
@@ -0,0 +1,335 @@
...
+          {
+            Systematic::cout() << "unpause internal" << Systematic::endl;
+            sync.unpause_incarnation++;
+            sync.yield(me);
+          }
+          else
+          {
+            Systematic::cout() << "unpause external" << Systematic::endl;
+            // Can be signalled from outside the runtime if external work is
+            // injected if this is a runtime thread, then yield.
```
Theodus, line 230:
```suggestion
            // injected. If this is a runtime thread, then yield.
```

Theodus (approved):
Looks good, just some suggestions

src/rt/sched/cown.h
mjp41: It's `e1`, `e2`, ..., so not sure it will be confusing.  But probably I should change, as otherwise I will curse you being right in two years time if I don't ;-)

mjp41:
```suggestion
      ss << 'E' << (external_id - 26);
```

src/rt/sched/threadsyncsystematic.h
```diff
@@ -0,0 +1,333 @@
...
+        if (result == start)
+        {
+          // The following note is for anyone wanting to add an assertion about
+          // the conditions that could be true at this point.
+          //
+          // Note, this could have zero external event sources, with the runtime
+          // waking up from pausing.  The external thread that dropped the event
+          // source, can still be `unpausing` the runtime.  So it appears all
+          // threads are going to sleep, but the external thread will continue to
+          // wake them until `runtime_pausing` is unset.
```
mjp41, line 95:
```suggestion
          // threads are going to sleep, but the external thread will continue
          // to wake them until `runtime_pausing` is unset.
```

PR #438, opened 2021-05-17
Update-dump normalised result file

rengolin:
This keeps popping in on my local tree whenever I run `ninja update-dump`. It doesn't change anything but it's good to have all dumps as the update dump expects so that this doesn't pop up in unrelated work.


mjp41:
Is it a line ending change?

rengolin:
> Is it a line ending change?

Yes

PR #439, opened 2021-05-24
Add LambdaBehaviours

marioskogias:


src/rt/test/func/lambdabehaviour/lambdabehaviour.cc
```diff
@@ -0,0 +1,29 @@
+// Copyright Microsoft and Project Verona Contributors.
+// SPDX-License-Identifier: MIT
```
mjp41, line 2: It would be good if the tests actually had state in the closures to exercise that aspect.

It would also be good if you captured a smart pointer, to check that the destructors are correctly called.

mjp41 (approved):
LGTM.  I think this can make some of the runtime test harness much nicer.  When this lands I might rewrite a few tests to use this, as it will reduce the amount of boiler plate code.

Theodus (approved):
LGTM

PR #442, opened 2021-05-26
Add tuple support in MLIR

rengolin:
This is a large change-set that adds support for tuples in MLIR.

Tuples are represented as anonymous structs. They're used in the AST as operands for function calls, so there's special logic in lowering calls to wrap/unwrap tuples into calls by checking the functions actual arguments. This allows us to differentiate, for example, `func(I32, I32)` from `func((I32, I32))` when calling, since both would call with a tuple `(I32, I32)`.

Some basic tests were added and they generate sane MLIR as well as LLVM IR, and the LLVM middle-end can easily optimise all of the generated kludge inside functions. Given tuples don't yet have accessors (ex. `tuple._1`), nor completely defined semantics, for now, their use is solely for the purpose of calling functions with more than one argument. Once we define the semantics for tuples, we can add the remaining logic.

This work is necessary for implementing arithmetic in Verona, since binary operator functions will have two arguments, and this is the only rationale for it being here at this stage. the next step is to implement arithmetic as functions in Verona, using intrinsics, and still compile to valid MLIR and LLVM IR, still having LLVM recognise the constant operations, propagating them and returning literal constants at the call sites (like we have today).

In addition, there are a number of cleanups and refactoring that was necessary to simplify the code so that tuples were added, mainly:
* Reduce the number of allocas (for arguments), loads (let assignment) and constant declarations
* Remove special `ReturnValue` structure and use plain `mlir::Value`s
* Split the generator into `consumer` (for AST logic) and `generator` (for MLIR logic)

Some other unrelated cleanups and fixups:
* Adding literal strings (for arithmetic calls)
* Fixed nested class lookup

Also added a design document with the intentions long term of the MLIR generator. This document is work in progress and shouldn't be treated as canon, but they explain the expectations of the AST in the future and the work-arounds we had to do meanwhile.


rengolin:
FWIW, this is the arithmetic branch that works because of this PR: https://github.com/microsoft/verona/commits/mlir-math

src/mlir/consumer.cc
```diff
@@ -0,0 +1,762 @@
...
+    if (auto funcOp = gen.lookupSymbol<FuncOp>(opName))
+    {
+      llvm::SmallVector<Value, 1> args;
+      // Here it's guaranteed the lhs is not a selector (handled above), so if
+      // there is one, it's the first argument of a function call.
+      if (lhs)
+      {
+        args.push_back(lhs);
+      }
+      if (rhs)
```
davidchisnall, line 414: I find it quite hard to follow this block.  It would help to have a block comment here explaining what it's trying to do.

src/mlir/consumer.h
```diff
@@ -0,0 +1,149 @@
...
+   * we start implementing classes fully, and dynamic dispatch.
+   */
+  struct FieldOffset
+  {
+    llvm::SmallVector<llvm::StringRef, 1> fields;
+    llvm::SmallVector<Type, 1> types;
+  };
+
+  /**
+   * AST Consumer
```
davidchisnall, line 34: This needs a bit more explanation.  I presume this is the thing that produces MLIR from an AST, but AST consumer sounds like a generic interface for anything that can consume AST nodes.

src/mlir/consumer.h
```diff
@@ -0,0 +1,149 @@
...
+   */
+  class ASTConsumer
+  {
+    ASTConsumer(MLIRContext* context) : gen(context) {}
+
+    /// MLIR Generator
+    MLIRGenerator gen;
+
+    /// Map for each type which fields does it have.
```
davidchisnall, line 43: From class name to a thing that contains a vector of field names and offsets?

src/mlir/consumer.h
```diff
@@ -0,0 +1,149 @@
...
+    std::map<llvm::StringRef, FieldOffset> classFields;
```
davidchisnall, line 44: Why `map` and not `unordered_map`, do we care about ordering here?

`llvm::StringRef` is quite dangerous to use as a map key, because it is a non-owning value and so there's an undocumented invariant here that the lifetime of every class name (wherever it comes from) must outlive the class.  I guess that's okay because it comes from the AST and this class is a transient thing that consumes an AST?

If you want to key on strings (which I'd prefer you didn't and instead keyed on interned string IDs), then `llvm::StringMap` is probably what you actually want.

src/mlir/consumer.h
```diff
@@ -0,0 +1,149 @@
...
+
+    /// Function scope, for mangling names. 3 because there will always be the
+    /// root module, the current module and a class, at the very least.
```
davidchisnall, line 47: That sounds like a good reason for picking a number like 5.  The size for `llvm::Small*` should capture the most common case, not the minimum case.

Same comment as above about lifetimes.  I presume these StringRefs are things from higher up the AST than the code that we're currently parsing, but if that's an invariant on which memory safety depends then please document it.

src/mlir/consumer.h
```diff
@@ -0,0 +1,149 @@
...
+    SymbolTableT& symbolTable()
+    {
+      return gen.getSymbolTable();
+    }
+
+    /// Get node as a shared pointer of a sub-type
+    template<class T>
+    ::verona::parser::Node<T> nodeAs(::verona::parser::Ast from)
+    {
+      return std::make_shared<T>(from->as<T>());
```
davidchisnall, line 80: I don't understand this use of `std::make_shared`.  Does this move or copy construct the node?  Each of these will construct a new refcounted `T`, so multiple calls to `nodeAs` will give different objects.  Is that actually the desired behaviour?

src/mlir/consumer.h
```diff
@@ -0,0 +1,149 @@
...
+    }
+
+    /// Get location of an ast node.
+    Location getLocation(Ast ast);
+
+    /// Mangle function names. If scope is not passed, use functionScope.
+    std::string mangleName(
+      llvm::StringRef name, llvm::ArrayRef<llvm::StringRef> scope = {});
```
davidchisnall, line 88:
```suggestion
    std::string mangleName(
      llvm::StringRef name, llvm::ArrayRef<llvm::StringRef> scope);
    std::string mangleName(
      llvm::StringRef name)
    {
      return mangleName(name, functionScope);
    }
```

(Avoids the need to check for the empty thing in the callee)

src/mlir/design.md
```diff
@@ -3,69 +3,225 @@
...
+To avoid order issues in the IR, we'll do multiple passes on the structures:
+1. Declaration pass: Defines reachable type structures and functions, including the entire standard library files.
+2. Lowering pass: Lower the reachable AST nodes using the structures above and defining the functions for the prototypes above.
+
+In theory, the AST could have been directly lowered to LLVM IR, but the LLVM IR is too low level for some analysis we want to do (ex. region aliasing) and it would be counter-productive to do so at that stage.
+
+## Classes & Modules
+
+Classes and modules are the same concept in Verona.
+Each source sub-directory containing Verona files is a module, and all its Verona files share the same namespace, named after the directory name.
```
davidchisnall, line 30: This is only half right.  There are two names for a module.  The name it shows up under in the source file where it is used and the name that tells the compiler where to find it.  I presume the AST handles the combination of these for you?

I also presume anything about visibility is handled before we get to this level?

src/mlir/design.md
```diff
@@ -3,69 +3,225 @@
...
+Functions are a native concept in MLIR and LLVM IR, and lowering free and static functions and calls to them is trivial.
+We'll declare all reachable functions beforehand, then lower the AST (implementing the prototypes as full functions).
+Dynamic resolution and selector colouring will be done at the parser level, and calls to dynamic methods in the AST will be a node with base pointer and offset.
+
+MLIR modules have a symbol table for functions, which we use to match declarations against definitions and calls.
+
+Arguments by value or reference have ABI reserved space (registers, stack), so we don't need to allocate them again (the `alloca` _trick_).
+Return values, for functions that do return a type, will be the last expression evaluated in the block.
+
+Lambdas will be lowered as an anonymous structure with captures as fields and `apply()` as the method for when the lambda is called.
```
davidchisnall, line 49: This lowering causes a lot of problems in clang.  In particular, it makes it incredibly hard to propagate information about aliasing state or mutability into the callee.  From the callee's perspective, all of the fields of the structure are just fields of a struct and so there's no way of propagating anything that we inferred about the variables that they capture into the lambda body or any clean way of running analyses that infer things across all uses.

It also makes some transforms tricky.  For example, if we discover after some inlining that one capture is always a constant, then we want to be able to remove it from the structure and constant propagate it into the lambda body.  This is very hard to do with the early lowering.  

There's ongoing work in clang to reverse this design decision there, but it's a multi-year engineering effort to undo after already going down the early-lowering path.  I was hoping that we'd be able to avoid this in Verona and keep lambdas abstract until LLVM IR generation.

src/mlir/design.md
```diff
@@ -3,69 +3,225 @@
...
+That transformation should have already been done by the parser and represented in the AST by the right basic constructs.
+At the MLIR level, they'll be allocated and dispatched in the usual way (stack/heap + call/behaviour dispatch).
+
+**QUESTION: How do we mark the `main` function?**
+Currently, if there is a function called `main`, it's lowered without scope to help testing execution, but we need more strict rules, or perhaps annotation, to mark the entry point.
+We could also have a class named `Main` which is restricted to only one per program, with a method `main(int, int**) : int`, which is the entry point.
```
davidchisnall, line 55: We could also adopt the OpenStep approach.  I'm somewhat biased here, but it's the one that I prefer.  Each program build has a property list that defines a bunch of metadata that doesn't live in any source files (author, version number, and so on) and this includes the name of the application class.  The entry point in the program is in this class.  This makes it fairly easy to build multiple programs with different entry points (e.g. for testing).

src/mlir/design.md
```diff
@@ -3,69 +3,225 @@
...
 MLIR doesn't support allocating stack memory with anything other than `memref` types, but it does support the LLVM dialect, which has full LLVM semantics.
-In this preliminary implementation, we use the LLVM dialect for alloca/load/store/GEP operations when dealing with stack objects.
-This approach seems very clean and sensible, and it could be used for accessing heap variables as well.
-
-For now, every function argument and local declaration (`let`, `var`) are allocated in the stack and its value stored there.
-This simplifies code generation (all variable symbols point to an address), but it doesn't help with value symbols and are really slow.
+We use the LLVM dialect for alloca/load/store/GEP operations when dealing with stack/heap objects.
 
-The LLVM pipeline, even at O1, recognises most of the argument, temporary and local allocas as bogus and optimises them away, so this isn't an immediately critical issue from a performance point of view, but we will want to clean up some of those repetitive patterns soon enough.
+Local declarations (`let`, `var`) without region information are allocated in the stack, via LLVM's `alloca` instruction and their types.
```
davidchisnall, line 75: This presumably means that we'll need to have a later pass to promote some of these to the heap if they are very large or if they escape.  The converse approach (which I think the Go compiler uses) is to model these in the early IR as a heap allocations and then to have a later pass that converts them to stack allocations where possible.

src/mlir/design.md
```diff
@@ -3,69 +3,225 @@
...
+But we will only do so if there is a clear path for optimisation.
 
-The Verona runtime is where the magic happens.
-Allocating regions, choosing the memory management strategy, scheduling behaviours, dispatching foreign language code, etc.
-The AST, however, has only the semantics of the language (with hints to the runtime behaviour), so we need to translate it at the MLIR lowering level.
+## Arithmetic and Builtins
 
-For example, heap allocation, which LLVM IR generally uses calls to `@llvm.malloc`, Verona uses `snmalloc`, which has a finer control over what we can do with the regions.
-In the end, however, it will be the same: calls to the runtime library in IR.
+Arithmetic and builtins are implemented in Verona via direct calls to intrinsics and MLIR/LLMV operations by name (string).
```
davidchisnall, line 83:
```suggestion
Arithmetic and builtins are implemented in Verona via direct calls to intrinsics and MLIR/LLVM operations by name (string).
```

src/mlir/design.md
```diff
@@ -3,69 +3,225 @@
...
+Each numeric class will have their operations defined in the standard library and the compiler will detect the call as _internal_ and create the appropriate intrinsic / operation.
+Builtins (number conversion, checks, cpu-specific instructions) will also be implemented in the standard Verona library.
 
-The generator knows what runtime calls are appropriate for each AST construct, and can insert those calls in the right places and make sure to declare the methods so that linking with the runtime library works.
+Adding two numbers (`a + b`) is the same as calling the method `+` on `a` over `b` (ex. `a.+(b)` or `+(a, b)`).
+These calls will be matched against the already implemented methods in those classes and behave like a normal call (dynamic or static).
 
-## Arithmetic
+These methods are small enough that the LLVM inliner pass will convert all calls into single operations.
+If any of them proves more complex than the inliner can handle, we can add the _always_inline_ attribute to them.
```
davidchisnall, line 91: It's probably worth doing this anyway.  We probably want to inline them even in debug builds (where we may only run the always-inline inliner) and we almost certainly want to run the always-inline inliner at the start of our LLVM pass pipeline to make sure that these things are inlined before we do anything else.  Most of the pre-inliner passes won't do very much good if they're run before they can see arithmetic.

src/mlir/design.md
```diff
@@ -3,69 +3,225 @@
...
-class Foo {
-  static +(a: U32, b: U32) { ... }
-}
-main() : U32 {
-  let a = 21 + 21; // This is simple arithmetic
-  let b = Foo::+(21, 21); // This calls the method
-}
-```
+Also, due to a constraint on the standard dialect in MLIR, we currently only support sign-less integers (neither signed nor unsigned).
+Arithmetic will be lowered in the LLVM dialect, so will contain full support for signed and unsigned as well as all intrisics.
```
davidchisnall, line 103: I'm not quite sure I understand what this means.  LLVM also only has signless integers, it has signed or unsigned operations on integers.  Does MLIR not have these?

src/mlir/design.md
```diff
@@ -3,69 +3,225 @@
...
+ 2. Allocating enough space when we see a type declaration (now the value is the address of the new memory)
+ 3. Accessing the address via references to the variable name (map lookup)
+
+Once more nodes have types (ex: `let`, `var`) we can skip the second step.
+
+## Type capabilities and regions
+
+The three main type capabilities are: `imm`, `mut` and `iso`. See https://microsoft.github.io/verona/explore.html#regions.
+
+Immutable regions will always be allocated on the same reference counted "region", via runtime calls.
```
davidchisnall, line 117: Immutable *regions* don't exist and objects are not ever allocated as immutable (at the abstract machine level, at least).  Immutable objects are created by freezing a mutable region.  The region in which objects are allocated prior to freezing may be any kind of region.

src/mlir/design.md
```diff
@@ -3,69 +3,225 @@
...
+They can only be read from, so there are no concurrency issues. They can be collected when there are no more references to them.
```
davidchisnall, line 118: No concurrency issues, other than needing atomic reference counting.

src/mlir/design.md
```diff
@@ -3,69 +3,225 @@
...
+
+Mutable regions will be allocated on the heap, but cannot be used directly, unless they're in a "forest", accessible through their roots (`iso`s).
```
davidchisnall, line 120: I am not sure what the 'forest' means here.  Regions are allocated in trees, the roots of these trees are cowns, there is a forest of cowns.

src/mlir/design.md
```diff
@@ -3,69 +3,225 @@
...
+
+`Cown` root will be scheduled via the `when` keywords and give no guarantee of completion time or order.
+The runtime calls will be just to create the `cown` and to push it into the scheduler.
+
+**QUESTION: Do stack roots also push behaviours to the scheduler or do they execute synchronously?**
+Either way, they should also yield similar runtime calls.
+
+Regions can also have two more properties:
```
davidchisnall, line 128: At least two more properties.  I suspect we will end up with a complex matrix of properties eventually.

src/mlir/consumer.cc
```diff
@@ -0,0 +1,762 @@
...
+      // Not found, continue searching
+      pos++;
+    }
+
+    // Not found, return false
+    return {0, Type(), false};
+  }
+
+  // ===================================================== AST -> MLIR
+  llvm::Error ASTConsumer::consumeRootModule(Ast ast)
```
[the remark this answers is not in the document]
rengolin: No reason. I think it came from the previous AST which was untyped. I can change that later.

src/mlir/consumer.cc
```diff
@@ -0,0 +1,762 @@
...
+          break;
+        }
+        case Kind::Field:
+        {
+          auto res = consumeField(sub);
+          if (auto err = res.takeError())
+            return err;
+
+          // Update class-field map
+          auto field = nodeAs<Field>(sub);
```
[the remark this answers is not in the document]
rengolin: right now, yes. soon, we'll have two passes. The first, declares all types and functions. The second implements all functions. We don't have tests for using fields before declarations or methods calling functions that weren't yet declared precisely because of that. this is future work.

src/mlir/consumer.cc
```diff
@@ -0,0 +1,762 @@
...
+
+    return llvm::Error::success();
+  }
+
+  llvm::Expected<Value> ASTConsumer::consumeNode(Ast ast)
+  {
+    switch (ast->kind())
+    {
+      case Kind::Lambda:
+        return consumeLambda(ast);
```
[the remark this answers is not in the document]
rengolin: This is a pattern that came from the old AST and I didn't optimise because I wasn't sure how long we'd have the AST like that. It seems long, now, so I'll do that work soon.

src/mlir/consumer.cc
```diff
@@ -0,0 +1,762 @@
...
+  {
+    auto lambda = nodeAs<Lambda>(ast);
+    assert(lambda && "Bad Node");
+
+    // Blocks add lexical context
+    SymbolScopeT var_scope{symbolTable()};
+
+    Value last;
+    llvm::SmallVector<Ast, 1> nodes;
+    for (auto sub : lambda->body)
```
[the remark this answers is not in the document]
rengolin: this is not a lambda per se, it's a `lambda` node, that happens to be the implementation of the function's body, which in itself, is a lambda, but inline. This is not what an actual lambda will look like in Verona, which will have already be converted to anonymous class with an `apply` method and will never get here in the first place.

src/mlir/design.md
rengolin: Precisely.

src/mlir/design.md
rengolin: I like the idea of annotating the entry point. The implementation is simpler and clearer. But this needs to come from the AST, which in turn, comes from the language syntax and user code.

I hate what Python does with the `__main__` thing. I hate even more what Perl does, starting from line 1.

I'd prefer either a static agreed name (like Java's `class Main` + `function main(...)`) or an annotation on the function, like:
```
func blergh(argc: I32, argv: [CharString]) : I32 # entry_point
{
  ...
}
```

rengolin: To be clear: I want to keep the question here for this PR. I don't mean to answer it right now, as this is a complex subject that needs discussion.

rengolin:
thanks for the review David. I'll merge this as interim work and we continue the discussion on the following PRs.

PR #443, opened 2021-06-02
Two small fixes for FreeBSD

davidchisnall:
One fixes finding clang-format, the other fixes building on any platform where there isn't an LLVM download.


mjp41 (approved):
LGTM.

rengolin:
sorry, missed that when I added the clang build...

davidchisnall:
We should be able to cross-build LLVM for FreeBSD quite easily on the Linux hosts.  When you're back, it would be good to have a little chat about how the current nightly LLVM builds work so I can try adding a FreeBSD binary to the set.

PR #444, opened 2021-06-07
Arithmetic as inline MLIR

rengolin:
This is the core of arithmetic in MLIR and it's extremely simple. With time, it'll get more complicated (signed vs unsigned, op vs builtin, special semantics, etc), but it should not change the core.

An example (more available in `arithmetic.verona`):
```
class I32
{
  +(lhs: I32, rhs: I32) : I32
  {
    let res : I32 = "std.addi"(lhs, rhs);
    return res;
  }
  ...
}
```

It builds on previous work to simplify constants and implement structures and string constants in LLVM dialect, and it delivers the simple logic:
* If the operation "looks like" a call to a literal string, it's a "native IR operation".
* If the compiler knows of an IR operation with the exact name as provided, it's a direct lowering to that operation, passing the operands.
* Otherwise, it will be a call to an intrinsic with that name.

The benefits are:
1. Extremely simple to lower.
2. Lets us write arithmetic in pure Verona (no need to special-case anything in the compiler).
3. Extensible, as we can change or add new operations just by calling new LLVM intrinsics or MLIR operations.

There are, however, fundamental problems with this approach:
1. It allows programmers to directly call dialect operations, potentially bypassing language restrictions (like _inline asm_).
2. It depends on the LLVM version and the exact name of the operations / intrinsics, decreasing maintainability.
3. The syntax is crude and may raise syntactic conflict with other string operations.

My view on the problems above:

First, we can restrict support for inline MLIR to the standard library only, not allowing regular users to call intrinsics. This is not a guarantee, as users can feel forced to clone the repo, update our runtime library and release their own toolchain. But this is at least equivalent if we had the logic inside the compiler, as users would have to clone and change that, pretty much in the same way.

Second, we control the LLVM version we pull from and we have no plans to support multiple versions of LLVM. The examples of projects I know that do that have a really hard time coping with the different versions, usually with `#ifdef` hell. Any move of LLVM versions will be followed by source update on our side, which will include any changes to arithmetic calls.

The only minor problem that can occur is if either MLIR or LLVM change the semantics of existing operations, creating hard-to-detect codegen problems and we don't notice until it's too late. Even in those cases, the fix is straightforward and easily backported, if needed to.

Third, this is a temporary syntax and has a small impact on how arithmetic is lowered. We just take the compile-time-constant value of the string (well, its nave even), so replacing it with a token (say `@std.addi` instead) will yield almost identical code to get the name from the operation, changing nothing in the remaining implementation.

All in all, I think it's a good direction, at least for the time being. But I'm releasing this an an RFC, to welcome the point of views from other people, to make sure I'm not missing anything obvious.


davidchisnall:
> First, we can restrict support for inline MLIR to the standard library only, not allowing regular users to call intrinsics

I think the right way to think about this is that we have a `Builtin` package that is provide by the compiler.  This is the lowest-level part of the standard library and contains things that cannot be implemented in pure Verona (e.g. primitive types, arrays).  The `Builtin` package may be implemented differently by two different Verona compilers, even if they share a standard library.

We choose (as an implementation detail that is not visible outside of the `Builtin` package) to use a Verona-like DSL to implement this package in our compiler.  This DSL is a superset of Verona, that adds a simple embedding of MLIR instructions in Verona.  Because this is used only in the `Builtin` package and the `Builtin` package is shipped and versioned with the compiler, there are no stability guarantees about this dialect *at all*: we can change it completely between minor revisions of the compiler.  We don't expose this DSL for any external Verona code (we may even compile the `Builtin` package to an AST as part of the compiler build process and ship that).

rengolin:
> We choose (as an implementation detail that is not visible outside of the `Builtin` package) to use a Verona-like DSL to implement this package in our compiler. This DSL is a superset of Verona, that adds a simple embedding of MLIR instructions in Verona. Because this is used only in the `Builtin` package and the `Builtin` package is shipped and versioned with the compiler, there are no stability guarantees about this dialect _at all_: we can change it completely between minor revisions of the compiler. We don't expose this DSL for any external Verona code (we may even compile the `Builtin` package to an AST as part of the compiler build process and ship that).

Right, this is a more elaborate way of describing the problem and how we can work around it, but still leaves the implementation of such calls implemented in the compiler.

There are a number of (usually bad) ways of restricting this to builtins only:
* (ugly) Implement a hidden flag (environment variable, etc) to allow builtins to be compiled, and use it only on our build of the builtin package.
* (horrible) Restrict the support for builtins from specific files.
* (disgusting) Compile the compiler with a special flag that enables builtins and use that to compile the builtins package only. (only really works with closed source compilers)

From my experience, hidden compiler flags (especially open source ones) have a horrible tendency of surfacing to users, generally (and unfortunately) by suggestions of compiler developers trying to avoid implementing something more complicated.

However, hidden compiler flags also have a very clear and universal notion of being completely unsupported "here-be-dragons" kind of thing, so any incompatibility in user code will be "user error". This is the least problematic from my point of view.

rengolin:
Ok, we seems to be in consensus that this is the direction we want to go, so I removed the RFC. I'll continue working on this direction and will merge this after a grace period.

src/mlir/generator.cc
```diff
@@ -395,31 +409,86 @@ namespace mlir::verona
...
+
+  Value MLIRGenerator::Arithmetic(Location loc, StringRef name, Value ops)
+  {
+    // FIXME: We already converted U32 to i32 so this "works". But we need
+    // to make sure we want that conversion as early as it is, and if not,
+    // we need to implement this as a standard select and convert that
+    // later. However, that would only work if U32 has a method named "+",
+    // or if we declare it on the fly and then clean up when we remove the
+    // call.
+    auto numOps = arithmetic[name];
```
plietar, line 449: Use `arithmetic.at(name)` instead. This will swallow invalid names and use `0` as the arity.

rengolin: Wait, this doesn't look right.
https://www.cplusplus.com/reference/map/map/at/ says:
```
Returns a reference to the mapped value of the element identified with key k.

If k does not match the key of any element in the container, the function throws an out_of_range exception.
```
But also that the operator `[]` creates an empty element if none exists, which is not what I want there either. 

I'm changing it to the common `find` pattern, which is what I should have done in the first place. :)
```
auto it = ...find(name);
assert(it != ...end());
auto numOps = it->second;
```

davidchisnall: `.at` is what you want here.  You then don't need the `assert`, because the uncaught exception will end the program.  If you compile without exceptions, [this aborts if it doesn't find the value](https://github.com/llvm/llvm-project/blob/bfbd73f87d7b2c1140b43192cea6b7fbbfc35cd0/libcxx/include/stdexcept#L270).

src/mlir/generator.h
```diff
@@ -44,6 +44,12 @@ namespace mlir::verona
     /// Symbol tables for variables.
     SymbolTableT symbolTable;
 
+    /// Set of known arithmetic operations and their number of operators
+    std::map<llvm::StringRef, size_t> arithmetic;
```
davidchisnall, line 48: This probably wants to be a `std::unordered_list` (lookup is faster and I don't think you're using the ordering).  [`llvm::StringMap`](https://llvm.org/doxygen/classllvm_1_1StringMap.html) may be even better.

src/mlir/generator.cc
```diff
@@ -12,6 +12,64 @@
...
 {
+  void MLIRGenerator::initializeArithmetic()
+  {
+    // FIXME: Using MLIR standard dialect comes with complications, since the
+    // arithmetic in MLIR is mostly for ML and doens't have the semantics we
+    // need for generic lowering. We should probably use the LLVM dialect + LLVM
+    // intrinsic calls for the rest.
+
+    // Floating-point arithmetic
+    arithmetic.emplace("std.absf", 1);
```
davidchisnall, line 23: Why not write this as an initialiser list in the constructor?  Something like:

```
arithmetic({ "std.absf", 1},
           { "std.ceilf", 1}, 
           ...)
```

Or, since this is a simple string-to-number map, remove this field entirely and just have a function that does the lookup and use `llvm::StringSwitch`, which typically generates much better code for this kind of thing.

rengolin: > Why not write this as an initialiser list in the constructor?

It grew from a few examples.

> Or, since this is a simple string-to-number map, remove this field entirely and just have a function that does the lookup and use `llvm::StringSwitch`, which typically generates much better code for this kind of thing.

I wasn't sure this wouldn't need multiple points of insertion. But I think we can cross that bridge when we get there.

I'll change to `StringSwitch`.

src/mlir/generator.h
rengolin: A `StringSwitch` like you proposed below would be even better.

src/mlir/generator.cc
rengolin: exception/aborting is worse than asserts because it provides no context. Though, now we're using `StringSwitch` and this isn't relevant any longer.

PR #445, opened 2021-06-08
Remove previous ASTs

rengolin:
We have moved to a new typed AST that will (hopefully) be our last, so
we should remove the previous iterations that led here, mainly to clean
up external dependencies (cpp-peglib).

If we need to go back and look at the code, we can always git checkout
before this commit, so we shouldn't move it to an `old` directory
instead.


davidchisnall:
Are we still using Pegmatite for anything or can that be GC'd as well?

rengolin:
> Are we still using Pegmatite for anything or can that be GC'd as well?

That's still used by `src/compiler` (the prototype) which is the only "functional" part of the project, currently.

I'm not sure how distant is the syntax and semantics of what the prototype implements versus what we are planning now, so I didn't include in this PR. But I'm not opposed to wiping that, too, if that's the general consensus.

@sylvanc @plietar should know better about that.

plietar:
Please don't remove `src/compiler` until you can run code with the new pipeline.

rengolin:
> Please don't remove `src/compiler` until you can run code with the new pipeline.

That's what I thought, thanks!

sylvanc (approved):

PR #447, opened 2021-06-10
Supporting signed/unsigned types, no auto-cast

rengolin:
This commit adds signed/unsigned integer types and moves all cast
operations to the standard library implementation. For now, it's
implemented on the tests themselves, but that implementation should move
to `stdlib` and the tests should include that.

From now on, we assume that all casts and type promotions will happen at
the AST level and we'll get the correct types for all arithmetic
operations, so this commit also adds a few asserts  to make sure that's
true.

See `testsuite/verona-mlir/arithmetic.verona` for an idea how arithmetic
functions will look like. Soon, inline MLIR calls will change syntax
(probably something like @std.addi instead of "std.addi") and we can
move the implementation to the standard library.

### DIFF
--- a/src/mlir/generator.cc
+++ b/src/mlir/generator.cc
@@ -189,66 +189,6 @@ namespace mlir::verona
 
   // ==================================================== Low level generators
 
-  Value MLIRGenerator::Convert(Value val, Type ty)
-  {
-    auto valTy = val.getType();
-    auto valSize = valTy.getIntOrFloatBitWidth();
-    auto tySize = ty.getIntOrFloatBitWidth();
-    if (valSize == tySize)
-      return val;
-
-    // Integer upcasts
-    // TODO: Consiger sign, too
-    auto valInt = valTy.dyn_cast<IntegerType>();
-    auto tyInt = ty.dyn_cast<IntegerType>();
-    if (valInt && tyInt)
-    {
-      if (valSize < tySize)
-        return builder.create<SignExtendIOp>(val.getLoc(), ty, val);
-      else
-        return builder.create<TruncateIOp>(val.getLoc(), ty, val);
-    }
-
-    // Floating point casts
-    auto valFP = valTy.dyn_cast<FloatType>();
-    auto tyFP = ty.dyn_cast<FloatType>();
-    if (valFP && tyFP)
-    {
-      if (valSize < tySize)
-        return builder.create<FPExtOp>(val.getLoc(), ty, val);
-      else
-        return builder.create<FPTruncOp>(val.getLoc(), ty, val);
-    }
-
-    // If not compatible, assert
-    assert(false && "Type casts between incompatible types");
-
-    // Appease MSVC warnings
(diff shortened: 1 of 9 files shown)


rengolin:
This is an example of:
* what an implementation is Verona will look like
* what I'm expecting the AST to give me in terms of type operations

Without type inference in the AST, differentiation between static calls and dynamic calls, and a proper syntax for inline MLIR calls, creating more examples won't add much. 

This seems to me like a good point to leave arithmetic hanging and look at other things I still can with this AST, for example:
* declaration order: implement a two-pass compiler (declare -> define)
* visitor pattern: simplify `consume` methods

Once we have the new AST, we should move the implementation of arithmetic to `stdlib/numbers` and get the tests to use them directly.

testsuite/verona-mlir/struct.verona
```diff
@@ -50,7 +57,7 @@ foo() : I32
   boop.d = PI;
 
   // Test reading / writing to fields
-  foo.a = 12;
+  foo.a = I64::trunc(12);
```
davidchisnall, line 60: I would expect type inference to start with a set of possible types for the integer literal and then to define the final type as `I32` based on the assignment.  If there is a bug preventing that in the front end, please can you leave a FIXME in the test so that it can be cleaned up later?

src/mlir/generator.cc
```diff
@@ -428,7 +368,8 @@ namespace mlir::verona
     return builder.create<LLVM::AddressOfOp>(builder.getUnknownLoc(), global);
   }
 
-  Value MLIRGenerator::Arithmetic(Location loc, StringRef name, Value ops)
+  Value
+  MLIRGenerator::Arithmetic(Location loc, StringRef name, Value ops, Type retTy)
   {
     // FIXME: We already converted U32 to i32 so this "works". But we need
```
davidchisnall, line 374: Is this comment stale now?

davidchisnall (approved):
Looks good to me, a couple of small questions inline.  We probably need to think a bit more about the names of the conversion methods for the version in the `Builtin` package, but that doesn't matter for the versions in the tests.

src/mlir/generator.cc
rengolin: No, that's not about unsigned vs. signed, just that the classes called `U32` and `I32` etc are converted to MLIR native types very early. This is only a problem if we end up with dynamic dispatch (ex. `a.+(b)`) without knowing the type of `a`. I'm not sure all the cases, but I imagine a union `I32 | F32` where both support `+` would be such a case.

Right now, because we can't implement arithmetic in Verona yet (need the planned changes in the AST for that), it should be fine.

testsuite/verona-mlir/struct.verona
rengolin: It will be, but for now, AST nodes don't have types, so I have to define some type, and I chose 64-bit types for both integer and floating point literals. There is a FIXME on the code for that already.

PR #448, opened 2021-06-15
[nightly.yml] Add a new stage for building Verona+LLVM at once on Linux

almiliMSFT:
The only functional change should be the addition of the new stage.

This PR also includes a syntactic change which breaks up the nightly pipeline into the following stages
  - Linux (Verona)
  - Linux (Verona+LLVM)
  - Windows (Verona)
  - macOS (Verona)

Having stages is convenient because the ADO GUI allows for selecting which stages to execute when enqueuing a new build (and so, it becomes easy to run partial validations).

### DIFF
--- a/devops/nightly.yml
+++ b/devops/nightly.yml
@@ -137,11 +137,6 @@ stages:
     - bash: |
         set -euo pipefail
 
-        # drop an AnyBuild.json config file and use it to
-        #   - ignore the CMAKE_BUILD_PARALLEL_LEVEL variable
-        #   - limit the number of concurrent cache lookups
-        echo '{"ActionCache": { "MaxParallelLookups": 10, "IgnoredEnvVars": [ "CMAKE_BUILD_PARALLEL_LEVEL" ] }}' | tee AnyBuild.json
-
         # remove '-march=native' which is incompatible with remote execution
         sed -i 's/ -march=native//g' src/rt/CMakeLists.txt
 
--- a/devops/nightly.yml
+++ b/devops/nightly.yml
@@ -137,19 +137,6 @@ stages:
     - bash: |
         set -euo pipefail
 
-        # remove '-march=native' which is incompatible with remote execution
-        sed -i 's/ -march=native//g' src/rt/CMakeLists.txt
-
-        echo "=============="
-        echo "Patch applied:"
-        echo "=============="
-        echo
-        git diff
-      displayName: Configure AnyBuild
-
-    - bash: |
-        set -euo pipefail
-
         mkdir -p build
         cd build
         cmake                                      \
--- a/devops/nightly.yml
+++ b/devops/nightly.yml
@@ -1,5 +1,5 @@
 parameters:
-  # number of parallel cmake jobs when building LLVM on Linux
+  # number of parallel cmake jobs when building with AnyBuild
   - name: AnyBuildParallelism
     type: number
     default: 48
--- a/devops/nightly.yml
+++ b/devops/nightly.yml
@@ -1,3 +1,9 @@
+parameters:
+  # number of parallel cmake jobs when building LLVM on Linux
+  - name: AnyBuildParallelism
+    type: number
+    default: 48
+
 resources:
 - repo: self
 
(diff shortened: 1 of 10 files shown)


devops/nightly.yml
```diff
@@ -13,182 +19,275 @@ schedules:
...
-
-  - script: |
-      set -eo pipefail
-      export ASAN_SYMBOLIZER_PATH=/usr/lib/llvm-6.0/bin/llvm-symbolizer
-      export ASAN_OPTIONS="alloc_dealloc_mismatch=0 symbolize=1"
-      ninja check
-    workingDirectory: build
-    displayName: 'Tests'
+stages:
+- stage: Linux
```
almiliMSFT, line 24: This stage should be exactly as it was before.  The only syntactic change is that I extracted some steps into templates because I wanted to reuse them in the definition of the Verona+LLVM stage.

devops/nightly.yml
```diff
@@ -13,182 +19,275 @@ schedules:
...
+        set -eo pipefail
+        ninja
+      workingDirectory: build
+      displayName: 'Compile'
+
+    - template: ./steps/linux/test.yml
+      parameters:
+        WorkingDirectory: build
+
+- stage: LinuxFull
```
almiliMSFT, line 98: This is the new stage which builds Verona+LLVM from sources in a single `cmake --build` invocation.

devops/nightly.yml
```diff
@@ -13,182 +19,275 @@ schedules:
...
+  displayName: Linux (Verona + LLVM)
+  dependsOn: []
+  jobs:
+  - job:
+    displayName: Linux
+    pool:
+      vmImage: 'ubuntu-20.04'
+    timeoutInMinutes: 120
+    strategy:
+      matrix:
```
almiliMSFT, line 108: We can add more configurations here, or use different ones.

devops/nightly.yml
```diff
@@ -13,182 +19,275 @@ schedules:
...
+
+    - template: ./steps/linux/dependencies.yml
+
+    - bash: |
+        set -euo pipefail
+
+        # drop an AnyBuild.json config file and use it to
+        #   - ignore the CMAKE_BUILD_PARALLEL_LEVEL variable
+        #   - limit the number of concurrent cache lookups
+        echo '{"ActionCache": { "MaxParallelLookups": 10, "IgnoredEnvVars": [ "CMAKE_BUILD_PARALLEL_LEVEL" ] }}' | tee AnyBuild.json
```
almiliMSFT, line 143: We could check in this `AnyBuild.json` file instead of dynamically generating it here, whatever you prefer.

devops/nightly.yml
```diff
@@ -13,182 +19,275 @@ schedules:
...
+
+        # remove '-march=native' which is incompatible with remote execution
+        sed -i 's/ -march=native//g' src/rt/CMakeLists.txt
```
almiliMSFT, line 146: This is more interesting.  Clearly, `-march=native` is incompatible with remote execution.  By default, AnyBuild ignores all compilation processes that include this switch and forces them to execute locally.  The problem with that is that processes that are executed locally (outside of AnyBuild) cannot be cached, and in this particular case that really hurts the end-to-end build time.

So you can decide whether you want to remove this switch from your `CMakeLists.txt` file(s) or we can remove them here (just for the sake of being able to execute everything with AnyBuild).

devops/nightly.yml
```diff
@@ -13,182 +19,275 @@ schedules:
...
+      parameters:
+        BuildName: Verona-$(BuildType)-$(CC)-Asan$(Asan)-AnyBuild
+        BuildCommandLine: env CMAKE_BUILD_PARALLEL_LEVEL=${{ parameters.AnyBuildParallelism }} cmake --build .
+        WorkingDirectory: build
+
+    - template: ./steps/linux/test.yml
+      parameters:
+        WorkingDirectory: build
 
 ############################################## Windows Builds
```
almiliMSFT, line 190: everything below should be exactly the same, just indented to accommodate the newly introduced scopes

external/CMakeLists.txt
```diff
@@ -63,6 +63,7 @@ if (VERONA_DOWNLOAD_LLVM)
     USES_TERMINAL_DOWNLOAD true
   )
 else()
+  separate_arguments(LLVM_EXTRA_CMAKE_ARGS UNIX_COMMAND ${LLVM_EXTRA_CMAKE_ARGS})
```
almiliMSFT, line 66: I'm not very familiar with cmake, but I think this was a legitimate bug with this spec.  Without this, whatever I would set via `-DLLVM_EXTRACMAKE_ARGS="..."` wasn't getting propagated to the cmake command used to configure the LLVM build.

devops/nightly.yml
rengolin: Would this be shared between different pipelines (ci, nightly)?

devops/nightly.yml
rengolin: IIRC, we use this because the default option (minimum common denominator) is really bad for performance. We discussed using a minimal set but that'd break builds on users' older machines. We don't really need anything special.

I guess in CI we don't really need such performance and we could drop the `march` entirely. @mjp41?

external/CMakeLists.txt
rengolin: We had that problem before, where some high-level flags weren't propagating to the sub-projects. @mjp41 knows more how that works.

devops/nightly.yml
almiliMSFT: Yes, I expect that for all different pipelines we'd use the same AnyBuild configuration.

For example, `MaxParallelLookups` really depends on the size of the VM where the build is running.  But for all the current pipelines I'm aware of, 10 should be a reasonable setting.

devops/steps/tar-and-upload.yml
```diff
@@ -5,13 +5,18 @@ parameters:
...
   type: string
 
 steps:
   - bash: |
       set -euo pipefail
-      readonly PKG_NAME="${{ parameters.PackageName }}"
+
+      readonly CommitLong="${{ parameters.PackageCommit }}"
+
+      PKG_NAME="${{ parameters.PackageName }}-${CommitLong:0:11}"
```
almiliMSFT, line 19: This is the change to shorten the commit SHA when constructing the package name

almiliMSFT:
I think this is ready to be checked in.  We can make incremental improvements as necessary.

devops/nightly.yml
rengolin: So perhaps this could be checked in as a separate json file and all users include the same.

rengolin:
> I think this is ready to be checked in. We can make incremental improvements as necessary.

Thanks Aleksandar, I'm just waiting @mjp41's comments on the two remaining discussions and it'll be good to go.

devops/nightly.yml
mjp41: We can drop it.  This is from quite early in our thinking, and was inherited from snmalloc, but now we don't have that on snmalloc.  I am happy for that to go.

external/CMakeLists.txt
mjp41: I thought it was fixed, but it is untested.  Was it broken for a single additional argument, or only for multiple.

mjp41:
Should we apply two changes I am tagged in to master as well?

rengolin:
> Should we apply two changes I am tagged in to master as well?

I think so. @almiliMSFT, if you feel comfortable, can you clean up those two issues like you proposed? It probably needs to touch other CMake files, but that should be fine.

external/CMakeLists.txt
rengolin: Only for multiple.

devops/nightly.yml
almiliMSFT: Remove `-march=native` from the `CmakeLists.txt` file.

almiliMSFT:
> > Should we apply two changes I am tagged in to master as well?
> 
> I think so. @almiliMSFT, if you feel comfortable, can you clean up those two issues like you proposed? It probably needs to touch other CMake files, but that should be fine.

I just updated `src/rt/CMakeLists.txt` to remove `-march=native` from it (and that's the only occurrence of it in this repo).  

Is there anything else?  I think that's it.

rengolin:
> Is there anything else? I think that's it.

That's it, yes, thank you!

rengolin (approved):

PR #451, opened 2021-06-20
Move MLIR Generator to AST Pass

rengolin:
This PR moves the previous monolithic generator to two AST passes (`pass.h`):
* `ASTDeclarations`, which goes through types, classes and functions and declares them on the main MLIR module.
* `ASTDefinitions`, which scans all functions and implements them in order.

This fixes the declaration order problem, as all methods and types are defined by the time of definition. It also allows us to use the error handling from the AST pass structure.

Right now, we're not using anything but `assert`s because every error in the current passes are internal logic errors, things that the AST should have done right beforehand, and so we want the execution to stop, making it easier for following it with a debugger.

There are two hacks to make it work now, both of them because the AST is still incomplete:
* Nodes don't have types. An assign address can contain the type for an expression that doesn't know it's own type (ex. arithmetic), so we need to set it on `pre(Assign)` and clear on `post(Select)`.
* Not all operations create temporaries. The traversal functions return `void`, so we must keep the internal temporaries (inline expression in another expression) by pushing them into a stack and making sure the wrapping expression uses them all.

Once the AST has types and creates temporaries for all expressions, leaving operands as simple references, we'll be able to remove these two hacks.

There still are the previous classes:
* `ASTConsumer`, which now is a driver for all AST passes and the place for AST helper functions. This may fade away in future versions, but we still need helper functions somewhere.
* `MLIRGenerator`, which still does all it did before, and more. It holds the owning reference of the MLIR module, so that the AST passes, the generators as well as future MLIR passes can all write to the same module, with the same builder, without having to pass along context in arguments.

Some tests were fixed to set the right type, or simplify the feature being tested without others getting in the way, but they still have the exact same meaning. All tests were generated all the way to LLVM IR and look correct on their optimised versions (correct expressions, constant folding to the expected values).


davidchisnall (approved):
Looks good, the new visitor pattern is a lot nicer than the older code.  Note that the Clang AST visitor allows the visitor to control recursion.  If our AST visitor allowed that then some of this could be simplified by pulling the function visitor into a separate (nested) class so that it could push and pop the symbol table in its constructor / destructor: rather than relying on a single visitor that keeps a stack, you could have a stack of visitors.

rengolin:
> Note that the Clang AST visitor allows the visitor to control recursion. If our AST visitor allowed that then some of this could be simplified by pulling the function visitor into a separate (nested) class so that it could push and pop the symbol table in its constructor / destructor: rather than relying on a single visitor that keeps a stack, you could have a stack of visitors.

I don't think we'll need that here, because the AST will be pure A-normal form and ultimately, the only context carried will be the symbol table and the MLIR module and context.

PR #453, opened 2021-06-21
Add some internal documents

rengolin:
A very crude set of documents, not unlike the others in the same directory. It may contains some blatant errors, as this is just what I could write from our previous meetings. Feel free to correct me and propose better examples and descriptions.

I'm not sure how to categorise them, so I just created some random documents that we can refer back to later.

I'm hoping in due time we'll have a better set of design documents, organised and clear, but I don't think this is the right time, nor that I'm the right person (right now).

I'd be happy to create such document later, but I'd rather have the hind-sight of experience.

### DIFF
--- a/docs/internal/arithmetic.md
+++ b/docs/internal/arithmetic.md
@@ -1,0 +1,51 @@
+# Arithmetic
+
+Numeric types in Verona are nothing but classes with methods that implement known arithmetic functions.
+
+For example:
+```
+  let x : I8 & mut = 42;
+  let y : I8 & mut = 24;
+
+  // This is what a programmer would write
+  let sum = x + y; // type inference to I8
+
+  // This is what it would turn to
+  let sum = x.+(y);
+  // Or this
+  let sum = I8::+(x, y);
+```
+
+Operations in Verona are nothing but function calls with numeric types.
+If an operation refers to a function that is not implemented, this is a syntax error.
+
+For example:
+```
+  let x : I8 & mut = 42;
+  let y : I16 & mut = 24;
+
+  // There is no such function I8::+(I8, I16);
+  let sum = x + y; // ERROR
+
+  // We can extend x, then add
+  let z : I16 & mut = x.toI16();
+  let sum = z + y; // type inference to I16
+
+  // This is what it would turn to
+  let sum = z.+(y);
+  // Or this
+  let sum = I16::+(z, y);
+```
+
+Verona does not have type casts, auto-promotion or guessing the types of arguments to functions.
+
+The implementation of `x + y` to `x.+(w)` isn't really a conversion, but a syntactic feature of the language.
+Prefix and infix operations are handled in the same way, so `a foo b` is the same as `a.foo(b)` and `ATy::foo(a, b)`.
+
+## Type-boxing
+
+To represent dynamic types, we'll use type boxing, which uses the fact that IEEE-754 64-bit floating point numbers have 53 bits left for NaN encodings (aka. NaN-boxing).
+With around 48 bits used to represent pointers, and any integer up to 32-bits fitting that space, too, we can use the left-over bits to tag which type the 64-bit pattern represents.
+Obviously, 64-bit integers cannot be represented in that manner, so it will need a longer representation (tag+payload) on different words.
+
+The implementation details are still under discussion, which will determine what type of boxing we want and which bits we'll use for what.
--- a/docs/internal/arithmetic.md
+++ b/docs/internal/arithmetic.md
@@ -4,8 +4,8 @@ Numeric types in Verona are nothing but classes with methods that implement know
 
 For example:
 ```
-  let x : I8 & mut = 42;
-  let y : I8 & mut = 24;
+  let x : I8 & imm = 42;
+  let y : I8 & imm = 24;
 
   // This is what a programmer would write
   let sum = x + y; // type inference to I8
@@ -21,14 +21,14 @@ If an operation refers to a function that is not implemented, this is a syntax e
 
 For example:
 ```
-  let x : I8 & mut = 42;
-  let y : I16 & mut = 24;
+  let x : I8 & imm = 42;
+  let y : I16 & imm = 24;
 
   // There is no such function I8::+(I8, I16);
   let sum = x + y; // ERROR
 
   // We can extend x, then add
-  let z : I16 & mut = x.toI16();
+  let z : I16 & imm = x.toI16();
   let sum = z + y; // type inference to I16
 
   // This is what it would turn to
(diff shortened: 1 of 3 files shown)


docs/internal/arithmetic.md
```diff
@@ -0,0 +1,51 @@
+# Arithmetic
+
+Numeric types in Verona are nothing but classes with methods that implement known arithmetic functions.
+
+For example:
+```
+  let x : I8 & mut = 42;
```
davidchisnall, line 7: I forgot to mention in the meeting, but I don't think `I8 & mut` actually makes sense as a thing.  `I8 & imm` should be a valid type.  You can't modify the value `42`, you can only modify the variable `x` (if it were `var`, not `let`) to refer to a different integer.  

Conceptually, we have one concrete instance of every possible number of every numeric type.  As an optimisation, we store the actual number, rather than a pointer to an object containing the number, in the variable.  As a result of this optimisation, we can't actually express `I8 & mut` unless we add an allocation.

For a concrete illustration of the difference:

```
var y : I8 & mut = 42;
var z = y;
z++;
```

Do you expect the value of `y` to change because you've modified the object that `z` refers to?

davidchisnall (commented):
Thanks for writing this up.  I'll finish reading it tomorrow, but submitting one quick comment so it doesn't get lost...

docs/internal/arithmetic.md
rengolin: You're right. I'm still confused as to whether this is a special case of numeric types or literal assignment, but in this case, it doesn't matter. I can easily replace it with `I8 & imm` and not worry about it in this section.

But perhaps it would be interesting to follow that thread in a separate section of the same document.

In your example, `z++` is `constexpr` so directly evaluated to `43` a compile time, but:
```
  func(a: I8) {
    var y : I8 & mut = 42;
    var z = y;
    z += a;
  }
```
where `a` cannot be inferred at compile time. Here, `z` first has a "copy" of `42` and reassign whatever the sum is next.

Still, the storage space is the same: bit-width. Dynamic dispatch on `z.+=(a)` would take the descriptor of `z`'s type, call the offset of `+=` and pass `z` and `a`, storing back on the same place the result. (assuming we have the `+=` operator).

I think the same is true for strings:
```
  var a : CString = "abc";
  var b : CString = a;
  a.concat("def");
```
Here, also `constexpr`, `a = "abc"` and `b = "abcdef"`. Those would be references to compile-time constant literals, and if we uniquefy strings, equality can be just a pointer comparison. If the operations cannot be constant folded, then comparison is a `strcmp`.

Does that make sense?

docs/internal/inheritance.md
```diff
@@ -0,0 +1,45 @@
...
+## Capabilities
+
+Like classes, capabilities can be sub-types of each other.
+This is used to allow objects with different capabilities to reference each other.
+
+Example:
+```
+  let x : Foo & mut = ...;
+
+  // This is fine, as `readonly` is a sub-type of `mut`
```
plietar, line 28: You got subtyping the wrong way round, `mut` would be subtype of `readonly`.

docs/internal/arithmetic.md
davidchisnall: Strings are an entirely separate thing and it's still not clear to me what the correct behaviour for string literals should be (possibly similar to integers, where it can be unified with any string type during type inference?), but the string example is also similar and has nothing to do with constant propagation or constant expression evaluation:

```
var a : SomeStringType & mut = SomeStringType("foo");
var b = a; // b and a are now pointers to the same string object.
b.append("bar");
// Both a and b refer to the same object that now contains the string "foobar"
```

Whether we evaluate the `append` call at compile time or run time, both `a` and `b` in this example will point to the same string and that string will have been mutated.

You definitely don't want this behaviour for integers, because it would require every integer variable to actually be a pointer to an integer, rather than simply an integer.  This example:

```
var a : U32 & mut = 1;
var b = a;
a++;
```

Would look something like this in C:

```
int *a = malloc(sizeof(int));
*a = 1;
int &b = a;
(*a)++;
```

If the types are `imm`, the naïve C equivalent is:

```
int *a = malloc(sizeof(int));
*a = 1;
int &b = a;
int *tmp = malloc(sizeof(int));
*tmp = (*a) + 1;
a = tmp;
```

Because the value of the integer in the allocation never changes; however, we can perform a simple optimisation and embed the value in the pointer.  Because Verona knows that the underlying type is `U32`, rather than a generic pointer-to-some-kind-of-object, we can avoid needing any boxing tricks and just make the underlying type of the variable a 32-bit integer.

rengolin: That's why I used `CString` (to mean `const char`) but perhaps I should have said `CString & imm`. A generic string object is just like any other object, reference, mutability and all. I was just talking about constant literal sequences of chars.

docs/internal/inheritance.md
rengolin: Ah, yes! I always mix that up. I instinctively see `readonly` as more restrictive, and therefore "less". 

But here, `mut` can be used as a `readonly` (by not changing it) but not the other way round, so `mut` is a subtype of `readonly`.

docs/internal/object-memory.md
```diff
@@ -0,0 +1,78 @@
...
+```
+  let a : Foo & iso = Foo();
+  let b : Foo & imm = Builtin::freeze(a);
+
+  // No guarantees some random behaviour is using our immutable data
+  let c : Foo & iso = b; // ERROR
+```
+
+Freezing a mutable object is an error, as we are not using the entry point (isolated reference).
+To freeze one mutable object, one needs to find the `iso` and freeze that.
```
davidchisnall, line 68: I'd prefer to phrase this as saying that you cannot freeze a single object, you must freeze an entire region.  This guarantees that there are no outstanding mutable references to an immutable object.

docs/internal/object-memory.md
```diff
@@ -0,0 +1,78 @@
...
+Freezing an `iso` freezes **all** objects reachable (dominated by) that `iso` reference.
+
+## Changing region types
+
+As stated, there are no automatic conversion, not even between region types.
+Creating an object in one region type and then moving into another requires users to code that explicitly.
+There isn't a syntax for selecting regions types yet, but the code must deep copy by hand.
+
+The idea is to create a new region, `when` both, and then copy element by element.
```
davidchisnall, line 77: I don't understand the `when` reference here.  You don't need a `when` to be able to access a region, only a `cown` and you would not create a new `cown` for this.

docs/internal/object-memory.md
```diff
@@ -0,0 +1,78 @@
...
+This is tedious and error-prone, but can be made better by coding deep-copies in the classes themselves.
```
davidchisnall, line 78: I don't understand what this means.

davidchisnall (approved):

PR #455, opened 2021-06-28
Use automatic cleanup for symbol stack pop

rengolin:
This allows us to return at any time from the post methods while making
sure we unwind the stack correctly on all cases.

For now there's only one case where this is beneficial, but it's a good pattern to keep.

### DIFF
--- a/src/mlir/consumer.cc
+++ b/src/mlir/consumer.cc
@@ -258,12 +258,12 @@ namespace mlir::verona
     /// Closes the function, checking the return value
     void post(Function& node)
     {
+      // Automatically clean up
+      ScopeCleanup([&]() { symbolTable().popScope(); });
+
       // Check if needs to return a value at all
       if (gen.hasTerminator(builder().getBlock()))
-      {
-        symbolTable().popScope();
         return;
-      }
 
       // Fetch the current function
       auto loc = con.getLocation(node);
@@ -274,9 +274,6 @@ namespace mlir::verona
       // Lower return value
       auto val = takeOperand(/*last=*/true);
       gen.Return(loc, func, val);
-
-      // Pop function's variable scope
-      symbolTable().popScope();
     }
 
     /// Local declarations (including temps) reserve a place on the symbol table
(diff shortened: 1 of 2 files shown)


src/mlir/consumer.cc
```diff
@@ -258,12 +258,12 @@ namespace mlir::verona
     /// Closes the function, checking the return value
     void post(Function& node)
     {
+      // Automatically clean up
+      ScopeCleanup([&]() { symbolTable().popScope(); });
```
davidchisnall, line 262: Does this actually work?  I was under the impression that this would create a new object that has the lifetime of the expression and so does the pop immediately.  I'd expect this to be more like:

```
ScopeCleanup pop([&]() { symbolTable().popScope(); });
```
I wrote a simple test that confirmed my suspicion:

```c++
ScopeCleanup([]() { foo(); });
baz();
```

Generates:

```llvm
  tail call void @_Z3foov() #2
  tail call void @_Z3bazv() #2
```

```suggestion
      ScopeCleanup pop([&]() { symbolTable().popScope(); });
```

src/mlir/consumer.h
```diff
@@ -86,4 +86,25 @@ namespace mlir::verona
     static llvm::Expected<OwningModuleRef>
     lower(MLIRContext* context, ::verona::parser::Ast ast);
   };
+
+  /*
+   * Scope Cleanup helper
+   *
+   * Automatically performs cleanup on destruction.
```
davidchisnall, line 93:
```suggestion
   * Automatically invokes the callable object passed to the constructor on destruction.
   * This class is intended to provide lexically scoped cleanups, for example:
   * ```c++
   * ScopedCleanup defer([&] { /* cleanup code here */});
   * ```
   * The code in the lambda will be invoked when `defer` goes out of scope.
```

davidchisnall (changes requested):

src/mlir/consumer.cc
rengolin: D'oh!

rengolin:
I think I have tests that check the stack not being popped, but I don't have one for it being popped too early.

src/mlir/consumer.cc
rengolin: I think I have tests that check the stack not being popped, but I don't have one for it being popped too early.

davidchisnall (approved):

PR #456, opened 2021-06-29
Adding types documentation

rengolin:
This is work in progress towards a tutorial like Pony (https://tutorial.ponylang.io/). But first, we need to make sure the documentation is accurate with the syntax and semantics of the language.

The overall idea is to write each of the "chapters" as a separate document and then when we're happy with their contents, we can collate them into a tutorial via some existing document handling software (like Pony did).

Note that this is now meant to be correct, so all examples must conform to the expected syntax and semantics and we should have no more hand-waving code. If any of the examples are syntactically or semantically wrong, we need to fix them now.

Also note that I'm not the best person to write this document, as there are a lot of the syntax and semantics that I don't know or am confused about, so please double check that what I wrote is correct. It very likely won't be.

If we're all happy with the content and the strategy of writing one file at a time, others will follow as soon as we merge this one.

PS: This also removes 1IR doc as we're not using. I'd be fine to do that on a separate PR, too.


docs/internal/types.md
```diff
@@ -0,0 +1,236 @@
...
+  // All derived classes will have the common field
+  var common : F64;
+
+  // Derived classes must implement this method
+  toString() : String;
+};
+
+class MyType : Serialisable
+{
+  // Inherits `common`
```
SeanTAllen, line 51: this is a little confusing. i think it might be better to show the "source you write" vs "the source you get" with the inclusion of common.

docs/internal/types.md
```diff
@@ -0,0 +1,236 @@
...
+  toString() : String
+  {
+    ...
+  }
+};
+```
+
+### Fields
+
+Define objects that belong to the class and will be instantiated (ie. use memory) on every instance of the class.
```
SeanTAllen, line 71: belong to a class or interface, not just class, right?

docs/internal/types.md
```diff
@@ -0,0 +1,236 @@
...
+
+Fields can be initialised in two ways:
+ * `var`: Must be initialised on construction, but can change after.
+ * `let`: Must be initialised on construction, and cannot change later.
```
SeanTAllen, line 75: not sure what `var` and `let` have to do with initalisation. I think you mean something different here.

docs/internal/types.md
```diff
@@ -0,0 +1,236 @@
...
+
+They can also be a private member, if its name starts with `_`.
```
SeanTAllen, line 77: I would change this to

"Fields whose name starts with `_` are private"

(but then what private means for verona needs to be defined somewhere, perhaps here.)

docs/internal/types.md
```diff
@@ -0,0 +1,236 @@
...
+
+They can also have a keyword `embed`, which means ...
```
SeanTAllen, line 79: Missing content here. The current writing makes it sound like I can do `let embed foo`.

docs/internal/types.md
```diff
@@ -0,0 +1,236 @@
...
+
+### Methods
+
+Methods are functions that belong to a class or interface. Like functions, they can have arguments and a return type.
```
SeanTAllen, line 83: there's nothing in this doc that discusses free standing functions so the "Like functions" is confusing.

docs/internal/types.md
```diff
@@ -0,0 +1,236 @@
...
+
+Methods can be called statically (via `Class::method()`) or dynamically (via `object.method()`).
+On the static case, the `self` pointer is a descriptor forthe type, while on the dynamic case, it's a descriptor to the instance (and its vtable, type info, etc).
```
SeanTAllen, line 86: typo: forthe

docs/internal/types.md
```diff
@@ -0,0 +1,236 @@
...
+
+Static calls cannot change the fields of the class unless an object is being created and returned, like in the case of constructors.
```
SeanTAllen, line 88: This is confusing.  I think this only makes sense when talking about what "self" is but that isn't really covered in a way that will make sense when you get to this; more "self" background is needed before this "methods" section is my takeaway.

docs/internal/types.md
```diff
@@ -0,0 +1,236 @@
...
+
+### Constructors and Destructors
+
+Verona distinguishes memory allocation with object construction.
```
SeanTAllen, line 92:
```suggestion
Verona distinguishes memory allocation from object construction.
```

docs/internal/types.md
```diff
@@ -0,0 +1,236 @@
...
+There are no `constructors` per se, in Verona, only static methods that can allocate memory (using `new`), build an object and return it.
+
+Example:
+```ts
+class Capibara
+{
+  let name : String;
+  var weight : F32; // Kgs
```
SeanTAllen, line 100: I think you should make these private in the example.

docs/internal/types.md
```diff
@@ -0,0 +1,236 @@
...
+
+  poop(self: Capibara)
+  {
+    self.weight -= 0.1; // Tiny poop!
+  }
+};
+```
+
+On the example above, the `newborn` method creates a `Capibara` with a specific name and a default weight.
+The `name` can't change, so we declare it with `let`, while the `weight` does change, so it's a `var`.
```
SeanTAllen, line 124: This is misleading. unless the String for name is immutable, it can still be changed with this `let` binding. The `let` is about the object the variable is bound to, not the mutability or lack thereof,

docs/internal/types.md
```diff
@@ -0,0 +1,236 @@
...
+
+There's nothing special about the method `newbord` in comparison with others, apart from the fact that it allocates, initialises and returns an object of the same type.
```
SeanTAllen, line 126:
```suggestion
There's nothing special about the method `newborn` in comparison with others, apart from the fact that it allocates, initialises and returns an object of the same type.
```

docs/internal/types.md
```diff
@@ -0,0 +1,236 @@
...
+### Methods
+
+Methods are functions that belong to a class or interface. Like functions, they can have arguments and a return type.
+
+Methods can be called statically (via `Class::method()`) or dynamically (via `object.method()`).
+On the static case, the `self` pointer is a descriptor forthe type, while on the dynamic case, it's a descriptor to the instance (and its vtable, type info, etc).
+
+Static calls cannot change the fields of the class unless an object is being created and returned, like in the case of constructors.
+
+### Constructors and Destructors
```
SeanTAllen, line 90: I think you should break constructors and destructors apart into their own sections.

docs/internal/types.md
```diff
@@ -0,0 +1,236 @@
...
+
+There's nothing special about the method `newbord` in comparison with others, apart from the fact that it allocates, initialises and returns an object of the same type.
+
+Destructors are still under debate, but they won't have to deallocate memory like in C++, just execute finalization tasks like closing file handles, running some cleanup routines, checking final consistency, etc.
+
+## Primitive Types
+
+Verona treats numeric types specially in the compiler but their representaiton isn't special in the language.
+
+Verona's numeric types are singleton types (no fields) that have their machine representaiton based on their names.
```
SeanTAllen, line 134: For me this would raise the question "so is a class without fields a singleton?" "how do i create my own singletons?"

SeanTAllen, line 134:
```suggestion
Verona's numeric types are singleton types (no fields) that have their machine representation based on their names.
```

docs/internal/types.md
```diff
@@ -0,0 +1,236 @@
...
+So, `I32` is a signed integer with 32-bit width (in LLVM's parlour, an `i32`) and so on.
```
SeanTAllen, line 135: I feel like you intended to have some other content before this. I'm not sure based on the "So" how this is connected to the previous statement about singleton types.

docs/internal/types.md
```diff
@@ -0,0 +1,236 @@
...
+
+The full list of primitive numeric types are:
+ * Unsigned integers: `U8`, `U16`, `U32`, `U64` and `U128`
+ * Signed integers: `I8`, `I16`, `I32`, `I64` and `I128`
+ * Floating Point: `F32` and `F64`
+ * System types: `ULength` and `ILength` (machine-word sized integers)
+
+Note, integer literals have no defined type and their types (and representation) will be inferred by the compiler depending on the type of the variable being assigned to.
```
SeanTAllen, line 143: Im not sure what this means. An example would be good.

`let x = 1`

what's the inferred type?

docs/internal/types.md
```diff
@@ -0,0 +1,236 @@
...
+
+## Type Unions
+
+Type unions means a variable can be either type on the list.
```
SeanTAllen, line 147:
```suggestion
Type unions means a variable can be any type in the list.
```

docs/internal/types.md
```diff
@@ -0,0 +1,236 @@
...
+  // This is fine
+  foobar(a);
+  foobar(b);
+
+  // This is not
+  foobar(d); // ERROR
+}
+```
+
+The imlementation of `foobar` will have to cope with all three types, either via a `match` statement or taking advantage of some behaviour that is common to all three (say, an interface they all implement).
```
SeanTAllen, line 171:
```suggestion
The implementation of `foobar` will have to cope with all three types, either via a `match` statement or taking advantage of some behaviour that is common to all three (say, an interface they all implement).
```

SeanTAllen: The aside about `match` seems out of place here. Its a detail not related to the type system and covers additional material that hasn't been covered in this document (match).

docs/internal/types.md
```diff
@@ -0,0 +1,236 @@
...
+
+The compiler is free, however, to optimise the calls by creating two versions of `foobar` and replace the calls to the generic `foobar` for their specific ones.
+If the semantics is defined at compile time, this can simplifies a `match` statement into their specific matches and even inline the call altogether.
```
SeanTAllen, line 174:
```suggestion
If the semantics are defined at compile time, this can simplifies a `match` statement into their specific matches and even inline the call altogether.
```

docs/internal/types.md
```diff
@@ -0,0 +1,236 @@
...
+foobar(a : (A | B | C)) { ... }
+
+// The specific specialisations used in defined cases
+foobar(a : A) { ... }
+foobar(b : B) { ... }
+```
+
+## Type Intersections
+
+Intersection means a type must implement all guarantees of all types involved.
```
SeanTAllen, line 188:
```suggestion
Intersection means a type must all types involved.
```

docs/internal/types.md
```diff
@@ -0,0 +1,236 @@
...
+
+main()
+{
+  let x : (A & Serialisable & mut) = ...;
+}
+```
+
+In the above example, the type of `x` must conform to both `A` and `Serialisable` guarantees (methods, fields) as well as the `mutable` capability.
+
+Every type in Verona must have a capability, so:
```
SeanTAllen, line 202: earlier you said that a capability is a type. here, with this phrasing you are saying that a capability is a thing that a type has.

docs/internal/types.md
```diff
@@ -0,0 +1,236 @@
...
+Type alias can be used to make it easier to use types, for example:
+```ts
+class _String { ... };
+
+// We mostly want strings as values
+type String = (_String & imm);
+
+// But we may want to change it
+type MutableString = (_String & mut);
+```
```
SeanTAllen, line 218: This seems unrelated to explaining type intersections.

docs/internal/types.md
```diff
@@ -0,0 +1,236 @@
...
+
+At a glance, Verona types are:
+ * Interfaces: types that can contain fields, methods and can be derived by other interfaces and classes.
+ * Classes: like interfaces, but classes are `final`, ie. cannot be derived further.
+ * Capabilities: behaviour guarantees, ex. `iso`, `mut`, `imm` and `readonly`.
+ * Unions: a type that can be two or more types, ex. `(A | B)`.
+ * Intersections: a type that must implement all types, ex. `(A & imm)`.
+
+Capabilities offer language and runtime guarantees:
+ * `iso`, or `isolated`, means this is a pointer to an exclusive area of memory that only this pointer has access to in the entire program.
```
davidchisnall, line 15:
```suggestion
 * `iso`, or `isolated`, means this is a pointer to an the sentinel object in a region.
   Every region has a single sentinel that dominates all of the live objects in the region.
   Exactly one `iso` pointer exists for any region and the behaviour accessing any mutable object must hold the `iso` pointer for the sentinel object of the corresponding region.
```

docs/internal/types.md
```diff
@@ -0,0 +1,236 @@
...
+ * `mut`, or `mutable`, means this object can be changed at will, which means it must be inside an area of memory you have exclusive access to, ie. traced back to an `iso` you hold the rights.
```
davidchisnall, line 16:
```suggestion
 * `mut`, or `mutable`, means this object can be modified.
   All mutable objects are contained by a region, the behaviour modifying the object must own the `iso` pointer that refers to the sentinel object of the region.
```

docs/internal/types.md
```diff
@@ -0,0 +1,236 @@
...
+ * `imm`, or `immutable`, means no one can write to, so anyone can read from at any time. You can make `iso` regions immutable but not the other way round, as there's no guarantees who's still reading from it.
+ * `readonly` is a local capability that is used when you want to make sure an external function can't modify, but you still hold a mutable reference.
```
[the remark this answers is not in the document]
davidchisnall:
```suggestion
  * `readonly` defines local immutability.
    No object reached via a `readonly` capability may be modified via that the object is not guaranteed to be immutable.
    `mut` and `readonly` capabilities to the same object may be visible at the same time.
```

docs/internal/types.md
```diff
@@ -0,0 +1,236 @@
...
+
+Note that type capabilities, unions and intersections are types themselves.
+So `(A & imm)` is completely different from `(A & mut)` or `(A & iso)`.
+Specifically, you cannot convert `mut` to `imm` (as you would with a `const` cast in C++), or vice versa.
+You can freeze an `iso` region into `imm`, but unfreeze an `imm` region.
```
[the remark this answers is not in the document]
davidchisnall:
```suggestion
A `freeze` operation consumes an `iso`, destroying the only copy of the capability that is required to mutate any object in a region, and returns an `imm` capability.
There is no corresponding `unfreze` operation to consume an `imm` capability and provide any kind of mutable capability.
```

docs/internal/types.md
```diff
@@ -0,0 +1,236 @@
...
+Most importantly, capabilities are not type traits like C++'s `const int * const`, they have specific non-trivial runtime semantics.
+
+The type system guarantees that:
+ * Structural soundness: Interfaces and classes form valid concrete types, operations and functions take in the correct (sometimes inferred) types.
+ * Capability guarantees are followed: The compiler checks operations on types, the runtime guarantees the semantics is followed.
+ * Runtime safety: A program that compiles is a program that runs safely, without crashes, traps or unhandled exceptions, including asynchronous behaviour, but not necessarily producing the right result or doing what you expect.
```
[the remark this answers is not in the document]
davidchisnall: I don't think the third point is true, unless something has changed.  We were planning on providing contracts that had fail-fast behaviour so that array access bounds checks were safe to hoist out of loops (permitting vectorisation) and we did not need to make array accesses return a union type.

docs/internal/types.md
```diff
@@ -0,0 +1,236 @@
...
+## Classes and Interfaces
+
+Classes and interfaces have a similar structure:
+ * Fields: elements of a specific type that have a runtime space allocated.
+ * Methods: functions that belong to a type.
+
+The declaration is trivial:
+```ts
+// Helps receivers know they can convert the object to a string representation
+interface Serialisable
```
davidchisnall, line 40: Calling this `Serialisable` is very confusing.

docs/internal/types.md
```diff
@@ -0,0 +1,236 @@
...
+{
+  // All derived classes will have the common field
+  var common : F64;
+
+  // Derived classes must implement this method
```
davidchisnall, line 45: The 'derived' terminology is confusing here.  Classes that implement this interface must implement this method and must provide the field.  Classes that inherit from this interface will get the field 'for free'.  Classes that inherit from this interface will need to implement `toString` if they wish to implement this interface (and may get a compile error if they don't).

docs/internal/types.md
davidchisnall: I believe the closest we came to consensus was that we have two kinds of visibility:

 - Public (anything can read / write the field or call the method)
 - Package (anything within the same package can read / write the field or call the method).

Protected makes little sense when subclassing is distinct from subtyping and private is not very important when you have a sensible module system, because it can be enforced by convention (and Package avoids the need for `friend` and similar).

docs/internal/types.md
davidchisnall: I don't believe we have free functions, because every module is a class and so every function is a method on some class.

docs/internal/types.md
davidchisnall: I agree, I think this section needs a bit of overview of the class / module model before it talks about methods.

docs/internal/types.md
davidchisnall: This is also somewhat tautological: Calls that do not take a self parameter have no object reference and so cannot modify fields of an object that they do not hold a reference to.

docs/internal/types.md
```diff
@@ -0,0 +1,236 @@
...
+
+Methods can be called statically (via `Class::method()`) or dynamically (via `object.method()`).
+On the static case, the `self` pointer is a descriptor forthe type, while on the dynamic case, it's a descriptor to the instance (and its vtable, type info, etc).
+
+Static calls cannot change the fields of the class unless an object is being created and returned, like in the case of constructors.
+
+### Constructors and Destructors
+
+Verona distinguishes memory allocation with object construction.
+There are no `constructors` per se, in Verona, only static methods that can allocate memory (using `new`), build an object and return it.
```
davidchisnall, line 93: I would phrase that as something more like:
```suggestion
Verona's built-in `new` function fully constructs an object and so must be called with values for all of the fields in the object.
It may be used only from static methods on the class and so is usually used by factory methods.
By convention, factory methods are called `create()` and there is syntactic sugar to allow `T()` to implicitly call `T::create()`.
```

docs/internal/types.md
```diff
@@ -0,0 +1,236 @@
...
+  let name : String;
+  var weight : F32; // Kgs
+
+  // Create a baby capibara
+  newborn(name : String) : Capibara
+  {
+    var self : Capibara = new Capibara;
+    self.name = name;
+    self.weight = 3; // Tiny baby!
+    self; // Returns `self`
```
davidchisnall, line 108:
```suggestion
    var self : Capibara = new Capibara(name, 3);
    self; // Returns `self`
```

docs/internal/types.md
```diff
@@ -0,0 +1,236 @@
...
+### Constructors and Destructors
+
+Verona distinguishes memory allocation with object construction.
+There are no `constructors` per se, in Verona, only static methods that can allocate memory (using `new`), build an object and return it.
+
+Example:
+```ts
+class Capibara
+{
+  let name : String;
```
davidchisnall, line 99:
```suggestion
  let name : String & imm;
```

docs/internal/types.md
```diff
@@ -0,0 +1,236 @@
...
+On the example above, the `newborn` method creates a `Capibara` with a specific name and a default weight.
+The `name` can't change, so we declare it with `let`, while the `weight` does change, so it's a `var`.
+
+There's nothing special about the method `newbord` in comparison with others, apart from the fact that it allocates, initialises and returns an object of the same type.
+
+Destructors are still under debate, but they won't have to deallocate memory like in C++, just execute finalization tasks like closing file handles, running some cleanup routines, checking final consistency, etc.
+
+## Primitive Types
+
+Verona treats numeric types specially in the compiler but their representaiton isn't special in the language.
```
davidchisnall, line 132: Typo: representaiton

docs/internal/types.md
davidchisnall: s/parlour/parlance/

Though I do like the idea of LLVM's parlour...

davidchisnall (commented):
I've mostly skipped the bits Sean commented on  in the second half - I'll do a more detailed read once his comments are resolved.

docs/internal/types.md
rengolin: IIUC, the idea is to make `let` similar to a `const` field that gets set on construction and cannot change, while `var` is your average field. But I may have expressed it wrong or worse, understood wrong.

docs/internal/types.md
rengolin: @SeanTAllen feel free to suggest better wording as a patch.

@davidchisnall I thought that, too, but then when talking about names this week, the `_` naming was mentioned to allow "private variables". This was a paragraph that was meant to raise discussion.

I'd be happy with no `_` rules and everything is package-private. But I'd also be happy with class-private via `_` in addition to package-private.

docs/internal/types.md
rengolin: On purpose, I'm still confused as to what `embed` will actually mean in the language.

docs/internal/types.md
rengolin: @davidchisnall correct, but then saying "like functions" means even less. What I meant is "like functions in other languages". I'm still confused with the class-private and if methods have access to them, or if there's no real difference between methods in a class and another outside that happens to take a "self" of that class.

docs/internal/types.md
rengolin: right, this perhaps should be in a different document, then.

docs/internal/types.md
rengolin: Happy to remove all...

docs/internal/types.md
rengolin: Ah, yes, David has corrected with `& imm`, thanks!

docs/internal/types.md
rengolin: I *always* do that typo "ation" -> "aiton" because of how "ai" in PT reads like "a" in GB. :(

docs/internal/types.md
rengolin: Good point. `let x = 1` is syntax error, but `let x : I32 = 1` and `let x : U64 = 1` infer different types for `1`.

docs/internal/types.md
rengolin: Right, it did feel out of place. But if similar parts about the behaviour (dynamic/static calls, etc) will be removed to their own section, we can move this one too.

docs/internal/types.md
rengolin: Your suggestion here seems to be missing something...

docs/internal/types.md
rengolin: Hm, vero. I'm not sure how to phrase this, but every type must be the intersection of a type and a capability, but there shouldn't be two capabilities, but capability is a type, and intersections between types are allowed. It's really confusing to me.

docs/internal/types.md
rengolin: They're not _explaining_ intersections, they're just demonstrating how you can use aliases to not have to declare capabilities on types that are commonly associated with one capability and not others.

docs/internal/types.md
```diff
@@ -0,0 +1,236 @@
...
+type String = (_String & imm);
+
+// But we may want to change it
+type MutableString = (_String & mut);
+```
+
+## Tuples
+
+Tuples allow unnamed types to encompass multiple objects of different types.
+Unlike classes, tuples don't have fields (or has unnamed fields), so you can't access them by name.
```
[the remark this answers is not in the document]
rengolin: A circle has no sides. But it can also have infinite sides. Same idea, different ways of seeing it.

A field is a named slot in the object representation that you can access by name. Tuples have slots but doesn't name them. 

Is that an unnamed field? Or no fields? I have no idea. Terminology is not my forte, I'm happy to accept whatever people are comfortable with.

docs/internal/types.md
```diff
@@ -0,0 +1,236 @@
...
+Here's a tuple example:
+```ts
+// Declare a tuple type
+var x : (String, U64);
+
+// Tuple literals
+x = ("hi", 42);
+
+// Access elements
+(var s, var i) = x;
```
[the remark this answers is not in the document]
rengolin: If x's types are `mut`, then I thought `s` and `i` are references to the internal elements. If they're `imm` or a primitive, then it's a copy. If they're `iso`, then it's a move?

This point wasn't well discussed, but we also talked about using Pony's `x._1` as accessor if nothing else works.

rengolin (commented):
Thank you both for the extensive reviews. I appreciate I'm not a writer and I am still missing a lot of context from the language semantics (and some of the syntax, too).

I have committed most of the suggestions and I'm still hoping some of the comments can turn to suggestions, as that'll reduce the back and forth of me trying to write something meaningful and failing, then repeating.

docs/internal/types.md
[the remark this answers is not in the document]
davidchisnall: We have three choices for arrays:

 1. Every array access must come with some proof that it's in-bounds.  For example, if we have index set types then we can check that every index comes from an index set that is derived from the array and has not been unioned with anything else (and the bounds haven't been changed without updating the index set), accesses are guaranteed to be in bounds.  This requires a lot of type system plumbing that we don't have and I've never seen a non-toy language implement this so I've no idea how usable it actually is.
 2. Every array access returns an option type.  For example, if I have `arr : Array[Foo]` then `arr(42)` is `Foo | ThingNotFound`.  This is pretty awful for the programmer because every array access has to be followed by a match expression to check if the access is in-bounds, even if the compiler can elide it.
 3. We provide contracts and say that the contract on `Array::apply(index : USize)` is that `index` is in bounds, if it's not then we have a contract violation.  This is nice for the compiler (contract violations may fast fail, if you have a loop accessing values in an array you can hoist the bounds check out of the loop), it's slightly bad for the system because any code has a capability to crash the program unless we can constrain contract violation to killing something smaller than a program.

In the absence of good choices, I thought we had decided 3 was the least bad.

docs/internal/types.md
davidchisnall: `Example`?

docs/internal/types.md
davidchisnall: Both `var` and `let` fields must be initialised at creation time.  We require complete object initialisation because we don't want every field to be an option type with a trivial default and we don't want UB from partial initialisation.

This is a big part of the motivation from splitting `new` (allocate and initialise) from public interfaces for construction.  You want to allow the class to run some code that calculates some default values for things so that you don't need the code that wants a new `Foo` to provide every field for `Foo`, but you *also* don't want to enable partially-constructed `Foo` instances to exist.

docs/internal/types.md
davidchisnall: I was being lazy in the discussion and saying private when I meant package, because I fundamentally don't believe sub-package private actually exists.

docs/internal/types.md
davidchisnall: Syntactically, I believe `embed` is used in place of `let` or `var`, though I'm not sure what this means with respect to capabilities (`embed thing : T & imm` probably doesn't make sense because you can't share a pointer to `thing` between `cown`s, can you `embed thing : T & readonly`?), but assuming that's sorted we will be able to do the equivalent of `var` vs `let` with capabilities on the `embed` field.

Semantically, it just means put the fields of the object here.

docs/internal/types.md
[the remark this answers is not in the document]
davidchisnall: I don't remember if we actually came up with a way of accessing tuple elements (@sylvanc?).  You can't have an accessor function because the type of the function depends on the value.  If you wanted to use the `apply` sugar then `x(n)` would have to have type `String | U64` (or probably `String | U64 | NotAThing`), so you'd then need to match on the result (which is annoying because for concrete values of `n` you know the type.  C++ uses `get<int>` to force the index to be a compile-time constant, which is quite ugly (and I'm not sure if Verona has enough support in generics to allow this).  As I recall, the Pony-inspired suggestion was to give names to the fields such as `_1`, `_2`, (which is how we got onto the discussion of `_` as a visibility specifier, so maybe `#1`, `#2`, and so on, possibly starting from `#0`).  

If any of the fields are `iso` then decomposition must be a destructive read that destroys the *entire* tuple because otherwise you have an `iso` field with no value, which would introduce UB.  It think it's really confusing is decomposition is sometimes destructive and sometimes not, so I'd be inclined to say that it's always destructive.

rengolin:
I'm trying to apply the changes @davidchisnall proposed but Github is saying "applying multiple changes to the same line is not supported" which I have no idea what it means. I'll probably have to merge that by hand... :(

davidchisnall:
> I'm trying to apply the changes @davidchisnall proposed but Github is saying "applying multiple changes to the same line is not supported" which I have no idea what it means. I'll probably have to merge that by hand... :(

This happens if you put two changes in the same line (which, I think, means line-of-diff, not line-of-source, so can apply to adjacent lines) into the same commit.  It normally works if you apply each change sequentially.  Your git history gets more ugly then (and it sends a few more emails), but you'll probably be tidying it up before the final merge, so it's not a big problem.

docs/internal/types.md
davidchisnall:
```vim
:set spell
:hi SpellBad ctermfg=Red ctermbg=black cterm=undercurl
:set spelllang=en
```
:-)

docs/internal/types.md
davidchisnall: I think we're quite fuzzy about what we mean when we say 'capability'.  To me, the things that we're calling capabilities are actually permissions.  A capability is an unforgeable, delegable, token of authority that grants rights to perform some action.  In a capability system you must present present a capability with the requisite rights when you attempt to perform an action.  A Verona pointer *is* a capability.  You cannot create one from nothing.  Owning a pointer allows you to do things to the object that it refers to.  Specific permissions on the pointer confer the ability to do specific things (e.g. `mut` or `iso` permit mutation, `iso` or `imm` permit sending to a different behaviour, `imm` permits sharing with other behaviours).

docs/internal/types.md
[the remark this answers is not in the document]
davidchisnall: I *think*, they do have names, but the names are assigned by the compiler and not by the programmer.

docs/internal/types.md
```diff
@@ -0,0 +1,236 @@
...
+ * Unions: a type that can be two or more types, ex. `(A | B)`.
+ * Intersections: a type that must implement all types, ex. `(A & imm)`.
+
+Capabilities offer language and runtime guarantees:
+ * `iso`, or `isolated`, means this is a pointer to an exclusive area of memory that only this pointer has access to in the entire program.
+ * `mut`, or `mutable`, means this object can be changed at will, which means it must be inside an area of memory you have exclusive access to, ie. traced back to an `iso` you hold the rights.
+ * `imm`, or `immutable`, means no one can write to, so anyone can read from at any time. You can make `iso` regions immutable but not the other way round, as there's no guarantees who's still reading from it.
+ * `readonly` is a local capability that is used when you want to make sure an external function can't modify, but you still hold a mutable reference.
+
+Note that type capabilities, unions and intersections are types themselves.
```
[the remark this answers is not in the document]
rengolin: I think I know what I wanted to say here: capabilities are not annotations that can be dropped easily with casts. But this is true of other types. My point was that in other languages you can abuse of dynamic casts or reflection to get around type system restrictions, but Verona won't allow that. 

The rest of the paragraph is the main point, not this single phrase.

docs/internal/types.md
rengolin: I don't like spell check on vim because it marks stupid things and I need to add a lot of stuff to the dictionary or the document looks like it's haemorrhaging. I usually run aspell from the terminal, but this time I forgot. :(

docs/internal/types.md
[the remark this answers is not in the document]
rengolin: Well, that's an implementation detail, not a language feature. If the programmer can't access those "names", then they don't exist from their point of view.

docs/internal/types.md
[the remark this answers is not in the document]
rengolin: Do you mean decomposition accessing all fields (like `(a, b) = c`) or also just one (like `a = c#1;`)?

If all, then it's ok for it to be destructive, but if it's also some, then what happens to the rest of the elements?

And if index access isn't destructive, then what happens if the field is an `iso`?

docs/internal/types.md
```diff
@@ -0,0 +1,246 @@
...
+ * `imm`, or `immutable`, means no one can write to, so anyone can read from at any time. You can make `iso` regions immutable but not the other way round, as there's no guarantees who's still reading from it.
+  * `readonly` defines local immutability.
+    No object reached via a `readonly` capability may be modified via that the object is not guaranteed to be immutable.
+    `mut` and `readonly` capabilities to the same object may be visible at the same time.
+
+Note that type capabilities, unions and intersections are types themselves.
+So `(A & imm)` is completely different from `(A & mut)` or `(A & iso)`.
+Specifically, you cannot convert `mut` to `imm` (as you would with a `const` cast in C++), or vice versa.
+A `freeze` operation consumes an `iso`, destroying the only copy of the capability that is required to mutate any object in a region, and returns an `imm` capability.
+There is no corresponding `unfreze` operation to consume an `imm` capability and provide any kind of mutable capability.
```
davidchisnall, line 33:
```suggestion
There is no corresponding `unfreeze` operation to consume an `imm` capability and provide any kind of mutable capability.
```

docs/internal/types.md
```diff
@@ -0,0 +1,246 @@
...
+Most importantly, capabilities are not type traits like C++'s `const int * const`, they have specific non-trivial runtime semantics.
```
davidchisnall, line 34: Possibly 'may have'?  

I'm possibly confused by 'run-time' vs 'runtime' here:  'run-time' is an adjective phrase and modifies semantics, so 'run-time semantics' refers to semantics that apply at the time the program runs.  'Runtime' is a noun that refers to the library that provides services to the program at run time, so 'runtime semantics', to me, means operations that the runtime provides.

Casting from `iso` to `mut` or `mut` to `readonly`, for example, change the set of operations that are permitted with the resulting capability (and are monotonic - you can't cast back) but (probably) don't involve the runtime doing anything.

docs/internal/types.md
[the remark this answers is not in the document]
davidchisnall: `a = c.#1` would have the same semantics for a tuple as `a = obj.field` would for any field.  I don't remember where we landed on `=` but this may give you a `mut` view of the `iso`.

My worry is that structured binding is used for multiple return values.  If I do:

```
example() : (mut & T, iso & X);
...
(var r1, var r2) = example();
```

I want `r2` to give me an `X & iso`.  I would be very confused if it gave me an `N & mut` and no way for me to recover the `iso` (presumably the region would be deallocate at the end of the stack frame and I wouldn't be allowed to store `r2` anywhere on the heap).  This means that I want it to extract the `iso` destructively.  

If the `iso` is extracted destructively, then the field is now invalid and I may not access it again.  I can't destroy a field of an object without destroying the object, so reading an `iso` out via structured binding would destroy the object.

Now imagine I have code like this:

```verona
var tuple : (T & mut, X & imm) = ...;
...

(var a, var b) = tuple;

...

foo = tuple.#1;
```

This is all fine if it's only reading an `iso` that makes this a destructive operation.  But now if I add an `iso` field to tuple, then my later `tuple.#1` becomes invalid *even though the field of the tuple that it references has not changed*.   

If we say that structured binding always destroys the object that it refers to then the above code is consistently disallowed.  I can see a few other valid options here.  For example:

 - Structured binding is not allowed on tuples with `iso` fields (I don't like this one because it eliminates a load of valuable uses for this, but it is a simple rule).
 - Structured binding on `iso` tuples destructively reads all fields, structured binding on `mut` and `imm` tuples does not, structured binding on non-`iso` tuples containing `iso` fields is not permitted.
 - Structured binding gives references to the fields, not copies of their values (I don't think we have a notion of reference-to-field anywhere else in the language though.  You might be able to fudge this with `embed` by making every tuple element an `embed` object with a single field, so you're actually taking a pointer to the object that contains one field but that feels very clunky).

The second of these would cause an error in the above example on the structured binding line, but that would already need modifying because the tuple structure changed.  This could then be fixed by making the tuple `iso`, which causes breakage in the same place but now it's a result of changing the type of the capability of `tuple`, not the type of a field.

docs/internal/types.md
rengolin: This is what capabilities meant to me, too. But in the type system, capabilities are also types that you can use with type inference, for example:
```
foo (a : imm) { ... }
```
Obviously, the type of `a` isn't just `imm` but this requires whatever type is passed to be viewed as having the `imm` capability. If `imm` wasn't a type, that statement would be invalid because `a` wouldn't have a type.

Also, if capabilities are types and we have type unions and intersections, then we can obviously use capabilities there, and then we naturally arrive at `Type & cap` that is required for all pointers.

I imagine capabilities can't be mixed in like other types (ex. `imm` & `mut`) not because they're special, but because their semantics (like other types can have too) don't mix. If you try to mix two types that have no intersection like `A & B`, then the resulting type is void.

Perhaps that's what it means to intersect capabilities?

But also, capabilities have intrinsic semantics that no other types have, so in a way, capabilities are special types. I don't know if this is a semantic difference (in the definition of the type system) or an implementation difference (like how numeric types are different than other singleton types).

docs/internal/types.md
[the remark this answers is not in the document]
rengolin: rewriting to simplify

docs/internal/types.md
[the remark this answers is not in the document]
rengolin: I think this is an interesting discussion but I don't want to write hypothetical stuff here. Once we define what accessing tuple elements do we can update the document.

rengolin:
Updated the document and ran spell check. :)

There are three pending issues, but otherwise I think I've applied all suggestions.

Please read again to make sure I have expressed as you intended.

docs/internal/types.md
```diff
@@ -0,0 +1,287 @@
...
+ * `let`: Must be initialised on construction, and cannot change later.
+
+They can also have a keyword `embed`, which means ...
+
+### Methods
+
+Methods are functions that belong to a class or interface which can have arguments and a return type.
+
+Methods can be called statically (via `Class::method()`) or dynamically (via `object.method()`).
+On the static case, the `self` pointer is a descriptor for the type, while on the dynamic case, it's a descriptor to the instance (and its vtable, type info, etc).
```
davidchisnall, line 124: I think this is conflating a few things.  If I have a class like this:

```
class Cls 
{
  method() {}
}
```
The following are both valid and equivalent:

```
Cls::method();
var obj : Cls = ...;
obj.method();
```

This is just about lookup.  Any method that does not take a `Self` parameter can be invoked either by specifying the class or by specifying an object of that class.  In both cases, this is static dispatch: the compiler knows that the method that is being invoked is `Cls::method` and so will generate a direct call to that function (which will probably be inlined).

This gets more interesting with this case:

```
interface Methodic
{
  method();
};
```

Now, `Cls` implements the interface `Methodic`.  I can *also* write this:

```
var ifc : Methodic = ...;
ifc.method();
```

If the concrete type of the current value of `ifc` is an instance of `Cls` then this will call the same method as above, but now we're doing it via dynamic dispatch. Without optimisation, this will generate a vtable lookup in the vtable for the object using the selector for `method()` and will then insert the call.  With optimisation, we may discover that (especially after some inlining) we can guarantee the concrete type of `ifc` and so insert the correct call.  In the trivial case, if `Cls` is the only class that implements `method()` then we can infer this without doing any clever dataflow things.

This is almost completely orthogonal to whether there's a `Self` parameter.  If we extend `Cls` to look like this:

```
class Cls 
{
  method() {}
  instance_method(self : Self & mut) {}
}
```

Now we can invoke this as either:

```
var obj : Cls = ...;
Cls::instance_method(obj);
obj.instance_method();
```

In both cases, we need to specify the receiver (the `Self` parameter, in either procedural or an OO syntactic style).  Again, both of these will call `Cls::instance_method` directly, *without* any dynamic dispatch.

If we extend `Methodic` to also include this instance method:

```
interface Methodic
{
  method();
  instance_method(self : Self);
};
```

Now we are able to call the instance method dynamically:

```
var ifc : Methodic = ...;
ifc.instance_method();
```

This will use the selector that corresponds to `instance_method(Self)`  to look up the method in the vtable of whatever object `ifc` is.  That method will then be invoked with `ifc` as the first argument.

docs/internal/types.md
```diff
@@ -0,0 +1,287 @@
...
+
+### Constructors
+
+Verona distinguishes memory allocation from object construction.
```
davidchisnall, line 128: I don't think this is true.  Objects are allocated fully constructed.  `new` allocates memory and fully constructs the object.

docs/internal/types.md
```diff
@@ -0,0 +1,287 @@
...
+
+Example:
+```ts
+class Capibara
+{
+  let name : String & imm;
+  var weight : F32; // Kgs
+
+  // Create a baby capibara
+  newborn(name : String) : Capibara
```
davidchisnall, line 141: Maybe rename this `create`?  It seems odd to say 'by convention these things are called `create`' and then not call this one `create`.  I'm not 100% sure on this, because I quite like the fact that it shows that factory methods don't *have* to be named `create`.

docs/internal/types.md
```diff
@@ -0,0 +1,287 @@
...
+Verona treats numeric types specially in the compiler but their representation isn't special in the language.
+
+Verona's numeric types are singleton types (no fields) that have their machine representation based on their names.
+So, `I32` is a signed integer with 32-bit width (in LLVM's parlance, an `i32`) and so on.
+
+The full list of primitive numeric types are:
+ * Unsigned integers: `U8`, `U16`, `U32`, `U64` and `U128`
+ * Signed integers: `I8`, `I16`, `I32`, `I64` and `I128`
+ * Floating Point: `F32` and `F64`
+ * System types: `ULength` and `ILength` (machine-word sized integers)
```
davidchisnall, line 179:
```suggestion
 * System types: `ULength` and `ILength` (size large enough to describe any offset into the largest array that it's possible to allocate)
```

I find 'machine word' confusing since Intel decided that word is 16 bits, a double-word is 32 and a quad word is 64.  Everyone seems to have a different definition of what a word means these days (when clearly it should be an arbitrary-length sequence of 6-bit EBCDIC characters).

docs/internal/types.md
```diff
@@ -0,0 +1,287 @@
...
+
+Type aliases can be used to make it easier to use types, for example:
+```ts
+class _String { ... };
+
+// We mostly want strings as values
+type String = (_String & imm);
+
+// But we may want to change it
+type MutableString = (_String & mut);
```
davidchisnall, line 269: Hmm, I'm not sure what this means.  Is `String` a public thing, even though `_String` is package-private?  What happens if you have a public type alias of a private type?

docs/internal/types.md
rengolin: You're right, I'm talking about `new` being the operator that constructs the object, but constructor methods having the startup logic.

I guess this is part of the same discussion about the finalisers. So, perhaps what I should say is that verona distinguishes construction (`new`) from initialisation (static methods that return objects of the class' type).

docs/internal/types.md
rengolin: Yeah, this is why I wanted it to not be `create`. I can add another method called `create` with no arguments for completeness.

docs/internal/types.md
rengolin: Removed the phrase as it wasn't necessary to understand the rest.

docs/internal/types.md
rengolin: I haven't touched public/private in this document, perhaps we should. Are public fields those prepended with `public`? Like `public class Foo {}` or `public type Foo = Bar`? I'm still unsure how we can expose some fields outside the module.

For example:
```
module 1 {
  class Private {
    public var foo; // is this legal?
  };
  public class Public {
    private var bar; // or this?
  };
```

If only types can public, then one way to expose would be to create a facade:
```
module 1 {
  public interface Public {
    create() : Public {
      var self = Private::create();
      return self;
    }
  };
  class Private : Public { ... };
}
```

but this is convoluted. It will depend on how we control visibility: type level, field/method level, etc.

davidchisnall: > I haven't touched public/private in this document, perhaps we should. Are public fields those prepended with public? Like public class Foo {} or public type Foo = Bar? I'm still unsure how we can expose some fields outside the module.

Either with `public` / `private` or with a naming convention (`_`-prefix = private to the package, anything else is public).  If a thing is public, it can be named outside the current package.  I think there are still a few corner cases though (more below).

> For example:

All of this is legal.

> If only types can public

Types, fields, and methods, can be public or (package) private.

```
module m1 // fake syntax, modules are classes in different places
{
  class Foo
  {
     var x : U64;
     var _privateField : U64;
     method(self : Foo) {...}
     _privateMethod(self :Foo) {...}
  }
}
// In another module, having imported m1 as m1
var f : Foo = ...; // This is fine
f.x  =12; //fine.
f._privateField = 12; //Nope
f.method(); //fine
f._privateMethod(); //Nope
```

This gets more interesting with interfaces:

```
module m1 // same one as above
{
  interface I1
  {
     var x : U64;
     var _privateField : U64;
  }
  getAnI1() : I1
  {
     var ret : Foo = ...;
     return ret;
  }
}
module m2
{
  // Assume m1 imported as m1
  var i1 = m1.getAnI1();
  i1.x = 42; // Yup
  i1._privateField = 12; // Maybe?  
  // Okay, what about this (not in m1)?
  interface I2
  {
     var _privateField : U64;
  }
  // Is this allowed?
  var i1 : I2 = i1;
}
```

I like a ClassBox-like approach, where all private selectors and type names are logically mangled with your local module ID, so `_privateField` as named in `m1` and in the package containing `I2` are different selectors and the last cast is not permitted.  It's less clear to me what this means for the `i1._privateField` access, it might simply mean that this access is statically not permitted because this is logically  the same as saying `i1._privateField@m2` and `I1` defines a field called `_privateField@m1`, not a field called `_privateField@m2` and so you may not access this.

There's one final case:

```
module m3
{
  interface I4 {...};
  class _Cls { ... } // Implements I4
  type Cls = _Cls;
  getI4() : I4 { /* return a _Cls instance */ }
}
//Elsewhere:
var i4=  m3::getI4();
match i4 { cls : m4::_Cls => ...} // Nope, you may not name m4::_Cls from here.
match i4 { cls : m4::Cls => ...} // Maybe?  Did the type alias make Cls public?
```

docs/internal/types.md
rengolin: Right, but I still have to explain where do we put the field when it's not an `embed`. ie. where is `here`? Is this like `static` fields in C++?

docs/internal/types.md
rengolin: Right, there are a few issues with that phrase. I didn't mean having a `self ` or not was the difference, only what descriptor was being used for. But I get that it was misleading (and also probably wrong).

IIUC, you can only modify the object is there is a self, statically or dynamically called, because otherwise, you have no base pointer on what to change.

What confuses me is where does `self` go.

```
class Foo {
  var field ...;
  create() { ... }
  method1(self : Self & mut) {
    self.field = ...;
  }
  method2() {
    // cannot touch field
    // does this mean I cannot call this as x.method2()?
  }
};
...
var x = Foo;
x.method1(); // -> method1(x);
x.method2(); // -> method2(x)??
```

rengolin: I have rewritten and  tried to give two examples. Hopefully this is clearer. 

I don't want to include the whole discussion. Although very pertinent, the document should only have what we agreed, not what we discussed.

docs/internal/types.md
rengolin: I think this deserves more discussions, but I've added a new line to the example to mean this is only about the same module, so there are no visibility concerns.

We should address visibility in another section or perhaps another document.

docs/internal/types.md
davidchisnall: All fields are stored in the object (unless they're class fields rather than instance fields [static fields, in C++ / Java terminology] - I don't know if we have syntax for that yet?

The difference between `let` / `var` and `embed` is the difference between a pointer and a value.  Here's some C++:

```c++
struct Point
{
  double x, y;
};

struct Example
{
  // let l : Point & mut
  const Point *l;
  // let li : Point & imm
  const Point * const li; // I think - C / C++ `const` placement confuses me
  // var v : Point
  Point *v;
  // embed e : Point
  Point e;
};
```

docs/internal/types.md
davidchisnall: I'd prefer that you didn't name the concrete type with an underscore prefix (or you named everything with an underscore prefix) because then there's no ambiguity.

docs/internal/types.md
rengolin: Ah! So much easier to understand when you put it into code. :)

So, this means that the structure layout of `Example` would be something like:
```
{ USize, USize, USize, F64, F64 }
```

And the compiler is in charge of encoding access (direct or indirect), but also making sure the field indexes are packed, aligned and available as an interface type that they implement (ie. same offsets?), according to the semantics of the language (and their modifiers).

I'm not sure your `const` comparison has the same semantics, so I won't include that in the document to avoid misleading readers. But I've added the description and an example in Verona.

docs/internal/types.md
rengolin: Fair enough, renamed.

docs/internal/types.md
```diff
@@ -0,0 +1,236 @@
+# Verona Type System
+
+Verona is a statically typed language.
+This means all types are known at compile time and the compiler can assert that the objects and operations have the correct type at all times.
```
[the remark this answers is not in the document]
rengolin: I think I managed to explain statically typed by stealing some of your text above. Let me know if that makes sense.

docs/internal/types.md
```diff
@@ -0,0 +1,359 @@
...
+Some types are inferred by the compiler, but there is no guessing: there's either one clear inferred type or it's an error.
+
+At a glance, Verona types are:
+ * Interfaces: abstract types that specify a set of fields and methods that an underlying concrete type must implement.
+   Interfaces may provide default implementations of methods that a class can adopt.
+ * Classes: concrete types.
+   A class may inherit from other classes but this does not make the class a subtype of the class from which it inherits.
+   Classes may explicitly adopt interfaces, in which case they adopt the default implementations from the interface, but conformance to an interface is entirely a property of the structural type.
+ * Capability rights: A Verona pointer is a *capability*: an unforgeable token of authority that permits access to an object. 
+   The set of rights that a capability holds are associated with a type, for example. `iso`, `mut`, `imm` and `readonly`.
```
nwf, line 36: This bullet seems a little out of nowhere.  It might not hurt to make line 11 read "a structural and algebraic type system which, additionally, tracks access rights" and add an early gloss paragraph paralleling lines 19-24.  Something like...

```
Verona programs are, by construction, devoid of data races.
As part of ensuring that property, references to objects in Verona carry a 'capability' annotation, which conveys access rights to that object and global properties of the program.
For example, a reference may be marked as `iso`-lated: the reference grants full read and write access and assurance that no concurrent accesses are possible.
Or a reference may be `imm`-utable: only reads are possible and the system guarantees that all live references to this object are similarly read-only.
```

docs/internal/types.md
```diff
@@ -0,0 +1,359 @@
...
+ * Unions: The concrete type of the object must fulfil all of the constraints of at least one out of two or more type expressions, for example the union type `(A | B)` must fulfil all of the requirements of either `A` or `B`.
+   Unions may refer to concrete types (for example, if `A` and `B` are classes then a variable of type `A | B` may be a reference to an instance of either class) or other type expressions, in which case the variable may refer to any concrete type that meets the requirements of either set of constraints.
+ * Intersections: The concrete type of the object must fulfil all of the constraints of two or more type expressions, for example the union type `(A & B)` must fulfil all of the requirements of both `A` and `B`, which may be interfaces or other type expressions.
+   No type satisfies the union type of two concrete types, so if `A` and `B` are both class types then `A` & `B` is a type that is not possible to satisfy.
```
nwf, line 40:
```suggestion
   No type satisfies the union type of two distinct concrete types, so if `A` and `B` are both class types then `A` & `B` is a type that is not possible to satisfy.
```

docs/internal/types.md
```diff
@@ -0,0 +1,359 @@
...
+
+Capabilities offer language and runtime guarantees:
```
nwf, line 42: Before diving into capabilities, it might be good to have a brief introduction to regions.

```suggestion
Verona manages memory by associating objects with 'regions'.
Each region has a single 'sentinel' object; all live objects in a region are reachable from its sentinel without crossing into another region.
To ensure data-race freedom, Verona limits programs to non-concurrent access to all objects of each region by carefully tracking the flow of references to the region's sentinel.

Capability types are the mechanism Verona uses to offer such language and runtime guarantees:
```

docs/internal/types.md
```diff
@@ -0,0 +1,359 @@
...
+The declaration is trivial:
+```ts
+// Helps receivers know they can call a function named increment
+interface Example
+{
+  // All derived classes will have the common field
+  var common : I64;
+
+  // Classes that implement this interface must implement this method and provide the field
+  // Classes that inherit from this interface must implement this method but get the field from the base class
```
nwf, line 79:
```suggestion
  // Classes that inherit from this interface must implement this method but get the field from the definition above
```

docs/internal/types.md
```diff
@@ -0,0 +1,359 @@
...
+  }
+
+  // Implements `increment`
+  increment()
+  {
+    self.common += 1;
+  }
+};
+
+// This class implements the interface, so it must provide the `common` field
```
nwf, line 102:
```suggestion
// This class aims to implement the interface without explicit inheritance, so it must provide the `common` field
```

Do we have a syntax to declare the _intent to ascribe_ without requesting field and method inheritance?

docs/internal/types.md
```diff
@@ -0,0 +1,359 @@
...
+For example
+```ts
+// This struct layout would be { F64, F64 }
+class Point
+{
+  embed var x : F64;
+  embed var y : F64;
+};
+
+// This struct layout would be { Ref[Point], Ref[Point] }, same as { USize, USize }
```
nwf, line 134:
```suggestion
// This struct layout would be { Ref[Point], Ref[Point] }, same as { UIntPtr, UIntPtr }
```

docs/internal/types.md
```diff
@@ -0,0 +1,359 @@
...
+  var orig : Point;
+  var dest : Point;
+};
+```
+
+### Methods
+
+Methods are functions that belong to a class or interface which can have arguments and a return type.
+
+Methods can be called statically (via `Class::method()`) if the compiler can infer the concrete type, or dynamically (via `object.method()`) otherwise.
```
nwf, line 147: I'm confused by "if the compiler can infer the concrete type"; ITYM "`Class::method(object, ...)` can be called if `object` is of type `Class`", though as you point out below, `Class::method()`s need not be given a `self` reference.

docs/internal/types.md
```diff
@@ -0,0 +1,359 @@
...
+{
+  method1() { ... }
+  method2(self : Self) { ... }
+}
+
+...
+
+// Object syntax, dynamic call
+var y : Bar = ...;
+y.method1(); // Could be Foo, but not sure, so use `y`'s vtable.
```
nwf, line 181:
```suggestion
y.method1(); // Could be Foo, but not sure, so use `y`'s vtable to call `Foo::method1()`
```

docs/internal/types.md
```diff
@@ -0,0 +1,359 @@
...
+y.method2(); // Same here, but Bar::method2(y);
```
nwf, line 182:
```suggestion
y.method2(); // Use `y`'s vtable to call `Foo::method2(y)`
```

docs/internal/types.md
```diff
@@ -0,0 +1,359 @@
...
+```
+
+The compiler will make sure that all classes that implement an interface will have the same methods to be deterministically positioned so that `Interface::method()` doesn't end up with multiple levels of indirection.
```
nwf, line 185:
```suggestion
The compiler will make sure that all classes that implement an interface will have consistent vtable layout so that `Interface::method()` doesn't end up with multiple levels of indirection.
```

docs/internal/types.md
```diff
@@ -0,0 +1,359 @@
...
+
+### Constructors
+
+Verona's built-in `new` function fully constructs an object and so must be called with values for all of the fields in the object.
+It may be used only from static methods on the class and so is usually used by factory methods.
```
nwf, line 190: What does "static" mean here?  I don't think you mean to preclude, for example,

```ts
class Mushroom
{
  let genus : String;

  clone(self: Mushroom) : Mushroom
  {
    return new Mushroom(self.genus);
  }

  create(genus : String) : Mushroom
  {
    return new Mushroom(genus);
  }
}

...
let c1 = Mushroom::clone(original);
let c2 = c1.clone();
...
```

docs/internal/types.md
```diff
@@ -0,0 +1,359 @@
...
+
+main()
+{
+  let x : (A & Serialisable & mut) = ...;
+}
+```
+
+In the above example, the type of `x` must conform to both `A` and `Serialisable` guarantees (methods, fields) as well as the `mutable` capability.
+
+Every type in Verona must be an intersection of a type and a capability, so:
```
nwf, line 322:
```suggestion
Every variable in Verona must be typed with an intersection of a type and a capability, so:
```

docs/internal/types.md
davidchisnall: Static here is used to mean methods that don't have a self parameter (i.e. class methods), but the doc doesn't mention what a self parameter is.  Actually, I believe there's no such restriction on `new` necessary: allowing an instance method to construct more instances directly isn't a problem (as you point out, for `clone`), but at least one class method must call `new` if you ever want to have an instance that reaches this point.

I'm not quite sure how `clone` interacts with `embed`.

docs/internal/types.md
```diff
@@ -0,0 +1,359 @@
...
+
+They can also have a keyword `embed`, which means they're placed in the struct directly, as opposed to the default of being a pointer to the data.
+
+For example
+```ts
+// This struct layout would be { F64, F64 }
+class Point
+{
+  embed var x : F64;
+  embed var y : F64;
```
davidchisnall, line 131: `F64` is a weird one here, because `F64` must be `F64 & imm` and the default representation of `F64` uses the space of the pointer to store the value.  This means that `embed x : F64` and `var x : F64` have the same layout (the same is true of all word-sized numeric types).

docs/internal/types.md
```diff
@@ -0,0 +1,359 @@
...
+};
+
+// This struct layout would be { Ref[Point], Ref[Point] }, same as { USize, USize }
+// With `Ref` being the notion of the "address of" an object.
```
davidchisnall, line 135:
```suggestion
// The following definition requires three separate memory allocations, one for each point and one for the line, which contains pointers to each point.
```

docs/internal/types.md
```diff
@@ -0,0 +1,359 @@
...
+class Line
+{
+  var orig : Point;
+  var dest : Point;
+};
```
davidchisnall, line 140:
```suggestion
}
// In contrast, this version requires a single memory allocation for the line, which contains two points:
class Line
{
  embed orig : Point;
  embed dest : Point;
}
```

docs/internal/types.md
rengolin: I agree, please suggest that as a patch.

docs/internal/types.md
rengolin: Maybe I could change to `F32`, when the embedded representation would be 32-bits?

docs/internal/types.md
rengolin: Ah, I see what you mean. The compiler doesn't need to infer the concrete type in `Class::method()`. I'll update this to include all 4 cases:
```
// Static
Class::method();
obj.method(); // with obj of a known concrete type -> Class::method(obj).

// Dynamic
Interface::method(obj); // which calls obj's concrete class' method
obj.method();
```

docs/internal/types.md
rengolin: I agree it's confusing. I don't know of a better name here. I'd be happy to change the nomenclature to whatever is more appropriate and less overloaded.

docs/internal/types.md
```diff
@@ -0,0 +1,384 @@
...
+Another example is a type union, where a variable can be one of many types, but the code must deal with all possible types.
+'Statically typed' means that the verification of these properties is done at compile time.
+
+The 'structural' bit means that open-world types can be defined by their shape: If an interface defines only a method `foo()` then any class that implements a method `foo()` implements this interface.
+This is in contrast to nominal typing, such as in Java, where an object implements an interface only if it adopts that interface.
+Objective-C supports both, with protocols defining nominal types but dynamic dispatch and explicit method checks providing structural typing.
+
+The 'algebraic' bit means that we can define types in terms of intersections or unions on other type expressions.
+For example, we can say `Integer | String` to define a set of constraints on a type that mean that it must either be an integer or a string (whatever they are: they may be concrete types, or they may be interfaces) or something like `Copyable & Hashable` to say any type that meets the constraints of the copyable and hashable interfaces.
+
```
nwf, line 28:
```suggestion

Verona programs are, by construction, devoid of data races.
As part of ensuring that property, references to objects in Verona carry a 'capability' annotation, which conveys access rights to that object and global properties of the program.
For example, a reference may be marked as `iso`-lated: the reference grants full read and write access and assurance that no concurrent accesses are possible.
Or a reference may be `imm`-utable: only reads are possible and the system guarantees that all live references to this object are similarly read-only.
```

docs/internal/types.md
rengolin: Replaced with F32 and added `imm` to complete the type

docs/internal/types.md
rengolin: I've added a short comment on the `self` argument to help, and with the new section on static and dynamic methods, it should be fine.

docs/internal/types.md
```diff
@@ -0,0 +1,359 @@
...
+All types should be known at compile time and the compiler can assert that the objects and operations have the correct type at all times.
+Some types are inferred by the compiler, but there is no guessing: there's either one clear inferred type or it's an error.
```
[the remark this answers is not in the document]
rengolin: I'm not sure what you mean by this. @sylvanc I think you were talking about this a few months ago, any insight?

rengolin:
Merged all changes into one commit, leaving the co-authored comments to make it clear this was a group effort.

docs/internal/types.md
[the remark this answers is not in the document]
nwf: Sorry, was too terse.  "having principal types" is a property of a type system, specifically, that every expression has exactly one most general type, in that every other type it can be said to have is a refinement of that most general type.  I was meaning to ask if this is true of Verona's type system in general (as is true of, say, SML), so there truly is only ever one answer (or the program is ill-typed), or if this is describing a policy of the type _inference_ system (as with Haskell with GADTs).  (Harper, who was my introduction to all this and so his vocabulary got stuck in my head, uses "elaboration" instead of "type inference", as [without putting too many words in his mouth, I hope] he wants people to think about it primarily as the transformation of a program with no or some type annotations into a program in which every term is annotated, with inference being a particular mechanism towards that end.)

nwf (approved):

docs/internal/types.md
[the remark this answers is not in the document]
rengolin: I may be wrong, but I believe Verona's type inference is a mechanism of validation, not of construction. So types are what they are and the inference only exposes the obvious, not creates new patterns. IIUC, this means Verona does have principal types. Though, this is something that Sylvan knows better, so we could just talk about this on our next meeting.

rengolin:
Thanks @nwf! I'd also like to have the formal approval from all participating in the review (@davidchisnall, @SeanTAllen) to make sure we're all reading the same document. Once we're happy with this one, I'll start working on the expressions document.

davidchisnall (approved):

rengolin:
Thanks all. If there are any left overs, feel free to submit a correcting PR.

PR #457, opened 2021-06-30
Fix the build with GCC:

davidchisnall:
- Remove a std::move that is redundant with RVO (GCC warning)
 - Fix an tautological compare (uintptr_t < 0)
 - Prefer std::experimental::source_location to std::source_location
   (the latter is only enabled for C++20 or later with libstdc++)
 - Disable two GCC warnings that produce false positives
   (always-inline may not be inlined - it probably is, function-pointer
   cast may be invalid - triggered by any use of dlfunc)

### DIFF
--- a/experiments/process_sandbox/CMakeLists.txt
+++ b/experiments/process_sandbox/CMakeLists.txt
@@ -14,6 +14,10 @@ set(CMAKE_CXX_STANDARD 17)
 if (MSVC)
 else ()
 	add_compile_options(-mcx16 -Wall -Wextra -Werror "$<$<CONFIG:DEBUG>:-fno-inline>")
+	# Disable GCC warnings that give a large number of false positives.
+	if (${CMAKE_CXX_COMPILER_ID} MATCHES "GNU")
+		add_compile_options(-Wno-attributes -Wno-cast-function-type)
+	endif()
 endif()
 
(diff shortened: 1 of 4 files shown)


rengolin (approved):

davidchisnall:
Looks like there are still some GCC-compat problems in the Linux-specific parts.  I'll try to fix those and then enable GCC in CI.

rengolin:
Where are you running those tests? All I can see on the CI passes, but there are no GCC sandbox texts on the CI, are there?

davidchisnall:
There are, and they [were run for this PR](https://dev.azure.com/ProjectVeronaCI/Project%20Verona/_build/results?buildId=2742&view=results).  They are run only if code in the process_sandbox directory has changed, so you won't see them in most commits.

rengolin:
These say "Clang", not GCC.

davidchisnall:
Oh, sorry, I misread.  We are not yet running CI tests on GCC, hence my comment 'I'll try to fix those *and then* enable GCC in CI.'  I have run them locally and have now submitted a PR that fixes the Linux-specific bits and enables GCC CI.

rengolin:
I was just curious because you realised they didn't fix the gcc tests after merge, which I assumed it was because some CI had failed. Nevermind.

PR #459, opened 2021-07-05
Sandbox cleanup

davidchisnall:


experiments/process_sandbox/include/process_sandbox/platform/safe_syscalls-linux.h
```diff
@@ -0,0 +1,128 @@
...
+      ;
+
+    static int constexpr faccessat2_number =
+#  ifdef __NR_faccessat2
+      __NR_faccessat2
+#  elif defined(__x86_64__)
+      // This system call was introduced with Linux 5.6, but Ubuntu 20.04
+      // doesn't provide headers that expose it, so hard-code the number
+      // for x86-64 until they catch up.
+      437
```
rengolin, line 44: Is it the same number?

experiments/process_sandbox/include/process_sandbox/path.h
```diff
@@ -0,0 +1,229 @@
...
+
+#include <list>
+#include <sstream>
+#include <string>
+#include <string_view>
+
+namespace sandbox
+{
+  /**
+   * Class encapsulating a path.  A path is route through a tree.  This class
```
rengolin, line 15:
```suggestion
   * Class encapsulating a path.  A path is a route through a tree.  This class
```

experiments/process_sandbox/include/process_sandbox/path.h
```diff
@@ -0,0 +1,229 @@
...
+   * works on an abstract filesystem namespace, it never accesses the
+   * filesystem.
+   */
+  class Path
```
rengolin, line 19: Why not use an existing library?

davidchisnall: Are there any good ones?  I've not seen one that does the things that we actually want.  The `std::filesystem::path` from the Filesystems TS is as much a joke as the rest of that TS, `boost::filesystem::path` expects to operate on the local machine's filesystem and so its canonicalisation also removes symlinks and so doesn't do the right thing for us.

experiments/process_sandbox/src/library_runner.cc
```diff
@@ -466,6 +467,134 @@ namespace
...
+     * An on-heap pointer for a copy of the pointee, allocated in the shared
+     * region.  If `original` is in the shared heap, this will be `nullptr`.
+     */
+    std::unique_ptr<T> copy;
+
+  public:
+    /**
+     * Constructor, captures `orig` and also allocates a copy of it on the
+     * heap.  The value of `orig` must not be modified for the lifetime of this
+     * object.
```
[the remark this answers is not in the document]
davidchisnall: This is not a type that's exposed to the library, if a multithreaded sandboxed library wants to race the thread using this class, it will send some nonsense data to the parent, but that's its problem.  The parent will always defensively copy this to protect against TOCTOU problems. 
 If you allocate this somewhere that's not in the sandbox, this copies it, and then you modify the original, then the changes that you've made won't be included in the RPC.  If it's an in-out parameter, then your changes will be clobbered on return.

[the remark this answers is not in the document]
rengolin: Perfect, thanks!

experiments/process_sandbox/include/process_sandbox/path.h
rengolin: Good ones? I doubt it.

LLVM has one (`llvm::sys::fs`), but it's strongly biased to compile unit handling and may not be what you want. I use it in the interop layer for `real_path` and `parent_path`, but that's mostly because the use there is similar to its bias.

The parser has one (`src/parser/path.h`), but it was created for the parser (imported from previous parser), so that, too, has local bias.

Either way, I'd try the LLVM one first, and if that's not good enough, I'd try to make our parser's `path` more generic, than create a third one in our codebase.

davidchisnall: This one has very specific requirements: It must operate on an abstract filesystem and not ever reference the real filesystem.  Given how simple it is, I'm far more comfortable having a stand-alone implementation here than I am auditing every change to another one to make sure that it does not accidentally provide a means for the sandboxed code to leak data about the filesystem contents.

experiments/process_sandbox/include/process_sandbox/platform/safe_syscalls-linux.h
davidchisnall: Ooops, good catch!

experiments/process_sandbox/include/process_sandbox/path.h
rengolin: Since this is an experiment, I don't mind much. But once we get this into the code proper, I'd like to fuse this with the parser's own implementation and make it generic. 

Given your requirements, using a third-party one is troublesome. But I'm sure we can have an implementation that we can guarantee doesn't violate our requirements for both parser and interop.

davidchisnall: It might be possible, but generally I'd want the version for the parser to end up with a capability abstraction, whereas this one is dealing entirely with names.

experiments/process_sandbox/src/library_runner.cc
```diff
@@ -481,50 +610,48 @@ namespace
...
+  intptr_t syscall_callback_helper(Args... args)
   {
-    auto args = std::make_unique<sandbox::CallbackArgs::Stat>();
-    unique_c_ptr<char> copy;
-    if (!is_inside_shared_memory(pathname))
+    typename ArgsStruct::rpc_type stack_args{args...};
+    CallbackPtrArg<decltype(stack_args)> argstruct{&stack_args};
+    auto ret = callback(ArgsStruct::kind, argstruct.get());
+    int result = static_cast<int>(ret.first);
+    if (ret.second.is_valid())
```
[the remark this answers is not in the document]
davidchisnall: I thought this was documented somewhere, but it doesn't seem to be.  Added.

rengolin (approved):

PR #460, opened 2021-07-07
Adding expressions documentation

rengolin:
Another document, now on expressions.

This document has a lot of assumptions that we discussed on meetings, but also some that we didn't and I took the liberty of picking something that is familiar to me, not necessarily to the rest of the team. This means that, if something is odd, it's very likely wrong.

There are also a few examples missing, where I'm still a little confused about the syntax, not necessarily the semantics. Please offer suggestions for examples where none are provided.

As the document on types, this is WORK IN PROGRESS and not meant as a tutorial or user facing documentation. It's just a collection of ideas we mostly agree on, not a list of things we will discuss in the future.

In places where we need more discussion and the section is incomplete, I've added _notes_ to the text to imply there's more to come.

@nwf, I can't seem to be able to add you to the reviewers list.

### DIFF
--- a/docs/internal/expressions.md
+++ b/docs/internal/expressions.md
@@ -1,0 +1,687 @@
+# Verona Expressions
+
+## WARNING
+
+This is a preliminary document under heavy discussion.
+The ideas here are mostly right, the details, not necessarily.
+There will be a tutorial document that will be correct in all counts, but this isn't it.
+
+## Overview
+
+Expressions are a sequence of operators and operands that specify a computation.
+
+Compared to most languages Verona has very few operators.
+Basically: calls, field access, `match`, `try`, `throw`, `catch` and `lambda`.
+All the other well known operations (ex. `if`, `while`, comparisons, arithmetic, etc) are build on top of those operators.
+This allows developers to modify the semantics of programs at a deeper level and makes it easy to develop DSLs on top of Verona.
+
+Verona operands, however, are similar to most languages.
+Basically: literals, variables, call arguments, return values.
+Counter-intuitively, there is no exception unwinding in Verona, exceptions thrown are just non-local return values (see below).
+
+## Literals
+
+Literals encode values of different types.
+Verona has literal of Boolean, numeric, character and array types.
+
+### Boolean
+
+Boolean literals are `true` and `false`.
+There is no coercion between numeric values and Boolean values, so one must initialise Boolean variables with their own literals, or as the result of Boolean expressions.
+
+For example:
+```ts
+// Direct assignment
+let a : Bool & imm = true;
+
+// This won't compile
+let b : Bool & imm = 42; // ERROR
+
+// Indirect assignment
+let c : U64 & imm = 42;
+let d : Bool & imm = (c == 0);
+
+// Using conversion from numeric types
+let e : Bool & imm = c.toBool();
+```
+
+### Numeric
+
+Both integer and floating point literals are type-inferenced by the type of the expression they're expected to match against.
+For example:
+
+```ts
+// This literal is 32-bits
+let a : I32 & mut = 42;
+
+// This one is 64-bits
+let b : F64 & mut = 3.1415;
+
+// This creates a temporary with the type that the `function` was declared with.
+function(42);
+```
+
+Numeric literals can be represented in binary (0b101010), decimal (42) or hexadecimal (0x2A) and will be treated in the same way.
+
+### Character literals
+
+Character literals are enclosed with single quotes (`'`).
+
+A few things are still under heavy discussion:
+* How we want character literals to natively interact with numeric types.
+* What is the core character type? A byte? A Unicode code point? A word?
+* How will string types interact with each other.
+
+For example:
+
+```ts
+// 'A' can be seen as 0x41, which is often OK as a number
+let a : U8 & imm = 'A';
+
+// This is pretty clear
+let b : U8 & imm = a + 1;
+
+// Is a Unicode code point with multiple bytes a "character"?
+let verona : U32 & imm = "🏟"; 
+
+// What does this mean?
+let treviso : U32 & imm = verona + 1; // ?!?
+
+// 'ABCD' is a bit pattern (0x41424344), what's its actual value?
+let b : U64 & imm = 'ABCD'; // little-endian? big-endian?
+```
+
+String literals are enclosed with double quotes (`"`).
+
+String types are still in discussion, but literals can be used for their construction.
+
+For example:
+```ts
+// A plain ASCii string, initialised with a string literal
+let str = "Hello World!\n"; // Creates a `String` object
+
+// An UTF-8 single character (🏟) string
+let verona = UTF8("\U1F3DF");
+
+// Or a whole UTF-16 string from an unicode string literal
+let intro = UTF16S "こんばんは、お元気ですか";
+
+// This is ok as long as the classes implement the methods
+let compat : UTF8 = intro.toUTF8();
+```
+
+### Array Literals
+
+We don't need array literals in Verona.
+We can create arrays of any types with the `Array` type constructor, like strings above.
+
+Example:
+```ts
+// An array with defined type
+let a = Array[F64](3.1415, 2.7183);
+
+// An array of a union of different types.
+let c = Array[U8 | String | F32](1, "2", 3.0);
+```
+
+## Variables
+
+Variables are placeholders for a reference to an object, to which they are bound.
+The compiler optimises the representation of certain variables (ex. numeric types) but the syntax is preserved as if they are pointers.
+
+### Let vs. Var
+
+A variable declared with `let` can only be assigned once.
+A variable declared with `var` can be reassigned multiple times.
+In both cases, this is independent of whether the object that the variable refers to can be modified.
+
+Example:
+```ts
+let a : U64 & mut = 42;
+a = 41; // This is a compiler error
+
+var b : U64 & mut = 42;
+b = 41; // This is perfectly fine
+
+// This is also fine. The variable still points to the same object
+// even if its contents have changed (it's a mutable object)
+let c : MyType & mut = MyType;
+c.updateMe();
+```
+
+### Value vs. Pointer
+
+All types in Verona are classes, so (at the abstract machine leve) all variables in Verona store pointers to objects.
+As an optimisation, small immutable objects may be stored inside the pointer representation, avoiding the need for heap allocation.
+This is the default case for any simple numeric types and may be extended to others in the future.
+Refer to the internal ABI document for more details on object representation, including singleton types.
+
+### Stack vs. Heap
+
+The lifetime of variables is either lexically scoped, when logically allocated on the stack, or determined by their regions' semantics, when allocated on the heap.
+
+All new objects are allocated in some region, made explicit by their declarations.
+There isn't yet consensus no the syntax of declaring region types, but it will be required for new regions.
+Other variables can be allocated on the same region as a previously declared one with the keyword `in`.
+
+Example:
+```ts
+// Creates a new region of type "FooBar" and allocates a "MyType" variable in it
+var a : MyType & iso = MyType "in region type FooBar"; // Pretend-syntax for regions
+// Now, `a` is a pointer to an element in that region
+
+// Reuses the same region as `a`
+var b : OtherType & mut = OtherType in a;
+// Now, `b` is a pointer to another element in the same region as `a`
+```
+
+The region is responsible for managing that memory. 
+The compiler and runtime guarantee that pointers can outlive the objects to which they point.
+Memory deallocation isn't necessarily done at the same time as invalidation of pointers (ex. garbage collected regions), but the compiler also guarantees that no code will follow pointers after becoming invalid.
+
+Objects that can be transferred from one behaviour to another must be allocated in a region and that region must be transferred in its entirety.
+The deallocation of those objects will depend on the policies defined by the type of region.
+
+Variables declared in functions have lifetimes defined by their lexical scope.
+If such a variable is captured by a `when` clause and sent to a new behaviour then the lifetime ends at the end of the lexical scope that captured the variable.
+This capturing involves a transfer of ownership and so the variable is no accessible in the original scope once the new scope has been defined.
+
+Example:
+```ts
+foo()
+{
+  // Declares a function-level variable
+  var a : U64 & mut = 42;
+
+  // Start a new lexical scope
+  {
+    // Create a new region of the default type
+    var b : MyType & iso = MyType;
+
+    // This only schedules the work, not executes it
+    when () {
+      // `b` is captured by the behaviour and the behaviour now owns the new region
+      b.doSomething();
+    }
+
+    // This is an error, as `b` was captured above
+    b.somethingElse();
+  }
+
+  // No `b` in this scope
+  b = ...; // ERROR
+
+  // Updates a, still valid
+  a = 38;
+} // `a` goes out of scope
+```
+
+### Object fields
+
+Like other variables, object fields store pointers to objects, unless declared `embed`, which makes them part of the structure.
+Refer to the types document for more information on the `embed` keyword.
+
+## Operators
+
+In Verona, operators and calls are treated equally.
+In-fix operators (ex. `a + b`) and pre-fix operators (ex. `+(a, b)`) are one and the same.
+
+All calls will invoke a method whose definition is scoped to some class.
+This class may be specified directly or inferred dynamically from the concrete type of an object.
+For example the type name (`Foo::method()`), the first argument of a pre-fix operator (`foo.method()` or `method(foo)`), or the left hand side of an infix operator (`foo method bar`).
+
+Example:
+```ts
+// Defines the type of `x`
+let x : U32 & imm = 42;
+
+// Infers the type of `y` and `10` from the type of `x`
+// Checks that a method matching `U32::+(U32, ?)` exists.
+// It does, with the full signature `U32::+(U32 & imm, U32 & imm) : U32 & imm)`
+// Unifies the literal `10` with the `U32 & imm` from the required type of the second argument.
+// Infers type of `y` from the return type (`U32 & imm`)
+// Calls U32::+(x, 10)
+let y = x + 10;
+
+// The above is identical to ones below
+let a = x.+(10);
+let b = U32::+(x, 10);
+```
+
+The Verona standard library will implement all operations for all appropriate types, including all arithmetic for all numeric types and string handling for all character and string types.
+Users can implement similar calls to specialised type (ex. `MyInt::+(MyInt, I32)`).
+
+There is no precedence in Verona, so operators must be wrapped in parenthesis to convey the right meaning.
+This means that, for example, `(a + b * c)` is a syntax error, whereas either `((a + b) * c)` or `(a + (b * c))` are expected.
+
+_Note: Can we avoid forcing parenthesis on DSLs?_
+
+## Arithmetic
+
+As detailed above, arithmetic in Verona is implemented as function calls on numeric types and operands.
+This means developers can extend the functionality of arithmetic to the existing types or their own new types naturally.
+It also means that, in a naïve implementation, every addition or subtraction would end up as a function call, possibly with dynamic dispatch, which could potentially be many orders of magnitude slower than the single hardware instruction required to implement the operation.
+
+Verona's arithmetic is designed to be easy to optimise to a single hardware instructions for any of the cases that would be expressible in languages such as Java or C++ that have primitive types that are distinct from object types.
+
+The reasons are:
+1. Numeric types are always immutable (you never, for example, modify the value `4`), which means that aliasing is not observable.
+   If two `U32` variables, for example, both refer to the value `4`, neither can modify that value (though either can be updated to refer to a different number) and so the compiler is free to embed the value inside the pointer.
+2. Computation often uses concrete types (ex. `U32` or `F64`) instead of interface types (ex. `IntegralType`) or type unions (ex. `(U32 | F64`)).
+   This means that the compiler does not have to employ NaN boxing, small integer boxing, or other tricks to store either a numeric value or a pointer in a variable.
+   Similarly, the compiler does not need to insert dynamic run-time checks that a particular variable contains a numeric type: if a variable is of type `U32`, it can store only a `U32`, never a pointer to some other type.
+3. The functions that implement arithmetic operations are trivial (often single instruction) functions that are ideal inlining candidates.
+
+The combination of these means that a `U32` can be stored as a 32-bit integer, rather than a pointer to a 32-bit integer stored elsewhere, that the `+` operation can be a direct call (no late binding or dynamic dispatch) to a built-in function that implements 32-bit integer addition, and this function can always be inlined.
+This comes without any loss of generality: you can still use dynamic dispatch over numeric types via interfaces or union types but you only pay for the additional overhead when you use it.
+
+This means `(a + b)` in Verona usually completely bypasses function calls and wrappers and just call a single `add` instruction.
+
+### Semantics
+
+There is no undefined behaviour in Verona.
+There are no traps or run-time exceptions from arithmetic in Verona.
+
+Verona's integer arithmetic has wrap-around semantics as defined behaviour.
+Adding unsigned values past their maximum values wraps around to zero, subtracting signed values past their minimum negative values wraps around to positive, etc.
+
+Integer division by zero is zero.
+Floating point division by zero is `Inf` (or `-Inf` if divided by `-0`).
+Verona will never trap or throw exceptions for divisions by zero.
+This involves checked code on architectures that do not provide that functionality.
+
+Users that want a different semantics for arithmetic can create their own wrappers and types.
+
+### Type Conversion
+
+Type conversion in Verona is always explicit, including numeric types.
+To operate on different types, users have to convert each type to the operation's expected type.
+
+Example:
+```ts
+let a : U32 & imm = 20;
+let b : U64 & imm = 22;
+
+// There is no such method U32::+(U32, U64)
+let x = a + b; // ERROR
+
+// Convert to U32 first
+let y = a + b.toU32();
+```
+
+This is valid for all conversions: Different sizes, signs, floating point.
+It allows for explicit semantics to make sure developers write their intentions clearly, and not have to remember complex rules on what casts to what and what's the wrapping semantics of the operations' types.
+
+## Control Structures
+
+Unlike many other programming languages, Verona does not have the traditional control flow structures like conditionals and loops.
+Verona follows Smalltalk-family languages in providing a minimal set of built-in flow control and implementing a richer set in the standard library.
+This allows complex flow control features such as for-all over a collection, to have the same syntax as more primitive constructs and makes user-defined control flow constructs as much first-class language features as the set provided by the standard library.
+Unlike Smalltalk, these features do not depend on dynamic dispatch or dynamic type information and are amenable to inlining in an ahead-of-time compiler.
+The Verona language defines only a minimal set of control-flow features that cannot be implemented as part of the standard library.
+
+### Core control structures
+
+TODO: Populate this entire section with examples, as they're super useful but I'm not clear on their full syntax.
+
+#### Match
+
+`match` is a conditional statement that allows matching an expression by type or value and execute an arbitrary lexical block of code for each match.
+
+#### Try/Catch/Throw
+
+Verona's exceptions are fully checked at compile time and are implemented as non-local returns.
+This means there are no hidden exceptions from the standard library or user code.
+
+Every function's return type must be checked against the variable or context they're being assigned to, including the exceptions they throw, via a `throws` keyword.
+
+More specifically:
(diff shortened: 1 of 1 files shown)


docs/internal/expressions.md
```diff
@@ -0,0 +1,597 @@
...
+
+// Singleton types are always equal to themselves but not other singleton types
+class None {}
+class All {}
+
+let always_eq = (None is None); // TRUE
+let never_eq = (None is All); // FALSE
+```
+
+Because singleton types (other than numeric types) have no machine representation, they always compare equal on the same types, which means all variables of the same singleton type are aliases to each other.
```
davidchisnall, line 524: I think this is saying that singletons are singletons and comparing two pointers to the same object will always yield true.

docs/internal/expressions.md
```diff
@@ -0,0 +1,597 @@
...
+
+Note that, even though literals have a machine representation for their values, the addresses that they are stored for `a` and `c` are different, and therefore they are not the _same variable_, even though they have the same value.
```
davidchisnall, line 526: I don't understand what this means.  Two `U32` things with the same value are the same object.  Comparison by pointer will yield equality because of the encoding tricks that we're doing with the pointer and because we guarantee that we always give you the same `42` for any `U32`.

docs/internal/expressions.md
```diff
@@ -0,0 +1,597 @@
...
+
+### Structural equality
+
+Structural equality, compared with the method `==` on appropriate types, is a comparison of the values in the object that the variable points to, including numeric literal values.
+It is the responsibility of each class to implement their own structural equality method.
+
+The standard library implements them for all classes that can be structurally compared.
```
davidchisnall, line 533: This is tautological.

docs/internal/expressions.md
```diff
@@ -0,0 +1,597 @@
...
+User objects can be compared structurally, simply by implementing the `==` method.
+Similar operators (`!=`, `<`, `>`, etc) are also implemented in the standard library and can be used by any user class in the same way.
+
+Example:
+```ts
+// Some variable
+var a : U32 & mut = 42;
```
davidchisnall, line 540: `U32 & mut` should be an error.

docs/internal/expressions.md
```diff
@@ -0,0 +1,597 @@
...
+// A new variable with different value
+let b = a + 1;
+// A new variable with the same value (but different address)
+var c : U32 & mut = 42;
```
davidchisnall, line 544: Error, because of the `mut`.  If we have two `U32 & imm`, they will refer to the same object if they have the same value.

docs/internal/expressions.md
```diff
@@ -0,0 +1,597 @@
...
+```
+
+_Note: There is nothing special about structural comparison methods.
+ They're methods like any other that just happens to "implement" comparisons.
+ You can do whatever you want with your own types for `==`._
+
+## Syntax Sugar
+
+In Verona, every operation is a method call, including lambda execution, object creation, destruction, etc.
+To create syntactically correct code, one must be explicit about everything.
```
davidchisnall, line 557: Everything?

docs/internal/expressions.md
```diff
@@ -0,0 +1,597 @@
...
+The two main sugars are:
+* `create`, for constructing objects, allowing you to call just the class name instead of the whole call syntax.
+* `apply`, for calling objects as functions, including lambdas, without the need to call it directly.
+
+The code above would look like:
+```ts
+let a : MyType & mut = MyType;
+a();
+```
+_Note: We also want the `update` sugar, but it may need a `Ref[T]` type to do `obj.update(field) = val` where `update()` returns `Ref[T]`._
```
davidchisnall, line 581: I thought the point of `update` sugar was to avoid the need for refs.

docs/internal/expressions.md
```diff
@@ -0,0 +1,597 @@
...
+  b = ...; // ERROR
+
+  // Updates a, still valid
+  a = 38;
+} // `a` goes out of scope
+```
+
+### Class fields
+
+Like other variables, class fields have the same value/pointer duality as before.
```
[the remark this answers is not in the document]
rengolin: ok

docs/internal/expressions.md
```diff
@@ -0,0 +1,597 @@
...
+Unless declared `embed`, class fields are pointers to the data.
+This is particularly useful for numeric fields.
+Refer to the types document for more information on the `embed` keyword.
+
+## Operators
+
+In Verona, operators and calls are treated equally.
+In-fix operators (ex. `a + b`) and pre-fix operators (ex. `+(a, b)`) are one and the same.
+All operators are calls on the descriptor object: the first argument of a pre-fix operator, or the left hand side of an infix operator (`a` above).
```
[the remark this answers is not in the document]
rengolin: expanded.

docs/internal/expressions.md
```diff
@@ -0,0 +1,597 @@
...
+Users can implement similar calls to specialised type (ex. `MyInt::+(MyInt, I32)`).
+
+There is no precedence in Verona, so operators must be wrapped in parenthesis to convey the right meaning.
+This means that, unlike other languages, `(a + b * c)` is not the same as `(a + (b * c))`.
+
+_Note: Do we want to force parenthesis in DSLs?_
+
+## Arithmetic
+
+As detailed above, arithmetic in Verona is equivalent to function calls on numeric types and operands.
```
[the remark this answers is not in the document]
rengolin: 'implemented as', hopefully clearer.

docs/internal/expressions.md
```diff
@@ -0,0 +1,597 @@
...
+
+But Verona's arithmetic is as fast as calling the right hardware instructions.
+
+The reasons are:
+1. As explained in the types document, numeric types are singleton classes, ie. they don't have fields, only methods.
+   The compiler treats them differently and uses their machine representation instead of a pointer to an allocated memory.
+2. Computation often uses concrete types (ex. `U32` or `F64`) instead of interface types (ex. `IntegralType`) or type unions (ex. `(U32 | F64`)).
+   The compiler can only match concrete types to their machine equivalent.
+   It can try to infer, match and separate concrete types from their declared collection types, but that's an optimisation, not a guarantee.
+3. Inline IR calls (ex. `@llvm.uadd.with.overflow.i8`) can be made in Verona, making explicit the exact instructions that each operation needs.
```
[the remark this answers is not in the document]
rengolin: ok, I'll change to "direct code generation" or something

docs/internal/expressions.md
```diff
@@ -0,0 +1,597 @@
...
+Floating point division by zero is `Inf` (or `-Inf` if divided by `-0`).
+Verona will never trap or throw exceptions for divisions by zero.
+This involves checked code on architectures that do not provide that functionality.
+
+Users that want a different semantics for arithmetic can create their own wrappers and types.
+
+### Type Conversion
+
+Type conversion in Verona is always explicit, including numeric types.
+Unlike most other languages, operations on different types is forbidden (by their own implementation in the standard library).
```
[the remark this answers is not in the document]
rengolin: removed the phrase as confusing

docs/internal/expressions.md
```diff
@@ -0,0 +1,597 @@
...
+let y = a + b.toU32();
+```
+
+This is valid for all conversions: Different sizes, signs, floating point.
+It allows for explicit semantics to make sure developers write their intentions clearly, and not have to remember complex rules on what casts to what and what's the wrapping semantics of the operations' types.
+
+## Control Structures
+
+Unlike most other programming languages, Verona does not have the traditional control flow structures like conditionals and loops.
+It does however, implement them using the core Verona control flow structures and make available to programmers, so that programming in Verona doesn't become a burden.
```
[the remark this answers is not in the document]
rengolin: I'm not well versed in SmallTalk enough to write a paragraph on it. Please suggest as a diff.

docs/internal/expressions.md
```diff
@@ -0,0 +1,597 @@
...
+}
+
+// On some yet another module
+using "MyModule";
+using "OtherModule";
+bar()
+{
+  foo(); // ERROR: which one?
+  MyModule::foo(); // Fully qualified, this is fine.
+}
```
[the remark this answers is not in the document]
rengolin: I agree, but the example is still valid.

Also, this problem happens even without importing both directly:
```
// File1 has
use "OpenFoo"; // provides `foo`

// File2 has
use "SuperFoo"; // also provides `foo`

// File3 has
use "File1"
use "File2"
```

which one you use above may bring different interfaces for the same functionality. Code may have both on non-intersecting places and you only need one file to include other files that end up including both to break on a non-local error message.

docs/internal/expressions.md
```diff
@@ -0,0 +1,597 @@
...
+bar()
+{
+  foo(); // Fine, `OtherModule`'s version is `Other::foo()`.
+  Other::foo(); // Fully qualified, using type alias, this is fine.
+}
+```
+
+### Calling methods
+
+Calling a method in Verona is similar to most other languages.
```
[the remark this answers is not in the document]
rengolin: removed

docs/internal/expressions.md
```diff
@@ -0,0 +1,597 @@
...
+The name must be fully qualifiable (not necessarily fully qualified) to identify the type that implements it, either at compile time (for static calls) or at run-time (for dynamic calls).
```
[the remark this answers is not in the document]
rengolin: clarified

docs/internal/expressions.md
rengolin: reworded

docs/internal/expressions.md
rengolin: This was one of the discussions in a recent meeting: Sylvan argued that two numeric types of the same value shouldn't compare equal, ie that the compiler shouldn't trick users into thinking that.

It wasn't a strong opinion, just an idea.

docs/internal/expressions.md
rengolin: removed

docs/internal/expressions.md
rengolin: changed

docs/internal/expressions.md
rengolin: not according to the new idea that `a` and `c` could be different objects with the same value.

we can discuss this on the next meeting and make a more general decision.

I'm personally happy either way.

docs/internal/expressions.md
rengolin: well, it's what I meant above between "qualified" vs "qualifiable". I don't know how to write this.

the compiler must be able to know that everything is correct, but the programmer can ignore a lot of stuff if the compiler "can" know itself.

So, the code must be explicit, even if the programmer isn't while writing code.

Makes sense?

docs/internal/expressions.md
rengolin: this is a snippet from one of @sylvanc meetings. I'll let him answer that.

docs/internal/expressions.md
```diff
@@ -0,0 +1,597 @@
...
+// Or a whole UTF-16 string from an UTF-16 literal
+let intro : UTF16String & imm = "こんばんは、お元気ですか";
+
+// This is ok as long as the classes implement the methods
+let compat : UTF8String & imm = intro.toUTF8();
+```
+
+### Array Literals
+
+The syntax for array literals hasn't been defined yet but we certainly want array literals with the usual semantics.
```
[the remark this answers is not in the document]
rengolin: Changed that section.

docs/internal/expressions.md
```diff
@@ -0,0 +1,597 @@
...
+```
+
+String literals are enclosed with double quotes (`"`).
+
+String types are still in discussion, but literals can be used for their construction.
+
+For example:
+```ts
+// A plain ASCii string, initialised with a string literal
+let str : CString & imm = "Hello World!\n";
```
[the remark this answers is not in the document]
rengolin: This is something I discussed with Sylvan this week and have changed the example on strings. The rest is a reply to your comment on the previous iteration.

The idea of an interface that defines what a string is can be interesting for modelling abstract interfaces, but in itself it's a bit void of context. Operations on different types of strings can be wildly different and not mean the same thing (for example `size` vs `length`, or `at`, `substr` and `replace`).

More concretely, interfaces on the _actions on strings_ (ex. `Iterable`, `HumanReadable` or `RandomAccess`) are more meaningful because they tell you what you need to provide in terms of functionality, not abstract relationships.

But the real driver behind this thinking, in my view, is simplicity of implementation. Systems programming needs very simple strings to manipulate, including using as keys to maps, etc. This must be as fast as possible and trumps speed concerns for user interfaces and logs. So, a simple string class is paramount.

In that sense, a string class would be something this simple:
```
class String {
  embed _contents : Array[U8];
}
```

and other strings can make use of that in their own representations:
```
class UTF8 {
  embed _contents : String;
  ...
}
```

I haven't had such an extensive experience in handling unicode strings across different languages, so I can't tell you how much of a burden this would be. But I have had enough experience working with simple strings (`char *`) to know that an array of bytes is very ameanable to all sorts of optimisations for systems programming.

This is still an open question, but one that we can take offline and change the document once it's clearer.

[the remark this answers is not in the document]
davidchisnall: > But the real driver behind this thinking, in my view, is simplicity of implementation. Systems programming needs very simple strings to manipulate, including using as keys to maps, etc. This must be as fast as possible and trumps speed concerns for user interfaces and logs. So, a simple string class is paramount

Note: Verona is *not* a systems programming language, it is an infrastructure programming language.  The use case of keys is important, but so is handling text protocols, and parsing JSON / XML.  

> I haven't had such an extensive experience in handling unicode strings across different languages, so I can't tell you how much of a burden this would be. But I have had enough experience working with simple strings (char *) to know that an array of bytes is very ameanable to all sorts of optimisations for systems programming.

Even for kernel programming, it's painful.  For example, a POSIX filesystem generally has to support unicode for filenames and should return the same file for different representations of the same string.  This means that you have to implement unicode canonicalisation in the kernel, which adds some complexity.  

An array of code units is something that we can definitely use as a building block, but requiring contiguous storage for a string ends up leaving a lot of potential optimisation on the floor for mutable strings.  Every C/C++ codebase that does string handling ends up building some kind of twine-like abstraction because contiguous storage for mutable strings is painful, but in LLVM you have a load of redundant copies because it has to deal with interfaces that expect a C/C++ string and so has to flatten the Twine at the end.  If you define your string interfaces to be suitably flexible at the start, then you can elide this copy.  

As soon as you require a particular representation for strings, you are optimising for a specific usage pattern and you'd better be sure that that covers 100% of your use cases.  When it doesn't, as in C/C++, you end up with a proliferation of incompatible higher-level string classes and then people end up building (also incompatible) higher-level abstractions such as [UText from ICU](https://unicode-org.github.io/icu/userguide/strings/utext.html) that can wrap all of these.  If you provide a clean string interface from the start, people create a lot of different implementations of this interface and all of the code that operates on strings works across all of them.

docs/internal/expressions.md
```diff
@@ -0,0 +1,597 @@
...
+```
+
+Numeric literals can be represented in binary (0b101010), decimal (42) or hexadecimal (0x2A) and will be treated in the same way.
+
+### Character literals
+
+Character literals are enclosed with single quotes (`'`).
+
+With the introduction of Unicode standards, characters are no longer single-bytes.
+However, the size of a character can be different depending on which Unicode standard you use (ex. `UTF-8` vs. `UTF-16`).
```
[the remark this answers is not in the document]
rengolin: > I think our character type should correspond to one code point / rune.

I think our basic "character type" can be a byte (U8) but our basic "rune type" can be more elaborate.

`String` objects must handle basic characters but not necessarily complex runes. Complex string types must handle their own rune types efficiently, but the conversion between them doesn't need to be.

I'm under the assumption that hot-code in programs won't have multiple string types being converted back and forth.

Again, this is a longer discussion than this PR, so we should talk about this on a meeting.

docs/internal/expressions.md
```diff
@@ -0,0 +1,597 @@
...
+  {
+    // Unless MyType::create() specifies the region, this is not allowed
+    var b : MyType & iso = MyType; // Possible ERROR
+
+    // Assuming `b` was created on the stack, this is not allowed
+    var cown : Cown(b);
+    // This only schedules the work, not executes it
+    when (cown) {
+      // This would probably execute *after* `b` has been deallocated
+      b.doSomething();
```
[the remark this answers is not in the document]
rengolin: Ah, right, changed.

docs/internal/expressions.md
davidchisnall: `a` and `c` are different pointers.  Making them `mut` makes sense *only* if they may refer to the same object such that changing one changes the other.  If you want that, you can create an object with a `U32 & imm` field and hold a `mut` reference to it in `a` and `b`.  If they are `imm` then it doesn't matter if they refer to the same or different objects: if they refer to objects with the same value then (as an optimisation) they can refer to the same object because we don't guarantee that `U32::+(U32 & imm, U32 & imm) : U32 & imm` returns a *new* immutable `U32`, it is allowed to return a pointer to an existing `U32 & imm` and (as an optimisation) we can therefore embed the value in the object.

If you have a `U32 & mut` then you have two choices:

 - It's a boxed thing (an object that has a word field) and you need to make sure it's in the right region and so on.
 - We have special casing of integers / floats that leaks all of the way into the abstract machine, rather than being purely an implementation artefact.

Neither of these makes me happy.

docs/internal/expressions.md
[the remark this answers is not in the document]
rengolin: I've updated the character part to make it more clear, but I don't think we should expand too much here. Let's take this discussion offline and update the string document once we're done.

docs/internal/expressions.md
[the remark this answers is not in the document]
rengolin: > The use case of keys is important, but so is handling text protocols, and parsing JSON / XML.

The way I see it is that an ASCii string is not the same thing as an UTF-8 string. They're fundamentally different classes of objects and we should be trying to common them up in an interface.

It's not because I can convert one to the other that they're similar, in the same way I can convert a floating point to an integer and vice versa, because there is often loss in the process.

Using a contiguous string of bytes for keys that have clear comparison rules is fundamental for performance, while trying to implement comparison on twine-like objects with variable sized grains is a nightmare.

But reading a JSON file as bytes and then matching symbols for parsing can be a nightmare. Not having a twine-like model here would also be a problem for streaming and incremental parsing.

So, my opinion is that we need:
* an class that is just a bag of bytes (whatever we call it), used for simple naming, map keys, symbols and low level stuff.
* a number of classes that implement human readable strings, which can themselves use the class above as storage and logic.
* potentially separate these implementations in compact / sparse modes.
* a set of explicit readers and writers (in classes like JSON and XML) on specific "string types" and not on an interface type.

> As soon as you require a particular representation for strings, you are optimising for a specific usage pattern and you'd better be sure that that covers 100% of your use cases.

Right, this is why my proposal is not to have a universal representation for "strings". But to have multiple string representations and conversion methods between them, so that developers can pick what suits them best for each case and only explicitly pay the cost of conversion.

docs/internal/expressions.md
rengolin: I have already replaced `mut` with `imm`

docs/internal/expressions.md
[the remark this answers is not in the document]
rengolin: I'm going to "resolve" this conversation, as it has strayed too far from the intent of the document, not because what this document states is correct. There is another document on Strings and we can continue to discuss the issues there.

docs/internal/expressions.md
rengolin: I have reworded here to make it more clear and avoid "everything".

docs/internal/expressions.md
```diff
@@ -0,0 +1,611 @@
...
+But Verona's arithmetic is often as fast as calling the right hardware instructions.
+
+The reasons are:
+1. As explained in the types document, numeric types are singleton classes, ie. they don't have fields, only methods.
+   The compiler treats them differently and uses their machine representation instead of a pointer to an allocated memory.
+2. Computation often uses concrete types (ex. `U32` or `F64`) instead of interface types (ex. `IntegralType`) or type unions (ex. `(U32 | F64`)).
+   The compiler can only match concrete types to their machine equivalent.
+   It can try to infer, match and separate concrete types from their declared collection types, but that's an optimisation, not a guarantee.
+3. Direct code generation of the hardware instructions for arithmetic functions.
+   Those functions are often short and simple and are always inlined by the compiler.
```
davidchisnall, line 275:
```suggestion
3. The functions that implement arithmetic operations are trivial (often single instruction) functions that are ideal inlining candidates.

The combination of these means that a `U32` can be stored as a 32-bit integer, rather than a pointer to a 32-bit integer stored elsewhere, that the `+` operation can be a direct call (no late binding or dynamic dispatch) to a built-in function that implements 32-bit integer addition, and this function can always be inlined.
This comes without any loss of generality: you can still use dynamic dispatch over numeric types via interfaces or union types but you only pay for the additional overhead when you use it.
```

docs/internal/expressions.md
```diff
@@ -0,0 +1,611 @@
...
+
+### Semantics
+
+There is no undefined behaviour in Verona.
+There are no traps or run-time exceptions from arithmetic in Verona.
+
+Verona's integer arithmetic has wrap-around semantics as defined behaviour.
+Adding unsigned values past their maximum values wraps around to zero, subtracting signed values past their minimum negative values wraps around to positive, etc.
+
+Integer division by zero is zero.
```
davidchisnall, line 287: Is there an efficient implementation of this?  I can see it for `x / 0 == 1` by doing `x / n = x / ((n == 0) ? x : n)`, which is a single extra instruction (maybe two) and is amenable to vectorisation, but I don't know how you get the 0 behaviour without a branch.

docs/internal/expressions.md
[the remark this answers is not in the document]
davidchisnall:
```suggestion
Verona follows Smalltalk-family languages in providing a minimal set of built-in flow control and implementing a richer set in the standard library.
This allows complex flow control features such as for-all over a collection, to have the same syntax as more primitive constructs and makes user-defined control flow constructs as much first-class language features as the set provided by the standard library.
Unlike Smalltalk, these features do not depend on dynamic dispatch or dynamic type information and are amenable to inlining in an ahead-of-time compiler.
The Verona language defines only a minimal set of control-flow features that cannot be implemented as part of the standard library.
```

docs/internal/expressions.md
```diff
@@ -0,0 +1,611 @@
...
+
+Each lambda gets its own type with the block inserted into the `apply()` method.
+Lambdas initialise the fields on creation and execute `apply()` when they're called.
+Their lifetime is bound to the region in which they're created.
+
+#### When
+
+A `when` block creates a behaviour that is executed asynchronously.
+The behaviour is a lambda that is scheduled for execution.
+The `when` block returns immediately, but the execution can be done at any time later.
```
davidchisnall, line 354:
```suggestion
The `when` expression evaluates immediately, but the body may execute later.
```

docs/internal/expressions.md
```diff
@@ -0,0 +1,611 @@
...
+
+To control which `when` block can work on which region, the argument is a list of regions (which can be empty for immediate dispatch), that the block needs to access.
```
davidchisnall, line 356:
```suggestion
To control which `when` block can work on which region, the argument is a list of `cown`s (which can be empty for immediate dispatch), that own the regions that the block needs to access.
A `when` block may also directly capture `iso` variables to transfer ownership directly into the new behaviour.
```

docs/internal/expressions.md
```diff
@@ -0,0 +1,597 @@
...
+The scheduler controls which behaviours will happen first (in causal order) and which will have to wait for others to finish (if they require regions that other behaviours are still working on).
+
+So, instead of a `fork/join` model, `when` has a dependency model which the runtime uses to decide which behaviour to execute first, or in parallel.
+
+### Emulated semantics
+
+Despite not having `if` and `while` as native statements, Verona uses the other control structures to implement them, arriving at the usual semantics.
```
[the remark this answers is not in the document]
davidchisnall:
```suggestion
The Verona standard library will provide a rich set of control flow constructs including those that are built-in constructs in C-family languages such as `if` statements, `while` and `for` loops.
```

docs/internal/expressions.md
```diff
@@ -0,0 +1,611 @@
...
+A `when` block creates a behaviour that is executed asynchronously.
+The behaviour is a lambda that is scheduled for execution.
+The `when` block returns immediately, but the execution can be done at any time later.
+
+To control which `when` block can work on which region, the argument is a list of regions (which can be empty for immediate dispatch), that the block needs to access.
+The scheduler controls which behaviours will happen first (in causal order) and which will have to wait for others to finish (if they require regions that other behaviours are still working on).
+
+So, instead of a `fork/join` model, `when` has a dependency model which the runtime uses to decide which behaviour to execute first, or in parallel.
+
+### Emulated semantics
```
davidchisnall, line 361:
```suggestion
### Standard-library control flow
```

docs/internal/expressions.md
```diff
@@ -0,0 +1,597 @@
...
+Methods are functions that were declared on a particular type.
+Like most other languages, methods can have arbitrary code inside, including creating asynchronous behaviours that could be executed long after the method has returned.
+
+### Declaring methods
+
+Methods can have any number of arguments (including none) and can return zero or one values.
+If a method needs to return more than one value, it can do so by returning a tuple of values.
+
+The types of the arguments and return values must be known at compile time, though they do not need to be concrete (classes).
+For non-concrete cases, their implementation needs to be complete for closed-world cases (ex. account for all types in a union) but not for open-world cases (ex. interfaces), as long as they only use features that the interfaces provides.
```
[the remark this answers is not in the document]
davidchisnall: I think the compiler will warn if you don't add a `{ _ => }` clause.

docs/internal/expressions.md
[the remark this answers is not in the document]
davidchisnall: I'm going to hit resolve here, because it's not a topic for this doc, but I have some thoughts about versioning and how we can try to address some of this that I think would be good to discuss.

docs/internal/expressions.md
```diff
@@ -0,0 +1,611 @@
...
+{
+  foo(); // Fine, `OtherModule`'s version is `Other::foo()`.
+  Other::foo(); // Fully qualified, using type alias, this is fine.
+}
+```
+
+### Calling methods
+
+Calling a method by name must allow the compiler to find the class or interface it belongs to at compile time.
+For concrete class types, the call will be static (direct call to the function), and for abstract interface or union types, a dynamic dispatch will be performed to the object's virtual table.
```
davidchisnall, line 463:
```suggestion
For concrete class types, the call will be static (direct call to the function), and for abstract interface or union types, a dynamic dispatch will be performed via the object's virtual table.
```

docs/internal/expressions.md
```diff
@@ -0,0 +1,611 @@
...
+
+_Note: It's still under review if we'll use named parameters or forwarding arguments for calls._
+
+_Note2: Check the type document for a review on static vs. dynamic methods, constructors and destructors._
+
+## Equality
+
+Verona has two types of equality: by address and structural.
+
+### Address equality
```
davidchisnall, line 500: Given this week's discussion, I think this section probably needs some significant rewrites.

docs/internal/expressions.md
```diff
@@ -0,0 +1,611 @@
...
+The syntax for object literals is still under discussion.
+
+Lambdas are an example of object literals, where the fields are the captures (and their types inferred from the objects captured) and the single method `apply()` is the body of the lambda.
+
+## Partial Application
+
+Partial application refers to the process of fixing some arguments of a call and returning a lambda with the remaining arguments to be passed to the lambda.
+As most of Verona's high-level functionality, it will be implemented in the Verona standard library as a method called `~`, using type lists for the arguments.
+
+TODO: Add an example here.
```
davidchisnall, line 611: Partial application is purely a standard-library thing, so we can probably drop it here.

davidchisnall (approved):
Lots of comments but they're almost all minor wording nits.

docs/internal/expressions.md
```diff
@@ -0,0 +1,611 @@
...
+
+### Object fields
+
+Like other variables, object fields store pointers to objects, unless declared `embed`, which makes them part of the structure.
+Refer to the types document for more information on the `embed` keyword.
+
+## Operators
+
+In Verona, operators and calls are treated equally.
+In-fix operators (ex. `a + b`) and pre-fix operators (ex. `+(a, b)`) are one and the same.
```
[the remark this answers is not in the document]
rengolin: I have to say I'm often confused about operators and operands in Verona, so I can't say for sure (maybe that's why I didn't say it here).

But I imagine `++a` and `a++`  are the same as `TypeOf(a)::++(a)`. So, unary and postfix operators are the same.

If they are, I won't mind adding a paragraph to the document, though we can always add on a new PR later.

docs/internal/expressions.md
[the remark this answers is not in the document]
rengolin: Ok, then that's what I mean by complete. I'll reword to make it clear.

docs/internal/expressions.md
rengolin: Yes, I'd like to get this right before merge, tbh.

PR #461, opened 2021-07-15
Adding capabilities document

rengolin:
This may be incomplete, but I'm not sure how to match the pony tutorial sections with Verona if some of the types are missing and the other's semantics are slightly different. For example, I haven't added a section on viewpoints because I'm not sure how that'll work.

I'd gladly to accept proposals for entire sections if you're feeling charitable.

### DIFF
--- a/docs/internal/capabilities.md
+++ b/docs/internal/capabilities.md
@@ -1,0 +1,240 @@
+# Reference Capabilities
+
+## WARNING
+
+This is a preliminary document under heavy discussion.
+The ideas here are mostly right, the details, not necessarily.
+There will be a tutorial document that will be correct in all counts, but this isn't it.
+
+## Overview
+
+Reference [capabilities](https://en.wikipedia.org/wiki/Capability-based_security) are unforgeable tokens of authority that confer the rights to access something.
+They can be delegated to other components of a system and must always respect the *principle of intentionality*: you must present a capability that confers the rights to perform an action when attempting to perform that action, it is not sufficient to simply own the capability.
+
+In Verona, pointers are capabilities: the ability to name an object confers the right to access the object in some way.
+Capabilities confer a range of different permissions, which are reflected in the type system.
+
+Variables represent pointers to objects and they must have rights associated that describes what you can do to the objects via those pointers.
+
+There are kinds of rights that can be granted to capabilities:
+* Isolated (`iso`).
+  Indicates that a capability is *the globally unique* pointer to its target.
+  Every `iso` capability is the entry point to a region.
+  Whichever running behaviour holds an `iso` reference to a region has the guarantee that no other running behaviour has any access to that region.
+  To assign an `iso` reference to another variable *moves* it, removing it from the source and transfering the ownership of the target region.
+* Mutable (`mut`).
+  Write access to data that is guaranteed to be dominated by an `iso` region.
+  Behaviours can freely write to `mut` pointers because they have the guarantee that they can only access them if they hold the corresponding `iso` pointer to their region.
+  Every `mut` pointer is associated with an `iso` pointer whose existence guarantees that the region is safe to access.
+  A `mut` pointer is invalidated once the type system can no longer guarantee that the associated `iso` pointer remains valid.
+* Read-only (`readonly`).
+  Behaviour-internal references to objects providing local immutability (ie. restrict write access) through this reference.
+  If the object is mutable then other `mut` pointers in the same behaviour may still modify the object.
+  When accessing an object via a `readonly` pointer, any read of a field declared `mut` or `iso` will return a `readonly` pointer: you may not use any pointer derived from a `readonly` pointer to modify any reachable object.
+* Immutable (`imm`).
+  Global immutability, guarantees that no behaviour will be able to write through references to the same objects.
+  Isolated regions can be entirely frozen into immutable regions.
+  Mutable references are frozen together with the region but can't themselves be frozen directly.
+  After freezing, only `imm` pointers can be created to any object from the region that was frozen.
+  If the frozen region contained objects with `iso` fields referring to child regions then the child regions are frozen at the same time.
+  When accessing an object via an `imm` pointer, any read of a field declared `mut` or `iso` will return an `imm` pointer: after freezing there is no distinction between objects that were allocated in the same or different regions.
+  `imm` pointers can be shared across behaviours because their globally immutable nature guarantees data-race freedom.
+
+## Regions
+
+Asynchronous tasks in Verona are performed by behaviours (`when`) on cowns, which own a tree of regions.
+The language semantics guarantee (via `iso` pointers) that, at any time, only a single behaviour has access to a region.
+
+Regions are portions of (not necessarily contiguous) memory dominated by an isolated (`iso`) pointer, with any number of mutable (`mut`) pointers in the same region, accessible only through the `iso` pointer.
+Regions can be implemented with different memory management strategies or execution models.
+
+## Assignment
+
+Capabilities change the semantics of assignments in the language.
+
+**Sentinel objects** in a region may have only a single pointer to them from outside of the region and that must be an `iso` pointer.
+The uniqueness guarantee of the `iso` pointer means that any read of an `iso` value that gives an `iso` capability is destructive.
+Verona pointers are not nullable and so an `iso` field must always hold a valid `iso` capability.
+Reading an `iso` field can produce a `mut` pointer bound to that `iso` but extracting the `iso` capability requires a swap operation that installs a new value at the same time as removing the old one.
+For stack variables, extracting the `iso` capability simply ends the scope of the variable: the variable no longer exists after the read, just as if the block in which it is declared had ended.
+
+```ts
+var foo : Foo & iso = Foo; // plus undecided syntax for "in new region type X"
+foo.doSomething();
+...
+var bar = foo; // bar now holds the *only* reference to foo
+bar.somethingElse(); // OK: bar is a reference to the same object as foo
+foo.cleanup(); // ERROR: foo is not a valid reference any more
+```
+
+Verona does not allow uninitialised values to be created or read, there is no such thing as a `nullptr`.
+Objects must be created with all values initialised and cannot lose their values as a result of an `iso` move semantics.
+Therefore, the example above would be invalid if `foo` was an object's field.
+
+```ts
+class Something { foo : Foo & iso; ... }
+...
+let some = Something; // Initialises `foo` with an `iso` reference
+...
+var bar = some.foo; // ERROR: This would `move` the `iso` into `bar` and let `foo` with an invalid value
+...
+doSomething(some); // What if this function does something with `foo`?
+```
+
+**Objects within a region** can hold `mut` pointers to each other, including aliased pointers to the same object.
+Reading a `mut` field creates a copy of the capability, which can be stored on the stack for as long as the type system can guarantee that the `iso` capability conveying the right to access the region is live and which may be stored into any suitably typed field of any object in the same region.
+Given that mutable references can be modified only by the one behaviour that holds their dominating `iso` reference, there can be no data races in any write or reassignment.
+
+```ts
+var foo : Foo & mut = Foo::create(10); // Returns a Foo object with value = 10
+print(foo.value); // prints "10"
+...
+// Binding a new variable to the same reference creates a new reference to the same object
+var bar = foo;
+bar.value = 42;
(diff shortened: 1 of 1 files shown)


docs/internal/capabilities.md
```diff
@@ -0,0 +1,210 @@
...
+```ts
+var foo : Foo & mut = Foo::create(10); // Returns a Foo object with value = 10
+print(foo.value); // prints "10"
+...
+var bar : Foo & readonly = foo; // OK: a read-only reference to a mutable object
+...
+var baz : Foo & mut = bar; // ERROR: cannot convert readonly to mutable
+```
+
+**Immutable objects** can also alias each other, but since none of those references can write through them, once created (frozen), an immutable region can only be read from.
```
davidchisnall, line 95:
```suggestion
**Immutable capabilities enforce global immutability**.
The `freeze` operation consumes an `iso` capability.
The type system therefore guarantees that there can be no pointers to objects in the region (or any child regions) anywhere other than as pointers within the regions being frozen.
The `freeze` operation returns an `imm` capability and *viewpoint adaptation* guarantees that this cannot be used to materialise any kind of mutable capability.
Once a region has been frozen, no capabilities exist that permit writes to any objects that were in that region.
```

docs/internal/capabilities.md
```diff
@@ -0,0 +1,210 @@
...
+...
+var ro : Foo & readonly = bar; // OK: imm is read-only anyway, so a noop
+var imm : Foo & imm = ro; // OK? also a noop but there's a sub-type relationship that I'm not sure is there
+```
+
+### Destructive reads
+
+Some reads invalidate the previous reference on assignment.
+By design, `iso` references have a destructive read, since the program cannot have two `iso` references to the same region.
+But behaviour captures also destroy the previous reference to avoid race conditions.
```
davidchisnall, line 112:
```suggestion
But behaviour captures also destroy the previous reference to avoid data races.
```

docs/internal/capabilities.md
```diff
@@ -0,0 +1,210 @@
...
+a : Sub = Sub;
+function(a); // OK, as Sub is a sub-type of Super
+```
+
+The notation is `Sub <: Super` that means `Sub` is a sub-type if `Super`.
+
+### Allowed Conversions
+
+These are the following capability sub-type relationships in Verona:
+* `iso <: mut`: Passing an `iso` in a (behaviour-local) place expecting a `mut` will _view_ the type as a mutable reference, because `iso` is strictly stringer than `mut`.
```
davidchisnall, line 156:
```suggestion
* `iso <: mut`: An `iso` capability grants the rights to access a sentinel object and to create `mut` capabilities to any object in that region (i.e. that are reachable by following only `mut` pointers from the sentinel object) that are valid for as long as you hold the `iso` capability.
  As such, an `iso` capability trivially authorises constructing a `mut` capability to the object (this is the degenerate case: follow 0 `mut` pointers from the sentinel).
  As with any other `mut` capability, this may be stored only in places where the `iso` capability can be statically proven to be held: in fields of objects in the same region or in lexically-scoped variables dominated by a variable holding the `iso`.
  For example, if one stack frame holds an `iso` capability conferring exclusive ownership of a region then a function called from that stack frame may take a `mut` view of the sentinel object as an argument.
```

docs/internal/capabilities.md
```diff
@@ -0,0 +1,210 @@
...
+* `mut <: readonly`: Passing a `mut` in a place expecting a `readonly` will _view_ the type as read-only and not allow writes to it.
```
davidchisnall, line 157:
```suggestion
* `mut <: readonly`: Passing a `mut` in a place expecting a `readonly` will _view_ the type as read-only and not allow writes to it.
  The `readonly` capability has the same lifetime restrictions as the `mut` capability from which it was derived.
```

docs/internal/capabilities.md
```diff
@@ -0,0 +1,210 @@
...
+* `imm <: readonly`:  Similar reasoning, but here it's a no-op, because the object was globally immutable to begin with.
```
davidchisnall, line 158:
```suggestion
* `imm <: readonly`:  An `imm` capability guarantees global immutability, `readonly` guarantees local immutability which is a strictly weaker property that is trivially satisfied by the stronger property.
```

I don't like the 'no-op' framing because that implies that the other conversions are non-trivial operations.

docs/internal/capabilities.md
```diff
@@ -0,0 +1,210 @@
...
+
+When the conversion between capabilities require more than just a cast, we must call an explicit builtin to perform that functionality.
```
davidchisnall, line 160:
```suggestion
Capability systems require that rights associated with a capability are never elevated except by interacting with a component that already owns these rights.
Any change to a capability involves an upgrade of rights or exchanging the existing rights for a mutually exclusive set then this must be an explicit operation.
```

docs/internal/capabilities.md
```diff
@@ -0,0 +1,210 @@
...
+
+* `iso -> imm` via `freeze`: Converts an `iso` region into an `imm` region by freezing it.
+The region will become globally immutable and cannot be unfrozen later, to avoid race conditions.
```
davidchisnall, line 163:
```suggestion
The region will become globally immutable and cannot be unfrozen later, to guarantee data-race freedom.
```

davidchisnall (commented):
@sylvanc, please can you review my review and make sure that we agree on terminology?

docs/internal/capabilities.md
```diff
@@ -0,0 +1,210 @@
...
+* Arena model: Objects are allocated and used, but only freed on the destruction of the region itself.
+
+They can also refer to how the code is executed:
+* Native: Verona code calling Verona code, fully checked and safe.
+* Sandbox: Verona code calling an external language module (ex. a C++ library), with process isolation.
+
+Sandboxes themselves can also be of different types, for example:
+* Child process: Forks a new process, shared memory, calls through RPC.
+* WebASM: Isolates OS functionality, memory access, etc.
+* Remote execution: For use with different technologies (ex. GPU) or architectures (ex. connected embedded board).
```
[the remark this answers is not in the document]
rengolin: I can remove this for now.

docs/internal/capabilities.md
```diff
@@ -0,0 +1,210 @@
...
+
+Capabilities change the semantics of assignments in the language.
+
+**Isolated objects** can only be used through a single reference, so assigning an `iso` region to a new variable moves the reference and invalidates the previous one.
+
+```ts
+var foo : Foo & iso = Foo; // plus undecided syntax for "in new region type X"
+foo.doSomething();
+...
+var bar = foo; // bar now holds the *only* reference to foo
```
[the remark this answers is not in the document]
rengolin: Not sure what the example would be, as I'm not entirely clear the semantics of that issue. Happy to take suggestions.

docs/internal/capabilities.md
```diff
@@ -0,0 +1,210 @@
...
+bar.somethingElse(); // OK: bar is a reference to the same object as foo
+foo.cleanup(); // ERROR: foo is not a valid reference any more
+```
+
+**Mutable objects** can alias each other, so assigning a `mut` variable to another creates a (different) reference to the same variable.
+Given that mutable references can only be modified by the one behaviour that holds their dominating `iso` reference, there are no race conditions in any write or reassignment.
+
+```ts
+var foo : Foo & mut = Foo::create(10); // Returns a Foo object with value = 10
```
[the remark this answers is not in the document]
rengolin: Without a syntax for creating regions, it's really hard to write examples.

rengolin (commented):
Thanks David! I've applied the non-controversial ones (typos, rewrites) and left the others to discuss more.

docs/internal/capabilities.md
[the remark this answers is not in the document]
davidchisnall: We have syntax for creating regions: `new` and functions that wrap `new` such as `create`.  These always return an object in a new region.  We don't have syntax for specifying any constraints on the region type.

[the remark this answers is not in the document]
rengolin: Right, I've been using `create` (assuming they'd wrap `new`) as object creation in a new region and `create ... in` for new object on an existing region. I don't think that's a reasonable syntax but I don't have a better one.

In our first MLIR dialect we had a distinction between `create region` and `create object`, which meant respectively: new object in a new region and new object in an existing region.

But none of that takes into account stack objects. If we want to create an object on the stack and `create` calls `new`, then do we need to create another method to do the same thing but not call `new`?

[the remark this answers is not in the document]
davidchisnall: I think objects created on the stack are in the abstract machine as objects created in a new region and then merged with the stack-frame region.  In the implementation, this does one of two things:

 - Allocates them with a heap allocator that cleans up on exit from the stack.
 - Promotes them to the stack.

We don't explicitly allocate objects on the stack, we allocate objects whose lifetimes are lexically scoped and let the compiler figure it out.

docs/internal/capabilities.md
```diff
@@ -0,0 +1,210 @@
...
+```ts
+var foo : Foo & iso = Foo; // plus undecided syntax for "in new region type X"
+foo.doSomething();
+...
+var bar = foo; // bar now holds the *only* reference to foo
+bar.somethingElse(); // OK: bar is a reference to the same object as foo
+foo.cleanup(); // ERROR: foo is not a valid reference any more
+```
+
+**Mutable objects** can alias each other, so assigning a `mut` variable to another creates a (different) reference to the same variable.
```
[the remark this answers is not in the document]
rengolin: This is where it gets fuzzy for me, because the stack has a different lifetime than regions.

There are two issues here:
1. Compile time guarantees by the type system: if the object is never accessed after it's invalidated (for whatever reason), the code is "correct" and regardless of how we implement that at run-time, it's fine.
2. Run-time semantics: if the compiler said no one access the object after it's been made invalid, then stack lifetime is irrelevant, ie. all stack frames can have the "arena" model.

What cannot happen is functions trying to access objects in other frames that they don't own. Previous frames could be fine because their lifetime is strictly longer than the current's, but it'd be up for the compiler to prove that the access is correct via its own region's lifetime (which I'm not sure it's always possible).

docs/internal/capabilities.md
rengolin: Should we explain or link to a page explaining "viewpoint adaptation"? The first hits on search engines are all about computer vision and object detection, which can be super hard to filter.

Also, the Pony tutorial talks about viewpoint adaptation, but the Pony types are so much more complex that I'm not sure how we'd relate in the same way.

docs/internal/capabilities.md
rengolin: Some of this is redundant with the description of `iso` and `sentinel objects` above. Perhaps we could elide some of the repetition here to just talk about the sub-type relationship?

rengolin (commented):
Merged a few more that looked obvious, and have some comments on the remaining ones.

@sylvanc, would be good if you had a look at this, to make sure we're all on the same page.

docs/internal/capabilities.md
```diff
@@ -0,0 +1,218 @@
...
+In Verona, pointers are capabilities: the ability to name an object confers the right to access the object in some way.
+Capabilities confer a range of different permissions, which are reflected in the type system.
+
+Variables represent pointers to objects and they must have rights associated that describes what you can do to the objects via those pointers.
+
+There are kinds of rights that can be granted to capabilities:
+* Isolated (`iso`).
+  Defines the entry point to a region.
+  Whichever behaviour holds a reference to a region has the guarantee that no other behaviour has access to the same region.
+  To assign an `iso` reference to another variable is to transfer the ownership of that region.
```
nwf, line 15:
```suggestion
  Indicates that a capability is *the globally unique* pointer to its target.
  Every `iso` capability is the entry point to a region.
  Whichever running behaviour holds an `iso` reference to a region has the guarantee that no other running behaviour has any access to that region.
  To assign an `iso` reference to another variable *moves* it, removing it from the source and transfering the ownership of the target region.
```

docs/internal/capabilities.md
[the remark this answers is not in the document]
rengolin: Do you mean that if a field is an `iso` and you assign to some other variable, the field will be left `null` and therefore not safe?

Like:
```
class Foo { field : Class & iso; create() { ... } };
...
let foo : Foo & mut = Foo; // now field has an isolated reference
let bar : Class & iso = foo.field; // ERROR: field cannot be nulled
```

davidchisnall (approved):

PR #462, opened 2021-07-15
Adding generics document

rengolin:
This one is simpler but I may have missed a lot of content that Verona has more elaborate than Pony.

Again, happy to take in large suggestions.

### DIFF
--- a/docs/internal/generics.md
+++ b/docs/internal/generics.md
@@ -1,0 +1,304 @@
+# Generics
+
+## WARNING
+
+This is a preliminary document under heavy discussion.
+The ideas here are mostly right, the details, not necessarily.
+There will be a tutorial document that will be correct in all counts, but this isn't it.
+
+## Overview
+
+Generics are a way to use type parameters to define or restrict other types.
+For example, a generic class can use its type parameters to define the type of fields, and a generic method can use its parameters to restrict what can be passed as an argument.
+
+Example:
+```ts
+class MyGenericType[MyType, SomeInterface]
+{
+  // MyType is the type of this field
+  var field: MyType;
+
+  // SomeInterface is the super type of anything that can be passed to this method
+  method(arg: SomeInterface)
+}
+
+// Some concrete types that implements Hashable
+class MyClass: Hashable { ... }
+class MyOtherClass: Hashable { ... }
+
+// This creates a concrete type
+let x = MyGenericType[U32, Hashable]::create();
+
+x.field = 42; // OK: The type of the field (U32) is compatible with integral literals
+
+let c = MyClass::create();
+x.method(c); // OK: MyClass implements Hashable
+let o = MyOtherClass::create();
+x.method(o); // OK: MyOtherClass also implements Hashable
+
+// This creates another concrete type, now with a more restrictive argument
+let y = MyGenericType[F64, MyClass]::create();
+
+y.field = "hi"; // ERROR: No known conversion between string literals and float types
+
+y.method(c); // OK: This is the expected type
+y.method(o); // ERROR: Even though MyOtherType implements the same interface, it's not the same type
+```
+
+In the example above, there is one generic class called `MyGenericClass` and two implementations (or specialisations) with types `[U32, Hashable]` and `[F64, MyClass]`.
+If all types are used correctly (ex. removing the ERROR lines above), the compiler will create two concrete classes from the one generic class: one for each specialisation.
+
+## Type Parameters
+
+Type parameters are the named types that can be used inside a generic declaration in place of actual types.
+In the example above, the generic class `MyGenericClass` has two type parameters: `MyType` and `Hashable`.
+Those names are used inside the class in place of existing (declared) types.
+In each specialisation, the type arguments (ex. `[U32, Hashable]`) replace the named arguments and the specialisation types are matched 1:1.
+
+Type parameters have the following properties:
+* A name: The name that will be used in the generic class as a substitute for the specialisation's passed declared name.
+* An optional super-type restriction: What requirements (e.g., interfaces or capabilities) must hold of the type being passed in each specialisation.
+* An optional default type: In case the type is omitted from the specialisation declaration, a default type is assumed.
+  If no default is provided, a type must be provided at specialisation.
+
+The syntax is: `[Name: Restriction = Default Value]`.
+
+On the example above, only the names were provided.
+Even though the names were `MyType` and `SomeInterface`, they could have been initialised with anything, included concrete types, interfaces and capabilities.
+It's up to the compiler's type checks to make sure the types passed by each specialisation make sense.
+For example, a field with just a capability type cannot be properly initialised, so that's a type error.
+
+Example declaration:
+```ts
+// A can be any type with any capability
+// B can be any type, but must have a mut capability
+// C can also be any type, but if none is passed, it's U32
+class Generic[A, B: mut, C = U32] { ... }
+```
+Example uses:
+```ts
+// A is unrestricted, and B requires it to be mut, while C has a default value
+let a = Generic[MyClass & mut, OtherClass & mut]::create();
+
+// Specifying C works too
+let b = Generic[MyClass & mut, OtherClass & mut, SomeOther & imm]::create();
+
+// ERROR: B has no default value
+let error = Generic[MyClass]::create();
+
+// ERROR: B has to be mut
+let error = Generic[MyClass, Otherclass & iso]::create();
+```
+
+In the example above, `C = U32` is a potentially misleading declaration.
+Having a concrete default type on a unrestricted generic type may lead the implementation to assume users will only pass numeric types.
+
+To correct that mistake, a complete declaration would help:
+```ts
+type Unsigned = (U8 | U16 | U32 | U64 | U128) & imm;
+class Generic[A, B: mut, C: Unsigned & imm = U32 & imm] { ... }
+```
+
+However, this is still misleading because `C` could be a union of any combination of the allowed types, for example, `(U8 | U64)` and methods like `+` wouldn't apply, because neither `U8`'s nor `U64`'s `+` method returns `(U8 | U64)`.
+
+To fix that, we can use interfaces to mean _any class that provides those characteristics or behaviours_.
+If we have an interface `Numeric` that all numeric types (integral and floating point) implement, and another (ex. `Integer`) that only integral ones do, and another (ex. `Unsigned`) that only unsigned integral types implement, then any type passed, including union types, will have to implement the interface and therefore the call will be a dynamic dispatch and we don't need a `match` on the receiver.
+
+```ts
+interface Numeric { }
+interface Integer: Numeric { }
+interface Unsigned: Integer { }
+...
+class U8: Unsigned { ... }
+...
+class U64: Unsigned { ... }
+...
+class Generic[A, B: mut, C: Unsigned = U32 & imm] { ... }
+```
+
+Now, in the `Generic` class above, the parameter `C` can only be one concrete type that implements `Unsigned` (and also `Integer` and also `Numeric`).
+
+Another way to fix this is to use `Self` types, and it shows how powerful the Verona type system is.
+If we create an interface that restricts what the `create` method can return to a `Self` type, then it restricts to only concrete types, as those types don't return union types in their constructors.
+
+```ts
+// Restricts constructors to only those that return the self concrete type
+interface OneOf
+{
+  create(): Self;
+}
+type OneUnsigned = OneOf & Unsigned;
+...
+class Generic[A, B: mut, C: OneUnsigned = U32 & imm] { ... }
+```
+
+This works because no `Unsigned` concrete class implements `create` methods which return type is a union, so `C` can only be a single concrete type.
+
+It does, however, have a limitation: The `create` method has no parameters.
+If the concrete type you want to use passes arguments to the constructor, it won't implement the `OneOf` interface.
+
+### Type Parameter Values
+
+It is still under discussion how we're going to implement values in type parameters.
+
+For example:
+```ts
+// A list of a particular type and a compile-time constant size (ex. a vector)
+class Vector[T, let size: U32]
+{
+  ...
+}
+
+// A dynamicaly allocated matrix of any size
+class Matrix[Type] { ... }
+
+// A function that takes in only lists of a particular type and size
+convolution3x3(image: Matrix[F64], filter: Vector[F64, 9]): Matrix[F64] { ... }
+
+// Initialise a matrix and a filter
+let matrix = Matrix[F64]::create(...);
+let vector = Vector[F64, 3]::create(1, 2, 3);
+let vector2 = Vector(1.0, 2.0, 3.0); // equivalent
+
+// ERROR: Types are different because the compile-time sizes are not the same
+convolution3x3(matrix, vector); // vector size is 3, expected size is 9
+```
+
+## Generic Methods
+
+Methods can declare additional type parameters to further specialise functionality.
+These are independent from the class specialisation and must be passed every time the method is called.
+
+Example:
+```ts
+class Generic[A, B: mut, C: Unsigned = U32 & imm]
+{
+  ...
+  // Converts some Unsigned to D
+  to[D: Unsigned](arg: C): D { ... }
+}
+...
+// Create a specialisation with default C = U32
+let obj = Generic[SomeType & imm, OtherType & mut]::create();
+
+let x: U8 & imm = 42;
+let y: (U8 | U16) & imm = ...;
+
+let z: U32 & imm = obj.to[U32](x); // OK: x is U8, return is U32
+
+let z: U16 & imm = obj.to[U8](x); // ERROR: return not U16
+
+let z: U32 & imm = obj.to(x); // OK: x is U8, return type inferred as U32
+
+let z: U32 & imm = obj.to(y); // ERROR: y is U8 | U16, not U8
+```
+
+## Specialisation
+
+Type parameters make classes generic to share implementation for a range of different types ([parametric polymorphism](https://en.wikipedia.org/wiki/Parametric_polymorphism)).
+But concrete classes need monomorphic types to be defined at compile time, so that we can have a concrete implementation of a parametric type.
+This conversion is called [reification](https://en.wikipedia.org/wiki/Reification_(computer_science)) and is done via a reachability analysis of all possible definitions of each polymorphic type.
+
+Example:
+```ts
+class Generic[A] { field: A & imm; ... };
+...
+// Here, `A` is `U32`
+let foo: Generic[U32] = ...;
+
+// Here, `A` is `MyClass`
+let bar: Generic[MyClass] = ...;
+```
+
+In the example above, there is one polymorphic class, `Generic[A]` where `A` is the polymorphic type parameter, and two concrete classes (specialisations): `Generic[U32]` and `Generic[MyClass]`.
+
+The compiler generates the required specialisations and replaces each definition with its respective concrete classes.
+Without loadable code (separate compilation processes), the compiler then removes the generic class.
+Since generic classes can only be used by specialisation, all other (potentially infinite) possibilities of types parameters don't exist in the program, so the generic class is elided and we're left with only the specialisations.
+The effect of loadable code is still uncertain in this design and under discussion.
+
+The end result would be:
+```ts
+// Definition of monomorphic classes only
+class Generic_U32 { field: U32 & imm; ... }
+class Generic_MyClass { field: MyClass & imm; ... }
+
+// Direct use of the concrete classes
+let foo : Generic_U32 = ...;
+let bar : Generic_MyClass = ...;
+```
+
+Note that type unions (`A | B`) are still monomorphic types.
+To resolve which implementation works with each inner type (`A` or `B`) one must use `match` inside the method.
+
+Example:
+```ts
+interface Addable[A]
+{
+  // Adds itself to other returning the same type of other
+  // All numeric types implement this interface
+  +(self, other: A): A;
+}
+class Generic[A] {
+  field: A & imm;
+  // Add matches on the implementation of a + method, via Addable
+  add(other: A & imm) {
+    match self.field
+    { x: Addable[A] => x + other }
+    { x: String => x.append(other) } // Some other notion of "adding"
+    { _ => throw Something } // Or some better error handling
+  }
+}
+...
+// This creates a monomorphic union type
+let foo: Generic[U32 | String] = ...;
+
+// This creates a monomorphic single type
+let bar: Generic[U32] = ...;
+
+// This will do dynamic dispatch of `field.+(other)`
+let baz: Generic[U32 | F64] = ...;
+```
+
+On the first case, `foo` is a (monomorphic) union between `U32` and `String`, which means, at run-time, either can be passed to `add` and we can't decide at compile time.
+So the implementation retains the `match` and will check if the types implement `Addable` (both do).
+Note that both `field` and `other` must have the same type (both are `A`).
+
+However, on the second case, `bar` is a compile-time single monomorphic type, which guarantees that it will always implement the `Numeric` interface and therefore there's only one `match` statement that fits.
+In this case, the compiler is free to elide the `match` altogether and emit only the case where the type is `Addable`, ie. `self.field + other`.
+In this case, the type is guaranteed to be `U32` and the call to `+` is a static call to `U32::+(U32, U32)`.
+
+Finally, in the last case, `baz`'s types are both `Addable`, but the `+` operator for each is different.
+When searching for which operator to call on `field + other`, the types aren't guaranteed to be the same, so `field` can be `U32` while `other` `F64`.
+In fact, because there is no restriction (`match`) for `other`, it's still `U32 | F64`, and there are no `+` methods on either `U32` or `F64` with the signature `+(Self & imm, (U32 | F64) & imm) : Self & imm`, so this is a type error.
+
+The correct part of the code above results in the following compile-time concrete types:
+```ts
+// (U32 | String) need different initialisation and add match
+class Generic_U32_String {
+  field: (U32 | String) & imm;
+  // Generic add
+  add(other : (U32 | String) & imm) {
+    // Different behaviour for each type
+    match self.field
+    { x: U32 => U32::+(field, other) }
+    { x: String => String::append(x, other) }
+    // All cases exhausted, no need for error handling
+  }
+}
+
+// Just U32 elides everything
+class Generic_U32 {
+  field : U32 & imm;
+  // Specific add
+  add(other : U32 & imm) {
+    U32::+(field, other)
+  }
+}
+
+// Then replace the definitions with the monomorphic types
+let foo : Generic_U32_String = ...;
+let bar : Generic_U32 = ...;
+```
+
+The compiler output will not contain any functions or data structures for the unspecialised generic polymorphic type `Generic[A]`.


docs/internal/generics.md
```diff
@@ -0,0 +1,114 @@
+# Generics
+
+Generics are a way to use type parameters to define or restrict other types.
+For example, a generic class can use its type parameters to define the type of fields, and a generic method can use its parameters to restrict what can be passed as an argument.
```
davidchisnall, line 4:
```suggestion
For example, a generic class can use its type parameters to define the type of fields, and a generic method can use its parameters to provide specialisations for different argument types.
```

I'm not really happy with this suggested change, but non-generic types also restrict what can be passed as an argument so it's difficult to understand what generics add with that phrasing.

docs/internal/generics.md
```diff
@@ -0,0 +1,114 @@
...
+
+Example:
+```ts
+class MyGenericType[MyType, SomeInterface]
+{
+  // MyType is the type of this field
+  field : MyType;
```
davidchisnall, line 11:
```suggestion
  var field : MyType;
```

I was debating whether to add an `& mut` in this suggestion.  If we don't add it, then the generic parameter must contain the capability types as well as the concrete type, if we do then the parameter can't specify something different.

docs/internal/generics.md
```diff
@@ -0,0 +1,114 @@
...
+
+let c = MyClass;
+x.method(c); // OK: MyClass implements Hashable
+let o = MyOtherClass;
+x.method(o); // OK: MyOtherClass also implements Hashable
+
+// This creates another concrete type, now with a more restrictive argument
+let y : MyGenericType[F64, MyClass] = MyGenericType;
+
+y.field = 42; // ERROR: No known conversion between integer literals and float types
```
davidchisnall, line 34: I missed this in the expressions document.  Is this actually true?  I'd expect integer literals to be able to match with any numeric type that has sufficient range to represent them.  That said, C programmer have probably been trained to write 42.0...

docs/internal/generics.md
```diff
@@ -0,0 +1,114 @@
...
+y.method(c); // OK: This is the expected type
+y.method(o); // ERROR: Even though MyOtherType implements the same interface, it's not he same type
+```
+
+In the example above, there is one generic class called `MyGenericClass` and two implementations (or specialisations) with types `[U32, Hashable]` and `[F64, MyClass]`.
+If all types are used correctly (ex. removing the ERROR lines above), the compiler will create two concrete classes from the one generic class: one for each specialisation.
+
+## Type Parameters
+
+Type parameters are the named types that can be used inside a generic declaration in place of actual types.
```
davidchisnall, line 45: Do we allow any kind of value parameters to generics?  

The simplest use for simple value-dependent types is to specify fixed-size arrays.  For example, `Array[U32, 10]` takes on generic parameter that is a type and one that is a value.  

C++ has always allowed templates to take primitive value types as parameters and with C++20 allows class literals to be template parameters (with the type defined by the type of the class and the byte pattern that represents a specific instance).  I don't think we'd want to take the same route as C++20 because it composes in very exciting ways with out desire to have arbitrary compile-time evaluation.  

You can trivially encode integers as church numerals to map from the value space to the type space, though it's incredibly clunky.  For example:

```verona
class Zero {}
class N[T] {}

// If programmers ever have to write this we have done something wrong!
var arr : Array[U32, N[N[N[N[N[N[N[N[N[N[Zero]]]]]]]]]]];
```

docs/internal/generics.md
```diff
@@ -0,0 +1,114 @@
...
+
+Type parameters have the following properties:
+* A name: The name that will be used in the generic class as a substitute for the specialisation's passed declared name.
+* A restriction: What are the requirements (ex. interfaces, super-types, capabilities) of the type being passed in each specialisation.
+* A default type: In case the type is omitted from the specialisation declaration, a default type is assumed.
+
+The syntax is: `[Name : Restriction = Default Value]`.
+
+On the example above, only the names were provided.
+Even though the names were `MyType` and `SomeInterface`, they could have been initialised with anything, included concrete types, interfaces and capabilities.
```
davidchisnall, line 58: An example here might be nice, for example:
```verona
class List[T, Cap : iso | imm = imm]
{
  var val : T & cap;
  ...
};
```

docs/internal/generics.md
```diff
@@ -0,0 +1,114 @@
...
+It's up to the compiler's type checks to make sure the types passed by each specialisation make sense.
+For example, a field with just a capability type cannot be properly initialised, so that's a type error.
```
davidchisnall, line 60: Is it a type error to specify it, or a type error to use it?  If it's an error to specify it then how do I express my above example where I want a list of some interface or class type and will provide a default capability but not a default value type?

docs/internal/generics.md
```diff
@@ -0,0 +1,114 @@
...
+
+Example:
+```ts
+// A can be any type with any capability
+// B can be any type, but must have a mut capability
+// C can also be any type, but if none is passed, it's U32
+class Generic[A, B : mut, C = U32] { ... }
+```
+
+In the example above, `C` is a dangerous declaration.
```
davidchisnall, line 70:
```suggestion
In the example above, `C` is a potentially misleading declaration.
```

docs/internal/generics.md
```diff
@@ -0,0 +1,114 @@
...
+Having a concrete default type on a unrestricted generic type may lead the implementation to assume users will only pass numeric types.
+
+To correct that mistake, a complete declaration would help:
+```ts
+class Generic[A, B : mut, C : (U8 | U16 | U32 | U64) & imm = U32 & imm] { ... }
```
davidchisnall, line 75: Or perhaps:

```suggestion
class Generic[A, B : mut, C : Numeric & imm = U32 & imm] { ... }
```

Where `Numeric` is an interface that defines the operations on numbers.  Users can then instantiate it with whatever numeric type they want.

In your version, is `Generic[Foo & imm, Bar & imm, (U8 | U16) & imm]` valid?  I believe it is, but the way it's written makes it look as if you expect a single concrete type.  Do we have syntax for expressing the constraint that it must be a concrete type?  Without that, I don't believe we can use `embed` on a generic parameter (what does it mean to embed an interface type?).

docs/internal/generics.md
```diff
@@ -0,0 +1,114 @@
...
+```
+
+Now we can see that the programmer only expected unsigned integers, which was even more restricted than previously thought.
+To make this easier to read, one can use type aliases:
+
+```ts
+type UInts = (U8 | U16 | U32 | U64) & imm;
+...
+class Generic[A, B : mut, C : UInts = U32 & imm] { ... }
+```
```
davidchisnall, line 85: Same subtyping question as above.  I don't think this means what you think it means.  In the trivial case, it would be allowable to instantiate this with `UInts` as the generic parameter and now you have dynamic dispatch in a place you presumably don't want it (you wouldn't restrict yourself to fixed-size integers if you didn't want arithmetic to be cheap and / or storage space to be small and well-defined).

davidchisnall (commented):
A few comments.  It would probably be good to have a short explanation of reification somewhere and, in particular, how you can implement custom specialisation for generic methods and methods of generic classes by using a `match` expression and relying on the compiler eliding all of the unmatched values for any given reification.

docs/internal/generics.md
rengolin: I personally think it's fine to require `42.0` here, or `U64(42).toF64()`.

docs/internal/generics.md
rengolin: Good point, I don't know how to restrict types to be concrete.

For the `Numeric` proposal, my idea was to show that type aliases can be quite useful here. Happy to take a better example.

docs/internal/generics.md
rengolin: I'm personally not against values as template parameters, but I'm not sure what types would they be and how would the type system cope with that: naturally or artificially through special cases.

docs/internal/generics.md
rengolin: There is an example below the next paragraph. I could add this one too.

docs/internal/generics.md
rengolin: It's a type error if the compiler has no way to infer it, regardless if you use it or not.

If you don't specify, but the compiler "knows what you mean", it's fine. If the imprecise declaration is ambiguous, then it's not fine.

For example:
```ts
main() : U32 {
  let x : imm;
  0
}
```

It doesn't matter if `x` is used or not, that's a type error because I have no idea what immutable thing we're talking about. Not to mention it's not properly initialised...

docs/internal/generics.md
rengolin: Is there a way to say: "only one of those" rather than "any combination of those"? Do we even want that?

I imagine @sylvanc would say: Well, if programmer pass a union type in the end, they'll get dynamic dispatch and it's their problem. If they want concrete behaviour they should have passed concrete types to begin with.

docs/internal/generics.md
rengolin: Ok, I think I've addressed this in my new change.

docs/internal/generics.md
nwf: Welcome to "dependent typing", where types can bind values; it's definitely the strangest dimension of the lambda cube.  I don't have an answer, but just some background (you've given me an excuse and I enjoy thinking about these things more than I would enjoy munging C++ at the moment)...

A Haskeller's view of, say, `MyClass[5]` is that that `5` is now a _type_, unrelated to any value `5` (in particular the value `5` does not have type `5`; in fact the type `5` has no values); one can introduce a notion of "kinds" of types and think of that type `5` as having _kind_ `int` and the _type_ constructor `MyClass` as having kind `int -> *`, where `*` is the kind of types that have values (there's also the _value_ constructor `MyClass` that takes values and builds a `MyClass`-typed value).  Because the type-level ints and value-level ints are different objects, while we can have a polymorphic "unlift" value which is the value `n` when instantiated at type `n`, its type (`forall _ : int . int`) would also be inhabited by, say, a constant `0` value, which might be dissatisfying.  On the other hand, because we've bolted type-level ints on to the side, the flavor of the type system is not hugely changed, and, notably, does not necessarily require a subtyping system, which may keep type elaboration/inference (and checking) tractable.

If we want to keep the notion of types as sets of values, we could employ singleton types, where the value `5` has type `5` (sometimes written `Sing 5` or such to make it clearer whether we mean the value or the type).  The type `U32` might then be something like an alias for the union `Sing 0 | Sing 1 | Sing 2 | ...` or might be another axiomatic type (that is, not defined algebraically, but presumably then taken to have a supertype relationship with `Sing 5`).  Value-level operators naturally lift across `Sing` and type algebra (e.g., `(Sing 0 | Sing 1) + (Sing 2) = (Sing 2 | Sing 3)`).  The analogous "unlift" polymorphic value could now be typed as `forall _ : Sing k . k`, which admits exactly the term we want and no other, which may be more satisfying.  On the other hand, this approach threatens to require a fully set-theoretic subtyping relationship, which in turn might mean that type inference is intractable.

Whether you think the "type ints aren't value ints" or "type ints are the singletons of value ints" approach is cleaner is probably mostly subjective or might depend on whether you believe that PLs are for programming computers and math is that neat thing over there or are for doing mathematics and computers are a fun afterthought -- or, at least, I'm not a sufficiently well-versed PL theorist to have a strong reason to prefer one to the other.  Both approaches run into challenges with increasing expressivity if the compiler is also required to terminate, so, probably, the question itself is undecidable; blame Turing and Gödel.

docs/internal/generics.md
```diff
@@ -0,0 +1,114 @@
...
+## Type Parameters
+
+Type parameters are the named types that can be used inside a generic declaration in place of actual types.
+In the example above, the generic class `MyGenericClass` has two type parameters: `MyType` and `Hashable`.
+Those names are used inside the class in place of existing (declared) types.
+In each specialisation, the type arguments (ex. `[U32, Hashable]`) replace the named arguments and the specialisation types are matched 1:1.
+
+Type parameters have the following properties:
+* A name: The name that will be used in the generic class as a substitute for the specialisation's passed declared name.
+* A restriction: What are the requirements (ex. interfaces, super-types, capabilities) of the type being passed in each specialisation.
```
nwf, line 60:
```suggestion
* An optional super-type restriction: What requirements (e.g., interfaces or capabilities) must hold of the type being passed in each specialisation.
```

Giving a concrete type as the super-type of a type parameter seems not very useful to me: concrete types don't have a subtyping relationship between them, right?  Is there a case where constraining a type parameter to be the union of concrete types, rather than an interface, would be useful?  (Could and would we do closed-world reasoning in that case?)

ETA: see some of the discussion below, too.

docs/internal/generics.md
```diff
@@ -0,0 +1,114 @@
...
+* A default type: In case the type is omitted from the specialisation declaration, a default type is assumed.
```
nwf, line 61:
```suggestion
* An optional default type: In case the type is omitted from the specialisation declaration, a default type is assumed.
  If no default is provided, a type must be provided at specialisation.
```

docs/internal/generics.md
rengolin: @nwf, thanks for the digression, that was most interesting!

I'm still unsure what the benefits of value parameters have with a compiler that can already do whole program analysis and create specialisations even when the user didn't request.

For example, what's the difference between:
```ts
// Imagine this is the syntax for value parameters
class List[Ty, Len = U32] {
  // Whatever Verona equivalent to: alignas(Ty) char data[Len * sizeof(Ty)];
  data : Ty[Len];
  // Constructor
  create() {
    data = new Ty[Len]; // whatever the syntax here... (**)
  }
};
...
// Creates a list with a data section size = 5 * sizeof(MyClass)
let array = List[MyClass, 5];
```

and

```ts
class List[Ty] {
  // This is now an array of unknown size until initialisation
  data : Ty[];
  // Constructor
  create(Len : U32) {
    data = new Ty[Len]; // whatever the syntax here... (**)
  }
};
...
// Creates a list with a data section size = 5 * sizeof(MyClass)
let array = List[MyClass](5);
```

Unless the initialisation syntax (**) is different for declared size vs undeclared size, these two are the same.

If whole-program reachability is allowed to create specialisations of known patterns and it recognises that both cases are the same at compile time, we can emit the exact same code, probably inlined, even if each went through a different concrete (anonymous) specialisation type in the higher level representation.

Am I missing anything?

nwf: It's possible that some (much?) of the _optimization_ opportunities nominally afforded by dependent typing could be realized instead by whole-program compilation, as you say.  That in and of itself is probably an interesting research topic.

But, to the example at hand, a variable whose type is `List[MyClass,5]` guarantees that it will _always_ have five elements, no matter how it is used in the source, while a variable of type `List[MyClass]` _might_ happen to always have five elements.  Whether or not the latter property could be discovered by the compiler and used for generating code, the former conveys something much more precise about programmer intent.

rengolin:
>It would probably be good to have a short explanation of reification somewhere and, in particular, how you can implement custom specialisation for generic methods and methods of generic classes by using a `match` expression and relying on the compiler eliding all of the unmatched values for any given reification.

I have added a new section called `Specialisation` which covers that. Let me know what you think. Thanks!

docs/internal/generics.md
davidchisnall: Think about the simplest case: fixed-size arrays.  Imagine two possible generic types for arrays:

```verona
class Array[T] {...}
class Array[T, Size : U32] {...}
```

In the first case, size is fixed at creation time, but I have no way of specifying in a function signature that I want an array of a fixed size.  I then need to add dynamic checks in my function and the caller needs to check the docs.  The compiler can optimise away the checks if it can do sufficient dataflow propagation, but consider something like this:

```verona
var x = Array[U32](1, 2, 3);
when (whatever)
{
  whatever.foo(x);
}
```

Now it's very difficult for the compiler to infer that the size that was set in one behaviour is guaranteed in the other.  First we need to inline `Array[U32]::Create` then we need to extract the information that a `let` field has a known value, then we need to propagate that to the function implementing the `when` clause.  Reachability helps a bit here, but it's a dataflow problem not a control flow problem so reachability is not a complete solution.

Now imagine that the `foo()` method was declared as:

```
foo(arr : Array[U32, 4] & readonly) : U32;
```

Now we know at type-checking time that this is an error, because I'm passing an `Array[U32,3]` to a function that expects an `Array[U32,4]`.  Once I correct the caller to add the missing element, the compiler doesn't need to do anything to propagate the bounds: they're part of the type.

docs/internal/generics.md
```diff
@@ -156,3 +156,110 @@ let z : U32 & imm = obj.convert(y); // ERROR: Not able to define what type y is
...
+let foo : Generic[U32] = ...;
+
+// Here, `A` is `MyClass`
+let bar : Generic[MyClass] = ...;
+```
+
+In the example above, there is one polymorphic class, `Generic[A]` where `A` is the polymorphic type parameter, and two concrete classes (specialisations): `Generic[U32]` and `Generic[MyClass]`.
+
+The compiler generates the required specialisations and replaces each definition with its respective concrete classes, then removes the generic class.
+Since generic classes can only be used by specialisation, all other (potentially infinite) possibilities of types parameters don't exist in the program, so the generic class is elided and we're left with only the specialisations.
```
davidchisnall, line 177: Note that this is not always going to be guaranteed by the language.  

Once we add support for loadable code, generics that are exposed across a program / plugin boundary will not be fully reified and will end up doing some dynamic dispatch.  

Before then, the compiler is always free to merge similar instantiations of generics and add some dynamic dispatch back.

docs/internal/generics.md
```diff
@@ -156,3 +156,110 @@ let z : U32 & imm = obj.convert(y); // ERROR: Not able to define what type y is
...
+let foo : Generic[U32 | String] = ...;
+
+// This creates a monomorphic single type
+let bar : Generic[U32] = ...;
+
+// This union of Numeric types isn't quite static
+let baz : Generic[U32 | F64] = ...;
+```
+
+On the first case, `foo` is a monomorphic union between `U32` and `String`, which means, at run-time, either can be passed to `add` and we can't decide at compile time.
```
davidchisnall, line 216: 'monomorphic union' feels like an oxymoron to me, but that might just be me.

docs/internal/generics.md
```diff
@@ -156,3 +156,110 @@ let z : U32 & imm = obj.convert(y); // ERROR: Not able to define what type y is
...
+So the implementation retains the `match` and will check if the types are `Numeric` or `String`.
+Note that both `field` and `other` must have the same type (both are `A`).
+
+However, on the second case, `bar` is a compile-time single monomorphic type, which guarantees that it will always implement the `Numeric` interface and therefore there's only one `match` statement that fits.
+In this case, the compiler is free to elide the `match` altogether and emit only the case where the type is `Numeric`, ie. `field + other`.
+In this case, the type is guaranteed to be `U32` to the call to `+` is a static call to `U32::+(U32, U32)`.
+
+Finally, in the last case, `baz`'s types are both `Numeric`, but the `+` operator for each is different.
+So, while the compiler is free to elide all other `match` cases, the dispatch of the `+` method is dynamic, ie. `field.+(other)`.
```
davidchisnall, line 225: I don't think this type checks.  `U32::+(U32 & imm) : U32 & imm` and `F64::+(F64 & imm) : F64 & imm` do not both satisfy the same interface.  The field is allowed to be a `U32` and the argument an `F64` (or vice versa) and the addition would then fail.

docs/internal/generics.md
```diff
@@ -156,3 +156,110 @@ let z : U32 & imm = obj.convert(y); // ERROR: Not able to define what type y is
...
+
+The code above results in the following compile-time concrete types:
+```ts
+// (U32 | String) need different initialisation and add match
+class Generic_U32_String {
+  field : (U32 | String) & imm;
+  // Implement constructor with match on types
+  add(other : (U32 | String) & imm) {
+    match (typeof(field))
+    { U32 => U32::+(field, other) }
```
davidchisnall, line 235: Nope, type checking error here too: `U32::+` is not defined to take a `(U32 | String) & imm` as the second argument.  I'm not sure about the `match` syntax: do we have flow sensitivity here, or do we need to provide a different variable name for `field` as a `U32`?

docs/internal/generics.md
```diff
@@ -156,3 +156,110 @@ let z : U32 & imm = obj.convert(y); // ERROR: Not able to define what type y is
...
+    { String => String::append(field, other) }
```
davidchisnall, line 236: `String::append` might be defined to allow a `U32` to be appended, but it probably isn't what you wanted to do here.

docs/internal/generics.md
```diff
@@ -156,3 +156,110 @@ let z : U32 & imm = obj.convert(y); // ERROR: Not able to define what type y is
...
+    U32::+(field, other)
+  }
+}
+
+// Both numeric, need dynamic dispatch
+class Generic_U32_F64 {
+  field : (U32 | F64) & imm;
+  // Implement constructor with match on types
+  add(other : (U32 | F64) & imm) {
+    field.+(other);
```
davidchisnall, line 255: Nope, this `+` doesn't exist in either of the concrete types.

docs/internal/generics.md
```diff
@@ -156,3 +156,110 @@ let z : U32 & imm = obj.convert(y); // ERROR: Not able to define what type y is
...
+  }
+}
+
+// Then replace the definitions with the monomorphic types
+let foo : Generic_U32_String = ...;
+let bar : Generic_U32 = ...;
+let baz : Generic_U32_F64 = ...;
+```
+
+In the final program, the generic polymorphic type `Generic[A]` doesn't exist anymore.
```
davidchisnall, line 265:
```suggestion
The compiler output will not contain anything functions or data structures for the unspecialised generic polymorphic type `Generic[A]`.
```

docs/internal/generics.md
rengolin: Right, but the main use case (a big example is LLVM's `SmallVector`) is a `List` that is at least 5, with higher cost for more.

What it does is allocate statically a list of N elements up front but allow it to grow if needed.

This avoid the high costs of using `std::vector` for "small" lists with power-of-two reallocation in the case:
```c++
std::vector<int> foo;
foo.push_back(1);
foo.push_back(2);
foo.push_back(3);
foo.push_back(4);
````
Which will go from 0->1->2->4 (worst case, 3 reallocs).

In C++, even if we use `resize`, the main difference is:
* `SmallVector[Ty, N]()` -> `char data[N * sizeof(Ty)] (constexpr)`
* `std::vector[Ty]().resize(N)` -> `char* data = malloc(N * sizeof(Ty)) + free()` (even if N is constexpr)

`SmallVector` allocates more space if you go past its limit (with new/delete) and check the index to retrieve/update values on either storage.

My point is that **/maybe/** all "interesting" cases to use values as template parameters could be reached by compiler analysis (not optimisation) via constant propagation, reachability, etc. Or, put another way, the uncommon / uninteresting cases can suffer some run-time penalties to avoid having to add values to type parameters.

I'm honestly just guessing, so feel free to correct me any time. :)

File: docs/internal/generics.md (hunk 12 above)
Comment (rengolin) [reply docs/internal/generics.md:177]: In my view, such a generic interface will be in a monomorphic (union) type with a `match` statement, so "fully reified".

File: docs/internal/generics.md (hunk 13 above)
Comment (rengolin) [reply docs/internal/generics.md:216]: Now, it does to me too, but I wanted to be clear. It is not obvious to people that haven't been looking at type systems for the past few years, as it certainly was't clear to me when I first started working on Verona.

File: docs/internal/generics.md (hunk 14 above)
Comment (rengolin) [reply docs/internal/generics.md:225]: What about `Self` types?

```ts
interface Numeric {
  +(Self & imm) : Self & imm;
};
```

Wouldn't `Self` be `U32` in one case and `F64` in the other, thus both satisfying the interface?

docs/internal/generics.md
davidchisnall: The main advantage is nothing to do with optimisation, it's to do with writing correct code:

In some cases, you want to write a function that takes a collection of *n* things, where *n* is an arbitrary number.  In other cases, you want to write a function that takes a collection of *n* things where *n* is a fixed value.

In C++, `std::vector` provides a convenient way (well, *a* way) of handling the first, `std::array` provides a convenient way of handling the second.  If I want to write a function that takes a 3x3 convolution matrix, for example, then it's *really* helpful if I can specify that I require a 9-element array in the signature of a function.  If the only tool that I have is a fixed-but-unknown-at-compile-time array, then I can't express this and I'm reliant on complex analyses to (hopefully) give me an error.

The fact that the compiler can now more easily elide bounds checks is a nice bonus but it's not the main motivation.

docs/internal/generics.md
rengolin: I'm not sure about the syntax either.

You're right, `typeof(field)` doesn't guarantee `typeof(other)` as it did when it matched the type parameter.

Can I match two conditions at the same type?

```ts
match (typeof(field) && typeof(other))
{ U32 && U32 => field + other }
...
```

or do I need to nest?

```ts
match (typeof(field))
{ U32 => match (typeof(other)) { U32 => field + other }
}
...
```

docs/internal/generics.md
rengolin: See my `Self` comment.

In cases like this, where both types are numeric and both have `+` implemented (on their own types), how can I make it so that I don't need to match against each? Or is that impossible?

docs/internal/generics.md
rengolin: Just to be clear, I wasn't advocating the performance/optimisation side only, though I understand why both of you thought so. I was just wondering if the compiler could *always* infer correctly without the need of value parameters.

I understand the difference between `List[I32, N]` and `List[I32](N)` in your examples and I agree they're fundamentally different for the type system. I'm not discussing that.

The main point I was missing is the function signature. I can easily create types that have static sizes, but I can't easily *require* types with a specific compile-time size if I can't specify its size at compile-time in a function signature.

docs/internal/generics.md
rengolin: put between parenthesis to make it clear that it's redundant.

docs/internal/generics.md
davidchisnall: This is not possible with separate compilation.

docs/internal/generics.md
davidchisnall: That's a fine definition for the interface, but it isn't used by the call in your `add` because it takes only a single argument.  The receiver is a concrete type that is one out of `U32` and `F64`.  The other argument is *also* a concrete type that is one out of `U32` and `F64` but there is no requirement that it be the same one.  Your `Numeric` should be defined something like this:

```verona
interface Numeric
{
  +(Self & imm, Self & imm) : Self & imm;
}
```

But this doesn't work because `Self` is a concrete type that is either `U32` or `F64` but the second argument is `U32 | F64`.  You could make `add` work by writing this:

```verona
interface Numeric[T]
{
  +(Self & imm, T & imm) : Self & imm;
}
```

But then neither `U32` nor `F64` would satisfy `Numeric[U32 | F64` because U32 neither implements a method that matches:

```verona
+(Self & imm, (U32 | F64) & imm) : Self & imm;
```

docs/internal/generics.md
davidchisnall: I think you'd need to do this with nested matches if you actually wanted to support that, but I don't think that anyone should ever write code like this.  The correct answer to trying to use `U32 | F64` with your generic is 'type error'.

docs/internal/generics.md
davidchisnall: If they both implement a `+` whose arguments are compatible with all of the types of the parameters that you pass, you don't need a `match`.  If they both implement `+` with argument types that are compatible with the concrete types that you have but not with the complete interface type that you're defining, then you need a match.

The problem here is that the concrete types *don't* implement a method that's compatible with your concrete types.  You example allows for `U32+F64` and `F64+U32` as the concrete types and neither is a defined operation.

docs/internal/generics.md
rengolin: I thought we wouldn't have separate compilation in Verona.

docs/internal/generics.md
rengolin: Ah, good point!

Furthermore, in the original generic implementation, the `Numeric` match won't catch `U32 | F64` either, so it would be throw.

What if we constrain that further:

```ts
    match (A)
    { Float => field + other }
    { Integer => field + other }
    { String => field.append(other) }
    { _ => Throw[None] } // Or some better error handling
```

Because one of the union types is `Integer` and the other is `Float`, there would only be one candidate for each dispatch, where `typeof(field) == typeof(other)`.

I'm not advocating to write that kind of code, which would fail if the union type was `U32 | U64`, but would that be a possible compiler transformation (to elide other cases and only lower the one match)?

docs/internal/generics.md
rengolin: Right, I'm convinced of that, too.

docs/internal/generics.md
rengolin: Right, perhaps I got the wrong types to be presenting this concept.

I didn't want to create completely new types that weren't familiar to readers because that's hard to connect.

I'll see if I can conjure up something better.

docs/internal/generics.md
rengolin: I've added an optional section on values as type parameters.

docs/internal/generics.md
rengolin: Marking this a type error with an explanation why.

docs/internal/generics.md
rengolin: Ok, I mention loadable code here so that there are no doubts. But loadable code is still not defined, so I want to keep the discussion of eliding the whole thing if static.

rengolin:
I believe all review comments were addressed, please check that the document is ready to merge. Thanks!

docs/internal/generics.md
```diff
@@ -0,0 +1,287 @@
...
+To make this easier to read, one can use type aliases:
+
+```ts
+type UInts = (U8 | U16 | U32 | U64) & imm;
+...
+class Generic[A, B : mut, C : UInts = U32 & imm] { ... }
+```
+
+However, this is still misleading because `C` could be a union of any combination of the allowed types, for example, `(U8 | U64)` and methods like `+` wouldn't apply.
+To fix that, we can use interfaces to mean _any class that provides this characteristics or behaviours_.
```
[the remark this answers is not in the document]
rengolin: I want it to mean the specific set of things required, so it should have been `these` not `this` but not `some` (arbitrary).

Makes sense?

docs/internal/generics.md
```diff
@@ -0,0 +1,287 @@
...
+If we have an interface `Numeric` that all numeric types (integral and floating point) implement, and another (ex. `Integer`) that only integral ones do, and another (ex. `Unsigned`) that only unsigned integral types implement, then it becomes clear that the code below only accepts one type:
```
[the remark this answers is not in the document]
rengolin: You are correct.  This was a misunderstanding of a previous case. What I should have said is something along the lines of:

`If we use an interface, then the types are guaranteed to implement that interface and the calls are dynamicaly dispatched, so they don't need a match on the receiver`.

Does that make more sense?

docs/internal/generics.md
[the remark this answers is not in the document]
nwf: Sounds good; quantifier scoping in English is a terrible catastrophe.

docs/internal/generics.md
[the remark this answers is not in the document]
nwf: Seems OK to me.

docs/internal/generics.md
```diff
@@ -0,0 +1,287 @@
...
+  // SomeInterface is the super type of anything that can be passed to this method
+  method(arg : SomeInterface)
+}
+
+// Some concrete types that implements Hashable
+class MyClass : Hashable { ... }
+class MyOtherClass : Hashable { ... }
+
+// This creates a concrete type
+let x : MyGenericType[U32, Hashable] & mut = MyGenericType;
```
sylvanc, line 30:
```suggestion
let x = MyGenericType[U32, Hashable]::create();
```

docs/internal/generics.md
```diff
@@ -0,0 +1,287 @@
...
+
+x.field = 42; // OK: The type of the field (U32) is compatible with integral literals
+
+let c = MyClass;
```
sylvanc, line 34:
```suggestion
let c = MyClass::create();
```

docs/internal/generics.md
```diff
@@ -0,0 +1,287 @@
...
+x.method(c); // OK: MyClass implements Hashable
+let o = MyOtherClass;
```
sylvanc, line 36:
```suggestion
let o = MyOtherClass::create();
```

docs/internal/generics.md
```diff
@@ -0,0 +1,287 @@
...
+x.method(o); // OK: MyOtherClass also implements Hashable
+
+// This creates another concrete type, now with a more restrictive argument
+let y : MyGenericType[F64, MyClass] = MyGenericType;
```
sylvanc, line 40:
```suggestion
let y = MyGenericType[F64, MyClass]::create();
```

docs/internal/generics.md
sylvanc: David is right.
```suggestion
y.field = "hi"; // ERROR: No known conversion between string literals and float types
```

docs/internal/generics.md
```diff
@@ -0,0 +1,287 @@
...
+```ts
+// A can be any type with any capability
+// B can be any type, but must have a mut capability
+// C can also be any type, but if none is passed, it's U32
+class Generic[A, B : mut, C = U32] { ... }
+```
+Example uses:
+```ts
+// A is unrestricted, and B requires it to be mut, while C has a default value
+let a : Generic[MyClass & mut, OtherClass & mut] = Generic;
```
sylvanc, line 81:
```suggestion
let a = Generic[MyClass & mut, OtherClass & mut]::create();
```

docs/internal/generics.md
```diff
@@ -0,0 +1,287 @@
...
+
+// Specifying C works too
+let b : Generic[MyClass & mut, OtherClass & mut, SomeOther & imm] = Generic;
```
sylvanc, line 84:
```suggestion
let b = Generic[MyClass & mut, OtherClass & mut, SomeOther & imm]::create();
```

docs/internal/generics.md
```diff
@@ -0,0 +1,287 @@
...
+
+// ERROR: B has no default value
+let error : Generic[MyClass] = Generic;
```
sylvanc, line 87:
```suggestion
let error = Generic[MyClass]::create();
```

docs/internal/generics.md
```diff
@@ -0,0 +1,287 @@
...
+
+// ERROR: B has to be mut
+let error : Generic[MyClass, Otherclass & iso] = Generic;
```
sylvanc, line 90:
```suggestion
let error = Generic[MyClass, Otherclass & iso]::create();
```

docs/internal/generics.md
```diff
@@ -0,0 +1,287 @@
...
+
+To correct that mistake, a complete declaration would help:
+```ts
+class Generic[A, B : mut, C : (U8 | U16 | U32 | U64) & imm = U32 & imm] { ... }
+```
+
+To make this easier to read, one can use type aliases:
+
+```ts
+type UInts = (U8 | U16 | U32 | U64) & imm;
```
sylvanc, line 104:
```suggestion
type Unsigned = (U8 | U16 | U32 | U64) & imm;
```

docs/internal/generics.md
```diff
@@ -0,0 +1,287 @@
...
+...
+class Generic[A, B : mut, C : UInts = U32 & imm] { ... }
```
sylvanc, line 106:
```suggestion
class Generic[A, B: mut, C: Unsigned = U32 & imm] { ... }
```

docs/internal/generics.md
```diff
@@ -0,0 +1,287 @@
...
+```
+
+Now, in the `Generic` class above, the parameter `C` can only be one concrete type that implements `Unsigned` (and also `Integer` and also `Numeric`).
+
+It is still under discussion if we'll allow values in type parameters, too.
+
+For example:
+```ts
+// A list of a particular type and a compile-time constant size (ex. a vector)
+class List[Type, Size : U32] { ... };
```
sylvanc, line 132:
```suggestion
class Vector[T, let size: U32]
{
  ...
}
```

docs/internal/generics.md
```diff
@@ -0,0 +1,287 @@
...
+
+// A dynamicaly allocated matrix of any size
+class Matrix[Type] { ... };
+
+// A function that takes in only lists of a particular type and size
+convolution3x3(image : Matrix[F64], filter : List[F64, 9]);
+
+// Initialise a matrix and a filter
+let matric : Matrix[F64] = ...;
```
sylvanc, line 141:
```suggestion
let matrix = Matrix[F64]::create(...);
```

docs/internal/generics.md
```diff
@@ -0,0 +1,287 @@
...
+let vector : List[F64, 3] = ...;
```
sylvanc, line 142:
```suggestion
let vector = Vector[F64, 3]::create(1, 2, 3);
let vector2 = Vector(1.0, 2.0, 3.0); // equivalent
```

docs/internal/generics.md
```diff
@@ -0,0 +1,287 @@
...
+For example:
+```ts
+// A list of a particular type and a compile-time constant size (ex. a vector)
+class List[Type, Size : U32] { ... };
+
+// A dynamicaly allocated matrix of any size
+class Matrix[Type] { ... };
+
+// A function that takes in only lists of a particular type and size
+convolution3x3(image : Matrix[F64], filter : List[F64, 9]);
```
sylvanc, line 138:
```suggestion
convolution3x3(image: Matrix[F64], filter: Vector[F64, 9]): Matrix[F64] { ... }
```

docs/internal/generics.md
```diff
@@ -0,0 +1,287 @@
...
+```
+
+## Generic Methods
+
+Methods can declare additional type parameters to further specialise functionality.
+These are independent from the class specialisation and must be passed every time the method is called.
+
+Example:
+```ts
+class Generic[A, B : mut, C : UInts = U32 & imm]
```
sylvanc, line 155:
```suggestion
class Generic[A, B: mut, C: Unsigned = U32 & imm]
```

docs/internal/generics.md
```diff
@@ -0,0 +1,287 @@
...
+{
+  ...
+  // Converts another uint to C
```
sylvanc, line 158:
```suggestion
  // Converts some Unsigned to D
```

docs/internal/generics.md
```diff
@@ -0,0 +1,287 @@
...
+  convert[D : UInt](arg : C) : D { ... }
```
sylvanc, line 159:
```suggestion
  to[D: Unsigned](arg: C): D { ... }
```

docs/internal/generics.md
```diff
@@ -0,0 +1,287 @@
...
+}
+...
+// Create a specialisation with default C = U32
+let obj = Generic[SomeType & imm, OtherType & mut] = Generic;
```
sylvanc, line 163:
```suggestion
let obj = Generic[SomeType & imm, OtherType & mut]::create();
```

docs/internal/generics.md
```diff
@@ -0,0 +1,287 @@
...
+
+let x : U8 & imm = 42;
+let y : (U8 | U16) & imm = ...;
+
+let z : U32 & imm = obj.convert[U8](x); // OK: x is U8, return is U32
```
sylvanc, line 168:
```suggestion
let z : U32 & imm = obj.to[U32](x); // OK: x is U8, return is U32
```

docs/internal/generics.md
```diff
@@ -0,0 +1,287 @@
...
+
+let z : U16 & imm = obj.convert[U8](x); // ERROR: return not U16
```
sylvanc, line 170:
```suggestion
let z : U16 & imm = obj.to[U8](x); // ERROR: return not U16
```

docs/internal/generics.md
```diff
@@ -0,0 +1,287 @@
...
+
+let z : U32 & imm = obj.convert(x); // OK: x is unambiguously defined as U8
```
sylvanc, line 172:
```suggestion
let z : U32 & imm = obj.to(x); // OK: x is U8, return type inferred as U32
```

docs/internal/generics.md
```diff
@@ -0,0 +1,287 @@
...
+
+let z : U32 & imm = obj.convert(y); // ERROR: Not able to define what type y is
```
sylvanc, line 174:
```suggestion
let z : U32 & imm = obj.to(y); // ERROR: y is U8 | U16, not U8
```

docs/internal/generics.md
```diff
@@ -0,0 +1,287 @@
...
+  }
+}
+...
+// This creates a monomorphic union type
+let foo : Generic[U32 | String] = ...;
+
+// This creates a monomorphic single type
+let bar : Generic[U32] = ...;
+
+// This union of Numeric types isn't quite static
```
sylvanc, line 234:
```suggestion
// This will do dynamic dispatch of `field.+(other)`
```

rengolin:
Squashed and merged after thorough review with Sylvan.

PR #465, opened 2021-07-23
Adding compiler steps document

rengolin:
This is more than a TODO list, it's a design document to make sure we're all in the same page as to what the passes are and what do we do in each stage.

I'm not expecting to be _right_ nor I'm expecting that everyone will _trust me_. I'm expecting people to have different views and promote discussion. But once we're in agreement, the document should reflect that agreement.

If later on we change our minds (amidst new evidence) we should update this document with a new PR.

### DIFF
--- a/docs/internal/compiler-steps.md
+++ b/docs/internal/compiler-steps.md
@@ -3,7 +3,7 @@
 ## Overview
 
 This is a preliminary study on each compiler step to produce executable objects from Verona code.
-The initial objective of the Verona compiler is to generate a single executable from a number of files, directories and existing foreing objects.
+The initial objective of the Verona compiler is to generate a single executable from a number of files, directories and existing foreign objects.
 Loadable code and separate compilation processes are not taken into account in this document but are under discussion and may be introduced in the future.
 
 ### Verona Sources
@@ -30,7 +30,7 @@ Interoperability code can be found at [`src/interop`](../../src/interop).
 
 The parser is responsible for:
 * Understanding how to navigate directory structures to include all reachable code
-* Recognising foreign calls and trigger parsing of foreing code
+* Recognising foreign calls and trigger parsing of foreign code
 * Check syntax and semantic correctness of all sources
 * Parse the code as one meta-module and emit A-normal form AST for the next stage
 
@@ -88,12 +88,21 @@ The MLIR code can be found at [`src/mlir`](../../src/mlir).
 
 ## LLVM
 
-This stage is mainly about managing the LLVM compiler to produce object code, but it will need creation of new passes and analises.
+This stage is mainly about managing the LLVM compiler to produce object code, but it will need creation of new passes and analyses.
 
-The Verona LLVM driver will need to:
+Library code will be compiled into LLVM IR and merged into the main IR module with the rest of the Verona code and compiler generated chunks, including:
+* The Verona runtime (scheduler, snmalloc, etc)
+* Builtin library written in C++
+* Sandbox RPC implementation
+
+In the future, we'll compartmentalise these chunks and run thin-LTO passes to clean up all of the unused code as well as inline and optimise it further.
+
+For now, we'll rely on existing passes as a single IR module.
+
+The Verona LLVM driver will then need to:
 * Create a new pass manager infrastructure
 * Create new passes to handle Verona-specific changes
-* Register the existing passes intercalated with our new passes
+* Register the existing passes interpolated with our new passes
 * Drive the code through the pipeline
 * Select the correct back-end options to emit appropriate object code
 
@@ -108,7 +117,7 @@ An initial take on which passes we'll need:
 The list and order of passes will probably change with time, but all of those above will need to be run one way or another.
 
 The back-end will need information from the environment and target flags:
-* Target triple defines the back-end to use as well as defaul options
+* Target triple defines the back-end to use as well as default options
 * Remaining target flags overwrite some defaults
 * Environment can fill other gaps when needed
 
@@ -127,19 +136,14 @@ The LLVM code can be found at [`src/mlir`](../../src/mlir).
 The final output of the compiler is a single object file containing the whole Verona meta-module and all its compile-time dependencies.
 Foreign code will be compiled separately in a shared object, including all its requirements (other libraries).
 
-In addition to those, Verona will have the following components to link:
-* The Verona runtime, declared at compile time but not implemented
-* Pre-compiled parts of the (builtin) standard library, if needed to be written in C++
-* Those components can be further optimised (inlined, cleaned up) by Link-Time optimisations
-
-The linker will then link all these objects together, doing the standard relocation fixups and cleanups, producing a final executable.
+The linker will then link all these objects together, doing the standard relocation fix-ups and cleanups, producing a final executable.
 
 The program will dynamically link all the necessary shared objects, including foreign code for sandboxing, at run time.
 
 The stages (and status) of the linker steps are:
 * (in progress) Verona runtime and snmalloc
 * (not started) Builtin library components in C++
-* (not started) Dynamic loading routines and sandbox code
+* (prototyping) Dynamic loading routines and sandbox code
 * (not started) Linking and producing an executable
 
 The runtime code can be found at [`src/rt`](../../src/rt).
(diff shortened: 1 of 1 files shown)


docs/internal/compiler-steps.md
```diff
@@ -0,0 +1,119 @@
...
+Loadable code and separate compilation processes are not taken into account in this document but are under discussion and may be introduced in the future.
+
+### Verona Sources
+
+The Verona compiler treats all files inside the same directory as a single module.
+Point the compiler to the directory and it will compile all (Verona) files together and include other modules (via `use` or `type`) in the same way.
+
+The end result is a single meta-module containing all modules: the main directory and all included modules from any module.
+This final module must contain a `Main` class with a `main(env : Env) : I32` method on any of the sub-modules, that will be the entry point for the program.
+Having zero or more than one `Main` classes is a compiler error.
```
davidchisnall, line 16: Not vitally important but it's nice if you can provide a compiler flag to specify a non-default main class, so that you can write stand-alone tests in a module and build an executable for each test by just telling the compiler that each test entry point is the one to use.

docs/internal/compiler-steps.md
```diff
@@ -0,0 +1,119 @@
...
+The final product of the parser is a fully checked concrete AST that the MLIR layer can consume as is (with potential additional helper structures) and generate plain MLIR without additional checks.
+
+The stages (and status) of the parser are:
+* (mostly done) Lex/Parse (+ syntax check)
+* (mostly done) Resolve symbols
+* (in progress) Type inference + Generics + Reification
+* (in progress) Reachability analysis & code elision
+* (mostly done) Emit A-normal form + Symbol table
+
+The parser code can be found at `src/parser`.
```
davidchisnall, line 45:
```suggestion
The parser code can be found at [`src/parser`](../../src/parser).
```

docs/internal/compiler-steps.md
```diff
@@ -0,0 +1,119 @@
...
+
+The MLIR stage takes in the AST and symbol table above and lowers to the following MLIR dialects:
+* Standard: Overall structure (functions, modules, etc) + existing operations
+* LLVM: Structures, memory management, remaining operations
+* Region (new): A dialect that tags objects into regions + creation & destruction semantics.
+
+With the (high-level) IR above, the following MLIR passes are executed:
+* Region and alias analysis
+* Heap to stack transformation
+* Foreign code RPC generation
```
davidchisnall, line 57: There are a few other things that we'd like to do here (not necessarily in MVP):

* Early inlining (so that we collapse user-defined control flow and elide lambdas early).  Inlining lambdas ideally happens while lambdas and their captures are still abstract and we don't have to go and re-infer things and then elide any lambda-specific memory-management operations.
* Post-inlining type inference, so we can refine the types after inlining.
* `match`-elision so that if we have more refined types we can remove match expressions that we now know are unreachable.

A couple more on the nice-to-have path:

- Union type representation selection (the later we do this, the smaller the set of types in the unions)
- Selector colouring (if we defer this until after we've done post-inlining type inference then we should end up with a smaller set to inlin.

docs/internal/compiler-steps.md
```diff
@@ -0,0 +1,119 @@
...
+
+The final lowering to LLVM IR is done via:
+* Partial lowering from the Region dialect to Std/LLVM dialects
+* Final lowering to LLVM dialect and LLVM IR
```
davidchisnall, line 61: Can we list the high-level IR features that we need to elide here?  For example, lowering an abstract method-invoke to a load from the type descriptor and an indirect call.

docs/internal/compiler-steps.md
```diff
@@ -0,0 +1,119 @@
...
+* (not started) Object code generation
+
+The LLVM code can be found at `src/mlir`.
+
+## Linker
+
+The final output of the compiler is a single object file containing the whole Verona meta-module and all its compile-time dependencies.
+
+In addition to those, Verona will have the following components to link:
+* The Verona runtime, declared at compile time but not implemented
```
davidchisnall, line 107: This should be linked as IR, not a .o file, so we can get inlining of the runtime into the rest of the system.

docs/internal/compiler-steps.md
```diff
@@ -0,0 +1,119 @@
...
+* Pre-compiled parts of the (builtin) standard library, if needed to be written in C++
```
davidchisnall, line 108: As should this.

docs/internal/compiler-steps.md
```diff
@@ -0,0 +1,119 @@
...
+* Any foreign module (archive) that need to be included in the final executable, along with dynamic loading routines for shared objects
```
davidchisnall, line 109: Foreign code will not be linked like this, it will be loaded by the sandboxing runtime (currently, it will be loaded in a child process as a .so), though that .so is generated by linking some veronac-generated wrappers with the .a / .so from the libraries used by that component.

docs/internal/compiler-steps.md
```diff
@@ -0,0 +1,119 @@
...
+
+The linker will then link all these objects together, doing the standard relocation fixups and cleanups, producing a final executable.
```
davidchisnall, line 111: The only things that the linker should be doing are:

 - Generating .so files for foreign code to be loaded in a sandbox.
 - Adding CSU files to the final build.

docs/internal/compiler-steps.md
rengolin: I think this is orthogonal. Whatever `Main` is, having zero or more than one is still a compiler error. That was the point I was trying to get across here.

docs/internal/compiler-steps.md
rengolin: > Early inlining (so that we collapse user-defined control flow and elide lambdas early).

I think part of this will be done by the parser (by reducing control flow to a small subset) and the lowering (by converting control flow into basic block flow.

> Post-inlining type inference

Currently, type inference is *completely* done at the parser. Translating the mechanism (or abstracting the API to be used in MLIR) may not be trivial. If the types are all concrete, then whatever is left can be handled by LLVM's own inlining later.

> Union type representation selection
> Selector colouring

Yes! I forgot about those ones. I'll add to the document. Though, selector colouring is currently being done in the parser, too.

docs/internal/compiler-steps.md
rengolin: That's done at the parser level. The AST will have two types of calls:
* A direct call, with a fully qualified name and arguments
* An indirect call, with a descriptor and an offset (vtable) and the arguments

Both types of calls are guaranteed to exist and argument/return to type check.

docs/internal/compiler-steps.md
rengolin: Both runtime and builtins are probably going to be large pieces, and if they're kept as IR, they're likely going to bloat the LLVM module, which has a single-threaded execution model. Bloating the IR and trying to do whole-program analysis can be disastrous for compile-time performance.

I imagine the runtime will have very clear and simple entry points (create region, schedule behaviour, garbage collect, etc) and so LTO would have an easy job at inlining them later on with greater ease and no bloat in the middle-end.

docs/internal/compiler-steps.md
rengolin: I meant the .a from those components. The .so will be loaded at runtime, so no need to mention here.

docs/internal/compiler-steps.md
rengolin: I forgot about the foreign SO generation. I think that's a separate linking process, called by the foreign interface and not part of the Verona code linking process.

docs/internal/compiler-steps.md
rengolin: Discussing this above

docs/internal/compiler-steps.md
rengolin: I've added a phrase to mention that.

docs/internal/compiler-steps.md
rengolin: Added the missing ones in the right places.

docs/internal/compiler-steps.md
rengolin: For the record, the `Region` dialect is something I just came up with now, as an example of a Verona-related-but-not-specific dialect. Since we can have as many dialects as we want, I think it makes more sense to have a `Region` dialect for the region analysis, and other dialects for different purposes than a `Verona` dialect for all those things.

Future dialects, once we move some of the logic into MLIR, would be for example `Type Algebra` (for type inference), `Call` dialect, for late dynamic vs static choice, along with selector colouring, etc. Those are interesting research projects on themselves, but shouldn't get in the way of reaching a functional compiler first.

docs/internal/compiler-steps.md
rengolin: Added a phrase to make explicit that the foreign code will be linked as an SO at run-time, not linking process.

docs/internal/compiler-steps.md
rengolin: Added sandbox code to the list here as well as above

docs/internal/compiler-steps.md
rengolin: Updated, mentioning thinLTO, dropped the link time libraries

PR #466, opened 2021-07-27
Adding module section, Self types

### LINKED ISSUE #349: Visibility specifications
We've had some internal discussions in the past about visibility and have a rough straw-robot proposal for wider discussion:

In terms of semantics, the proposal was to limit ourselves to public and package, for various reasons:

 - In the absence of subclassing, `protected` visibility does not mean anything.
 - A package will always be compiled together.  Anything that is broken by changes to internal implementation details can be refactored at the same time.  
 - Supporting only two kinds of visibility makes it very easy to decide which to use.

Open questions:

Can interfaces contain package-local methods?  If so, what are the restrictions on casting a concrete type to an interface that contains package-local methods (is the cast allowed only in the package that defines the class?  Is it allowed only if the interface is, itself, package-local?  Are there some more complex rules?).  If not, can generic methods be sufficiently expressive within a package for this not to be a problem?

What is the syntax for describing private types / methods?  Different case for the start (used by Go and, by convention, C#)?  Underscore at the start of package-local (informally used by Python and some C styles)?  Explicit `package` / `public` keywords (used by most Simula-like languages)?

rengolin:
Adding module section, talking about visibility. Fixes #349?

Also adding a short description on `Self` types.

Also fixing typo

### DIFF
--- a/docs/internal/types.md
+++ b/docs/internal/types.md
@@ -67,7 +67,84 @@ Most importantly, capabilities are not type traits like C++'s `const int * const
 
 ## Module / Class model
 
-TODO
+In Verona, modules are classes.
+Classes and functions inside modules are just nested types and methods of the module's class.
+
+Modules are composed of all Verona files in a specific directory.
+But there is no special relationship between modules in hierarchical directory structure.
+Sub-directories are not part of the module, nor visible in any special way to the module in the parent directory.
+
+For example, the following directory structure:
+```
+myproject +-- main.verona
+          |-- driver.verona
+          `-- packages +-- Module1 +-- foo.verona
+                       |           `-- bar.verona
+                       `-- Module2 +-- baz.verona
+                                   |-- ber.verona
+                                   `-- Module3 +-- mod3.verona
+```
+
+translates to the following module structure:
+```ts
+class myproject {
+  // both main and driver inserted here
+}
+class Module1 {
+  // both foo and bar inserted here
+}
+class Module2 {
+  // both baz and ber inserted here
+}
+class Module3 {
+  // mod3 inserted here
+}
+```
+
+The structure of modules in sub-directories should be used for organisational purposes and won't change the visibility of their definitions.
+
+The discovery path for modules is done in the following order:
+1. **Local path.**
+   For example, `using Module3` inside `Module2` will find it directly.
+2. **The project's `packages` directory.**
+   For example, `using Module2` inside `Module1` will find it via `packages`.
+3. **The compiler's standard library location.**
+   For example, `using numbers` inside any package will find the standard library implementation.
+   The location of the standard library is up to the compiler and irrelevant from the users' perspective.
+
+If a module cannot be found in any of these three locations, the compiler emits an error.
+
+The use of relative paths (example `using ../../Module1` inside `Module3`) is allowed.
+However, the package maintainer is responsible to making sure the structure holds on any user environment.
+This usually means you should not go beyond the boundaries of your own directory structure.
+Specifically, it is dangerous for `Module2`'s implementation to refer to `Module1` in that fashion, which could cause lookup failure on ill-formed directory structures.
+
+Mechanisms for importing packages using URIs (ex. `github:developer:project:version`) is under discussion but we have not finalised how it would look like.
+The consensus is to have a closed interpretation of how to handle each URI to guarantee integrity and security of the modules.
+New handlers should be incorporated into the compiler via pull requests, and not via third-party plugins, which could potentially be exploitable.
+
+### Visibility
+
+There are only two types of visibility: `public` and `module private`.
+Module private limits the visibility to classes and files on the same module, while public extends the visibility to all modules.
+
+Verona does not provide any equivalent of `private` or `protected` from C++, Java, and so on.
+Visibility is primarily a software-engineering abstraction to prevent implementation details leaking past abstraction boundaries.
+Verona treats a module as an abstraction and packaging boundary.
+
+Any form of protection within an abstraction boundary is purely advisory: the author of the module can trivially modify the code to override a specifier.
+As such, we do not believe that class-private has any value in Verona: any code that is able to modify an object's package-private fields is part of the same abstraction and distribution unit and so can be updated at the same time as any changes to that field's behaviour, without affecting any external APIs.
+
+Verona does not allow subtyping of concrete types and so has no need of a `protected` visibility specifier.
+
+The concept of visibility is different depending on what we refer to.
+
+If public:
+* Classes can be used as types by other modules.
+* Methods can be called by other modules.
+* Fields can be directly read and written to by other modules.
+
+Since sub-directories have no special relationship with their parent directories, `module private` does not permit modules in parent/child directory to see each others' private types.
 
 ## Classes and Interfaces
 
@@ -159,7 +236,7 @@ class Line
 Methods are functions that belong to a class or interface which can have arguments and a return type.
 
 Methods can be called statically or dynamically.
-Static calls are either directl (like `Class::method()`) or via an object (like `obj.method()`) if the compiler can infer the concrete type.
+Static calls are either directly (like `Class::method()`) or via an object (like `obj.method()`) if the compiler can infer the concrete type.
 Dynamic calls are either direct (via `Interface::method(object)`) or via an object (like `obj.method()`) otherwise.
 Methods that change the fields in the class must receive a `self : Self` argument and use as the field accessor (ex. `self.field`).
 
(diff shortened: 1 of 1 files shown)


docs/internal/types.md
```diff
@@ -382,3 +432,44 @@ x = ("hi", 42); // Type inference looks at `x`'s type for literals
...
+
+  // Implements the Addable interface
+  +lhs : MyClass & readonly, rhs : MyClass & readonly) : MyClass & imm
+  {
+    lhs.field + rhs.field
+  }
+}
+```
+
+In the example above, the `MyClass` argument type _matches_ the `Self` type in the interface declaration because it is the same type as the class declaring it.
```
davidchisnall, line 475: I think it would be useful to show why this is a problem without `Self`.  For example, if you defined:

```Verona
interface EqualComparable
{
  ==(EqualComparable & readonly, EqualComparable & readonly) : Bool;
}
```

Most classes would not conform to it because they can't be compared against *anything* that is equallity-comparable.  You'd need to define something like this:

```Verona
interface EqualComparable[T]
{
  ==(T & readonly, T & readonly) : Bool;
}
```

Except that this also isn't right: a concrete instance of the interface would define a single type that could be compared against itself and you want the interface to be generic over all concrete instances (`MyClass` would implement `EqualityComparable[MyClass]`, not `EqualityComparable` and if you have to specify the concrete type that implements the interface whenever you use the interface then there's no point in providing the interface).

docs/internal/types.md
```diff
@@ -67,7 +67,57 @@ Most importantly, capabilities are not type traits like C++'s `const int * const
 
 ## Module / Class model
 
-TODO
+In Verona, modules are classes.
+Classes and functions inside modules are just sub-types and methods of the module's class.
+
+However, unlike classes, modules cannot be created or destroyed, so they cannot be allocated in some region.
```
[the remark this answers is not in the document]
rengolin: I thought classes in modules would be instantiated, not modules themselves.

docs/internal/types.md
```diff
@@ -67,7 +67,57 @@ Most importantly, capabilities are not type traits like C++'s `const int * const
...
+Therefore, modules cannot have fields (which must be initialised upon construction), and so Verona has no global variables.
```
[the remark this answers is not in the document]
rengolin: If modules can be instantiated, then they should have the same semantics as classes. However, the root module (where Main is) is technically instantiated at startup and now we need a way to specify instantiation of its fields, and allow mutable fields that act like global variables.

Of course, due to regions, only a single behaviour can update those "global variables" at one given time, but now methods can have side-effects if they change global context without making it explicit.

I think the fact that I originally had a very weak understanding of this concept and that now I think I understand even less, is a testament that we need more discussions in this area.

docs/internal/types.md
rengolin: The first case obviously doesn't work, as you say.

The second case, as I understand, is not that _"there is no point"_, but that you'd have to use the interface as a generic, too, and you wouldn't be able to implement the method as the interface.

```ts
function(Interface[T]) : Bool {
  // I can't make it generic here because T has to be concrete of *any* type
}
...
class Foo : Interface[Foo] { ... }
class Bar : Interface[Bar] { ... }
foo = Foo...;
bar = Bar...;
...
function(foo) // matches function(Interface[Foo]);
function(bar) // matches function(Interface[Bar]);
```

Makes sense?

davidchisnall: Yes, you can't use the interface in its non-generic form in the second case, so it doesn't let you do the thing.

docs/internal/types.md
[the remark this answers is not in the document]
davidchisnall: They're just classes.  If the `Main` function is like a C++ static (doesn't have a parameter that is a pointer to the module) then it won't need (or be allowed to take) a pointer to an instance of the module as a parameter, so you won't need to instantiate the module to invoke `Main`.  The `Main` method may instantiate a module, but it may also instantiate any other class.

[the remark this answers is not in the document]
rengolin: ok, I have removed that entire paragraph.

docs/internal/types.md
```diff
@@ -473,3 +477,23 @@ class MyClass : EqualComparable, Addable
...
+
+Example:
+```ts
+interface Addable
+{
+  +(Addable & readonly, Addable & readonly) : Addable & imm;
+}
+class Foo : Addable { ... }
+class Bar : Addable { ... }
+class Baz : Addable { ... }
```
davidchisnall, line 491: Explicitly stating that you conform to an interface that you don't conform to is a bit confusing.  It would help if `Foo` implemented `+(Foo, Foo) : Foo` to show that this doesn't match `+(Addable, Addable) : Addable`.

docs/internal/types.md
```diff
@@ -473,3 +477,23 @@ class MyClass : EqualComparable, Addable
...
+
+function() {
+  foo = Foo;
+  bar = Bar;
+  // This matches the interface, but not Foo's (or Bar's) implementation
+  baz : Baz & imm = Addable::+(foo, bar);
```
davidchisnall, line 497: This seems to be calling a method that doesn't have an implementation.

docs/internal/types.md
```diff
@@ -473,3 +477,23 @@ class MyClass : EqualComparable, Addable
...
+{
+  +(Addable & readonly, Addable & readonly) : Addable & imm;
+}
+class Foo : Addable { ... }
+class Bar : Addable { ... }
+class Baz : Addable { ... }
+
+function() {
+  foo = Foo;
+  bar = Bar;
```
davidchisnall, line 495:
```suggestion
  var foo = Foo;
  var bar = Bar;
  // Type error.  Foo's + implementation takes parameters of type `Foo` not arbitrary addable things.
  var addable : Addable = foo;
```

docs/internal/types.md
rengolin: That was implied by the three dots, I didn't want to bloat the example. But I can add if you think that's clearer.

docs/internal/types.md
rengolin: I'm still a bit fuzzy on the syntax for calls through interfaces.

I didn't want to call `foo.+(bar)` or `foo + bar` because that has a lot of context that is irrelevant to the example.

Is there a clear way to show that this is the call that the interface mandates? Calling `Foo::+(foo, bar)` isn't really the same thing.

Perhaps I should just show two concrete classes, where one implements `+(One, One)` (which works) and the other `+(One, Two)` which also works but isn't what you meant.

rengolin: Ok, so now I've gone some implementations and the final example should be clearer.

docs/internal/types.md
```diff
@@ -486,14 +486,29 @@ interface Addable
 {
   +(Addable & readonly, Addable & readonly) : Addable & imm;
 }
-class Foo : Addable { ... }
-class Bar : Addable { ... }
-class Baz : Addable { ... }
+class Foo : Addable
+{
+  // Implements interface and operates on Foo only (as expected)
```
davidchisnall, line 491: It does not implement the interface.  This method does not have the signature from the interface.

docs/internal/types.md
```diff
@@ -486,14 +486,29 @@ interface Addable
...
+class Bar : Addable
+{
+  // Implements interface and operates on Bar only (as expected)
+  +(Bar & readonly, Bar & readonly) : Bar & imm { ... }
+}
+class Baz : Addable
+{
+  // Implements interface but operates on different types
+  // This is perfectly valid for the interface but it isn't what we want
+  +(Foo & readonly, Bar & readonly) : Baz & imm { ... }
```
davidchisnall, line 503: This also does not implement the interface.

docs/internal/types.md
```diff
@@ -486,14 +486,29 @@ interface Addable
...
 function() {
-  foo = Foo;
-  bar = Bar;
+  var foo = Foo;
+  var bar = Bar;
+  // Type error.  Foo's + implementation takes parameters of type `Foo` not arbitrary addable things.
+  var addable : Addable = foo;
   // This matches the interface, but not Foo's (or Bar's) implementation
-  baz : Baz & imm = Addable::+(foo, bar);
+  var baz : Baz & imm = foo + bar;
```
davidchisnall, line 512: This doesn't look like it works either.  You need to be calling `Baz::+`, but you're going to call `Foo::+`, which doesn't support these types.

rengolin: IIUC, this calls `Foo::+(foo, bar)` because the descriptor comes from the left-hand side, not the assignment type.

docs/internal/types.md
```diff
@@ -67,7 +67,64 @@ Most importantly, capabilities are not type traits like C++'s `const int * const
...
+class Module1 {
+  // both foo and bar inserted here
+}
+class Module2 {
+  // both baz and ber inserted here
+}
+class Module3 {
+  // mod3 inserted here
+}
+```
```
nwf, line 102: After today's deeply rabbit-holed meeting, I think it might be worth expanding the "inserted here"s with some example text and, perhaps, giving examples of paths that do and don't reach, from within each module, the other modules.  Examples like "Inside `Module2/baz.verona`, the name `"Module3"` finds `Module3` (under the `./` search path) and the name `"Module1"` finds `Module1` (under the `$project_root` search path) which is also found by the name `"../Module1"` (under the `./` search path)."

rengolin: I've updated the document with what seems to have been the consensus on the meeting today. Package discovery and relative paths (and its perils).

PR #467, opened 2021-07-28
Adding ABI document

rengolin:
This is a dump of three sources:
* What we discussed on Teams over time
* The [previous ABI document](https://github.com/microsoft/verona/pull/341)
* The [unmerged NaN-boxing document](https://github.com/microsoft/verona/blob/sylvanc-typecheck/docs/NaN-Boxing.md)

It's also a dump from *my* point of view. It is biased and it can definitely be wrong.

It's not as small as I expected, but at least I think it has all of the points I was confused about until now.

As with other documents, this is a starting point, not the end goal. It's a list of things we agree on, not points for further discussions.

We'll invariably have more discussions, and any change to the ABI should be an update to this document.

In addition, there are some "further optimisation" points on the things that can perform really badly, for example wide packing. I think it's an important part of this document, even if we only do that much later on. It keeps us focused on getting things right first without losing sight of the end goal, which is a high-performing language.


docs/internal/abi.md
```diff
@@ -0,0 +1,409 @@
...
+* Pointers to `cown`s are sealed with a type reserved for the runtime.
+  This ensures that `cown`s in immutable objects cannot be dereferenced by unsafe (sandboxed) code and provides some defense in depth against compiler bugs that would accidentally dereference a `cown` pointer.
+* Pointers to immutable objects are not represented by read-only capabilities.
+  Doing so would require expensive operations for reference-count manipulation.
+  When a pointer to an immutable object is passed via the foreign-code layer, it must have the load-storable-capabilities permission removed, such that no reachable capability will ever provide write permission.
+
+### Classes and interfaces
+
+Concrete classes and interfaces each have their own unique header, with a unique descriptor.
+Following a pointer and decoding the header gives you the type of the object and therefore there is no need to treat interfaces and classes differently.
```
rengolin, line 56: Are pointer to object going to be at the beginning of the object's data (so we read the header backwards) or to the header directly (so all element pointers are +2)?

docs/internal/abi.md
```diff
@@ -0,0 +1,409 @@
...
+
+Luckily, by using LLVM, we don't need to worry about any calling convention features that "just work" and only need to look at issues if they prove to be a case common enough to show up in profiles.
+
+This section discusses some of those opportunities.
+
+### Concrete Types
+
+FIXME: The wording below is confusing, please suggest a better one.
+
+Concrete types, including pointers, are always passed by value to functions.
```
rengolin, line 363: I think this paragraph is correct, but poorly worded. I may not be using the correct Verona terminology here.

docs/internal/abi.md
```diff
@@ -0,0 +1,437 @@
...
+Machine-word types are refereed to by value as an optimisation by the compiler.
+
+For that reason, they can only _contain_ the following representations:
+* Machine-word values.
+* Pointers to objects.
+
+To identify which type the run-time object has, we need a discriminator flag.
+Because the values take up all their storage, we need a wider storage to bundle more than one type in the same representation.
+
+Objects have a descriptor that identifies their types, so pointers don't need a special descriptor in the union representation.
```
davidchisnall, line 183:
```suggestion
Objects have a pointer to their type descriptor, which uniquely identifies their concrete types, so pointers implicitly carry a discriminator in the pointee when stored in the union representation.
```

docs/internal/abi.md
```diff
@@ -0,0 +1,437 @@
...
+So a union of pointers is represented as just a pointer and the discrimination will happen at the object level.
+However, numeric types don't have discriminators, so they need special handling when inside union.
```
davidchisnall, line 185:
```suggestion
In contrast, machine-word types don't contain anything other than their data and so have no internal state that can serve as a discriminator.
As such, they need special handling when inside union.
```

docs/internal/abi.md
```diff
@@ -0,0 +1,437 @@
...
+Furthermore, mixing pointers and numeric types creates the need to differentiate pointers from the rest.
+
+The type representation will be composed of a _descriptor_ and a _payload_ with the size of the largest object.
+There are two ways of storing the descriptor: `wide packing` and `NaN-boxing`.
```
davidchisnall, line 189:
```suggestion
The type representation will be composed of a _discriminator_ and a _payload_ with the size of the largest object.
There are two ways of storing the discriminator: `wide packing` and `NaN-boxing`.
```

Avoid confusion between a type descriptor (which includes the dispatch table and is all that's needed for any form of dynamic dispatch) and a type discriminator (which discriminates between a small fixed set of types).

docs/internal/abi.md
```diff
@@ -0,0 +1,437 @@
...
+63     62          51          0
+```
+So:
+* 1 sign bit
+* 11 exponent bits
+* 52 mantissa bits
+
+A `NaN` (not-a-number) is encoded in IEEE-754 as setting the exponent to _all ones_.
+All other bits are ignored.
+If we set the exponent to _all ones_, we can use the remaining 53 bits to encode a variety of smaller types.
```
davidchisnall, line 224: There's a little bit of subtlety here around qNaNs vs sNaNs, but it's probably fine for now.

docs/internal/abi.md
```diff
@@ -0,0 +1,437 @@
...
+[0][11111111111][0001][16*][32 data bits] = F32
+[0][11111111111][0010][40*][8 data bits] = I8
+[0][11111111111][0011][32*][16 data bits] = I16
+[0][11111111111][0100][16*][32 data bits] = I32
+[0][11111111111][0101][16*][32 data bits] = ISize (32)
+[0][11111111111][0110][40*][8 data bits] = U8
+[0][11111111111][0111][32*][16 data bits] = U16
+[0][11111111111][1000][16*][32 data bits] = U32
+[0][11111111111][1001][16*][32 data bits] = USize (32)
+```
```
davidchisnall, line 264: I wonder if it's useful to juggle this encoding a bit so that there's a single bit for signed vs unsigned?  If we sign- or zero-extend all signed / unsigned values in the payload, a cheap 'is this signed' check will let us use the same arithemetic paths for all signed / unsigned values.

docs/internal/abi.md
```diff
@@ -0,0 +1,437 @@
...
+
+To access the 7-th element, the unpacking does:
+```ts
+// Takes the 7th descriptor (in pseudo tuple notation starting at _0)
+descrip = Array[0]._6;
+
+// Takes the 7th element (at 8th position starting from 0)
+payload = Array[6+1];
+```
```
davidchisnall, line 372:
```suggestion

To facilitate vectorisation, it might be more efficient to store parallel arrays of values and arrays of discriminators.
It is possible to optimise this further for cases where the total number of machine-word types in the array is limited.
For example `Array[Foo | U64]` requires only one bit of discriminator state to identify the type.
This could be compressed by providing a two-element map (16 bits of state in the array object) that indexes from a dense type representation to a the wide encoding.
In this form, an `Array[Foo | U64]` would be encoded as a structure of the form:

```llvm
array = {{ ptr, ptr }, i1*, ptr*, i64, i8[2] }
```

The first two pointers are the object header, the next is a pointer to an array of one-bit discriminators for the different types, then a pointer to an array of payloads, followed by the length of the array and a map from the `i1` discriminators to the wide discriminators.
Accessing array element 4 would require the following steps:

 - Load the element at offset 3 in the payload array.
 - Load the bit at offset 3 in the discriminator array.
 - Load the real discriminator from the start of the discriminator map + the value loaded from the discriminator array.

In a loop, all of these operations have a trivial mapping from the loop induction variable and so are amenable to vectorisation.
```

Comment (davidchisnall) [reply docs/internal/abi.md:372]: Github doesn't do nice nesting of code things, annoyingly, so the above suggestion is mangled.

File: docs/internal/abi.md [hunk 9]
```diff
@@ -0,0 +1,437 @@
...
+Luckily, by using LLVM, we don't need to worry about any calling convention features that "just work" and only need to look at issues if they prove to be a case common enough to show up in profiles.
+
+This section discusses some of those opportunities.
+
+### Concrete Types
+
+FIXME: The wording below is confusing, please suggest a better one.
+
+Concrete types, including pointers, are always passed by value to functions.
+A machine-word numeric type will create a copy of the value to be used inside the function.
```
davidchisnall, line 392:
```suggestion
A machine-word numeric type will create a copy of the value to be used inside the function.
Machine-word types are immutable and so the fact that a copy is made is not visible in the abstract machine.
```

docs/internal/abi.md
rengolin: Right, so I'm assuming we don't care about that because we're ignoring it in the boxing and we don't want any traps or signals in the run time, so all NaNs are quiet?

docs/internal/abi.md
rengolin: Ah, yes, I thought of that, too. I think we can probably do it, but then bool and float would invariably have a sign, which would be weird.

docs/internal/abi.md
rengolin: That's a good idea, too, though I think the example is too confusing for this document. It would require a longer explanation. I think we can just leave the description and let the actual representation to be discussed later.

wrt formatting, I'll copy locally, check formatting and commit in separate.

docs/internal/abi.md
```diff
@@ -0,0 +1,444 @@
...
+    embed var b : F64 & imm;
+    var other : Other & mut;
+    create() { ... }
+}
+```
+
+A naive layout could be:
+```ts
+ { { ptr, ptr }, i32, f64,  ptr   }
+      header,    U32, F64, Other*
```
nwf, line 118:
```suggestion
      header,      a,   b, other*
```

`embed` doesn't seem to be doing anything here, tho'.

Do you mean for that F64 to be unnaturally aligned on 64-bit machines, are you suppressing padding, should the example say "on a 32-bit machine", or am I missing something?

docs/internal/abi.md
```diff
@@ -0,0 +1,444 @@
...
+ { { ptr, ptr }, f64,   ptr,  i32 }
+      header,    F64, Other*, U32
+```
+
+Because `f64` aligns stronger than `i32` and `ptr` can be either.
+On CHERI, the 128-bit `ptr` would be the first field after the header.
+
+### Dispatch Tables & Selector Colouring
+
+Object headers have a pointer to their type's dispatch table.
```
nwf, line 139:
```suggestion
Object headers have a pointer to their type descriptor, which includes a dispatch table.
```

docs/internal/abi.md
```diff
@@ -0,0 +1,444 @@
...
+A dispatch table contains pointers to the functions that the type provides, at specific offsets.
+Those pointers are called when it's not possible to determine the actual function being called at compile time.
+There is only one dispatch table per type, not per object.
+
+The offset of each method is calculated at compile time.
+The address at each offset has a pointer to the actual functions.
+Dynamic dispatch is done by taking the pointer at the offset from a table pointer and calling that.
+If an interface provides a method `foo`, all classes that implement that interface will have a method `foo` and their dispatch tables will have an entry for it.
+
+To avoid each type to have a different calculation for each method's offset, the compiler will do **selector colouring**.
```
nwf, line 149:
```suggestion
To avoid each type having a different calculation for each method's offset, the compiler will do **selector colouring**.
```

docs/internal/abi.md
```diff
@@ -0,0 +1,444 @@
...
+This process finds all common methods across all types and ensure the same methods all have the same offsets on all tables.
```
nwf, line 150:
```suggestion
This process finds all common methods across all types and ensures the same methods all have the same offsets on all tables.
```

docs/internal/abi.md
```diff
@@ -0,0 +1,444 @@
...
+This applies to concrete classes that implement specific interfaces, but it's not limited by it.
+Any two classes that have the same methods (signature) will end up with the same offset.
+
+With multiple interfaces and classes implementing similar methods, this can lead to a number of gaps between offsets.
+The tables will be compacted (either during colouring or afterwards) to minimise that.
```
nwf, line 155: By aliasing disjoint sets of selectors?

docs/internal/abi.md
```diff
@@ -0,0 +1,444 @@
...
+function(arg : U8 | F32) { ... }
+...
+var a : U8 & imm = 10;
+function(a); // Passes a 1-byte integer
+var b : F32 & imm = 3.14;
+function(b); // Passes 32-bit floating point
+```
+
+All objects are referred to by pointers.
+Machine-word types are refereed to by value as an optimisation by the compiler.
```
nwf, line 175:
```suggestion
Machine-word types are referred to by value as an optimisation by the compiler.
```

docs/internal/abi.md
```diff
@@ -0,0 +1,444 @@
...
+
+For that reason, they can only _contain_ the following representations:
```
nwf, line 177:
```suggestion
For that reason, union types can _contain_ only the following representations:
```

docs/internal/abi.md
```diff
@@ -0,0 +1,444 @@
...
+* Machine-word values.
+* Pointers to objects.
+
+To identify which type the run-time object has, we need a discriminator flag.
```
nwf, line 181:
```suggestion
To identify which type the run-time object has, we need a discriminator value (e.g., a flag or small bit field).
```

docs/internal/abi.md
```diff
@@ -0,0 +1,444 @@
...
+Because the values take up all their storage, we need a wider storage to bundle more than one type in the same representation.
```
nwf, line 182:
```suggestion
Because values generally take up all their storage, we usually need storage wider than the largest union member to bundle more than one type in the same representation.
```

docs/internal/abi.md
```diff
@@ -0,0 +1,444 @@
...
+[0][11111111111][0001][16*][32 data bits] = F32
+[0][11111111111][0010][40*][8 data bits] = I8
+[0][11111111111][0011][32*][16 data bits] = I16
+[0][11111111111][0100][16*][32 data bits] = I32
+[0][11111111111][0101][16*][32 data bits] = ISize (32)
+[0][11111111111][0110][40*][8 data bits] = U8
+[0][11111111111][0111][32*][16 data bits] = U16
+[0][11111111111][1000][16*][32 data bits] = U32
+[0][11111111111][1001][16*][32 data bits] = USize (32)
+```
```
nwf, line 266: Does Verona's `F64` type exclude `NaN` values?  (If not, presumably we need at least one "`NaN`-boxed" `NaN` to serve as that value, and whenever we commit a `F64` to the heap we need to canonicalize any possible FPU-generated `NaN`s to that/one of those `NaN`s?)

docs/internal/abi.md
```diff
@@ -0,0 +1,444 @@
...
+[0][**][1][0101] USize (on a 64 bit platform)
+```
+
+This bit pattern is also arbitrary and can change in the actual implementation.
+
+Note that, in 64-bit architectures, pointers (and `F64`) have the same width on both native and NaN-boxed representations.
+All other types increase the width to at least 64-bit, even if they were `U8` to begin with.
+
+Also note that the pattern above is an extra byte, so all values will have the whole 64/128 bits of the payload to be encoded.
+
```
nwf, line 302: When resorting to wide-boxing in the heap, we presumably pad to alignment (which, on CHERI, means unions are really uncomfortably big, with a 16-byte header... OTOH, for some small solace, there's more room for cap-packing non-pointer types)?  When upcasting to a union type as part of a function call, do we follow SROA and pass the discriminator and payload in potentially different registers if they fit?

nwf (commented):
Some nits, but looking good as far as I got!

docs/internal/abi.md
```diff
@@ -0,0 +1,444 @@
...
+Immutable references (`imm`) point to immutable objects outside of any region.
+The Verona runtime guarantees no more than one behaviour has access to the same region at any time, avoiding concurrent mutation problems.
+
+### Regions
+
+A region is a tree of related objects, not necessarily contiguous in memory.
+The reason for grouping objects in the same region is to they can be accessed together by the same behaviour.
+However, some regions (ex. sandboxes) do require memory to be contiguous (for range access protection).
+
+Two regions can have overlapping memory addresses.
```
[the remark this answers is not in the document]
rengolin: I meant what you understood, but that's not good enough. :)

I'm going to change this to a more clear: "addresses are not necessarily contiguous in memory"

docs/internal/abi.md
```diff
@@ -0,0 +1,444 @@
...
+However, dynamic arrays can still have dynamic out-of-bounds access, which should trigger a run time error.
+
+Objects have their sizes known statically and are allocated directly and assigned to a region.
+Dynamic arrays can grow their storage size, to add more elements at run time, but each element still has a static size.
+
+### Forests
+
+Regions can point to other regions (via their sentinel objects), recursively.
+Sentinel objects can only have a single owning reference, so when another region acquires the owning reference, it is moved from the previous region to the new one.
+This forms a forest, or a tree of trees of objects (not a DAG).
```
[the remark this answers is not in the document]
rengolin: My thoughts exactly, but the name Sylvan keeps repeating is "tree", so I used it.

Probably best if I just drop any attempts at nomenclature and just describe what things are.

docs/internal/abi.md
```diff
@@ -0,0 +1,444 @@
...
+* 128-bit CHERI capabilities granting access to an address range.
+
+All Verona objects are at least 8-byte aligned and so the low three bits of a pointer will always be zero.
+When targeting a system that uses integer addresses, all pointers have the same representation.
+When targeting a CHERI system, the following additional rules apply:
+* Pointers to `cown`s are sealed with a type reserved for the runtime.
+  This ensures that `cown`s in immutable objects cannot be dereferenced by unsafe (sandboxed) code and provides some defense in depth against compiler bugs that would accidentally dereference a `cown` pointer.
+* Pointers to immutable objects are not represented by read-only capabilities.
+  Doing so would require expensive operations for reference-count manipulation.
+  When a pointer to an immutable object is passed via the foreign-code layer, it must have the load-storable-capabilities permission removed, such that no reachable capability will ever provide write permission.
```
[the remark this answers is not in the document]
rengolin: @davidchisnall ^^

docs/internal/abi.md
```diff
@@ -0,0 +1,444 @@
...
+Classes and interfaces are stored similar to a _C structure_.
+The general layout is:
+* A header.
+* The list of fields (with `embed` values or pointers to objects).
+
+The header is a pair of values:
+* The region meta-data, a pointer-sized value containing information for the runtime library.
+* A pointer to the type descriptor (see below).
+
+Each object points to a descriptor that uniquely identifies its type.
```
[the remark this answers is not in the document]
rengolin: The object's descriptor contains its dispatch table, so I think the compiler would also have to prove, statically, that none of its methods is ever called dynamically.

docs/internal/abi.md
rengolin: Sorry, that was a bad example made worse. `embed` did something before my latest change but it was the wrong something. I'll try to come up with a better example.

I was not suggesting padding and used the term "naive" to show unaligned F64 on purpose. the paragraph below explains why.

docs/internal/abi.md
rengolin: It's unclear to me all methods by which we can compact the table, and which of them have to be part of the original colouring algorithm or done after the colouring is finished. My initial pessimistic assumption is that any meaningful compacting will happen during colouring, which makes it much more complex.

docs/internal/abi.md
rengolin: I thought about that but didn't discuss on meetings or wrote anything to that effect here.

We can have a 4bit-pattern for "actual NaNs", like `[1111]`.

Alternatively, actual `NaN`s can (almost) always be converted to all-zeroes on the mantissa before storing in a union.
* If the type is `F64`, it's still a `NaN`.
* If the type was a pointer, it's a NULL pointer and thus invalid in Verona.
* If the type is a `Bool`, then it could be a `false`, so we could distinguish `false` by setting some other bit to 1.
* Any other type won't allow all-zeroes as a representation.

But this adds a check on every floating point before storing in a union. Ideally, we'd be able to differentiate betweenn quiet and trapping `NaN`s if the hardware only emits one, we use the other for all other values. This may not work everywhere (or even anywhere), but it would remove the need to check every FP value, including most often non-`NaN` values.

docs/internal/abi.md
rengolin: > When resorting to wide-boxing in the heap, we presumably pad to alignment (which, on CHERI, means unions are really uncomfortably big, with a 16-byte header...

Correct.

> OTOH, for some small solace, there's more room for cap-packing non-pointer types)? When upcasting to a union type as part of a function call, do we follow SROA and pass the discriminator and payload in potentially different registers if they fit?

There are some discussions on SROA and alternatives at the end of the document, clearly inspired by this glaring issue. :)

docs/internal/abi.md
rengolin: Actually, if we made `Bool`'s pattern `[0001]` and made sure no pattern is `[0000]`, it would just work without hacks.

docs/internal/abi.md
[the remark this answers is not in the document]
davidchisnall: I think this is due to confusion over cowns and regions:

 - A region contains a *graph* of objects where every live object is dominated by the sentinel (and every non-live object is unreachable from the sentinel).
 - Regions are a arranged in a *tree* where an `iso` pointer in one region points to a sentinel in a child region.  A single region can have an unbounded (except by physical constraints of the machine) number of child regions but because there may only be one `iso` pointer from outside the region to the sentinel, regions may not form any graph structure other than a tree.
 - Each `cown` forms the root of a tree of regions.  The set of cowns defines a *forest* of regions.

docs/internal/abi.md
davidchisnall: I think the key point is that, when we're doing wide packing, the fact that we don't have concurrent mutation means that there's no need to keep the data and the discriminator together.  When doing class / array layout, we can store them separately and pack the discriminator just as we would any U16 field.

docs/internal/abi.md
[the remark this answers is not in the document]
rengolin: Excellent! I'll add that! Thanks!

docs/internal/abi.md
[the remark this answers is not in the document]
rengolin: I'll add some wording about optimisations over the "abstract machine" implementation.

docs/internal/abi.md
rengolin: Right, this is what I meant.

Colouring being done early makes it harder to compact later because trying to find matching sets in a sea of overlapping sets is a practically unbound problem. Not to mention having to change the indexes on flattened IR code after-the-fact.

So, compacting is done by colouring, but how far do we go during colouring or in which graph we run it is a challenge.

docs/internal/abi.md
rengolin: Which is explained below, in this document.

docs/internal/abi.md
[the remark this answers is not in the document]
nwf: OK, so maybe
```suggestion
  When a pointer to an immutable object is passed via the foreign-code layer, it must have store and load-storeable-capabilities permission removed, such that no reachable capability will ever provide write permission.
```

Untrusted code is presumed not to acquire new references to immutable objects, and so does not need to update reference counts, and all references passed in at invocation are retracted after the sandbox returns or is abandoned?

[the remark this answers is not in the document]
rengolin: I read the original as David said he mentioned above.

I don't know the CHERI model close enough to understand how the proposed diff clarifies it.

@davidchisnall, if the change looks good, feel free to commit the suggestion (or thumb-up, so I can do that).

PR #469, opened 2021-08-04
Adding dialect document

rengolin:
This is a prototype created on top of my expectation of what the AST will look like when type inference is done, so there may be mismatches between what I wrote and what will eventually be.

Would be nice to spot obvious mistakes now, but I'm not expecting to finalise the design until we actually implement and see the problems first hand.

### DIFF
--- a/docs/internal/dialect.md
+++ b/docs/internal/dialect.md
@@ -64,20 +64,19 @@ This is so that we can carry the capability semantics throughout the MLIR passes
 ## Union Types
 
 Union types will be represented by `!join<A, B>`.
-Verona types are algebraic, so `!join<A, !join<B, C>>` is equivalent to `!join<A, B, C>` and will be represented as the latter in MLIR.
+Verona types are algebraic, so `!join<A, !join<B, C>>` is equivalent to `!join<A, B, C>`.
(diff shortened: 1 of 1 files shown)


docs/internal/dialect.md
```diff
@@ -0,0 +1,116 @@
...
+* `!ro<Type>` (Read-Only)
+
+This is so that we can carry the capability semantics throughout the MLIR passes and avoid emitting too many run-time checks at final lowering (see below).
+
+## Union Types
+
+Union types will be represented by `!join<A, B>`.
+Verona types are algebraic, so `!join<A, !join<B, C>>` is equivalent to `!join<A, B, C>` and will be represented as the latter in MLIR.
+
+We need to keep union types represented in such a way because this will allow us to simplify calls and checks after optimisations.
```
jemc, line 69: Does Verona normalize intersections of unions to [Disjunctive Normal Form](https://en.wikipedia.org/wiki/Disjunctive_normal_form)?

If so, it might be worth explicitly stating that here.

rengolin: Yes it does! that's a good point, I'll add to the document. Thanks!

rengolin:
This is definitely work in progress and will change once I start implementing the lowering, so I'll merge this as is and we can revisit with PRs later on.

PR #471, opened 2021-08-13
Upgrade LLVM to a19747ea739

rengolin:
Brings LLVM forward 4 months, across versions, and has no noticeable differences in the API (clang or MLIR), which is incredible.


rengolin:
This should silence the error on Windows: https://reviews.llvm.org/D108124

PR #472, opened 2021-08-18
Clarify size comment on object

rengolin:
Clarifying a comment on the object header after me assuming the region metadata was 64-bits when in fact it's pointer sized.

### DIFF
--- a/src/rt/object/object.h
+++ b/src/rt/object/object.h
@@ -14,7 +14,7 @@ namespace verona::rt
    * Object representation
    * =====================
    *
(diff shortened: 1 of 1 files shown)


davidchisnall (approved):
In theory, a sufficiently weird platform might have a larger `size_t` than a pointer, but I've never seen one.

rengolin:
> In theory, a sufficiently weird platform might have a larger `size_t` than a pointer, but I've never seen one.

We can also change `size_t` to `uintptr_t` on the Header...

PR #473, opened 2021-08-20
Add version straw-man versions doc

davidchisnall:
This is a write-up of the proposal that we discussed a couple of weeks ago.  There are still a few unanswered questions for me and so probably more for everyone else.

### DIFF
--- a/docs/internal/versioning.md
+++ b/docs/internal/versioning.md
@@ -83,7 +83,8 @@ This represents any work-in-progress changes since the last release and exists s
 
 By making the API version distinct from the implementation version, we allow:
 
- - A program to consume two different API versions of the same package (while consuming a single implementation)
+ - A program to consume two different API versions of the same package (while consuming a single implementation).
+   See below for some corner cases that need to be addressed for this to work reliably.
  - A package to deprecate and remove APIs gradually over time without needing two versions of the code.
  - A package to add new APIs without breaking any consumers.
 
--- a/docs/internal/versioning.md
+++ b/docs/internal/versioning.md
@@ -43,8 +43,8 @@ Even within the category of interface breakage, there are two key subcategories:
  - Changes that affect things that are visible in the Verona type system and can cause a consumer of a package to fail to compile.
  - Changes to behaviour documented via API contracts that can cause a consumer of a package to behave unexpectedly.
 
-For example, consider a map defines an iterator that returns the elements in insertion order.
-If a later version of the same package redefines this API to return the elements sorted by a total ordering on the keys, this is a different type.
+For example, consider a map defining an iterator that returns the elements in insertion order.
+If a later version of the same package redefines this API to return the elements sorted by a total ordering on the keys, this is the latter type.
 The API in both versions appears identical at the level of an abstract syntax tree but moving to the new version will break most consumers.
 We aim to provide tools to help developers avoid this kind of problem but we cannot (and do not try to) detect it automatically.
 
@@ -53,7 +53,7 @@ These are [*Note* This should be an exhaustive list, but probably isn't.  We sho
 
  - Removing a public method or field from a class that is public.
    This breaks any code that tries to call the method / access the field or that tries to cast a pointer to that concrete type to an interface type defining that member.
- - Removing a public method field from a private class that is returned as an interface type from a public method.
+ - Removing a public method or field from a private class that is returned (as an interface type) from a public method.
    Code outside the package may cast this to an interface defining the member and expose it.
  - Adding a public method or field to a public class.
    This will potentially allow the class to match interfaces that it did not previously and may affect the behaviour of code that assumes specific semantics from classes that match the interface.
@@ -61,7 +61,7 @@ These are [*Note* This should be an exhaustive list, but probably isn't.  We sho
  - Adding any top-level public field, method, or type.
    This will break any file that imports the module into its top-level scope and defines something with the same name (or imports something else into its top-level scope that defines something with the same name).
 
-In most packages, it will be important to make this kind of breaking change periodically.
+In most packages, it will be important to make this kind of breaking change occasionally.
 The simplest way for a package author to do this is simply to bump the library version and require everyone to fix any breakages in their code.
 This is not desirable because the only way that this allows *graceful deprecation*, which provides a window in which old and new versions of the APIs exist, is to ship two versions of the library.
 Shipping two versions of the library causes several problems, most notably:
--- a/docs/internal/versioning.md
+++ b/docs/internal/versioning.md
@@ -1,0 +1,244 @@
+Versioning in Verona
+====================
+
+**WARNING** This is an under-discussion proposal, not a final design for the language.
+
+Verona, as an infrastructure language, aims to support large-scale software engineering, where programs may be built from different components with different release schedules and that are maintained by different teams.
+A Verona programmer should have the expectation that updating to a new version of a package will not break their code.
+A Verona programmer publishing a library should have tooling that improves their ability to make this guarantee for their downstream consumers.
+
+The key goals for versioning are:
+
+ - A programmer should be able to ship breaking changes to an API with backwards-compatibility shims and graceful deprecation.
+ - Updating to a new version of a dependency should not cause code to fail to compile unless your code depends on a version of the interface that has been explicitly dropped in the new version.
+ - Developers should be given tooling at warns them of API breakages and allows them to provide compatibility interfaces.
+ - If a library ships a new version of an interface, downstream consumers should be able to gradually upgrade their codebase in the common case (unless the new API provides fundamentally different abstractions).
+
+Visibility is a software-engineering feature
+--------------------------------------------
+
+Many object-oriented languages provide some forms of restrictions on when the fields or methods of an object can be accessed.
+The Java security model depends on these.
+In Verona, visibility is regarded as purely a software-engineering tool for defining API-stability guarantees.
+As such, Verona provides only two levels of visibility:
+
+ - Public: may be accessed by anything
+ - Package-private: may be accessed only by code written in the same package
+
+These visibility rules apply to methods, fields, and types.
+They exist to establish a boundary for API-breaking changes.
+Adding or removing a private field or method does not constitute an API break.
+Changing the behaviour or a private method does not constitute an API break.
+Adding or removing a private type does not constitute an API break.
+All of these are true because no code outside of the package may name the types, fields, or methods that have changed.
+
+API-breaking changes
+--------------------
+
+Several kinds of change in Verona can introduce API-breaking changes.
+Note that we differentiate between *API breakage* and *implementation breakage*.
+Nothing that we do can prevent a buggy version of a package from being shipped.
+Even within the category of interface breakage, there are two key subcategories:
+
+ - Changes that affect things that are visible in the Verona type system and can cause a consumer of a package to fail to compile.
+ - Changes to behaviour documented via API contracts that can cause a consumer of a package to behave unexpectedly.
+
+For example, consider a map defines an iterator that returns the elements in insertion order.
+If a later version of the same package redefines this API to return the elements sorted by a total ordering on the keys, this is a different type.
+The API in both versions appears identical at the level of an abstract syntax tree but moving to the new version will break most consumers.
+We aim to provide tools to help developers avoid this kind of problem but we cannot (and do not try to) detect it automatically.
+
+There are a lot of ways of breaking a Verona API such that some consumers will fail to compile.
+These are [*Note* This should be an exhaustive list, but probably isn't.  We should add everything else we discover here]:
+
+ - Removing a public method or field from a class that is public.
+   This breaks any code that tries to call the method / access the field or that tries to cast a pointer to that concrete type to an interface type defining that member.
+ - Removing a public method field from a private class that is returned as an interface type from a public method.
+   Code outside the package may cast this to an interface defining the member and expose it.
+ - Adding a public method or field to a public class.
+   This will potentially allow the class to match interfaces that it did not previously and may affect the behaviour of code that assumes specific semantics from classes that match the interface.
+   It will also cause compile failures for any code that uses implementation inheritance to reuse the code from that class and defines methods or fields of the same name.
+ - Adding any top-level public field, method, or type.
+   This will break any file that imports the module into its top-level scope and defines something with the same name (or imports something else into its top-level scope that defines something with the same name).
+
+In most packages, it will be important to make this kind of breaking change periodically.
+The simplest way for a package author to do this is simply to bump the library version and require everyone to fix any breakages in their code.
+This is not desirable because the only way that this allows *graceful deprecation*, which provides a window in which old and new versions of the APIs exist, is to ship two versions of the library.
+Shipping two versions of the library causes several problems, most notably:
+
+ - Security fixes must be back-ported to the old code.
+   This is somewhat tractable when you have two versions but if this is needed for every API change then it can rapidly become dozens of versions.
+ - If a project uses package A and also package B that uses package A, then it is possible to end up with a dependency on two incompatible versions of package A.
+
+API Versioning
+--------------
+
+In Verona, we differentiate between the package (implementation) version and the API version.
+A package version is a simple monotonic counter of package releases.
+For each release, the package supports a set of API versions.
+API versions use semantic versioning, where the major version is incremented to indicate the removal of types or methods and the minor version to indicate the addition of new classes or methods.
+Both indicate potentially breaking changes for Verona consumers but it should be possible for a package to support all minor versions of an API with minimal effort by the package author.
+There is also one special API version: `unstable`.
+This represents any work-in-progress changes since the last release and exists so that developers can import the current snapshot of a library to test whether in-development changes are likely to cause problems.
+
+By making the API version distinct from the implementation version, we allow:
+
+ - A program to consume two different API versions of the same package (while consuming a single implementation)
+ - A package to deprecate and remove APIs gradually over time without needing two versions of the code.
+ - A package to add new APIs without breaking any consumers.
+
+When a package is imported, the string describing the package should include the API version, for example:
+
+```verona
+import "some/package@2.0" as p;
+```
+
+The build tooling should pick the most recent version of the package that supports all of the versions that are used in the project and its dependencies and warn if that is not the most recent available version.
+**Open question: Should pinning to specific implementations be part of this or should it be separate build metadata such as vcpkg.json?**
+
+When the package is imported with this version, every class, interface, method, or field that is declared as public but is *not* exposed by this version is implicitly package-private.
+This means that it cannot name any type or field.
+
+Implementation
+--------------
+
+Package-private interface and class types are simply not nameable and therefore not usable in other packages.
+The same is true if they are implicitly hidden by belonging to a version that is not in use.
+This is purely a compile-time check.
+
+Methods and fields are slightly more complex.
+Methods or fields accessed on concrete types are trivially allowed or disallowed depending on whether the caller can see an API version that includes them.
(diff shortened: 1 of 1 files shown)


docs/internal/versioning.md
```diff
@@ -0,0 +1,244 @@
...
+   This is somewhat tractable when you have two versions but if this is needed for every API change then it can rapidly become dozens of versions.
+ - If a project uses package A and also package B that uses package A, then it is possible to end up with a dependency on two incompatible versions of package A.
+
+API Versioning
+--------------
+
+In Verona, we differentiate between the package (implementation) version and the API version.
+A package version is a simple monotonic counter of package releases.
+For each release, the package supports a set of API versions.
+API versions use semantic versioning, where the major version is incremented to indicate the removal of types or methods and the minor version to indicate the addition of new classes or methods.
```
rengolin, line 79:
```suggestion
API versions use semantic versioning, where the major version is incremented to indicate the removal or change in behaviour of types or methods and the minor version to indicate the addition of new classes or methods.
```

docs/internal/versioning.md
```diff
@@ -0,0 +1,244 @@
...
+ - A package to add new APIs without breaking any consumers.
+
+When a package is imported, the string describing the package should include the API version, for example:
+
+```verona
+import "some/package@2.0" as p;
+```
+
+The build tooling should pick the most recent version of the package that supports all of the versions that are used in the project and its dependencies and warn if that is not the most recent available version.
+**Open question: Should pinning to specific implementations be part of this or should it be separate build metadata such as vcpkg.json?**
```
rengolin, line 97: I think there are two cases:
1. The code doesn't require a specific implementation but wants to avoid upgrade breakages. This is best done at a configuration file level.
2. The code can only work with a specific version. This is best done at the import level.

If the code imports multiple versions that are incompatible (either through config or `import`), the runtime should import both and separate them by symbol.

This could create the situation where the mangling is different after an upgrade, which would need recompilation of some parts that weren't necessarily part of the original upgrade (ie. other imports that moved to an intermediary API of the same library).

I think the migration path that ties API with code and data structure migration should take care of that, but the update can end up much larger than expected.

docs/internal/versioning.md
```diff
@@ -0,0 +1,244 @@
...
+The same is true if they are implicitly hidden by belonging to a version that is not in use.
+This is purely a compile-time check.
+
+Methods and fields are slightly more complex.
+Methods or fields accessed on concrete types are trivially allowed or disallowed depending on whether the caller can see an API version that includes them.
+Method or field accesses via interfaces depend on whether the selector used in the calling package unifies with the versioned selector in the callee's package.
+
+Package-private members effectively introduce an anonymous (or, at least, unnameable) namespace for selectors such that `_foo` in one package and `_foo` in another are different selectors and there is no way of code in one package naming the other selector.
+Versions simply generalise this slightly.
+Versioned members are identified by a selector that includes the version and package of origin, which has the same value as a selector of the same name and arrity in a package that imports them and is permitted to see that version.
```
rengolin, line 115: This will probably require some mangling rules, which each different mangled symbol pointing to a different selector.

docs/internal/versioning.md
```diff
@@ -0,0 +1,244 @@
...
+By making the API version distinct from the implementation version, we allow:
+
+ - A program to consume two different API versions of the same package (while consuming a single implementation)
+ - A package to deprecate and remove APIs gradually over time without needing two versions of the code.
+ - A package to add new APIs without breaking any consumers.
+
+When a package is imported, the string describing the package should include the API version, for example:
+
+```verona
+import "some/package@2.0" as p;
```
rengolin, line 93: Would be nice to have a more complete example, like including the version annotation seen in the corner cases section below.

docs/internal/versioning.md
```diff
@@ -0,0 +1,244 @@
...
+If `a` is passed to another package containing a similar interface, should it match?
+There are two obvious answers:
+
+ - If the other package imports `A` with a version <2.0, then it should not see the method.
+ - If the other package imports `A` with a version >=2.0, then it should see the method.
+
+The third case is the most interesting.
+What is the correct behaviour if the other package does not import `A` at all?
+Note that, without full dataflow analysis, it is not possible to tell in the general case which package the object was allocated in (for example, it might be stored in an array of `Any` allocated by a fourth module and populated by many other things in the system) and so the question reduces to selector identity.
+Specifically: Is the selector `isSomethingElse/1` in a module that does not import `A` the same as the selector `isSomethingElse@2.0/1` in `A`?
```
rengolin, line 158: Could we create symbol aliases for each method, field and type that points to the most concrete implementation and so each upgrade only needs to add its new aliases, because all previous ones (that are still supported) should be identical to previous implementations.

docs/internal/versioning.md
```diff
@@ -0,0 +1,244 @@
...
+{
+  match (a) {
+    Fooable => a;
+    _ => None
+  }
+}
+
+// In module C
+import "A@1.0" as A;
+import "B@2.0@ as B;
```
rengolin, line 195:
```suggestion
import "B@2.0" as B;
```

docs/internal/versioning.md
```diff
@@ -0,0 +1,244 @@
...
+
+// Does this return `None` or a `Fooable`?
+var a = B.getFooable(A.A());
+```
+
+If `a` at the end of this example is a `Fooable` then this is probably fine.
+The surrounding code knows that `foo` is a method on an interface that can be returned from this function, even if it didn't get one with the previous version of package `A`.
+
+The slightly more confusing outcome occurs if module `B` imports `A` with version 1.0, but `C` is updated to import version 1.1.
```
rengolin, line 204: What would be a valid use case for a dynamic update on the version of the same package on the same run time?

I was under the impression that library updates would only occur at code update, which would be replaced (together with data migration shims) and so won't see each other.

rengolin (commented):
Some overall comments:
* It's hard to understand many of the descriptive examples. Perhaps some code or diagrams would help.
* The document doesn't mention live upgrades, which we also discussed, I'm guessing this is intentional (and another document will cover that)?
* The implementation section is too vague. I'd like to have a straw-man idea of how we'll mangle symbols, alias them to others, and make sure the calls on user code points to the right symbol

docs/internal/versioning.md
davidchisnall: This is about API versions, not library versions.  Imagine the case where you have a package `A` that you're using both directly and a package `B` that uses `A` indirectly.  If you update your code to use the new API, that doesn't necessarily mean that package `B` is also upgraded.  Conversely, I may pick up a new version of `B` that doesn't change its public interface at all, but has moved to using the new APIs exposed by `A`, while I am still using the old ones.

This is one of the key design goals: If `A` introduces a new API, then you don't need all of the things that depend on `A` to move to the new API at the same time.

rengolin: I see, so we also need to define what happens (and how to warn users) in cases where the user doesn't specify anything but versions break.

In your example, when `B` updates `A`, if my `A` doesn't have a version number, it could be confusing to users to be warned that their code breaks the API without modification.

```
Before:
  MyCode[0.1] -> A[any]
  B[1.0] -> A[any]
  A[1.0]

After:
  MyCode[0.1] -> A[any]
  B[2.0] -> A[2.0+]
  A[2.0] // Because "any"
```
If `MyCode` doesn't work with the new major version, and I didn't know that `B` used `A` at all, I'll get confusing error messages.

The error message I'm expecting is something like: `MyCode` doesn't work with the new version of `A (2.0)` installed as a requirement from `B(2.0)`. Please specify the version on `MyCode` to restrict it to the previous version, ie `1.0`.

Instead, if not clear, the error message could be something horrible like: `A::foobar` not found, from `MyCode::functionX()`. Even though the only thing you did was "tool update B".

But this requires a knowledge of the previous environment and the reasons why all versions are what they are, which isn't trivial to communicate effectively to users.

Also, expecting users to manage that, even when using tools to dump the state, on large projects, can be daunting.

rengolin: If we instead, say: any module that upgrades another, bumps the local version but leaves other versions intact.

For example:
```
Before:
  MyCode[0.1] -> A[any(weak: 1.0)]
  B[1.0] -> A[any(weak: 1.0)]
  A[(weak: 1.0)]

After:
  MyCode[0.1] -> A[any(strong: 1.0)]
  B[2.0] -> A[forced:2.0+)]
  A[(strong:1.0) + (forced:2.0)] // Because local upgrade
```

This will never break, but can lead to a huge number of variants without actual need.

davidchisnall: I am not sure what the numbers refer to in your snippet.  Are these the implementation versions or API versions?  If they are the API versions, which package's API versions are they referring to?

docs/internal/versioning.md
```diff
@@ -0,0 +1,244 @@
...
+
+interface Any {}
+
+getFooable(a : Any & mut) : Fooable | None
+{
+  match (a) {
+    Fooable => a;
+    _ => None
+  }
+}
```
nwf, line 191: I recognize that this is probably just the Haskell programmer in me speaking, but this really feels like uncomfortably excessive duck typing.  I think I would rather have a first-class notion of selectors, so that `B`'s `Fooable::foo` really was imported from `A`, at a particular version, at that.

The downside of of this kind of approach, which is more or less what Haskell does, is that it becomes impossible to name a type without picking up a dependency on the defining package.  The ecosystem has responded by either using exclusively compositions of `base` types (eschewing `newtype` or `data` definitions entirely, e.g., in the case of `lens`), or by splitting packages into what amount to interface, usually dependency-less packages (`foo-types`) and implementation, dependency-carrying packages (`foo`).

nwf: David suggested in our 1:1 today that, rather than making selectors generative (and so, nominal) rather than applicative (to borrow terminology from ML, I think?), we could have a generalized "selector attribute" mechanism, with generative (nominal) attributes.  Specifically,
* Concrete types continue to define (at most) one implementation of a given method name and arity (or name and argument type vector?) and may stick a set of named attributes on this definition.
* Interface definitions continue to enumerate entries as now, but they may additionally specify constraints on any would-be matching selector.

This generalizes the version scheme, in that versions are now just generic attributes on concrete types' methods (that is, each method is annotated with all the versions it was present in) and interfaces require the appropriate version label.  It also serves as a kind of nominal proxy for semantics: if David and I have different ideas about what kind of thing goes inside the result of `std::string serialize(Self)` (say), then I can annotate my methods with an annotation from a package I control and David his with one he controls and there won't be risk of confusion.

docs/internal/versioning.md
rengolin: Sorry, it was a bit confusing. The numbers are API versions with the following rules:
* Just numbers (ex. `MyCode[1.0]`) means a fixed version, by the user
* An arrow means "needs", so `MyCode[1.0] -> A[...]` means `MyCode` needs some `A`
* `any` means I don't care which version, any should work (or so my ignorance says so)
* `weak` means the user didn't request specific versions and the compiler inferred something
* `strong` means the compiler was forced to choose a specific version due to constraints
* `forced` means the user forced the compiler to choose a specific version

So:
```
// MyCode version 0.1 just needs A, doesn't care what, compiler picks 1.0 because it's what's available
MyCode[0.1] -> A[any(weak: 1.0)]

// B version 1.0 needs A in the same way, same choice
B[1.0] -> A[any(weak: 1.0)]

// A is imported in version "weak" 1.0 (ie, no constraints)
A[(weak: 1.0)]

// =========== We migrate B to version 2.0, requiring A version 2.0+

// B was upgraded to 2.0 and it now requires A version 2.0 or greater
// We must import that, too, but there could be others (MyCode) that still use 1.0
B[2.0] -> A[forced:2.0+)]

// MyCode version 1.0 still depends on A, but we don't know if it works with 2.0
// and the user didn't say anything so we add a strong requirement to force the 1.0
// Unless the policy is to upgrade and let users handle subtle breakages on their own?
MyCode[0.1] -> A[any(strong: 1.0)]

// Now we have two versions of A
A[(strong:1.0) + (forced:2.0)] // From B and MyCode
```

davidchisnall: The proposal for API versions is that you must *always* specify an API version when you import a package (including the `Builtin` package).  The problems happen if one package imports another (with a version), and then passes pointers to concrete types from that package as interface types to another.

rengolin: I see.

It can be cumbersome to specify all versions when you don't care about it, or there's only ever going to be one (I control it). So I imagine that, in theory, the helper tool that creates the config file would do that for you, and you just check that everything is right.

Then API version migration would generate a different config file that would show up in a `git diff`, making it easy for developers to know what changed and take action if needed.

docs/internal/versioning.md
```diff
@@ -0,0 +1,244 @@
...
+Visibility is a software-engineering feature
+--------------------------------------------
+
+Many object-oriented languages provide some forms of restrictions on when the fields or methods of an object can be accessed.
+The Java security model depends on these.
+In Verona, visibility is regarded as purely a software-engineering tool for defining API-stability guarantees.
+As such, Verona provides only two levels of visibility:
+
+ - Public: may be accessed by anything
+ - Package-private: may be accessed only by code written in the same package
```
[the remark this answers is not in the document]
davidchisnall: We don't think it's useful.  There are two reasons for type-private:

 - You want to build some stronger security properties on top and do very fine-grained isolation that depends on this.
 - You want to guarantee API stability by preventing anything from depending on the implementation details.

The first of these was embraced by Java and leads to the kind of 'defence in breadth' design that the JVM ended up with, where a bug in any part of the implementation can violate the security properties.  We don't want that.

For the second part, type-level private always ends up needing escape hatches like `friend` declarations in C++ or undocumented (or documented 'do-not-use') methods in Smalltalk-family languages because it's actually fine for code that is always shipped along with a class to depend on implementation details of that class.  You end up with cognitive load (remember trying to explain what `public`, `private`, and `package` mean and when they should be used to undergrads?) for a thing that provides no real value: it doesn't add security and it doesn't make software engineering easier.

docs/internal/versioning.md
```diff
@@ -0,0 +1,244 @@
...
+There are a lot of ways of breaking a Verona API such that some consumers will fail to compile.
+These are [*Note* This should be an exhaustive list, but probably isn't.  We should add everything else we discover here]:
+
+ - Removing a public method or field from a class that is public.
+   This breaks any code that tries to call the method / access the field or that tries to cast a pointer to that concrete type to an interface type defining that member.
+ - Removing a public method field from a private class that is returned as an interface type from a public method.
+   Code outside the package may cast this to an interface defining the member and expose it.
+ - Adding a public method or field to a public class.
+   This will potentially allow the class to match interfaces that it did not previously and may affect the behaviour of code that assumes specific semantics from classes that match the interface.
+   It will also cause compile failures for any code that uses implementation inheritance to reuse the code from that class and defines methods or fields of the same name.
```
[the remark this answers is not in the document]
davidchisnall: If I understand the question correctly:

 - If you add a private field, an inheriting class from a different package doesn't see that field and so there's no conflict and because subclassing (implementation inheritance) does not define a subtyping relationship then there's no way that your class will ever be passed to something that expects the other type.  If you define an interface including the private field then the private field is actually defined by {method name, originating package} and so such an interface defined in the two packages will see different private fields.
 - If you add a public field, it's in the same namespace as fields of the inheriting class and so name conflicts can arrise.

I think this introduces a slight variation of the corner case below though.  Consider:

```verona
// Package A
class Base
{
  var x : U32;
  [[version=2]]
  var y : U64;
}

// Package B
import "A@1" as A;

class Sub : A.Base
{
  var y : U64;
}
```

If an instance of `Sub` is passed (as, for example an `Any`) to a third module that does import `A` buy defines an interface requiring a `var y : U64`, which does it see?  I think this one is probably the easiest to solve because there's no subtyping relationship between `Base` and `Sub`, so `Sub`'s definition defines which things are private and which things are public and in this case `Sub` interprets `Base`'s `y` as private and so does not expose it to *any* interface defined outside of package `B`.

docs/internal/versioning.md
```diff
@@ -0,0 +1,244 @@
...
+A package version is a simple monotonic counter of package releases.
+For each release, the package supports a set of API versions.
+API versions use semantic versioning, where the major version is incremented to indicate the removal or change in behaviour of types or methods and the minor version to indicate the addition of new classes or methods.
+Both indicate potentially breaking changes for Verona consumers but it should be possible for a package to support all minor versions of an API with minimal effort by the package author.
+There is also one special API version: `unstable`.
+This represents any work-in-progress changes since the last release and exists so that developers can import the current snapshot of a library to test whether in-development changes are likely to cause problems.
+
+By making the API version distinct from the implementation version, we allow:
+
+ - A program to consume two different API versions of the same package (while consuming a single implementation)
```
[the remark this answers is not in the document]
davidchisnall:
```suggestion
 - A program to consume two different API versions of the same package (while consuming a single implementation).
   See below for some corner cases that need to be addressed for this to work reliably.
```

docs/internal/versioning.md
davidchisnall: Separating them by symbol in the presence of structural typing is not possible, at least in the general case.

docs/internal/versioning.md
davidchisnall: It's not just mangling, it will also require some unification rules to define *which* mangling each package sees for unversioned selectors.

docs/internal/versioning.md
davidchisnall: Different mangling and aliases is part of the solution but that's the easy bit.  The unification rules and how this interacts with structural typing are the difficult bits.

docs/internal/versioning.md
davidchisnall: To add a bit more detail:

We discussed Wes' solution a bit.  The core problem is that it doesn't allow defining open-world interface types (you must specify the concrete types that expose selectors, which means that you're defining a closed-world set and at that point you may as well just use union types).  

We discussed a bit the way that nominal typing is used in Smalltalk-family languages (Java, Objective-C) as a proxy for semantics.  I recalled a random ArXiv paper I read about 15 years ago from someone reimagining Objective-C by providing a central table of selectors that came with contracts for the implementation and using that, rather than simple name-based lookup, for dispatch (I don't think it was ever published anywhere and I never heard anything about the project afterwards).  This felt a bit closer to what I wanted than nominal typing.  

The suggestion that we came up with was to introduce explicit annotations on selectors (which would be first-class concrete types, so are namespaced and exported from individual packages not just ad-hoc strings), with the ability to express a set of them on concrete implementations and refer to them in interface definitions with some simple set-theory operations.  I think this gives me the things I want from nominal typing (where I don't actually want nominal typing, it's just the thing I often have), without introducing a parallel nominal type system (everything is still structural, it's just that now you have a way of unambiguously expressing intentionality in the structural type system). 

I *think* the versioning information could *almost* be treated as a special case if we had this general mechanism, but we should probably discuss them separately and if we want both then consider how they can be merged later.

docs/internal/versioning.md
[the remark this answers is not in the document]
rengolin: In our other (type inheritance) discussion, I thought the semantics was that a derived type cannot override a base's private field, even though it cannot see it. Like the `mixin` semantics, you type check the derived signature against the base class signature (that includes the private types).

With versioning, this becomes much harder, because API version 2 introduces new private fields (even if public, they're "private" to previous versions), but they're not really private, not in the class signature.

Does this mean API versioning will essentially change the type semantics? We'll need to carry version information around to the type checks to make sure we give meaningful error messages to users.

docs/internal/versioning.md
rengolin: Right, as discussed, I was conflating library version with API version.

There's still the case where I only need an intersection of API versions 2 and 5, so don't care which API version I'm using on that range, but this is only problematic if we have to rely on third-party package distribution, where some OSs have the old packages and others the new ones. 

If we use our own package distribution (which comes with its own problems), we can always take the latest version.

PR #474, opened 2021-08-20
Add more detail to the foreign code doc.

davidchisnall:


docs/internal/foreign_code.md
```diff
@@ -62,32 +65,210 @@ We hope that the region abstraction will make programming with sandboxes natural
...
+{
+  void x();
+};
+```
+
+In Verona, given an instance of `Sub`, you could call `Super::x()` like this:
+
+```verona
+var s = CXXLib.Sub();
+CXXLib.Super.x(CXXLib.static_cast[CXXLib.Sub](s));
```
rengolin, line 211: This is clear, but not trivial. It involves users understanding both Verona and C++ concepts to a deeper level than I assume most people trying to interface the languages will have.

I'm guessing the majority of people using C++ inside Verona will be only well versed in one of the two languages. For example:
* A Verona programmer trying to speed up the program by using an optimised C++ library
* A C++ programmer trying to create a safer facade in Verona for a low-level C++ library

On the other hand, trying to sugar this will probably lead to many corner cases and impedance mismatch.

docs/internal/foreign_code.md
```diff
@@ -62,32 +65,210 @@ We hope that the region abstraction will make programming with sandboxes natural
...
+s.x();
+CXXLib.Sub.x(s);
 ```
 
-This includes two attributes that tell the compiler the format of the library (native code) and how to parse the header (which, in this example, would have to be something provided by the user that included the desired headers).
-Once C++ module support is more widespread, module imports can take the place of header inclusions.
+### Templates
+
+C++ templates are exposed as Verona generics.
+Unlike normal Verona generics, C++ templates can take values as type parameters.
```
rengolin, line 220:
```suggestion
Unlike Verona generics, C++ templates can take values as type parameters.
```

docs/internal/foreign_code.md
```diff
@@ -62,32 +65,210 @@ We hope that the region abstraction will make programming with sandboxes natural
...
+For the MVP, we should not support C++20 object literals and so support only types and numeric types as template parameters.
+This is sufficient for existing rich C++ APIs and requires approximately the same parsing logic as is already needed for Verona arrays.
+
+Unlike Verona generics, a C++ template may parse but not be possible to instantiate with some of the types that it permits as arguments.
+For example, `std::vector<T>` requires that `T` is copy or move constructable but the only constraint on `T` is that it is a type.
+To avoid propagating this problem into Verona code, we require that all template arguments are concrete types *before* reification.
+A Verona generic can wrap a C++ template by providing a closed set of permitted values for the template instantiation.
+If any of these fail, the generic will not valid Verona code, if all of them succeed then the generic is valid for all possible instantiations.
```
rengolin, line 228:
```suggestion
If any of these fail, the generic will not be valid Verona code, if all of them succeed then the generic is valid for all possible instantiations.
```

docs/internal/foreign_code.md
```diff
@@ -62,32 +65,210 @@ We hope that the region abstraction will make programming with sandboxes natural
...
+var s = CXXLib.Sub();
+CXXLib.Super.x(CXXLib.static_cast[CXXLib.Sub](s));
+// The following two are equivalent 
+s.x();
+CXXLib.Sub.x(s);
 ```
 
-This includes two attributes that tell the compiler the format of the library (native code) and how to parse the header (which, in this example, would have to be something provided by the user that included the desired headers).
-Once C++ module support is more widespread, module imports can take the place of header inclusions.
+### Templates
```
rengolin, line 217: I think we also need to make a strong restriction here:
* Using Verona types in C++ containers isn't permitted, they would be passed as opaque pointers and C++ code could do nothing with them
* Using C++ types in Verona containers is less of a problem, but could also be restricted in the first iteration

docs/internal/foreign_code.md
```diff
@@ -62,32 +65,210 @@ We hope that the region abstraction will make programming with sandboxes natural
...
-C++ namespaces and classes will appear as nested classes within the Verona `LLVM` class.
-For example, `LLVM.llvm.SmallVector[LLVM.int, 12]` corresponds to `llvm::SmallVector<int, 12>`.
+### Overloaded operators
 
-Instantiating this class would require the user to specify a sandboxing policy, for example (again, straw-man syntax):
+Verona allows overloading most of the operators that C++ provides and so these can be trivially mapped just like any other methods.
+The main exceptions relate to assignment, though Verona's `apply` sugar provides similar functionality.
+Classes from C++ need to opt out of `create` and `apply` sugar because these reflect patterns that don't exist in C++ codebases and so it's possible that apply sugar could be applied for C++ assignment operators.
+
+### Casts
```
rengolin, line 236: This section could come before `Classes`, as it has an example of `static_cast` and I assumed other things would fall through the net, which isn't true.

docs/internal/foreign_code.md
```diff
@@ -62,32 +65,210 @@ We hope that the region abstraction will make programming with sandboxes natural
...
-This would create a new process that loaded the LLVM library and had RPC invocations for any of the LLVM functions that this program called.
-The `llvm` object in this example is a new isolated region.
-The programmer can then construct LLVM types within that region.
+Note that the last line of this is explicitly violating type safety.
+All of these operations are restricted to values that are within a sandboxed foreign region where type safety is explicitly not guaranteed.
+
+### Exceptions
+
+**This doesn't make me very happy, better suggestions welcome**
+Every sandboxed function may throw a `SandboxError` exception if the sandbox has failed and so all invocations of sandboxed code must be wrapped in a `try` block.
```
rengolin, line 265: This may not be necessary when interfacing with languages that do not have unchecked exceptions... Are there any?

docs/internal/foreign_code.md
```diff
@@ -62,32 +65,210 @@ We hope that the region abstraction will make programming with sandboxes natural
...
+
+C++ has unchecked exceptions and so any function not explicitly declared as `noexcept` may throw any type.
+Verona, in contrast, has only checked exceptions.
+This provides us with a way of identifying the expected exceptions:
+Any exception of a type from a particular sandboxed library that the code invoking the foreign code either explicitly catches or advertises that it may throw is assumed to potentially originate in the sandbox.
```
rengolin, line 270: It may be hard to know which exceptions can be thrown, with standard exceptions being thrown from standard library functions (like `bad_alloc`).

docs/internal/foreign_code.md
```diff
@@ -62,32 +65,210 @@ We hope that the region abstraction will make programming with sandboxes natural
...
+C structures do not contain a type descriptor and, even if they did, it cannot be trusted by Verona, but it would be possible to box them.
+Without any native support, it is possible to create a Verona generic class that holds an external reference to a C structure of the type of the generic parameter.
+
+Nested structures in C are exposed as objects with `embed` fields in Verona if they are named.
+Anonymous structure or union members in C simply add their members directly into the class.
+
+Unions in C are *not* mapped to union types.
+There is no way of implementing a `match` expression on a `C` union.
+C unions are not type safe and some code depends on writing to one union field and reading from another.
+In this way, they are closer to a Verona object type that provides set and get methods of different types and uses internal storage that allows it to materialise a value of the requested type from whatever was set.
```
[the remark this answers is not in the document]
davidchisnall: I see the former as an optimisation on the latter.  We always call the function to access a field, if we can statically prove that it's safe then we can move it to the Verona side of the RPC line.

docs/internal/foreign_code.md
```diff
@@ -62,32 +65,210 @@ We hope that the region abstraction will make programming with sandboxes natural
...
+
+All built-in C types (for example, `int` and `long`) are surfaced as type aliases on the class that is exposed for the library.
+A single program may include sandboxes with 32- and 64-bit ABIs for foreign code and so the size of primitive types must be defined per-sandbox-instance.
```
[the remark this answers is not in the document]
davidchisnall: I'm not particularly wedded to this, but my assumption was that `one.int` would be an alias for `SandboxType.int` so both would exist but `one.int` is more convenient for code that wants to be generic over sandbox.

docs/internal/foreign_code.md
davidchisnall: Good point.  I'm not sure exactly who would work with this.  If it ends up being anything like the Python ecosystem, then there will be people writing Verona wrappers who will deal with this and so we'll end up with:

 - People who just want to grab some C/C++ code quickly working here with some quick-and-dirty code that won't really care much about the difficult corner cases because they're using fairly simple interfaces.
 - People writing idiomatic Verona wrappers around popular C/C++ libraries.  They'll hopefully be pretty knowledgeable about both Verona and C++ and we need to give them tools to write good bindings because they're going to end up writing things that a lot of people use.
 - People writing C/C++ specifically for use in the Verona context because they can write something faster in C/C++.  I hope there aren't many of these because in most cases I'd hope that the Verona compiler can do as good a job as the C/C++ one, but there may be some things where if you're willing to aggressively violate type safety then you can pool memory and make something faster in C/C++ than Verona.

The middle category are the folks I care the most about.  I want to make sure that they have to write the absolute minimum of boilerplate to be able to write idiomatic Verona wrappers around C/C++ libraries.  In an ideal world, they'll be able to quickly write a Verona wrapper that's nicer to use than the underlying library and so people writing new code would prefer to use libWhatever from Verona than from C/C++.

docs/internal/foreign_code.md
davidchisnall: I'm not sure what you mean by 'container' here, since that isn't a language construct, but I think I more-or-less agree:

 - If you pass a Verona object to a foreign library, the foreign code sees a thing that looks like a pointer but that it can't actually dereference (in the non-CHERI case, this is actually an index into a table [possibly encoded in such a way that if you look at the bits it's a plausible pointer so that it doesn't clash with things that do tricks like use numbers < 64 KiB as special values] that contains a pointer to the Verona object, in the CHERI case it's probably a sealed capability to the object).  It has no methods.  You can store it in a `std::vector<void*>` or whatever, but it isn't actually very useful.
 - A C++ pointer looks in Verona like an external reference.  You can store it in a collection only if that collection can handle external references (doing anything with an external reference other than [perhaps] identity comparison requires presenting the region that contains the object).  Any method you call on it will (logically, at least) be an RPC.

docs/internal/foreign_code.md
davidchisnall: Java / .NET, for example.  If we called gcj-compiled Java or AoT-compiled C# then we'd be able to get a complete list of all possible exceptions (the checked list plus runtime exceptions like `NullPointerDereference`).  

I can imagine wanting to build higher-level interfaces on top of the C/C++ interop layer, for example something talking to CPython to provide a direct Verona <-> Python bridge that doesn't require the Verona programmer to think about the CPython implementations, but that's very much a v2 feature.

docs/internal/foreign_code.md
davidchisnall: Right, and without checked exceptions a C++ function will happily propagate any exceptions that are thrown.  This is much worse in the general case with the Itanium ABI because a C++ function may call a gcj / Ada / whatever function that throws an exception that the C++ program doesn't know how to handle.  These can always be caught with `catch(...)` though, which means that we can always wrap calls in that and propagate the fact that an exception occurred (even if we don't know what the exception was) to Verona.

docs/internal/foreign_code.md
```diff
@@ -62,32 +65,210 @@ We hope that the region abstraction will make programming with sandboxes natural
...
+A package is referred to in Verona by a path.
+Normally this refers to a directory that contains Verona source code.
+For foreign modules, this provides a JSON (subject to change) file that contains (for C/C++):
+
+ - The header file describing the interface.
+ - The compiler flags required to parse the header file (for example, predefined macros, language dialect).
+ - The library file containing the implementation (initially this will be a .so / .a file, it may be a .wasm file for WebAssembly sandboxing once that is implemented).
+ - Any other metadata that we require (TBD).
+
+This file should be trivial to generate from existing build systems.
```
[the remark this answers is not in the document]
davidchisnall: Yes.  I should probably add a rough example of this.

docs/internal/foreign_code.md
davidchisnall: At least in the absence of CHERI, I don't really see any reason for pointers to Verona objects to show up as anything other than `void*` on the C++ side.  I'll add some explanation there.  

`Array[C++Type]` is a bit more interesting because I *can* imagine wanting to have `Array[SomeSandbox.int]`, for example.  If `C++Type` is a C++ class then it must be created in a sandboxed region, but `Array` can't be (probably - we might want to allow `Array` to be created in sandboxed regions so that we have a consistent class for 'contiguous blob of data', but I think creating a C array and doing an explicit copy probably makes more sense), so you can't actually express that type (or, rather, you can, but you can only use it to create a 0-element array).  You can express `Array[external_ref & C++Type` (or whatever the syntax is for external references), but this will then be a pair of a pointer to the sandbox and a pointer to the object and needs explicitly converting to a pointer to the object before it can be used, so there's a hint in the language that you don't want to be doing this.

docs/internal/foreign_code.md
davidchisnall: I think this probably needs an example.

docs/internal/foreign_code.md
rengolin: Exactly. Both cases of mixing lead to complicated semantics which I think we should avoid. Not so much for the compiler, but for the cognitive load on developers.

`Array[Sandbox.int]` is an interesting case, but since that's an alias to a Verona type, it's trivial. 

What I meant by `C++Type` was C++ classes, and as you say, ends up being an external reference, which needs not only pointer conversion but RPC-calling after that.

rengolin (approved):
You have addressed all of my points, this looks good to me now. Thanks!

PR #475, opened 2021-08-25
Add basic support for promises

marioskogias:
This is an one-shot promise with a single producer single consumer.


mjp41 (approved):
LGTM. 

At some point I think we actually want to add support to the runtime for this.  So that certain optimisations for message send can occur, but this gives a good step to using Promises on top of the runtime, and we can change the inners of this, without affecting clients.

PR #477, opened 2021-08-26
User friendly LLVM option

rengolin:
When asking to build LLVM or re-use an already built directory, CMake
forced users to turn off VERBOSE_LLVM_DOWNLOAD, which was on by default
and emitted a confusing error message.

To avoid having to pass -DVERBOSE_LLVM_DOWNLOAD=OFF on the other two
ways to build Verona, we just assume that, when users set those
variables, they really don't need to download it, too.

Keeping the error message when both build and reuse options are set, as
neither are set by default and it would clearly be an error.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -12,13 +12,17 @@ macro (option_top VAR HELP DEFAULT)
   set_cache_top(${VAR} ${HELP} BOOL ${DEFAULT})
 endmacro ()
 
-option_top (VERONA_DOWNLOAD_LLVM "Download cached version of LLVM. Do not set with LLVM_EXTRA_CMAKE_ARGS or VERONS_EXTERNAL_LLVM_DIR" ON)
-option_top (VERBOSE_LLVM_DOWNLOAD "Verbose LLVM/MLIR download step" OFF)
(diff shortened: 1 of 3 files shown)


CMakeLists.txt
```diff
@@ -76,46 +80,72 @@ if (NOT CMAKE_BUILD_TYPE)
...
     -DCMAKE_BUILD_TYPE=$<CONFIG>)
 
-  if (NOT VERONA_DOWNLOAD_LLVM)
+  # Extra options for building or downloading LLVM
+  if(VERONA_BUILD_LLVM_SUBMODULE)
     list (APPEND EXTERNAL_EXTRA_CMAKE_ARGS
       -DLLVM_EXTRA_CMAKE_ARGS=${LLVM_EXTRA_CMAKE_ARGS}
     )
-  else ()
+  else(VERONA_DOWNLOAD_LLVM)
```
ihaller, line 144: Should this not be elseif? From the CMake documentation `Per legacy, the else() and endif() commands admit an optional <condition> argument. If used, it must be a verbatim repeat of the argument of the opening if command.`

rengolin: It should. And I thought I had fixed that... :(

rengolin: Fixed it. Thanks for checking!

ihaller:
LGTM

PR #478, opened 2021-08-27
Added flag to be used by custom platforms to integrate their own threading APIs

ihaller:
A custom platform developer can add the define VERONA_EXTERNAL_THREADING, which will trigger the inclusion of `<verona_external_threading.h>` which can be provided by the developer in the include path.

A template of this file is located in `src\rt\test\func\external_threading`. Methods marked as deleted are not used by Verona, but can exist if the platform developer desires as such.
As a summary, the necessary functionality is: public interfaces from `cpu.h`, minimal subset of `std::thread`, `FlushProcessWriteBuffers`.

Over time the code in `barrier.h` and `cpu.h` could be transitioned into PAL classes using this template. I can even do that change now if people feel that it is the right time to do such a PAL.

Another change is that the ThreadPoolBuilder will only create `N-1` threads, where `N` is the desired number of scheduler threads. The `N`-th scheduler thread will be the same thread used to create the thread pool. Something to be mindful is that some scheduler thread state might not be cleaned today as the expectation is that a new execution will get a fresh thread, which is not the case anymore. I reset `Scheduler::local()`, but there might be other state I am unaware of.


src/rt/sched/cown.h
```diff
@@ -11,6 +11,8 @@
 #include "priority.h"
 #include "schedulerthread.h"
 
+#include <algorithm>
```
mjp41, line 14: Why is this now needed?

src/rt/test/func/external_threading/build_test.cc
```diff
@@ -0,0 +1,13 @@
...
+#define VERONA_EXTERNAL_THREADING
+#include <verona.h>
+
+/**
+ * This test if only a build test to confirm that the template of
+ * verona_external_threading.h matches the requirements of the Verona run-time.
+ */
+int main()
+{
+  return 0;
```
mjp41, line 12: Mildly concerned that template instantiation could hide compile bugs as this doesn't actually call the runtime.

mjp41 (approved):
LGTM.  This seems a reasonable approach until we factor the Pal out properly.

src/rt/sched/cown.h
ihaller: std::any_of in this file needs it and for the current platforms this is picked up from cpu.h, but for the new platforms we need an explicit include.

PR #480, opened 2021-08-31
Deprecate old compiler/interpreter

rengolin:
The old compiler/interpreter had test instabilities and were
implementing an old standard for the language that has changed
considerably in the last year (and probably will change again soon).

This moves them to a deprecated directory to make it clear to drop-in
developers that they're not supposed to implement the current standard
encoded by the design documents.

We're still building the binaries and putting them on `dist`, so people
can still play with them, but they no longer break CI builds or
misdirect people where the compiler really is.

@mjp41, is that what you meant?


rengolin:
This is basically a CMake change only. Most of the changes are just moving things into a directory called `deprecated` in `src` and `testsuite`. The CMake changes are minor, just working around the move.

The tests are not run anymore, but they're still there. There is nothing in our current development that touches those two pieces of code, so nothing we do can ever change them. 

Because they were passing for the past year, it means they're supposed to work with the state the tree is now (external dependencies included). Because we still build them, any compile-time dependency will still be tested in CI. If there's any problem in the tests, it will be because `snmalloc` got upgraded or something, which is minor.

This will also make it a lot easier to remove the components once we have an execution environment for the new compiler: just remove the `deprecated` directories and their dependencies.

rengolin:
I'm working on enabling sanitizers for all programs, not just the run-time. The two programs here don't pass either ASAN or UBSAN, so before we do that we must deprecate them (and their tests), or the build will never pass.

PR #484, opened 2021-09-01
Fix issues in verona-interop found by ASAN

rengolin:
While turning ASAN on all programs, I found these issues with `verona-interop`.

Before turning ASAN/UBSAN builds in CI we need to fix them all.


rengolin:
ASAN runs pass and solutions break UBSAN. I'll merge both changes into one package and submit another PR.

PR #486, opened 2021-09-01
Generic sanitizer run on CI

rengolin:
Change USE_ASAN option to SANITIZER, so that we can choose any sanitizer
to run on local builds or CI.

Runs ASAN on both commit CI and nightly, so we can catch errors earlier.

Not running UBSAN yet because some errors in the parser are still
failing and there may be some spurious run-time errors that need fixing
before turning it on.

But from this commit onwards, anyone can use any sanitizer (even
combining them with -DSANITIZER="address,undefined,leak") on a local
build for all programs, not just ASAN on the run-time.


rengolin:
Passed [nightly](https://dev.azure.com/ProjectVeronaCI/Project%20Verona/_build/results?buildId=3067&view=results) on Linux and Windows. Mac is still breaking but that's unrelated to this PR.

mjp41 (approved):
LGTM

PR #491, opened 2021-09-03
Encapsulate signal safety thread

mjp41:
During signals very few methods are allowed to be called. The code
creates a new thread to wait on a condition variable that condition variable
is notified during a signal, and the thread then handles it asynchronously.

Prior to this commit, this had an issue that Dtors and `atexit` order could be bad.
In particular, the mutex could be being torn down during the `atexit` handling. 

This now encapsulates everything in a single class that only uses C++ standard
library features. This means the destruction semantics are "obvious".

The first two commits are useful for debugging, and were used to find this bug,
but are otherwise unrelated.


mjp41:
@theodus, I tagged you as I thought you might be interested as your were involved in the original code.

src/rt/test/systematic.h
```diff
@@ -503,51 +504,14 @@ namespace Systematic
...
-    }
-    while (true)
-    {
-      sleep(1000);
-    }
-  }
 
   inline static void crash_dump()
   {
     std::unique_lock<std::mutex> lock(mutx);
```
ihaller, line 514: Do you still need this mutex if it is only ever called via ThreadPing?

src/rt/CMakeLists.txt
```diff
@@ -155,6 +155,15 @@ if(NOT DEFINED VERONA_RT_ONLY_HEADER_LIBRARY)
     set (TOP_SEED 20)
   endif ()
 
+  # Look for lldb 
+  find_program(LLDB NAMES lldb)
+
+  # For CI builds attempt to run inside LLDB to get more information
+  # for crashes.
+  if (VERONA_CI_BUILD AND (NOT LLDB STREQUAL LLDB-NOTFOUND))
```
rengolin, line 163: Running every test inside a debugger will incur in performance penalty for normal runs. Perhaps we could make this a separate flag and only check if LLDB exists and use it with that flag, regardless of where we build.

src/rt/test/systematic.h
mjp41: That lock is protecting the `stack` stuff.  As we only do this once, it should be fine without.  If we want to output repeatedly, then we need a lock so that the backtrace isn't raced on.  I'll add some comments.

src/rt/test/harness.h
```diff
@@ -135,6 +135,8 @@ class SystematicTestHarness
                 << duration_cast<seconds>((t1 - start)).count() << " seconds"
                 << std::endl;
     }
+
+    std::cout << "Finished!" << std::endl;
```
rengolin, line 139: What finished? Here we know, but in the middle of a log, it may be harder to pinpoint.

Specific messages like "Execution of test <test> finished" makes it easier, as well as to find in code when looking for a break.

src/rt/test/systematic.h
ihaller: Sorry, I see that it is still needed to synchronize with signal_handler.

src/rt/CMakeLists.txt
mjp41: The problem is if there is ever a crash, we want as much information immediately.  If everything is super repeatably, then it is fine.  As far as I understand this is a fixed cost per test.  To the runtime, they all run a reasonable amount inside.  I'll measure the time difference though, to see what we are paying for this.

src/rt/test/threadping.h
```diff
@@ -0,0 +1,87 @@
+// Copyright Microsoft and Project Verona Contributors.
+// SPDX-License-Identifier: MIT
+#include <condition_variable>
+#include <thread>
```
ihaller, line 4: Would it be possible to add support for VERONA_EXTERNAL_THREADING to this file?
```#if !defined(VERONA_EXTERNAL_THREADING)

#  include <thread>

namespace verona::rt
{
  using InternalThread = std::thread;
}

#else

#  include <verona_external_threading.h>

#endif
```

and using ```verona::InternalThread``` instead of ```std::thread```.

PR #495, opened 2021-09-08
[cmake] Allow LLVM_EXTRA_CMAKE_ARGS to override CMAKE_BUILD_TYPE

almiliMSFT:
Otherwise, when building Verona + LLVM all together, the build type for LLVM is set to the same build type as specified for Verona.  Consequently, it's not possible to at once build Verona in Debug and LLVM in Release.

### DIFF
--- a/external/CMakeLists.txt
+++ b/external/CMakeLists.txt
@@ -64,8 +64,15 @@ if (VERONA_DOWNLOAD_LLVM)
   )
 else()
   separate_arguments(LLVM_EXTRA_CMAKE_ARGS UNIX_COMMAND ${LLVM_EXTRA_CMAKE_ARGS})
-  list (APPEND LLVM_EXTRA_CMAKE_ARGS
+  set (LLVM_CMAKE_ARGS
+    # set a default CMAKE_BUILD_TYPE to be the same as for Verona
     -DCMAKE_BUILD_TYPE=${CMAKE_BUILD_TYPE}
+
+    # append LLVM_EXTRA_CMAKE_ARGS (which may override CMAKE_BUILD_TYPE)
+    ${LLVM_EXTRA_CMAKE_ARGS}
+
+    # set mandatory cmake args below
+
     -DCMAKE_INSTALL_PREFIX=${LLVM_INSTALL}/install  # Add install as tar has install prefix. Outer project knows about this.
     # Note: We must use the SEMICOLON generator expression here because CMake
     # uses semicolons as list separators and this must be a literal semicolon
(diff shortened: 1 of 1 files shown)


davidchisnall:
Is that desirable?  LLVM has a different ABI in release and debug builds and including the LLVM headers without `NDEBUG` defined is generally a great way of having silent memory corruption.

I've been working in the vcpkg-wip branch on removing a lot of the custom CMake logic and using vcpkg to provide cacheable debug / release builds, which should allow us to easily add more platforms (e.g. FreeBSD, Morello) to our CI matrix (CheriBSD + Morello is going to be a requirement in January) and use generic caching infrastructure for the dependencies.  vcpkg builds both debug and release builds of all of the dependencies to avoid any mismatches like this.

external/CMakeLists.txt
```diff
@@ -64,8 +64,15 @@ if (VERONA_DOWNLOAD_LLVM)
   )
 else()
   separate_arguments(LLVM_EXTRA_CMAKE_ARGS UNIX_COMMAND ${LLVM_EXTRA_CMAKE_ARGS})
-  list (APPEND LLVM_EXTRA_CMAKE_ARGS
+  set (LLVM_CMAKE_ARGS
+    # set a default CMAKE_BUILD_TYPE to be the same as for Verona
     -DCMAKE_BUILD_TYPE=${CMAKE_BUILD_TYPE}
+
+    # append LLVM_EXTRA_CMAKE_ARGS (which may override CMAKE_BUILD_TYPE)
+    ${LLVM_EXTRA_CMAKE_ARGS}
```
rengolin, line 72: Move this to the end of the list, so that it overrides all args, not just build type

rengolin (commented):
I think this makes sense from a build system point of view, but the append should be after all args, not just build type. So user arguments override everything that is default, not just the build type.

@davidchisnall, we had hit this problem a long time ago when downloading LLVM's debug builds, which fill up the free VMs for any platform. For better or worse, this is how we've been building Verona since we introduced downloads, about a year ago. That's the reason why we don't have debug LLVM blobs.

On my machine, a debug build + install of LLVM+clang+MLIR has more than 90GB. Those disks have 100GB, I think. Just downloading a debug blob fills the disk, too.

Long term, using vcpkg may be our only hope, but for now, I think this fixes the build.

external/CMakeLists.txt
almiliMSFT: moved

rengolin (approved):
LGTM, thanks!

rengolin:
Builds are green again, thanks @almiliMSFT!

PR #497, opened 2021-09-14
Changed std::thread usage in harness to verona::InternalThread

ihaller:
This allows custom platforms to reuse the harness code. Updated comments to hopefully address confusion from the internal/external naming.

### DIFF
--- a/src/rt/pal/cpu.h
+++ b/src/rt/pal/cpu.h
@@ -29,6 +29,7 @@
 #endif
 
 #include <algorithm>
+#include <functional>
 #include <vector>
 
 namespace verona::rt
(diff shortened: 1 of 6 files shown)


src/rt/pal/cpu.h
```diff
@@ -29,6 +29,7 @@
 #endif
 
 #include <algorithm>
+#include <functional>
```
ihaller, line 32: This was being pulled in from `<thread>` earlier. `std::function` is used in this file, so it is best to make sure the header is also included here.

mjp41 (approved):
Given that comment I wonder if `InternalThread` should be `PlatformThread`

ihaller:
Great suggestion about the name, updated it.

mjp41 (approved):

PR #499, opened 2021-09-22
Dialect changes from current semantic discussions

rengolin:
Recent changes in semantics have inspired additions to the dialect, to allow optimisations and easy representation in IR.

This change adds more detail to existing operations as well as new types (`!throw`, `!region`) and new operations (`throw`, `catch`, `match`, etc) which will all be used for high-level optimisations. It also adds some examples of optimisations and how those operations help.

Not really final, by a long shot, but I think we're getting close to a semantics that connects Verona AST with LLVM IR really well.


rengolin:
A rendered version: https://github.com/rengolin/verona/blob/dialect-doc/docs/internal/dialect.md

PR #502, opened 2021-09-28
Add initial RFC for handling low memory in Verona.

davidchisnall:


docs/internal/constrained-memory.md
```diff
@@ -0,0 +1,175 @@
...
+Note that this is a global property, not necessarily visible to the thread that cause memory exhaustion.
+
+The next thing that it will try to do is flush some things to the backing store.
+In a system with a unified buffer cache, there is no difference between anonymous memory backed by swap and memory-mapped files.
+Both can be written back to some backing store to free physical memory.
+This is somewhat more complex for the kind of deployment where we expect to see Verona: cloud services that may not have local storage.
+Swapping in general often requires some allocation, swapping to a remote blob store by sending some network messages may require a lot of memory, especially given that everything that is swapped will typically require encrypting.
+
+If this still fails, the kernel will often look for things to kill.
+Linux, for example, identifies the process with the most unsaved state of value to the user and kills it.
```
nwf, line 41: While not, exactly, wrong...

docs/internal/constrained-memory.md
```diff
@@ -0,0 +1,175 @@
...
+### Low-memory noticeboards
+
+XNU, Android, and Windows all provide some form of low-memory notifications.
+The most useful forms of these are when memory is fairly low but allocation remains possible.
+These can be used by software to avoid allocation or free existing objects when memory is low but not yet exhausted.
+These are particularly useful in garbage-collected systems where it's often beneficial to avoid running the GC until the system starts to experience some memory pressure.
+
+In a Verona world, traced regions are only collected in response to explicit invocation of the collector.
+A cown may wish to avoid tracing until memory starts to be constrained.
+Signalling a noticeboard and scheduling the cowns that have requested notifications does not require memory allocation and so providing a low-memory noticeboard would allow a cown to perform some cleanup when memory is low, without requiring any additional allocation.
```
nwf, line 132: But again, as above, this message may arrive much later, after additional (de)allocation... so it seems like this may be more of a "memory was low in the past" vs. "memory is low now" thing, which seems less useful?

docs/internal/constrained-memory.md
```diff
@@ -0,0 +1,175 @@
...
+This would add it to a list in the runtime.
+When the runtime detects an out-of-memory condition then it would walk the list of cowns that support sudden termination and free any that are not currently strongly referenced.
+
+Cown references are not required to be option types and so there is no way of making cowns that can go away unexpectedly other than by ensuring that all references to it are weak references.
+Scheduling a behaviour on a cown implicitly references it and so a cown that supports sudden termination will not be deallocated if it has pending behaviours to process or if it is running.
+
+When a `cown` is deallocated (either explicitly or when killed by low-memory conditions) then it will need to be removed from the list.
+This is likely to be a fairly uncommon operation and so it would probably be acceptable to have a simple doubly linked list protected by a lock.
+
+In the current implementation, `cown`s are not deallocated while they still have strong references but all objects that they point to are.
```
nwf, line 149: That seems sketchy; I must be misunderstanding... what happens to pending behaviours or what's given to behaviours that get scheduled?  Or is it just objects weakly held that are discarded?

docs/internal/constrained-memory.md
davidchisnall: Is that not what their algorithm does?

docs/internal/constrained-memory.md
davidchisnall: Kind of.  I'd missed that when @mjp41 fixed my code, we'd lost the property that notifications were unordered with respect to behaviours, though even with that the cown would still need to get to the front of the scheduler queue before it could handle the notification.  This is why I think this is a lot more useful as a 'memory is a bit low' notification than an aggressive 'you need to free things now!' message.

That said, notifications are both edge triggered and coalescing.  If you register to receive this notification then your notification callback will receive the value of the noticeboard *at the time when the callback is invoked*.  This may lead to some confusion, where you receive a low-memory notification telling you that memory is not low, but as long as your notification callback guards that properly then it can very quickly discard notifications that memory was low in the past.

docs/internal/constrained-memory.md
davidchisnall: Sorry, terminology problem:

Cowns in the abstract machine are dealloated as soon as the last strong reference to them goes away.

The cown data structure in the runtime is not deallocated until all weak references to it have gone away.  

This is basically the same as `std::shared_ptr` created with `std::make_shared`: the refcounting control block is part of the allocation and so it can't be deallocated until the object is deallocated.  The finaliser runs and it's gone from the abstract machine, but the memory is not reclaimed.

docs/internal/constrained-memory.md
nwf: It is what their algorithm achieves, perhaps.

PR #506, opened 2021-09-30
Enable cown early release in multimessage

marioskogias:
Give users the ability to release a cown that is part of a multi-message before the end of the currently executing behavior. This will enable other scheduler threads to execute pending behaviors on the early released cown earlier.


src/rt/sched/cown.h
```diff
@@ -1199,10 +1204,16 @@ namespace verona::rt
 
         // Reschedule the other cowns.
         for (size_t s = 0; s < (senders_count - 1); s++)
-          senders[s]->schedule();
+          if (senders[s])
```
nwf, line 1207: Some braces might not hurt here, or at least would have stopped me from tripping up.  Yes, `for() if() s;` generates the right AST, but `for() { if() s; }` is a bit clearer to read.

src/rt/sched/cown.h
```diff
@@ -1199,10 +1204,18 @@ namespace verona::rt
 
         // Reschedule the other cowns.
         for (size_t s = 0; s < (senders_count - 1); s++)
-          senders[s]->schedule();
+        {
+          if (senders[s])
+            senders[s]->schedule();
+        }
+
+        bool should_break = senders[senders_count - 1] == nullptr;
```
mjp41, line 1212: I don't think `should_break` can be false now?  I think this is dead code from previous implementation.

mjp41 (approved):
LGTM. Couple of minor bits, but can be merged once they are addressed.

mjp41 (commented):
You've committed a change to the snmalloc version.

mjp41 (approved):

PR #507, opened 2021-10-01
Update the process sandbox code for snmalloc2.

davidchisnall:
Snmalloc now has a single point for providing configurations, which
cleans up some of this code.  This is now used for both of our uses of
snmalloc.

In this implementation, the in-sandbox allocator uses most of the
generic snmalloc infrastructure but uses RPC to get chunks of address
space to use.  There's scope for some caching here but it's not
currently done.

The parent process has a customised snmalloc implementation that
does not do any of the lazy initialisation.  This implementation
allocates `Metaslab`s outside of the shared memory region so all
metadata is now trusted.  This implementation should check any pointers
it walks in allocatable memory (free lists and message queues) but does
not yet.

### DIFF
--- a/experiments/process_sandbox/include/process_sandbox/sandbox.h
+++ b/experiments/process_sandbox/include/process_sandbox/sandbox.h
@@ -4,6 +4,7 @@
 #pragma once
 
 #include <memory>
+#include <mutex>
 #include <string.h>
 #include <tuple>
 #include <vector>
@@ -20,40 +21,13 @@
 
 #include <snmalloc.h>
 
-#ifndef SANDBOX_PAGEMAP
-#  ifdef SNMALLOC_DEFAULT_PAGEMAP
-#    define SANDBOX_PAGEMAP SNMALLOC_DEFAULT_PAGEMAP
-#  else
-#    define SANDBOX_PAGEMAP snmalloc::SuperslabMap
-#  endif
-#endif
-
-namespace snmalloc
-{
-  template<class T>
-  class MemoryProviderStateMixin;
-  template<class T>
-  class PALPlainMixin;
-  template<
-    bool (*NeedsInitialisation)(void*),
-    void* (*InitThreadAllocator)(function_ref<void*(void*)>),
-    class MemoryProvider,
-    class ChunkMap,
-    bool IsQueueInline>
-  class Allocator;
-  template<typename T>
-  struct SuperslabMap;
-  void* no_replacement(void*);
-}
-
 namespace sandbox
 {
   struct SharedMemoryRegion;
-  struct SharedPagemapAdaptor;
-  struct MemoryProviderBumpPointerState;
   class CallbackDispatcher;
   class ExportedFileTree;
   struct CallbackHandlerBase;
+  class Library;
 
   /**
    * An snmalloc Platform Abstraction Layer (PAL) that cannot be used to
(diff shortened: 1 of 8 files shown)


experiments/process_sandbox/include/process_sandbox/sandbox.h
```diff
@@ -20,95 +21,383 @@
...
+   * Globals used by the snmalloc instances that allocate sandbox memory from
+   * the outside.
    */
-  class SharedMemoryProvider
-  : public snmalloc::MemoryProviderStateMixin<NoOpPal>
+  struct SharedAllocConfig : public snmalloc::CommonConfig
   {
     /**
-     * Base address of the shared memory region.
+     * Forward definition of `Pagemap` so that `LocalState` can refer to it.
```
ihaller, line 51: Is this still the case, since I cannot see the reference in the code, but I might be missing it?

experiments/process_sandbox/src/library_runner.cc
```diff
@@ -910,7 +870,7 @@ int main()
   close(PageMapPage);
   callbackSocket.reset(FDSocket);
 
-#ifndef NDEBUG
+  //#ifndef NDEBUG
```
ihaller, line 873: Is this commented line still relevant?

ihaller:
This comment is not related to this PR in particular, but I had a hard time figuring out which FlagPagemap instance owned the pagemap memory and which does not as it depends on the specific `init` method used. Once I understood that I need to look at the `init` usage it all made sense, but it cannot be signaled at the declaration at the moment.

davidchisnall:
> This comment is not related to this PR in particular, but I had a hard time figuring out which FlagPagemap instance owned the pagemap memory and which does not as it depends on the specific `init` method used. Once I understood that I need to look at the `init` usage it all made sense, but it cannot be signaled at the declaration at the moment.

I'm not sure I understand this comment but it sounds like something that we should try to address.  Can you give a bit more detail?

experiments/process_sandbox/include/process_sandbox/sandbox.h
```diff
@@ -20,95 +21,383 @@
...
+    {
+      /**
+       * Concrete instance of a pagemap.  This is updated only with
+       * `pagemap_lock` held.
+       */
+      inline static snmalloc::FlatPagemap<
+        snmalloc::MIN_CHUNK_BITS,
+        snmalloc::MetaEntry,
+        NoOpPal,
+        /*fixed range*/ false>
```
ihaller, line 168: Maybe a comment here to clarify that this pagemap does have a fixed range defined by the base and top, but uses the non-fixed range implementation.

ihaller:
SharedAllocConfig::Pagemap::pagemap, SnmallocGlobals::Pagemap::pagemap, SnmallocGlobals::PrivatePagemap::pagemap all have the same type. But the last one one represents a full pagemap (and manages its memory), while the first two are references to a pagemap memory range, whose lifetime is managed externally. But I have no way to infer this information by looking at the declaration, I need to see the usage with `init`, if it takes a pointer or not.

ihaller:
> > This comment is not related to this PR in particular, but I had a hard time figuring out which FlagPagemap instance owned the pagemap memory and which does not as it depends on the specific `init` method used. Once I understood that I need to look at the `init` usage it all made sense, but it cannot be signaled at the declaration at the moment.
> 
> I'm not sure I understand this comment but it sounds like something that we should try to address. Can you give a bit more detail?

SharedAllocConfig::Pagemap::pagemap, SnmallocGlobals::Pagemap::pagemap, SnmallocGlobals::PrivatePagemap::pagemap all have the same type. But the last one one represents a full pagemap (and manages its memory), while the first two are references to a pagemap memory range, whose lifetime is managed externally. But I have no way to infer this information by looking at the declaration, I need to see the usage with init, if it takes a pointer or not.

experiments/process_sandbox/include/process_sandbox/sandbox.h
davidchisnall: Yup, dead code.  Thanks!

experiments/process_sandbox/include/process_sandbox/sandbox.h
davidchisnall: Added, thanks.

experiments/process_sandbox/src/library_runner.cc
davidchisnall: Ooops, I commented out this block while refactoring and missed readding it.  Thanks!

davidchisnall:
I see.  I've added some more comments here, let me know what you think.

experiments/process_sandbox/src/library_runner.cc
```diff
@@ -168,123 +170,90 @@ namespace
...
+  }
+
+  std::pair<snmalloc::CapPtr<void, snmalloc::CBChunk>, snmalloc::Metaslab*>
+  SnmallocGlobals::alloc_chunk(
+    SnmallocGlobals::LocalState*,
+    size_t size,
+    snmalloc::RemoteAllocator* remote,
+    snmalloc::sizeclass_t sizeclass)
+  {
+    // TODO: Check this doesn't actually write to the pagemap.
```
nwf, line 244: It looks like it's OK if it does, now that it's pointing at the private pagemap?

experiments/process_sandbox/src/libsandbox.cc
```diff
@@ -215,65 +219,48 @@ namespace sandbox
...
-        }
+        range_size = snmalloc::sizeclass_to_size(m.get_sizeclass());
       }
-      return safe;
+      // Metadata updates must refer to addresses associated with the sandbox.
+      if (!s->contains(base_address, range_size))
+      {
+        return false;
+      }
+      // The message queue must be in the allocator.
```
nwf, line 244:
```suggestion
      // The message queue must be in the sandbox.
```

experiments/process_sandbox/src/libsandbox.cc
```diff
@@ -692,20 +571,44 @@ namespace sandbox
   Library::~Library()
   {
     wait_for_child_exit();
+    {
+      auto [g, pm] = SharedAllocConfig::Pagemap::get_pagemap_writeable();
+      snmalloc::address_t base =
+        snmalloc::address_cast(memory_provider.get_base());
+      auto top = snmalloc::address_cast(memory_provider.top_address());
+      // Scan the pagemap for all memory associated with this and deallocate
+      // the metaslabs.  Note that we don't need to do any cleanup for the
```
nwf, line 580:
```suggestion
      // our metaslabs.  Note that we don't need to do any cleanup for the
```

experiments/process_sandbox/src/libsandbox.cc
```diff
@@ -692,20 +571,44 @@ namespace sandbox
...
+      // memory referenced by these metaslabs: it will all go away when the
+      // shared memory region is deallocated.
```
nwf, line 582: And similarly for any metaslabs (and their slabs' memory) not in the shared region.

experiments/process_sandbox/src/libsandbox.cc
```diff
@@ -692,20 +571,44 @@ namespace sandbox
...
+      {
+        auto& meta =
+          SharedAllocConfig::Pagemap::get_metaentry(&memory_provider, a);
+        auto* remote = meta.get_remote();
+        if (
+          (remote != nullptr) &&
+          !contains(remote, sizeof(snmalloc::RemoteAllocator)))
+        {
+          delete meta.get_metaslab();
+        }
```
nwf, line 593: Will we not construct large allocations in the shared region?  Those will have a fake remote (i.e., `nullptr`) but will have `ChunkRecord`s pointed to by their `MetaEntry`.  I think this leaks those, if they exist?

ihaller:
> I see. I've added some more comments here, let me know what you think.

Looks good, thanks for the clarifying comments!

experiments/process_sandbox/include/process_sandbox/sandbox.h
```diff
@@ -20,95 +21,383 @@
...
+      }
+
+      /**
+       * Predicate to test whether an object of size `sz` starting at `ptr`
+       * is within the region managed by this memory provider.
+       */
+      bool contains(const void* ptr, size_t sz)
+      {
+        // We shouldn't need the const cast here, but pointer_offset doesn't
+        // correctly handle const pointers yet.
```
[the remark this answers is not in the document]
mjp41: I was more worried about `reinterpret_cast` being hidden from the caller.  In `pointer_offset` this is inherently a pretty unsafe thing to do, so I am less concerned.

experiments/process_sandbox/include/process_sandbox/sandbox.h
```diff
@@ -20,95 +21,383 @@
...
+        return (ptr >= base) &&
+          (pointer_offset(const_cast<void*>(ptr), sz) < top);
```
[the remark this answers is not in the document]
mjp41: I always prefer calculating the diff, e.g.
```
pointer_diff(ptr, top) >= sz
```
style check, as it is as efficient, and doesn't have explicit overflow checking required.

experiments/process_sandbox/include/process_sandbox/sandbox.h
```diff
@@ -20,95 +21,382 @@
...
+       */
+      static auto& get_pagemap_handle()
+      {
+        return pagemap_mem->get_handle();
+      }
+
+      /**
+       * Reserve a chunk of memory in the shared address space for this sandbox.
+       */
+      static void* reserve(LocalState& local_state, size_t size)
```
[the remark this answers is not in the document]
davidchisnall: The convention is to use pointers only for nullable things and references where they're not optional. This is an internal API that can be used only when the `LocalState` exists.

PR #509, opened 2021-10-06
Add allocation support for reference counted region

jacob-hughes:
This provides a way to allocate new objects in an rc region; the first of many commits needed for full RC support.

A fundamental property of reference counting is deterministic destruction: once an object is decrefed to zero, it is finalized and
deallocated immediately.  Using a singly linked list 'ring' similar to the traced region would not be efficient here, since an O(n) scan over the region would be needed on each removal.

Instead, the rc region keeps track of its objects using "region vectors": which contain a pointer to each object, and its reference count as a pair. A region has two region vectors for tracking both trivial and non-trivial objects separately.

Unlike a stack, removal from the region vector can happen at any index. It is also O(1) because a removed item will leave a hole in the vector rather than shifting remaining items down. To reduce fragmentation which can occur with deallocation churn, the region vector maintains a freelist which is threaded through the holes left in the vector. Insertion of new items will first query the freelist to see if a hole can be reused, otherwise items are bump allocated.

This commit primarily consists of the implementation of the region vectors and their interaction with allocation on the rc region.

### DIFF
--- a/src/rt/region/region_rc.h
+++ b/src/rt/region/region_rc.h
@@ -69,10 +69,11 @@ namespace verona::rt
     // Index of the full dummy block
     // Due to pointer arithmetic with nullptr being undefined behaviour
     // we use a statically allocated null block.
-    static constexpr MaybeElem* null_index = &(null_block.data[BLOCK_COUNT - 1]);
+    static constexpr MaybeElem* null_index =
+      &(null_block.data[BLOCK_COUNT - 1]);
 
     /// Mask to access the index component of the pointer to a block.
-    static constexpr uintptr_t INDEX_MASK = (ITEM_COUNT- 1) * sizeof(T);
+    static constexpr uintptr_t INDEX_MASK = (ITEM_COUNT - 1) * sizeof(T);
 
     /// Pointer into a block.  As the blocks are strongly aligned
     /// the bits 9-3 represent the element in the block, with 0 being
@@ -123,8 +124,8 @@ namespace verona::rt
 
     ALWAYSINLINE void remove(Elem* index)
     {
-      MaybeElem* hole = (MaybeElem*) index;
-      hole->hole_ptr = (MaybeElem*) ((uintptr_t) next_free | EMPTY_MASK);
+      MaybeElem* hole = (MaybeElem*)index;
+      hole->hole_ptr = (MaybeElem*)((uintptr_t)next_free | EMPTY_MASK);
       next_free = hole;
     }
 
@@ -149,7 +150,7 @@ namespace verona::rt
         MaybeElem* prev = next_free->hole_ptr;
         next_free->item = item;
         MaybeElem* cur = next_free;
-        next_free = (MaybeElem*)((uintptr_t) prev & ~EMPTY_MASK);
+        next_free = (MaybeElem*)((uintptr_t)prev & ~EMPTY_MASK);
         return &(cur->item);
       }
       if (!is_full(index))
@@ -198,7 +199,7 @@ namespace verona::rt
       iterator operator++()
       {
         ptr--;
-        auto maybe_ptr = (MaybeElem*) ptr;
+        auto maybe_ptr = (MaybeElem*)ptr;
         if (maybe_ptr != vec->null_index)
         {
           if (!vec->is_empty(maybe_ptr))
@@ -381,7 +382,7 @@ namespace verona::rt
         oc = reg->get_trivial_vec()->push({o, 1}, alloc);
       else
         oc = reg->get_non_trivial_vec()->push({o, 1}, alloc);
-      o->set_rv_index((Object*) oc);
+      o->set_rv_index((Object*)oc);
 
       // GC heuristics.
       reg->use_memory(desc->size);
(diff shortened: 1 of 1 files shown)


jacob-hughes:
This will teach me for only running a debug build locally. `-Wstrict-aliasing` only warns when `-On` >1 or `-fstrict-aliasing` is explicitly enabled. I'll push a fix for the objectionable type punning I'm doing shortly.

src/rt/region/region_rc.h
```diff
@@ -0,0 +1,391 @@
...
+        auto prev = get_block(prev_ptr);
+        alloc.template dealloc<sizeof(Block)>(local_block);
+        local_block = prev;
+      }
+      index = null_index;
+    }
+
+    ALWAYSINLINE void remove(T* index)
+    {
+      *(uintptr_t*)index = ((uintptr_t)next_free | EMPTY_MASK);
```
mjp41, line 109: Assert that the `next_free` does not have the bit set, or perhaps better `index`

src/rt/region/region_rc.h
```diff
@@ -0,0 +1,391 @@
...
+    {
+      friend class RegionVector;
+
+    public:
+      iterator(RegionVector<T, Alloc>* vec) : vec(vec)
+      {
+        // If the vec is empty then there is nothing to iterate over.
+        if (vec->empty())
+        {
+          ptr = nullptr;
```
mjp41, line 172: Can you use `null_index` here?

src/rt/region/region_rc.h
```diff
@@ -0,0 +1,391 @@
...
+        return ptr;
+      }
+      inline iterator begin()
+      {
+        return {vec};
+      }
+
+      inline iterator end()
+      {
+        return {vec, nullptr};
```
mjp41, line 223: Can this be `null_index`

src/rt/region/region_rc.h
```diff
@@ -0,0 +1,391 @@
...
+
+      iterator operator++()
+      {
+        ptr--;
+        if (ptr != vec->null_index)
+        {
+          if (!vec->is_empty(ptr))
+          {
+            next_non_empty();
+            return *this;
```
mjp41, line 189: Is this effectively 
```
return this++;
```

src/rt/region/region_rc.h
```diff
@@ -0,0 +1,391 @@
...
+          }
+          ptr = *(T**)&(vec->get_block(ptr)->prev);
+          if (ptr != vec->null_index)
+          {
+            next_non_empty();
+            return *this;
+          }
+        }
+        ptr = nullptr;
```
mjp41, line 198: I think this can be dropped.

mjp41 (commented):
Some comments on the Vector class.

src/rt/object/object.h
```diff
@@ -77,6 +77,12 @@ namespace verona::rt
   class Object;
   class RegionBase;
 
+  struct ObjectCount
```
mjp41, line 80: Could you add a comment saying what this class is for?

src/rt/test/func/memory/memory_rc.h
```diff
@@ -0,0 +1,102 @@
...
+        rm2 = object_count;
+
+      if (count == 100)
+        rm3 = object_count;
+      count++;
+    }
+
+    assert(count == NUM_OBJECTS);
+
+    reg->get_trivial_vec()->remove(rm3);
```
mjp41, line 54: Could we expose a more useful primitive on the region.
```
RegionRC::decref(RegionRC*, Object*)
RegionRC::incref(RegionRC*, Object*)
```
The current API requires too much knowledge of the implementation

mjp41: This also makes me less certain of separating the two stacks.  If we have to branch on which stack for each incref and decref, then this is going to potentially cost way more than we save at teardown.

src/rt/region/region_rc.h
jacob-hughes: Yeah, there is definitely an implicit assumption that `T` starts with a pointer. I think a constraint here is a good idea.

src/rt/region/region_rc.h
jacob-hughes: Ack

src/rt/region/region_rc.h
```diff
@@ -0,0 +1,391 @@
...
+        }
+
+        ptr = vec->peek();
+      }
+
+      iterator(RegionVector<T, Alloc>* vec, T* p) : vec(vec), ptr(p) {}
+
+      iterator operator++()
+      {
+        ptr--;
```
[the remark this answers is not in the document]
jacob-hughes: Ack. Good Idea.

src/rt/test/func/memory/memory_rc.h
mjp41: Your right.  I guess it is only for decref, where you hit empty, and then you have to put it on the relevant free list, but then you have to inspect the header anyway.

src/rt/region/region_rc.h
jacob-hughes: I've reworked this now to be less generic and use unions to specify that the vector might have holes in it (fd32d6).

PR #511, opened 2021-10-08
replace a few relative links with absolute ones

akkartik:
Links should now work from both https://microsoft.github.io/verona/explore.html and https://github.com/microsoft/verona/blob/master/docs/explore.md


rengolin (approved):
Thanks!

PR #512, opened 2021-10-11
Add bag data structure

jacob-hughes:
A bag is an vector-like data structure using aligned block which is optimised for constant time insertion and removal.

Removal is O(1) because a removed item will leave a hole in the bag rather than shifting remaining items down like in a bag. To reduce fragmentation which can occur with deallocation churn, the bag maintains a freelist which is threaded through the holes left in the bag. Insertion of new items will first query the freelist to see if a hole can be reused, otherwise items are bump allocated.

This data structure is useful for tracking objects in regions -- such as one which uses reference counting -- where efficient alloc/dealloc is necessary.

### DIFF
--- a/src/rt/ds/bag.h
+++ b/src/rt/ds/bag.h
@@ -1,0 +1,234 @@
+// Copyright Microsoft and Project Verona Contributors.
+// SPDX-License-Identifier: MIT
+#pragma once
+
+#include <cassert>
+#include <snmalloc.h>
+
+namespace verona::rt
+{
+  /**
+   * This class contains the core functionality for a bag using aligned blocks
+   * which is optimised for constant time insertion and removal.
+   *
+   * Removal is O(1) because a removed item will leave a hole in the bag rather
+   * than shifting remaining items down like in a bag. To reduce fragmentation
+   * which can occur with deallocation churn, the bag maintains a freelist which
+   * is threaded through the holes left in the bag.  Insertion of new items will
+   * first query the freelist to see if a hole can be reused, otherwise items
+   * are bump allocated.
+   *
+   * To maintain an internal freelist with no additional space requirements, the
+   * item `T` must be at least 1 machine word in size.
+   */
+  template<class T, class U, class Alloc>
+  class Bag
+  {
+  public:
+    struct Elem
+    {
+      T* object;
+      U metadata;
+    };
+
+  private:
+    static constexpr size_t ITEM_COUNT = 32;
+    static_assert(
+      snmalloc::bits::next_pow2_const(ITEM_COUNT) == ITEM_COUNT,
+      "Should be power of 2 for alignment.");
+
+    static constexpr size_t BLOCK_COUNT = ITEM_COUNT - 1;
+    static constexpr uintptr_t EMPTY_MASK = 1 << 0;
+
+    // An element in the Bag may be empty. Where this is the case, a
+    // freelist is threaded through holes so that the element's slot can be
+    // reused. This is encoded as the `hole_ptr` field in the union.
+    union MaybeElem
+    {
+      MaybeElem* hole_ptr;
+      Elem item;
+    };
+
+    struct alignas(ITEM_COUNT * sizeof(MaybeElem)) Block
+    {
+      MaybeElem prev;
+      MaybeElem data[BLOCK_COUNT];
+    };
+
+    // Dummy block to effectively allow pointer arithmetic on nullptr
+    // which is undefined behaviour.  So we statically allocate a block
+    // to represent the end of the bag.
+    inline static Block null_block{};
+
+    // Index of the full dummy block
+    // Due to pointer arithmetic with nullptr being undefined behaviour
+    // we use a statically allocated null block.
+    static constexpr MaybeElem* null_index =
+      &(null_block.data[BLOCK_COUNT - 1]);
+
+    /// Mask to access the index component of the pointer to a block.
+    static constexpr uintptr_t INDEX_MASK =
+      (ITEM_COUNT - 1) * sizeof(MaybeElem);
+
+    /// Pointer into a block.  As the blocks are strongly aligned
+    /// the bits 9-3 represent the element in the block, with 0 being
+    /// a pointer to the `prev` pointer, and implying the empty block.
+    MaybeElem* index;
+
+    // Used to thread a freelist pointer through the bag.
+    MaybeElem* next_free;
+
+    /// Takes an index and returns the pointer to the Block
+    static Block* get_block(MaybeElem* ptr)
+    {
+      return snmalloc::pointer_align_down<sizeof(Block), Block>(ptr);
+    }
+
+    /// Checks if an index into a block means the block is empty.
+    static bool is_first_block_elem(MaybeElem* ptr)
+    {
+      return ((uintptr_t)ptr & INDEX_MASK) == 0;
+    }
+
+    /// Checks if an index into a block means the block has no space.
+    static bool is_last_block_elem(MaybeElem* index)
+    {
+      return ((uintptr_t)index & INDEX_MASK) == INDEX_MASK;
+    }
+
+  public:
+    Bag<T, U, Alloc>() : index(null_index), next_free(nullptr)
+    {
+      static_assert(
+        sizeof(*this) == sizeof(void*) * 2,
+        "Stack should contain only the index and freelist pointer");
+    }
+
+    /// Deallocate the linked blocks for this bag.
+    void dealloc(Alloc& alloc)
+    {
+      auto local_block = get_block(index);
+      while (local_block != &null_block)
+      {
+        auto prev = get_block(local_block->prev.hole_ptr);
+        alloc.template dealloc<sizeof(Block)>(local_block);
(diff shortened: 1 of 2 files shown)


jacob-hughes:
This PR is a WIP because I'm still working through a bug with the `INDEX_MASK`. Once this is fixed I'll push additional tests too. But I think it's useful @mjp41 for you to comment on the new union based approach for elements inside the bag.

src/rt/ds/bag.h
```diff
@@ -0,0 +1,255 @@
...
+      return &(index->item);
+    }
+
+    /// Insert an element into the bag.
+    ALWAYSINLINE Elem* insert(Elem item, Alloc& alloc)
+    {
+      if (next_free != nullptr)
+      {
+        MaybeElem* prev = next_free->hole_ptr;
+        next_free->item = item;
```
mjp41, line 146: Can you assert here that the bit is set.

src/rt/ds/bag.h
```diff
@@ -0,0 +1,246 @@
...
+        return ptr;
+      }
+      inline iterator begin()
+      {
+        return {bag};
+      }
+
+      inline iterator end()
+      {
+        return {bag, nullptr};
```
mjp41, line 238:
```suggestion
        return {bag, bag->null_index};
```

src/rt/ds/bag.h
```diff
@@ -0,0 +1,246 @@
...
+      iterator operator++()
+      {
+        auto p = (MaybeElem*)ptr;
+        do
+        {
+          p--;
+          if (p->hole_ptr == bag->null_index)
+          {
+            ptr = nullptr;
+            return *this;
```
mjp41, line 204:
```suggestion
            break;
```

src/rt/ds/bag.h
```diff
@@ -0,0 +1,246 @@
...
+          }
+
+          if (bag->is_empty(p))
+          {
+            p = bag->get_block(p)->prev.hole_ptr;
+          }
+        } while ((uintptr_t)p->hole_ptr & EMPTY_MASK);
+
```
mjp41, line 212:
```suggestion
        assert(
          (p == bag.null_index) || 
          (((uintptr_t)p->hole_ptr & EMPTY_MASK) == 0));
```

src/rt/ds/bag.h
```diff
@@ -0,0 +1,246 @@
...
+      iterator(Bag<T, U, Alloc>* bag, Elem* p) : bag(bag), ptr(p) {}
+
+      iterator operator++()
+      {
+        auto p = (MaybeElem*)ptr;
+        do
+        {
+          p--;
+          if (p->hole_ptr == bag->null_index)
```
mjp41, line 201:
```suggestion
          if (p->hole_ptr == Bag::null_index)
```

src/rt/test/func/bag/bag.cc
```diff
@@ -0,0 +1,84 @@
...
+  {
+    UNUSED(item);
+    count++;
+  }
+  assert(count == NUM_OBJECTS - 3);
+
+  // Check that the freelist is used to fill existing holes before bump
+  // allocation.
+  auto idx1 = bag.insert({nullptr, 123}, alloc);
+  assert((uintptr_t)rm1 == (uintptr_t)idx1);
```
mjp41, line 69: This is very strict test that imposes a stack like ordering of the free list.  Is there any easy way to make it only impose the constraints we really want.

How about effectively saying the sets are equal? `{rm1, rm2, rm3} == {idx1,idx2,idx3}`

src/rt/test/func/bag/bag.cc
```diff
@@ -0,0 +1,84 @@
...
+
+  auto idx2 = bag.insert({nullptr, 456}, alloc);
+  assert((uintptr_t)rm2 == (uintptr_t)idx2);
+
+  auto idx3 = bag.insert({nullptr, 789}, alloc);
+  assert((uintptr_t)rm3 == (uintptr_t)idx3);
+
```
mjp41, line 79: Can you allocate one more that won't be using free list at this point to get the coverage for changing mode.

src/rt/test/func/bag/bag.cc
```diff
@@ -0,0 +1,87 @@
...
+  bag.remove(rm1);
+
+  // Check that the iterator skips over holes left in the region vector
+  count = 0;
+  for (auto item : iter)
+  {
+    UNUSED(item);
+    count++;
+  }
+  assert(count == NUM_OBJECTS - 3);
```
mjp41, line 64: Don't use assert in tests.  This wants to be checked on release builds too.   We have a macro that does `check`s:
```suggestion
  check(count == NUM_OBJECTS - 3);
```

src/rt/ds/bag.h
jacob-hughes: Why do we want to return `null_index` here? I thought the iterator advancement will return an `Elem*`, and when no more elements exist will return `nullptr`. Isn't exposing `null_index` (which is a `MaybeElem*`) an implementation detail?

src/rt/test/func/bag/bag.cc
jacob-hughes: What about something like this ? 57cbf756da87743cd0237dfb6989bff8a9462279

src/rt/test/func/bag/bag.cc
jacob-hughes: Added as part of 57cbf756da87743cd0237dfb6989bff8a9462279

src/rt/ds/bag.h
```diff
@@ -0,0 +1,250 @@
...
+    public:
+      iterator(Bag<T, U, Alloc>* bag) : bag(bag)
+      {
+        // If the bag is empty then there is nothing to iterate over.
+        if (bag->empty())
+        {
+          ptr = nullptr;
+          return;
+        }
+
```
mjp41, line 189: If you used `null_index` as end, then this code would go away.

src/rt/ds/bag.h
mjp41: If you do this, you will end up with less.  Code, the bag implementation uses `null_index` to signify when it has reached the end, but your implementation of iterator uses a different constant to signify the end.

This is not being exposed, it is the value of a private field.

src/rt/test/func/bag/bag.cc
```diff
@@ -0,0 +1,95 @@
...
+#include "test/log.h"
+#include "verona.h"
+
+using namespace snmalloc;
+using namespace verona::rt;
+
+void test()
+{
+  auto& alloc = ThreadAlloc::get();
+
```
mjp41, line 16: Perhaps add
```C++
using B = Bag<uintptr_t, uintptr_t, Alloc>;
using E = B::Elem*;
```
and then replace the uses below.

src/rt/test/func/bag/bag.cc
```diff
@@ -0,0 +1,95 @@
...
+      rm2 = entry;
+
+    if (count == 100)
+      rm3 = entry;
+    count++;
+  }
+
+  check(count == NUM_OBJECTS);
+
+  std::set<uintptr_t> removed = {
```
mjp41, line 54: If you add the shorthands above, then this can be:
```suggestion
  std::set<E> removed = {
```
and then you can drop all the casts to `uintptr_t`.  Or is this because you need a less defined on the type?  Does `unordered_set` help here?  You aren't traversing in an order anyway?

src/rt/ds/bag.h
```diff
@@ -0,0 +1,250 @@
...
+      }
+
+      inline iterator end()
+      {
+        return {bag, nullptr};
+      }
+
+    private:
+      Bag<T, U, Alloc>* bag;
+      Elem* ptr;
```
mjp41, line 247: I guess given my suggestion this would have to be
```suggestion
      MaybeElem* ptr;
```
```suggestion
      Elem* ptr;
```

src/rt/ds/bag.h
```diff
@@ -0,0 +1,250 @@
...
+      }
+
+      inline bool operator==(const iterator& other) const
+      {
+        return ptr == other.ptr;
+      }
+
+      inline Elem* operator*() const
+      {
+        return ptr;
```
mjp41, line 233: This would have to change to
```suggestion
        return ptr->item;
```
```suggestion
        return ptr;
```

src/rt/test/func/bag/bag.cc
jacob-hughes: I need a less defined type here as I just want to compare the pointer addresses. I've changed it to an `unordered_set` though: 85ae978cb22f4acb0bf1ffd956f62d9e95b4282f

jacob-hughes:
Addressed comments here 85ae978cb22f4acb0bf1ffd956f62d9e95b4282f and refactored the pointer while loop into a `next_non_empty` function. I've also removed the `peek` function because I think it's misleading as it implies stack like behaviour on the Bag: we make no such ordering guarantees so the last inserted item may not be what `peek` returns.

jacob-hughes:
I've reworked the `next_non_empty` method to use a `step` method instead now, and removed the misleading `empty` method.

PR #513, opened 2021-10-13
Fix Topology to work with repeated runtime starts

mjp41:
The topology was not working correctly if it was run twice.  This change
makes it static, and thus does not need to query the CPUs repeatedly.

While writing some new tests I observed that after the first run, when the
runtime restarted, all the threads had been affinitized to a single core. This
makes the topology static, and just initialises it on first use.

I also made the vector an embedded field for simpler memory/lifetime
management.

### DIFF
--- a/src/rt/pal/cpu.h
+++ b/src/rt/pal/cpu.h
@@ -87,7 +87,7 @@ namespace verona::rt
       }
     };
 
-    std::vector<CPU>* cpus = nullptr;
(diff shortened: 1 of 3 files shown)


mjp41:
@plietar just tagging you as I think you wrote the code originally.

mjp41:
> Would it make sense to make it a Singleton with late initialization instead of acquire?

I've used the `snmalloc::Singleton` now.

PR #518, opened 2021-10-20
Add basic support for a reference counted region

jacob-hughes:
This includes: allocation, deallocation, and finalisation of objects in the reference counted region; incref, and decref methods; and support for releasing the entire region (and its subregions).

Merging, root swapping, and reclamation of cyclic data structures are not yet implemented.

A fundamental property of reference counting is deterministic destruction: once an object is decrefed to zero, it is finalized and deallocated immediately. Using a singly linked list 'ring' similar to the traced region would not be efficient here, since an O(n) scan over the region would be needed on each removal.

Instead, the rc region keeps track of its objects using the bag data structure [1]. Each element in the bag contain a pointer to each object in the region, along with its reference count. A spare bit in the object pointer is used to quickly deduce whether an object is trivial or non-trivial.

[1]: https://github.com/microsoft/verona/pull/512

### DIFF
--- a/src/rt/object/object.h
+++ b/src/rt/object/object.h
@@ -2,6 +2,7 @@
 // SPDX-License-Identifier: MIT
 #pragma once
 
+#include "../ds/bag.h"
 #include "../ds/stack.h"
 #include "../test/systematic.h"
 
@@ -76,6 +77,9 @@ namespace verona::rt
   using namespace snmalloc;
   class Object;
   class RegionBase;
+
+  using RefCounts = Bag<Object, uintptr_t, Alloc>;
+  using RefCount = RefCounts::Elem;
 
   using ObjectStack = Stack<Object, Alloc>;
   static constexpr size_t descriptor_alignment =
@@ -423,6 +427,7 @@ namespace verona::rt
     friend class RegionBase;
     friend class RegionTrace;
     friend class RegionArena;
+    friend class RegionRc;
     friend class RememberedSet;
     friend class ExternalReferenceTable;
     template<typename Entry>
@@ -475,6 +480,22 @@ namespace verona::rt
         (get_class() == RegionMD::PENDING) || (get_class() == RegionMD::ISO));
 
       return (Object*)(get_header().bits & ~MASK);
+    }
+
+    inline RefCount* get_ref_count()
+    {
+      assert(
+        (get_class() == RegionMD::MARKED) ||
+        (get_class() == RegionMD::UNMARKED));
+      return (RefCount*)(get_header().bits & ~MASK);
+    }
+
+    inline void set_ref_count(RefCount* rc)
+    {
+      // The MARKED/UNMARKED tags are not used by RegionRc, so we don't need to
+      // bother restoring the bit, and can simply leave it unset (i.e.
+      // UNMARKED).
+      get_header().next = (Object*)rc;
     }
 
     inline void set_next(Object* o)
--- a/src/rt/region/region_rc.h
+++ b/src/rt/region/region_rc.h
@@ -1,0 +1,428 @@
+// Copyright Microsoft and Project Verona Contributors.
+// SPDX-License-Identifier: MIT
+#pragma once
+
+#include "../object/object.h"
+#include "region_arena.h"
+#include "region_base.h"
+#include "region_trace.h"
+
+namespace verona::rt
+{
+  using namespace snmalloc;
+
+  /**
+   * Please see region.h for the full documentation.
+   *
+   * This is a concrete implementation of a region, specifically one with
+   * reference counting. This class inherits from RegionBase, but it cannot call
+   * any of the static methods in Region.
+   *
+   * In a rc region, all objects are tracked using a bag; where each element
+   * contains a pointer to the object and its current reference count. The bag
+   * is updated each time an object is allocated or deallacted.
+   *
+   * The RegionRc uses a spare bit in each object pointer in the bag
+   * (`FINALIZER_MASK`) to quickly deduce whether an object is trivial or
+   * non-trivial.
+   *
+   **/
+  class RegionRc : public RegionBase
+  {
+    friend class Freeze;
+    friend class Region;
+    friend class RegionTrace;
+
+  private:
+    static constexpr uintptr_t FINALISER_MASK = 1 << 1;
+
+    RefCounts counts{};
+
+    RefCount* entry_point_count = nullptr;
+
+    // Memory usage in the region.
+    size_t current_memory_used = 0;
+
+    RegionRc() : RegionBase() {}
+
+    static const Descriptor* desc()
+    {
+      static constexpr Descriptor desc = {
+        vsizeof<RegionRc>, nullptr, nullptr, nullptr};
+
+      return &desc;
+    }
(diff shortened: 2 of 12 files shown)


src/rt/test/func/memory/memory_iterator.h
```diff
@@ -280,14 +280,107 @@ namespace memory_iterator
...
+      auto* reg = RegionRc::get(r);
+      /* s1.insert(r); */
+      /* s2.insert(r); // Note: might need to change this if o's type is
+       * changed. */
+
+      // Add some objects.
+      oc = new (alloc, r) C;
+      s1.insert(oc);
+      s2.insert(oc);
+
```
mjp41, line 306: Don't we leak `oc` here, as we aren't handling the RC correctly?

src/rt/test/func/memory/memory_rc.h
```diff
@@ -0,0 +1,147 @@
...
+    RegionRc::incref(o);
+    o5->f1 = o;
+
+    check(Region::debug_size(o) == 6);
+    RegionRc::decref(alloc, o, o);
+    check(Region::debug_size(o) == 6);
+
+    // Break the cycle
+    RegionRc::decref(alloc, o, o);
+    check(Region::debug_size(o) == 0);
```
mjp41, line 137: Perhaps, enable this test, but comment out the assertion here saying that it will be enabled when cycle detection is enabled.

src/rt/region/region_rc.h
```diff
@@ -0,0 +1,408 @@
...
+      o->init_iso();
+      o->set_region(reg);
+
+      assert(Object::debug_is_aligned(o));
+      return o;
+    }
+
+    /**
+     * Allocates an object `o` of type `desc` in the region represented by the
+     * Iso object `in`, and adds it to the appropriate ring. Returns a pointer
```
mjp41, line 103: Comment out of date? "appropriate ring"

src/rt/object/object.h
```diff
@@ -477,6 +482,16 @@ namespace verona::rt
...
     }
 
+    inline RefCount* get_ref_count()
+    {
+      return (RefCount*)get_header().next;
+    }
+
+    inline void set_ref_count(RefCount* rc)
+    {
+      get_header().next = (Object*)rc;
```
mjp41, line 492: Do we need to mask bits here?

src/rt/object/object.h
```diff
@@ -477,6 +482,16 @@ namespace verona::rt
       return (Object*)(get_header().bits & ~MASK);
     }
 
+    inline RefCount* get_ref_count()
+    {
+      return (RefCount*)get_header().next;
```
mjp41, line 487: Do we need to mask bits here?

src/rt/region/region_rc.h
```diff
@@ -0,0 +1,408 @@
...
+    {
+      if (decref_inner(o))
+      {
+        dealloc_object(alloc, o, in);
+        return true;
+      }
+      return false;
+    }
+
+    static uintptr_t debug_get_ref_count(Object* o)
```
mjp41, line 150: Is this only to be used in "debug".  It looks safe to use in a non-debug case.  It should be concurrency safe, as this is purely single-threaded region code.

src/rt/region/region_rc.h
```diff
@@ -0,0 +1,408 @@
...
+      {
+        Object* p = dfs.pop();
+        switch (p->get_class())
+        {
+          case Object::ISO:
+            sub_regions.push(p);
+            break;
+          case Object::MARKED:
+          case Object::UNMARKED:
+          case Object::SCC_PTR:
```
mjp41, line 189: I am not convinced SCC_PTR, RC and Cown should be decreffed immediately.  Also, don't they need a different decref?

src/rt/region/region_rc.h
```diff
@@ -0,0 +1,408 @@
...
+
+      flzr_q.push(o);
+
+      while (!dfs.empty())
+      {
+        Object* p = dfs.pop();
+        switch (p->get_class())
+        {
+          case Object::ISO:
+            sub_regions.push(p);
```
mjp41, line 185: Can this reference the Entry point of the current region?

src/rt/region/region_rc.h
```diff
@@ -0,0 +1,408 @@
...
+      {
+        return true;
+      }
+      rc->metadata -= 1;
+      return false;
+    }
+
+    static void dealloc_object(Alloc& alloc, Object* o, Object* in)
+    {
+      // We first need to decref -- and potentially deallocate -- any object
```
mjp41, line 170: Does this need to remove from additional roots?

src/rt/region/region_rc.h
```diff
@@ -0,0 +1,408 @@
...
+    friend class Region;
+    friend class RegionTrace;
+
+  private:
+    static constexpr uintptr_t FINALISER_MASK = 1 << 1;
+
+    RefCounts counts{};
+
+    // Stack of stack based entry points into the region.
+    StackThin<Object, Alloc> additional_entry_points{};
```
mjp41, line 42: Where is this maintained?  I don't see anything being added to it?

src/rt/region/region_rc.h
jacob-hughes: I think you're right. What about this fd926797dec3db8dd8214fa80f5577f2c147ccf1? Is there something in particular that needs to happen for SCC objects too?

src/rt/region/region_rc.h
jacob-hughes: I only used this for debuggin/test code so I prefixed it with debug. I suppose you're right that it might be a useful API for non-debug usage. Fixed here: 9120678b2abdd330de73e156be48caa5537e689a

src/rt/test/func/memory/memory_iterator.h
jacob-hughes: I'm not quite sure I understand. Really this particular test is orthogonal to RC and is just checking that the iterator goes through every object in the bag. I can manually create a link and add an incref for `oc`, but I don't think it will prevent leaks because the entire region gets released after the test anyway.

src/rt/region/region_rc.h
mjp41: I think we are talking at cross purposes:

```
root.f -> o
o.f -> root
```
Let's drop the reference root.f, `root.f = nullptr`. This means `o` has a ref count of `0`, and thus starts to tear down what it can see. But one of its fields is an `iso`, `o.f` so this gets added to the `sub_regions`, but that is not correct, as there could still be references to this object from the stack.   I think you need to check if it is the current entry point, and if it isn't add it, otherwise you need to ignore it, until you support decref on the entry point.

src/rt/region/region_rc.h
```diff
@@ -185,15 +185,17 @@ namespace verona::rt
...
               flzr_q.push(p);
               p->trace(dfs);
             }
             break;
+          case Object::SCC_PTR:
+            break;
+          case Object::RC:
+          case Object::COWN:
+            p->decref();
           default:
```
mjp41, line 200: You need a break here as you are going to hit the assert.

PR #519, opened 2021-10-28
Add `BagThin` data structure

jacob-hughes:
The `BagThin` data structure is a bag where each element contains only a `T*`. Unlike `Bag`, where elements can have an additional field, `ThinBag` is optimised to be space-efficient when this is not necessary.

This change refactors the original bag implementation into a common base class which is then composed into concrete `Bag` 
 and `BagThin` implementations.

### DIFF
--- a/src/rt/ds/bag.h
+++ b/src/rt/ds/bag.h
@@ -7,32 +7,16 @@
 
 namespace verona::rt
 {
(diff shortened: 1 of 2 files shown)


mjp41 (approved):
LGTM

PR #521, opened 2021-11-09
Enhance promises

marioskogias:
This PR adds some functionality to the promises implementation in the verona runtime.

It manages the case in which the promise returns a smart pointer and it allows getting the raw promise pointer out of the read endpoint.


mjp41 (approved):
LGTM

PR #522, opened 2021-11-10
Add external event source usecase

marioskogias:
This PR exercises the path of adding and removing external event sources to keep the runtime up and running similarly to an external IO device.


mjp41:
I don't think Windows is going to like the Linux system calls.

src/rt/test/func/external_event_source/external_event_source.cc
```diff
@@ -0,0 +1,117 @@
...
+    else
+      empty_count++;
+
+    if (empty_count < 10)
+      schedule_lambda(this, [=]() { main(); });
+    else
+    {
+      // add external source and enable notifications
+      Scheduler::add_external_event_source();
+      should_notify = true;
```
mjp41, line 46: After `should_notify` you need to recheck the `Noticeboard` to see if new data has arrived, otherwise you can go to sleep here, and never be woken up.

mjp41:
@lukecheeseman do you have any comments on this?  If not, I'll merge this.

mjp41 (approved):

PR #523, opened 2021-11-12
Add network support to sandboxes.

davidchisnall:
This proxies the three POSIX socket functions that access the global
namespace:

 - bind
 - connect
 - getaddrinfo (which could be used to implement the legact IPv4
   interfaces)

These are passed through a policy layer that allows sandbox consumers to
either:

 - Block them unconditionally.
 - Allow them to work with any address.
 - Add an interposer that implements the function based on some custom
   policy.

This makes it possible to enforce policies such as 'this sandbox may
establish outbound connections to microsoft.com:80 but nothing else' by
interposing on the `getaddrinfo` call to record the valid addresses for
this address / service combination and allowing `connect` only for those
addresses.

The callback infrastructure was starting to be a bit unwieldy, so this
has been refactored to use tuples indexed by the callback kind and do
compile-time checks if entries are missing.  The system call entries now
map from callback kind to system call number (with a compile-time check
that there is an entry for every callback kind that represents a system
call and that the entries are in the correct order) and the dispatch is
done via a recursive template instantiation rather than by copying and
pasting code.  This makes it impossible to forget to implement syscall
callbacks on either side of the interface.  These compile-time checks
showed some bugs, which are now fixed.

The callback interface now has to be able to take a `Handle` in either
direction.  This was already supported by the transport but not used by
anything and so not exposed higher up.  Both `connect` and `bind` need
to pass a socket to the parent and have the parent modify it.  This is
now exposed generically.

Sometimes, if the tests crash they can leave orphan sandbox processes
running indefinitely.  This is now fixed on FreeBSD when using vfork (it
was already fixed with `pdfork`.  It is, unfortunately, not possible yet
to fix on Linux because the Linux `pdfork` analog does not have
kill-on-close behaviour and Linux's parent-death signal is triggered
when the *thread* that created the child process exits, not when the
*process* exits and so cannot be safely used within a library.

### DIFF
--- a/experiments/process_sandbox/include/process_sandbox/netpolicy.h
+++ b/experiments/process_sandbox/include/process_sandbox/netpolicy.h
@@ -1,0 +1,245 @@
+// Copyright Microsoft and Project Verona Contributors.
+// SPDX-License-Identifier: MIT
+
+/**
+ * This file contains the abstractions required for granting network access in
+ * a sandbox.  Sandboxes are not permitted to do anything directly that touches
+ * a global namespace and so may not connect or bind sockets.  Any attempt to
+ * use the network must be proxied.
+ */
+
+#pragma once
+#include <netdb.h>
+#include <optional>
+#include <string>
+#include <sys/socket.h>
+#include <sys/types.h>
+#include <tuple>
+#include <type_traits>
+#include <unordered_map>
+#include <variant>
+
+namespace sandbox
+{
+  /**
+   * Class encapsulating a network access policy for a sandbox.  This allows
+   * specific access to some network resources to be granted to a sandbox.
+   */
+  class NetworkPolicy
+  {
+    /**
+     * C++17 made noexcept part of the type system, so `R(Args...) noexcept` is
+     * not the same as `R(Args...)`.  Unfortunately, `std::function` remains
+     * specialised only for `R(Args...)` and so you cannot create a
+     * `std::function<R(Args...) noexcept>`.  Fortunately, the `noexcept`
+     * version is a subtype of the potentially throwing version (throwing 0
+     * exceptions is a stronger constraint than throwing 0 or more exceptions)
+     * and so we can just strip `noexcept` from the types.
+     *
+     * This helper provides a mechanism for stripping the `noexcept` qualifier
+     * from a function type.
+     */
+    template<typename Res, typename... ArgTypes>
+    struct remove_noexcept
+    {};
+
+    /**
+     * Specialisation for types that are already not-`noexcept`, just forwards
+     * the type.
+     */
+    template<typename Res, typename... ArgTypes>
+    struct remove_noexcept<Res(ArgTypes...)>
+    {
+      /**
+       * The type (already) without `noexcept`.
+       */
+      using type = Res(ArgTypes...);
+    };
+
+    /**
+     * Specialisation for types that are `noexcept`, forwards the type without
+     * `noexcept`.
+     */
+    template<typename Res, typename... ArgTypes>
+    struct remove_noexcept<Res(ArgTypes...) noexcept>
+    {
+      /**
+       * The type without `noexcept`.
+       */
+      using type = Res(ArgTypes...);
+    };
+
+    /**
+     * Helper for avoiding the need to write `::type` after every use of
+     * `remove_noexcept`.
+     */
+    template<typename Res, typename... ArgTypes>
+    using remove_noexcept_t = typename remove_noexcept<Res, ArgTypes...>::type;
+
+  public:
+    /**
+     * Enumeration describing the set of network operations that may be
+     * controlled by this policy.
+     */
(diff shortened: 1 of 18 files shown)


experiments/process_sandbox/include/process_sandbox/netpolicy.h
```diff
@@ -0,0 +1,245 @@
...
+  class NetworkPolicy
+  {
+    /**
+     * C++17 made noexcept part of the type system, so `R(Args...) noexcept` is
+     * not the same as `R(Args...)`.  Unfortunately, `std::function` remains
+     * specialised only for `R(Args...)` and so you cannot create a
+     * `std::function<R(Args...) noexcept>`.  Fortunately, the `noexcept`
+     * version is a subtype of the potentially throwing version (throwing 0
+     * exceptions is a stronger constraint than throwing 0 or more exceptions)
+     * and so we can just strip `noexcept` from the types.
```
rengolin, line 37: I'd expect the type system to do that for you. Is there a defect reported on this?

experiments/process_sandbox/include/process_sandbox/netpolicy.h
```diff
@@ -0,0 +1,245 @@
...
+       * access to the global namespace but in implementation it usually
+       * requires the ability ot query DNS or similar.
+       */
+      GetAddrInfo,
+      /**
+       * Connect a socket to a remote address.  Requires access to the global
+       * namespace to be authorised to connect to the specific remote address.
+       */
+      Connect
+    };
```
rengolin, line 102: Maybe not for this PR, but shouldn't the policy structure be independent of what the policy applies to?

I'd imagine policies with call backs would be useful for all sorts of operations, not just network.

rengolin:
Finished looking at the definitions, and it seems sensible. I have some comments and will continue looking at the implementation and tests later.

experiments/process_sandbox/include/process_sandbox/netpolicy.h
davidchisnall: Possibly.  We need some kind of policy for any global namespace that a sandbox will be granted conditional access to.  I think there are four kinds of things that we need to expose to sandboxes:

 - A filesystem
 - The network
 - Local IPC
 - User-defined callbacks

We expose the filesystem declaratively by allowing the user to expose the set of files / directories that the sandbox may access.  This PR exposes the network in a more imperative style, providing a filtering policy.  It would also be possible to wrap this in a declarative policy that let the programmer register something that looks more like a firewall rule set (e.g. 'Allow all outbound requests on the http or https service' or 'Allow any connection to microsoft.com').

I don't know what the best way of exposing local IPC is.  This may be something we just disable, or we could expose it via the filesystem namespace.  On Linux, most local IPC actually lives in the filesystem namespace (e.g. `shm_open` is just openning a file on a tmpfs filesystem).

For user-defined callbacks, the user can provide whatever policy they want.  

In terms of how these are surfaced in Verona, my expectation is that:

 - The `FileTree` structure would be exposed in Verona as something like `Builtin.Sandbox.FileTree` so that users can explicitly define a filesystem namespace for sandboxes before creating them.
 - The `NetworkPolicy` structure would be exposed as something that took an interface type in Verona as a delegate so that users can interpose on these functions via explicit callbacks.  Anything that looks more like a firewall interface would be a Verona standard-library class that implemented a policy.
 - ???
 - Lambdas / functions passed from Verona to foreign code would be wrapped for invocation on the other side and the user wouldn't be expected to think about this.

experiments/process_sandbox/include/process_sandbox/netpolicy.h
davidchisnall: I posted a comment about this in the MSR Cambridge Engineering teams channel, but here's a summary:

Prior to C++17, `noexcept` was not part of the type system.  It is now, but `R(Args...) noexcept` is a subtype of `R(Args...)` because throwing zero exceptions is a strictly weaker constraint than throwing zero or more exceptions.  The call method on `std::function` is allowed to throw if on `std::function<R(Args...)>(nullptr)`, so capturing whether the wrapped function is `noexcept` or not is not of value to `std::function` (the caller must assume that it can throw in any situation).  As a result, `std::function<R(Args...) noexcept>` would have to behave exactly as `std::function<R(Args...)>` and so there is no benefit in allowing the first specialisation.  Because the argument of the first is a subtype of the argument of the second, we can always just implicitly cast away the `noexcept` when creating a `std::function`.    As a result, this is not viewed as a defect in the C++17 standard.

This is only a problem when the type of the `std::function` is deduced from the initialiser.  In this case, the deduced type contains `noexcept` and (at least with libstdc++) the instantiation of `std::function` fails.  

I think the only real defect here is that there isn't a `std::remove_noexcept` in `<type_traits>`.

aghosn (commented):
The libcurl example is great!

PR #524, opened 2021-11-15
process_sandbox documentation & build

aghosn:
Adding build instructions, short descriptions of end-to-end tests, and a small walkthrough example to the `process_sandbox` README.

Modifying  `CMakeLists.txt`  to toggle the `USE_MEMFD` flag on Linux platforms.


experiments/process_sandbox/README.md
```diff
@@ -330,3 +329,184 @@ Code layout
...
+
+* [test-sandbox-callback-basic](tests/sandbox-callback-basic.cc): Creates a simple sandbox and registers a callback function that returns an `int`. The sandbox library simply invokes the callback and asserts it gets the correct value (`42`). 
+
+* [test-sandbox-callback-recursive](tests/sandbox-callback-recursive.cc): Creates a single sandbox that recursively invokes 6 nested levels of callbacks, i.e., a callback triggers a call into the sandbox that itself invokes a callback etc. 
+
+* [test-sandbox-crash](tests/sandbox-crash.cc): Creates 3 sandboxes meant to crash (calls `abort`). The crash triggers a `runtime_error` exception caught in the parent process. 
+
+* [test-sandbox-fake-open](tests/sandbox-fake-open.cc): Creates a shared memory file and a sandbox. The sandbox opens the file in two different ways: 1) with a raw system call, and 2) with the `open` function that performs a callback to the host. 
+
+* [test-sandbox-modify-pagemap](tests/sandbox-modify-pagemap.cc): Creates a sandbox attempting to modify the access rights of the pagemap using `mprotect`. The `mprotect` call should fail with a `permission denied`.
```
davidchisnall, line 372:
```suggestion
* [test-sandbox-modify-pagemap](tests/sandbox-modify-pagemap.cc): Creates a sandbox attempting to modify the access rights of the pagemap using `mprotect`.
  The `mprotect` call should fail with a `permission denied`.
```

(One sentence per line makes later diffs easier)

experiments/process_sandbox/README.md
```diff
@@ -330,3 +329,184 @@ Code layout
...
+cmake .. -GNinja -DCMAKE_BUILD_TYPE=Debug
+ninja
+```
+
+Tests
+------
+
+The previous steps build 15 tests, 8 unit and regression tests, and 7 end-to-end sandbox tests:
+
+* [test-sandbox-basic](tests/sandbox-basic.cc): Creates 3 sandboxes wrapping around a small library that exposes a function returning the sum of two `int`. Each sandbox is called twice.
```
davidchisnall, line 362:
```suggestion
* [test-sandbox-basic](tests/sandbox-basic.cc): Creates 3 sandboxes wrapping around a small library that exposes a function returning the sum of two `int`.
  Each sandbox is called twice.
```
(One sentence per line makes diffs easier)

experiments/process_sandbox/README.md
```diff
@@ -330,3 +329,184 @@ Code layout
...
+
+* [test-sandbox-callback-basic](tests/sandbox-callback-basic.cc): Creates a simple sandbox and registers a callback function that returns an `int`. The sandbox library simply invokes the callback and asserts it gets the correct value (`42`).
```
davidchisnall, line 364:
```suggestion
* [test-sandbox-callback-basic](tests/sandbox-callback-basic.cc): Creates a simple sandbox and registers a callback function that returns an `int`.
   The sandbox library simply invokes the callback and asserts it gets the correct value (`42`). 
```

experiments/process_sandbox/README.md
```diff
@@ -330,3 +329,184 @@ Code layout
...
+
+The previous steps build 15 tests, 8 unit and regression tests, and 7 end-to-end sandbox tests:
+
+* [test-sandbox-basic](tests/sandbox-basic.cc): Creates 3 sandboxes wrapping around a small library that exposes a function returning the sum of two `int`. Each sandbox is called twice.
+
+* [test-sandbox-callback-basic](tests/sandbox-callback-basic.cc): Creates a simple sandbox and registers a callback function that returns an `int`. The sandbox library simply invokes the callback and asserts it gets the correct value (`42`). 
+
+* [test-sandbox-callback-recursive](tests/sandbox-callback-recursive.cc): Creates a single sandbox that recursively invokes 6 nested levels of callbacks, i.e., a callback triggers a call into the sandbox that itself invokes a callback etc. 
+
+* [test-sandbox-crash](tests/sandbox-crash.cc): Creates 3 sandboxes meant to crash (calls `abort`). The crash triggers a `runtime_error` exception caught in the parent process.
```
davidchisnall, line 368:
```suggestion
* [test-sandbox-crash](tests/sandbox-crash.cc): Creates 3 sandboxes meant to crash (calls `abort`).
  The crash triggers a `runtime_error` exception caught in the parent process. 
```

experiments/process_sandbox/README.md
```diff
@@ -330,3 +329,184 @@ Code layout
...
+
+* [test-sandbox-basic](tests/sandbox-basic.cc): Creates 3 sandboxes wrapping around a small library that exposes a function returning the sum of two `int`. Each sandbox is called twice.
+
+* [test-sandbox-callback-basic](tests/sandbox-callback-basic.cc): Creates a simple sandbox and registers a callback function that returns an `int`. The sandbox library simply invokes the callback and asserts it gets the correct value (`42`). 
+
+* [test-sandbox-callback-recursive](tests/sandbox-callback-recursive.cc): Creates a single sandbox that recursively invokes 6 nested levels of callbacks, i.e., a callback triggers a call into the sandbox that itself invokes a callback etc. 
+
+* [test-sandbox-crash](tests/sandbox-crash.cc): Creates 3 sandboxes meant to crash (calls `abort`). The crash triggers a `runtime_error` exception caught in the parent process. 
+
+* [test-sandbox-fake-open](tests/sandbox-fake-open.cc): Creates a shared memory file and a sandbox. The sandbox opens the file in two different ways: 1) with a raw system call, and 2) with the `open` function that performs a callback to the host.
```
davidchisnall, line 370:
```suggestion
* [test-sandbox-fake-open](tests/sandbox-fake-open.cc): Creates a shared memory file and a sandbox. The sandbox opens the file in two different ways:
  1. With a raw system call, and
  2. With the `open` function that performs a callback to the host.
```

experiments/process_sandbox/README.md
```diff
@@ -330,3 +329,184 @@ Code layout
...
+
+* [test-sandbox-callback-recursive](tests/sandbox-callback-recursive.cc): Creates a single sandbox that recursively invokes 6 nested levels of callbacks, i.e., a callback triggers a call into the sandbox that itself invokes a callback etc. 
+
+* [test-sandbox-crash](tests/sandbox-crash.cc): Creates 3 sandboxes meant to crash (calls `abort`). The crash triggers a `runtime_error` exception caught in the parent process. 
+
+* [test-sandbox-fake-open](tests/sandbox-fake-open.cc): Creates a shared memory file and a sandbox. The sandbox opens the file in two different ways: 1) with a raw system call, and 2) with the `open` function that performs a callback to the host. 
+
+* [test-sandbox-modify-pagemap](tests/sandbox-modify-pagemap.cc): Creates a sandbox attempting to modify the access rights of the pagemap using `mprotect`. The `mprotect` call should fail with a `permission denied`.
+
+* [test-sandbox-zlib](tests/sandbox-zlib.cc): Uses `zlib` to `deflate`, i.e., compress, the program's file. This test compares the output produced by a sandboxed and an unsandboxed version of zlib.
```
davidchisnall, line 374:
```suggestion
* [test-sandbox-zlib](tests/sandbox-zlib.cc): Uses `zlib` to `deflate`, i.e., compress, the program's file.
  This test compares the output produced by a sandboxed and an unsandboxed version of zlib. 
```

experiments/process_sandbox/README.md
```diff
@@ -330,3 +329,184 @@ Code layout
...
+
+* [test-sandbox-fake-open](tests/sandbox-fake-open.cc): Creates a shared memory file and a sandbox. The sandbox opens the file in two different ways: 1) with a raw system call, and 2) with the `open` function that performs a callback to the host. 
+
+* [test-sandbox-modify-pagemap](tests/sandbox-modify-pagemap.cc): Creates a sandbox attempting to modify the access rights of the pagemap using `mprotect`. The `mprotect` call should fail with a `permission denied`.
+
+* [test-sandbox-zlib](tests/sandbox-zlib.cc): Uses `zlib` to `deflate`, i.e., compress, the program's file. This test compares the output produced by a sandboxed and an unsandboxed version of zlib. 
+
+* [test-sandbox-network](tests/test-sandbox-network.cc): Example of how network policies can be used to restrict the sandbox's ability to perform network operations. 
+This test creates a server sandbox and gradually enables network operations (1) `getaddrinfo`, (2) `bind`ing to a socket, (3) the ability to receive messages on the socket from the parent, and (4) the ability for asandboxed client to `connect` to the server.
```
davidchisnall, line 377:
```suggestion
  This test creates a server sandbox and gradually enables network operations that are blocked by default for sandboxes:
  - `getaddrinfo` for name lookup (requires network access for DNS lookups and so implicitly accesses a global namespace).
  - `bind`ing to socket to a local address (the local address is part of a global namespace.
  - `connect`ing to a remote address (the remote address forms part of a global namespace)
```

experiments/process_sandbox/README.md
```diff
@@ -330,3 +329,184 @@ Code layout
...
+Each sandbox runs in a separate process and is called twice.
+
+
+### Runtime Walkthrough
+
+Below is the process tree generated by the test:
+
+```
+test-sandbox-basic─┬─3*[library_runner]
+                   └─{test-sandbox-basic}
```
davidchisnall, line 416: What is this line?  It looks like `test-sandbox-basic` is a child of itself?

experiments/process_sandbox/README.md
```diff
@@ -330,3 +329,184 @@ Code layout
...
+
+```
+
+The test creates three children processes and one thread.
+
+* *Thread:* It is created as part of the `sandbox::Library::MemoryServiceProvider` defined in [src/libsandbox.cc](src/libsandbox.cc) and handles pagemap updates from children.
```
davidchisnall, line 422:
```suggestion
The first time a sandbox is created the library creates a background thread (`sandbox::Library::MemoryServiceProvider`, defined in [src/libsandbox.cc](src/libsandbox.cc)).
This thread listens for RPCs from the child processes requesting updates to the pagemap.
It is responsible for validating that the requests correspond to parts of the address space owned by the child and that the requested updates are well formed.
The updates made by this are then visible to the children via the shared memory object that backs the pagemap.
```

experiments/process_sandbox/README.md
```diff
@@ -330,3 +329,184 @@ Code layout
...
+
+* *Processes:* Each process runs the [library_runner](src/library_runner.cc) which dynamically loads the sandboxed library `sandboxed-basic.so`.
```
davidchisnall, line 424:
```suggestion
Each sandbox manages a child process that `execve`s [library_runner](src/library_runner.cc) with reduced privileges (no access to any global namespace).
This process then loads the library that contains the in-sandbox code, in this case `sandboxed-basic.so`.
```

experiments/process_sandbox/README.md
```diff
@@ -330,3 +329,184 @@ Code layout
...
+                                                 |                                              |        PageMap                               |
+                                                 |                                              |                                              |
+                                                 |        PageMap                               |                                              | 
+  |-- 0x7ffe80000000 -- 0x7ffec0000000 [rw-s]    |                                              |                                              |-- child 3 region [rw-s]
+                                                 |                                              |                                            
+  |-- 0x7fff00000000 -- 0x7fff40000000 [rw-s]    |                                              |-- child 2 region [rw-s]                     
+                                                 |                                                                                           
+  |-- 0x7fff80000000 -- 0x7fffc0000000 [rw-s]    |-- child 1 region [rw-s]                                                                   
+
```
davidchisnall, line 473: I think this picture would be easier to understand if the ranges were shown as ranges.  It confused me that there's a big range shown for the pagemap but then the thing that looks like the end of the pagemap is actually the shared memory object.  The location of the pagemap in the child's memory is not very important, the fact that each sandbox's heap is in the same location as it is in the parent is the key thing that this diagram needs to convey.

experiments/process_sandbox/README.md
```diff
@@ -330,3 +329,184 @@ Code layout
...
+```
+
+The `library_runner` is responsible, at initialization time, for ensuring that both the `PageMap` and the child's memory region (i.e., heap) are correctly mapped in the process's address space.
+For that purposes, it uses the `memfd` file descriptors inherited from the parent.
```
davidchisnall, line 477:
```suggestion
For that purposes, it uses the shared-memory file descriptors inherited from the parent.
```
It would be good to avoid Linux-specific descriptions as much as possible here.

experiments/process_sandbox/README.md
```diff
@@ -330,3 +329,184 @@ Code layout
...
+It relies on `mmap` to install the heap as a shared memory region that is both readable and writable.
+For the `PageMap`, it resizes the region and drops write access to the corresponding file descriptor using `mmap`.
```
davidchisnall, line 479: Does it?  It shouldn't resize the region at all, and the parent should be dropping write access when it sets up the sandbox policy.

On platforms with well-designed security mechanisms, [it doesn't have the rights to do anything other than a read-only mapping of the shared-memory object](https://github.com/microsoft/verona/blob/b5be27ea4730301bc893d8746fa758741b2e0ab4/experiments/process_sandbox/include/process_sandbox/platform/sandbox_capsicum.h#L77).  On Linux, [it can't map the memory read-write](https://github.com/microsoft/verona/blob/b5be27ea4730301bc893d8746fa758741b2e0ab4/experiments/process_sandbox/include/process_sandbox/platform/sandbox_seccomp-bpf.h#L52) (and there's a [test that checks that it can't](https://github.com/microsoft/verona/blob/b5be27ea4730301bc893d8746fa758741b2e0ab4/experiments/process_sandbox/tests/sandboxlib-modify-pagemap.cc#L22)).  It can resize the file on Linux and so we close the file descriptor before we run any untrusted code.  We should fix this using the sealing mechanism on memfd, but we'd still need to close the file descriptor because we fall back to POSIX shared memory if `memfd` doesn't work.

experiments/process_sandbox/README.md
```diff
@@ -330,3 +329,184 @@ Code layout
...
+Despite observable overlaps between the `PageMap` memory regions and heaps of different children, this initialization step ensures that the corresponding virtual memory is backed by separate physical pages in the different children address spaces.
```
davidchisnall, line 480: I'm not sure what this means.  The key thing is that the same physical memory backs the pagemap, it doesn't matter where the pagemap is in the child's address space because it's accessed via a pointer.  This is an implementation detail that's subject to change: we may want to put the pagemap at the same place in all processes in the future (and we did in the pre-snmalloc2 version).

experiments/process_sandbox/README.md
```diff
@@ -330,3 +329,184 @@ Code layout
...
+It relies on `mmap` to install the heap as a shared memory region that is both readable and writable.
+For the `PageMap`, it resizes the region and drops write access to the corresponding file descriptor using `mmap`.
+Despite observable overlaps between the `PageMap` memory regions and heaps of different children, this initialization step ensures that the corresponding virtual memory is backed by separate physical pages in the different children address spaces. 
+
+#### Invoking the sandbox
+
+We follow the execution of the program and show how, on Linux, the parent invokes the `sum` function defined in [sandboxlib-basic](tests/sandboxlib-basic.cc).
+
+
+On the parent side, the sandbox library setups a `msg_buffer`, i.e.,  call frame containing the arguments to the function call and a return element.
```
davidchisnall, line 487:
```suggestion
On the parent side, the sandbox library sets up a call frame containing the arguments to the function call and a return element in the shared memory region and points the `msg_buffer` field in the header of the shared memory region to this buffer.
```

('set up' is verb phrase, 'setup' is a noun.  You can't noun a thing).

experiments/process_sandbox/README.md
```diff
@@ -330,3 +329,184 @@ Code layout
...
+It passes the pointer to that call frame as well as the index of the library's function invoked via shared memory to the child that it wakes up.
+On Linux, semaphores (implemented with the underlying `futex` system call) are used to communciate with the child process.
```
davidchisnall, line 489:
```suggestion
Simple one-bit in-memory semaphores are used to communicate with the child process.
These are implemented with the platform's low-level primitive, `futex` on Linux and `_umtx_op` on FreeBSD.
```

experiments/process_sandbox/README.md
```diff
@@ -330,3 +329,184 @@ Code layout
...
+#0  syscall () at ../sysdeps/unix/sysv/linux/x86_64/syscall.S:38
+#1  0x00007ffff7f3b611 in sandbox::platform::FutexOneBitSem::futex_op (this=0x7fff800000c0, futex_op=1, val=1, timeout=0x0) at ../include/process_sandbox/platform/onebitsem_futex.h:29
+#2  0x00007ffff7f3b884 in sandbox::platform::FutexOneBitSem::wake (this=0x7fff800000c0) at ../include/process_sandbox/platform/onebitsem_futex.h:46
+#3  0x00007ffff7f38d77 in sandbox::Library::send (this=0x7fffffffd1b0, idx=1, ptr=0x7fff80004020) at ../src/libsandbox.cc:774
+#4  0x000055555557e0b0 in sandbox::Function<int, int, int>::operator() (this=0x7fffffffd660, args#0=1, args#1=2) at ../include/process_sandbox/cxxsandbox.h:103
+#5  0x000055555557cd65 in test_sum (sb=..., a=1, b=2) at ../tests/sandbox-basic.cc:31
+#6  0x000055555557cf18 in main () at ../tests/sandbox-basic.cc:40
+```
+
+On the child's side, the process stuck waiting on the futex is woken up, extracts the index and `msg_buffer`, and invokes the corresponding sandboxed function.
```
davidchisnall, line 502:
```suggestion
On the child's side, the process that was blocked waiting on the one-bit semaphore is woken up, then extracts the index and call frame, and invokes the corresponding sandboxed function. 
```

experiments/process_sandbox/README.md
aghosn: I will clarify that part of the text, I mixed things up.
For the `PageMap`, as you pointed out, we get the following trace in the child:

```
openat(AT_FDCWD, "/proc/self/fd/4", O_RDONLY) = 8
dup2(8, 4)                              = 4
close(8)                                = 0
```

And later on, we map the content with:

```
mmap(NULL, 274877906944, PROT_READ, MAP_SHARED, 4, 0) = 0x7fbff7a1a000
```

Reformulation: 

```
The `library_runner` opens the `memfd` file descriptor that corresponds to the `PageMap` object and maps its content read-only in the child's address space.
```

experiments/process_sandbox/docs/walkthrough.md
```diff
@@ -0,0 +1,135 @@
...
+  |-- 0x7ffe80000000 -- 0x7ffec0000000 [rw-s]    |                                              |                                              |-- child 3 region [rw-s]
+                                                 |                                              |                                            
+  |-- 0x7fff00000000 -- 0x7fff40000000 [rw-s]    |                                              |-- child 2 region [rw-s]                     
+                                                 |                                                                                           
+  |-- 0x7fff80000000 -- 0x7fffc0000000 [rw-s]    |-- child 1 region [rw-s]                                                                   
+
+```
+
+The `library_runner` is responsible, at initialization time, for ensuring that both the `PageMap` and the child's memory region (i.e., heap) are correctly mapped in the process's address space.
+For that purposes, it uses the `memfd` file descriptors inherited from the parent.
```
davidchisnall, line 100:
```suggestion
For that purposes, it uses the file descriptors referring to shared memory objects that it inherits from the parent.
```

experiments/process_sandbox/docs/walkthrough.md
```diff
@@ -0,0 +1,135 @@
...
+It relies on `mmap` to install the heap as a shared memory region that is both readable and writable.
+The `library_runner` opens the `memfd` file descriptor that corresponds to the `PageMap` object and maps its content read-only in the child's address space.
```
davidchisnall, line 102:
```suggestion
The `library_runner` opens the file descriptor that corresponds to the shared memory object that backs the `PageMap` and maps its content read-only in the child's address space.
The platform-specific sandboxing code is responsible for ensuring that the child process cannot perform a read-write mapping of this file descriptor.
```

experiments/process_sandbox/docs/walkthrough.md
```diff
@@ -0,0 +1,135 @@
...
+
+#### Invoking the sandbox
+
+We follow the execution of the program and show how, on Linux, the parent invokes the `sum` function defined in [sandboxlib-basic](../tests/sandboxlib-basic.cc).
+
+
+On the parent side, the sandbox library sets up a call frame containing the arguments to the function call and a return element in the shared memory region and points the `msg_buffer` field in the header of the shared memory region to this buffer.
```
davidchisnall, line 109:
```suggestion
On the parent side, the sandbox library creates an object representing a call frame, containing the arguments to the function call and space for any return element, in the shared memory region and points the `msg_buffer` field in the header of the shared memory region to this object.
```

experiments/process_sandbox/docs/walkthrough.md
```diff
@@ -0,0 +1,135 @@
...
+It passes the pointer to that call frame as well as the index of the library's function invoked via shared memory to the child that it wakes up.
```
davidchisnall, line 110:
```suggestion
It also stores the index of the library's function to be invoked in the `function_index` field.
```

experiments/process_sandbox/docs/walkthrough.md
```diff
@@ -0,0 +1,135 @@
...
+#0  syscall () at ../sysdeps/unix/sysv/linux/x86_64/syscall.S:38
+#1  0x00007ffff7f3b611 in sandbox::platform::FutexOneBitSem::futex_op (this=0x7fff800000c0, futex_op=1, val=1, timeout=0x0) at ../include/process_sandbox/platform/onebitsem_futex.h:29
+#2  0x00007ffff7f3b884 in sandbox::platform::FutexOneBitSem::wake (this=0x7fff800000c0) at ../include/process_sandbox/platform/onebitsem_futex.h:46
+#3  0x00007ffff7f38d77 in sandbox::Library::send (this=0x7fffffffd1b0, idx=1, ptr=0x7fff80004020) at ../src/libsandbox.cc:774
+#4  0x000055555557e0b0 in sandbox::Function<int, int, int>::operator() (this=0x7fffffffd660, args#0=1, args#1=2) at ../include/process_sandbox/cxxsandbox.h:103
+#5  0x000055555557cd65 in test_sum (sb=..., a=1, b=2) at ../tests/sandbox-basic.cc:31
+#6  0x000055555557cf18 in main () at ../tests/sandbox-basic.cc:40
+```
+
+On the child's side, the process that was blocked waiting on the one-bit semaphore is woken up, then extracts the index and call frame, and invokes the corresponding sandboxed function.
```
davidchisnall, line 125:
```suggestion
The pointer to the call frame object and the function index are not used again after this point and so a reentrant call (if the child calls back into the parent, which then calls back into the child) can reuse them.
```

experiments/process_sandbox/tests/README.md
```diff
@@ -0,0 +1,34 @@
...
+
+Tests
+------
+
+This folder contains 17 tests, 8 unit and regression tests, and 9 end-to-end sandbox tests.
+
+## End-to-End sandbox tests
+
+* [test-sandbox-basic](sandbox-basic.cc): Creates 3 sandboxes wrapping around a small library that exposes a function returning the sum of two `int`.
+  Each sandbox is called twice.
```
davidchisnall, line 11:
```suggestion
  This test ensures that it is possible to create multiple instances of a sandbox.
```

experiments/process_sandbox/tests/README.md
```diff
@@ -0,0 +1,34 @@
...
+
+* [test-sandbox-callback-basic](sandbox-callback-basic.cc): Creates a simple sandbox and registers a callback function that returns an `int`.
+   The sandbox library simply invokes the callback and asserts it gets the correct value (`42`). 
+
+* [test-sandbox-callback-recursive](sandbox-callback-recursive.cc): Creates a single sandbox that recursively invokes 6 nested levels of callbacks, i.e., a callback triggers a call into the sandbox that itself invokes a callback etc.
```
davidchisnall, line 16:
```suggestion
  This test ensures that the sandboxes are reentrant and that the stack stitching (jumping between the stacks in the parent and child processes) works correctly.
```

experiments/process_sandbox/tests/README.md
```diff
@@ -0,0 +1,34 @@
...
+
+* [test-sandbox-callback-basic](sandbox-callback-basic.cc): Creates a simple sandbox and registers a callback function that returns an `int`.
+   The sandbox library simply invokes the callback and asserts it gets the correct value (`42`). 
+
+* [test-sandbox-callback-recursive](sandbox-callback-recursive.cc): Creates a single sandbox that recursively invokes 6 nested levels of callbacks, i.e., a callback triggers a call into the sandbox that itself invokes a callback etc. 
+
+* [test-sandbox-crash](sandbox-crash.cc): Creates 3 sandboxes meant to crash (calls `abort`).
+  The crash triggers a `runtime_error` exception caught in the parent process. 
+
+* [test-sandbox-fake-open](sandbox-fake-open.cc): Creates a shared memory file and a sandbox. The sandbox opens the file in two different ways: 1) with a raw system call, and 2) with the `open` function that performs a callback to the host.
```
davidchisnall, line 21:
```suggestion
* [test-sandbox-fake-open](sandbox-fake-open.cc): Creates a shared memory file and a sandbox. 
  The sandbox opens the file in two different ways: 1) with a raw system call, and 2) with the `open` function that performs a callback to the host.
  This ensures that the child process correctly interposes on both the libc wrappers (by providing a strong definition of the functions in `library_runner`) and on the raw system call (by catching a signal delivered by the underlying sandboxing technology and emulating it).
```

experiments/process_sandbox/tests/README.md
```diff
@@ -0,0 +1,34 @@
...
+
+* [test-sandbox-crash](sandbox-crash.cc): Creates 3 sandboxes meant to crash (calls `abort`).
+  The crash triggers a `runtime_error` exception caught in the parent process. 
+
+* [test-sandbox-fake-open](sandbox-fake-open.cc): Creates a shared memory file and a sandbox. The sandbox opens the file in two different ways: 1) with a raw system call, and 2) with the `open` function that performs a callback to the host. 
+
+* [test-sandbox-modify-pagemap](sandbox-modify-pagemap.cc): Creates a sandbox attempting to modify the access rights of the pagemap using `mprotect`.
+  The `mprotect` call should fail with a `permission denied`.
+
+* [test-sandbox-zlib](sandbox-zlib.cc): Uses `zlib` to `deflate`, i.e., compress, the program's file. This test compares the output produced by a sandboxed and an unsandboxed version of zlib.
```
davidchisnall, line 26:
```suggestion
* [test-sandbox-zlib](sandbox-zlib.cc): Uses `zlib` to `deflate`, i.e., compress, the program's file.
  This test compares the output produced by a sandboxed and an unsandboxed version of zlib.
  This is an integration test that ensure that it is possible to sandbox a (simple) real library.
```

experiments/process_sandbox/README.md
aghosn: This is the MemoryServiceProvider:
https://github.com/microsoft/verona/blob/master/experiments/process_sandbox/src/libsandbox.cc#L280

aghosn (commented):
LGTM

PR #526, opened 2021-11-17
Simplify promise error handling

marioskogias:
This PR removes the extra lambda passed in the then() function to handle cases in which the promise is never fulfilled. Instead, we pass a single lambda that does error handling.


src/rt/test/func/promise/promise.cc
```diff
@@ -16,15 +16,33 @@ void promise_test()
...
-  rp.then(
-    [](int val) { Systematic::cout() << val << std::endl; },
-    []() { Systematic::cout() << "Error\n"; });
-  rp2.then(
-    [](int val) { Systematic::cout() << val << std::endl; },
-    []() { Systematic::cout() << "Error\n"; });
+  rp.then([](std::variant<int, Promise<int>::PromiseErr> val) {
+    if (std::holds_alternative<int>(val))
+    {
+      auto v = std::get<int>(std::move(val));
```
mjp41, line 22: It would be useful to know which bits are expected and not expected to be reached in this example.  Could you put `abort` on the cases you don't expect to be called.

mjp41 (approved):
LGTM

src/rt/test/func/promise/promise.cc
marioskogias: Good idea. The promise_no_writer case will become cleaner this way as well. I'm adding them now.

PR #527, opened 2021-12-01
Make unpause signal safe

mjp41:
Prior to this PR, `unpause` would take a lock to ensure that wake-ups weren't lost.  This is unfortunately a problem if an external signal/interrupt causes the runtime to `unpause`, while the thread it signalled/interrupted is holding the lock.

This code changes the implementation to use a custom spin lock and semaphores to enables `unpause` to not require the lock, but instead notifies the thread holding the lock that it must wake all threads up on exit from the lock.

### DIFF
--- a/src/rt/sched/threadsync.h
+++ b/src/rt/sched/threadsync.h
@@ -2,10 +2,8 @@
 // SPDX-License-Identifier: MIT
 
 #pragma once
-#include "test/systematic.h"
-
 #include "../pal/semaphore.h"
-#include <mutex>
+#include "test/systematic.h"
 
 /**
  * This file contains the synchronisation implementation for suspending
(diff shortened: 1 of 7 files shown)


mjp41:
This is draft as had to more to C++20 for <semaphore> support.  This can be abstracted away to call posix `<semaphore.h>` and the appropriate windows calls. But I want eyes on the design, before I go there.

src/rt/sched/threadsync.h
```diff
@@ -2,32 +2,143 @@
...
+      if (!lock.unlock())
+      {
+        Systematic::cout() << "Pending unpause" << Systematic::endl;
+
+        auto* curr = waiters;
+        waiters = nullptr;
+        // Subsequent unpause requests can be ignored as there are no more
+        // waiters as we have held the lock continuously.
+        lock.unlock_unpause();
+        // Don't need to hold the lock to wake up the waiters.
```
davidchisnall, line 119: If I read this correctly, it is to ensure that the state is updated before any waiters resume?

davidchisnall (approved):
There's some subtle ordering from the fact that the waiter count and the state need to be in different places, which wouldn't be necessary on a futex-based implementation.  The probably doesn't matter for new platform bring-up because you can trivially build a semaphore if you have a futex but I wonder what the code complexity is from having three implementations of a Semaphore, rather than 3/4 thin futex wrappers plus a simpler data structure on top.

src/rt/sched/threadsync.h
mjp41: Yeah, it is to ensure atomicity of the update with respect to waiters.

ihaller:
LGTM

PR #528, opened 2021-12-01
Acquire all cowns at the beginning

marioskogias:
This PR changes the way the verona runtime acquires cowns. When scheduling a behaviour, the runtime acquires all the idle cowns and enqueues the multimessage on the queues of the busy ones. The multimessage has an atomic counter that gets decrement every time a new cown is available. The scheduler thread that last decrements this counter in run_step() also executes the behaviour.


rengolin:
What do we want? All cowns!

When do we want? At the beginning!

mjp41:
@lukecheeseman this is an experiment to implement causal ordering using overlap rather than subset.

PR #533, opened 2021-12-15
First pass at creating a region API.

mjp41:
Much is still needed.
  * Debug features are not part of API.
  * More operations need lifting into the API.

Primarily PRing this as it will help @jacob-hughes with his outstanding PRs.

### DIFF
--- a/src/rt/region/region_api.h
+++ b/src/rt/region/region_api.h
@@ -205,6 +205,8 @@ namespace verona::rt::api
         return RegionRc::alloc(
           ThreadAlloc::get(), (RegionRc*)RegionContext::get_region(), d);
     }
+    // Unreachable as case is exhaustive
+    abort();
   }
 
   inline void add_reference(Object*)
--- a/src/rt/region/region_api.h
+++ b/src/rt/region/region_api.h
@@ -224,13 +224,13 @@ namespace verona::rt::api
 
   inline void incref(Object* o)
   {
-    assert(Region::get_type(RegionContext::get_region() == RegionType::Rc));
+    assert(Region::get_type(RegionContext::get_region()) == RegionType::Rc);
     RegionRc::incref(o);
   }
 
   inline void decref(Object* o)
   {
-    assert(Region::get_type(RegionContext::get_region() == RegionType::Rc));
+    assert(Region::get_type(RegionContext::get_region()) == RegionType::Rc);
     RegionRc::decref(
       ThreadAlloc::get(), o, (RegionRc*)RegionContext::get_region());
   }
--- a/src/rt/region/region_api.h
+++ b/src/rt/region/region_api.h
@@ -5,7 +5,6 @@
 #include "freeze.h"
 #include "region.h"
 
-
 namespace verona::rt::api
 {
   namespace internal
@@ -76,6 +75,7 @@ namespace verona::rt::api
    */
   inline void open_region(Object* r)
   {
+    assert(r->debug_is_iso());
     auto md = r->get_region();
     RegionContext::push(r, md);
     switch (Region::get_type(md))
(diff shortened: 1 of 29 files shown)


src/rt/region/region_api.h
```diff
@@ -0,0 +1,332 @@
...
+  inline void region_collect()
+  {
+    switch (Region::get_type(RegionContext::get_region()))
+    {
+      case RegionType::Trace:
+      // Other roots?
+        RegionTrace::gc(ThreadAlloc::get(), RegionContext::get_entry_point());
+        break;
+      case RegionType::Arena:
+        break; // TODO?
```
jacob-hughes, line 288: Why `TODO`, won't this always be a no-op?

src/rt/region/region_api.h
```diff
@@ -0,0 +1,332 @@
...
+  inline Object* use_external_reference(ExternalRef* e)
+  {
+    assert(is_external_reference_valid(e));
+    return e->get();
+  }
+
+  /**
+   * Create object in current region
+   */
+  inline Object* create_object(Descriptor* d)
```
jacob-hughes, line 193: Is this intended to replace `Region::alloc` in `region.h`? If not, why would I use one over the other?

src/rt/region/region_api.h
```diff
@@ -0,0 +1,332 @@
...
+        break;
+      default:
+        abort();
+    }
+  }
+
+  /**
+   * Close current region
+   */
+  inline void close_region()
```
jacob-hughes, line 97: This will crash if called without a current region in thread local storage right? Shall we add an assert on entry that the top of the stack is not a nullptr?

src/rt/region/region_api.h
```diff
@@ -0,0 +1,332 @@
...
+  {
+    // TODO
+  }
+
+  inline void remove_reference(Object*)
+  {
+    // TODO
+  }
+
+  inline void incref(Object* o)
```
jacob-hughes, line 222: Should we switch on region type and abort if != `RegionRc` like we do in decref?

src/rt/region/region_api.h
```diff
@@ -0,0 +1,332 @@
...
+        entry_point = RegionTrace::create(ThreadAlloc::get(), d);
+        break;
+      case RegionType::Arena:
+        entry_point = RegionArena::create(ThreadAlloc::get(), d);
+        break;
+      case RegionType::Rc:
+        entry_point = RegionRc::create(ThreadAlloc::get(), d);
+        break;
+    }
+    return {reinterpret_cast<T*>(entry_point)};
```
jacob-hughes, line 259: Why is this templatised to `T*`? What could `T` be other than `Object`?

jacob-hughes (commented):
Looks good. I like this new API a lot. A few questions about the design but nothing major.

src/rt/region/region_api.h
mjp41: Should be removed.  I have not been dropping stuff unless it doesn't compile initially.

src/rt/region/region_api.h
```diff
@@ -0,0 +1,332 @@
...
+    Freeze::apply(ThreadAlloc::get(), r);
+    return r;
+  }
+
+  /**
+   * Add supplied region to the current region
+   * and return the entry point.
+   */
+  template<typename T = Object>
+  inline T* merge(T* r)
```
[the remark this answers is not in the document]
mjp41: To be decided.  I think we can change still at this point.  Currently we only have merge on the same kind of regions.

src/rt/region/region_api.h
mjp41: Can be a subtype of `Object*`, so then we can build a more type safe interface.

mjp41:
Going to merge early to add @jacob-hughes next PR.

PR #536, opened 2022-01-25
Change validation of new objects

mjp41:
New `Object`s should only be created by the runtime through the region
API.  The previous check required the constructor to be run for
`Object`.  This refactoring no longer requires the constructor for
`Object` to be run, but still detects if objects are created outside of
the runtime API such as the `new` overload provided in `vobject.h`.


mjp41:
@aghosn this should address some of the issues you were having with the region API.

PR #541, opened 2022-02-03
Factor out a simplified SysTesting runtime.

mjp41:
The systematic testing runtime was intertwinned with the implementation
of thread sync.

This PR makes a more independent systematic testing runtime. The runtime
only requires a mechanism to sleep and wake a thread, and no longer
depends on condition variable or mutexs.


ihaller:
LGTM

PR #542, opened 2022-02-07
Separate runtime tests.

mjp41:


rengolin:
Is the plan to have a pipeline flag that forces the RT test, so that you can trigger a manual build with RT tests without having to change the source in the runtime?

mjp41:
> Is the plan to have a pipeline flag that forces the RT test, so that you can trigger a manual build with RT tests without having to change the source in the runtime?

I was just looking at what the time would be if I put the runtime build separate from the other stuff.  I wasn't planning to add a specific flag.

mjp41:
This PR improves the for runtime changes to pass CI. It was about 40mins before this PR, and is down to 26mins after the change.

rengolin:
Is this faster just because it's more parallel? I'm assuming your timing is wall clock for the whole build, not each stage. 

Won't the higher parallelism increase usage of the free vms and that, in turn, increase pressure on other builds? 

We now have duplicate build instructions for each platform, can we common them up somehow? The AnyBuild files have a model how to do that.

PR #549, opened 2022-03-16
Add a document for future optimisations.

davidchisnall:
This is a place to collect optimisations that are not necessary for
MVP but that we'd like to consider as possible future extensions when
designing the compiler.

### DIFF
--- a/docs/internal/future_optimisations.md
+++ b/docs/internal/future_optimisations.md
@@ -1,7 +1,7 @@
 Future Optimisation Opportunities
 =================================
 
-The Verona language affords a number of optimisation opportunities that do not need to be implemented for a minimum viable product.
+The Verona language affords a number of optimisation opportunities that do not need to be implemented for a minimum viable product (MVP).
 This document exists to capture them so that they can be implemented in a future version and so that the design of the compiler doesn't preclude their implementation.
 They are not in any particular order.
 
@@ -34,7 +34,7 @@ someFunction(object : Interface)
 	match (object)
 	{
 	  _MyPrivateConcreteType => fastPath(object);
-	   => slowPath(object;
+	   => slowPath(object);
 	}
 }
 ```
--- a/docs/internal/future_optimisations.md
+++ b/docs/internal/future_optimisations.md
@@ -1,0 +1,163 @@
+Future Optimisation Opportunities
+=================================
+
+The Verona language affords a number of optimisation opportunities that do not need to be implemented for a minimum viable product.
+This document exists to capture them so that they can be implemented in a future version and so that the design of the compiler doesn't preclude their implementation.
+They are not in any particular order.
+
+Late selector colouring
+-----------------------
+
+In the MVP compiler, selector colouring is done in the front end.
+This means that we need to assign descriptor state even for interface calls that are possible to elide.
+The front end can make some early decisions:
+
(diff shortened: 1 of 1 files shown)


docs/internal/future_optimisations.md
```diff
@@ -0,0 +1,163 @@
...
+```
+
+This transform is often done by dynamic language VMs because it both plays well with branch predictors and because it exposes inlining opportunities.
+It may be beneficial to inline one or both of the methods, which improves locality of the code and then exposes further optimisation opportunities from specialising the method for this specific call site.
+
+Descriptor Elision
+------------------
+
+The descriptor is used for dynamic dispatch and for `match` expressions.
+If a class is never cast to an interface type (including a union type) then it should be possible to completely elide its descriptor.
```
rengolin, line 69: I thought this could be caught by the front end, but perhaps not through an interface, before inlining.

If that's the only case it's worth looking, then a comment here would be clearer.

docs/internal/future_optimisations.md
```diff
@@ -0,0 +1,163 @@
...
+	embed _nanoseconds : Nanoseconds;
+	// Methods here.
+	<(self : Self, other : TimeInterval) { ... }
+}
+```
+
+This class stores 16 bytes of state but would require 96 bytes in the MVP compiler (on a non-CHERI 64-bit system).
+Both `_seconds` and `_nanoseconds` are used by only this class and so we should be able to elide both of their descriptors and region metadata fields, shrinking this class to 32 bytes.
+If `TimeInterval` is never used in an interface or union type, then we can shrink it to 24 bytes.
+If nothing uses `TimeInterval` for region identification then we can shrink `embed` fields of type `TimeInterval` to 16 bytes.
```
rengolin, line 115: If enough inlining and static analysis is done, we can even reduce all of that to `U64` native operations ('+', '<', etc) and not have a class structure, descriptor, metadata, etc., reducing the "size" of that class' members to 8+8 bytes in total.

docs/internal/future_optimisations.md
```diff
@@ -0,0 +1,163 @@
...
+As a further optimisation, if another object from the same region is visible in all callers of the function that access the region and the first two conditions are satisfied, then an additional region parameter could be added to the functions to propagate the region information.
+
+Tagged unions for closed-world types
+------------------------------------
+
+A union type with a small number of concrete types (for example `A | B`) will be represented in the MVP compiler as an interface that contains all of the methods in `A` and all of the methods in `B` (and possibly some others generated by the compiler to support later casts).
+Casting back to `A` or `B` requires comparing the descriptor pointer.
+This requires materialising the global addresses of the descriptors and loading from the object.
+
+We could alternatively represent these using the low bit(s) as a discriminator.
```
rengolin, line 137: I thought we were using the low bits for snmalloc purposes.

If not, then we could only apply this for small closed-world unions (up to 4).

rengolin (commented):
Another opportunities that I remember discussing and was not included in the doc is region metadata for alias analysis.

If we do that, we need to juggle between eliminating the metadata for size purposes during inlining, but also keeping them for alias analysis, especially after inlining, which won't be a trivial process.

Perhaps we could "mark the metadata for removal" but not remove it, and only after we're sure we've done all we can to annotate `noalias` everywhere we can, we can safely drop the metadata.

PR #550, opened 2022-03-16
Atomic Enqueue - Split Execution

marioskogias:
This PR builds on top of the acquire-all semantics and enables enqueueing multiple behaviours atomically while their execution can still happen in parallel.


mjp41:
New PR on Verona-rt superseeds this.

PR #572, opened 2022-04-21
C++ When support

mjp41:
This PR adds a new C++ API for using Cowns and When.  It is designed to look like Verona syntax but uses C++ style smart pointers.

There are two examples using this.


mjp41:
@lukecheeseman I can't tag you as a reviewer, but tagging you here.

mjp41:
@marioskogias @ihaller tagging so you know this has happened.

src/rt/cpp/cown.h
```diff
@@ -0,0 +1,165 @@
+// Copyright Microsoft and Project Verona Contributors.
+// SPDX-License-Identifier: MIT
+
+#include <functional>
+#include <tuple>
+#include <utility>
+#include <verona.h>
+
+class cown_ptr_trait
```
davidchisnall, line 9: What is this for?

src/rt/cpp/cown.h
```diff
@@ -0,0 +1,165 @@
...
+  /**
+   * Internal Verona runtime cown for this type.
+   */
+  class ActualCown : public VCown<ActualCown>
+  {
+  private:
+    T value;
+
+  public:
+    ActualCown(T&& t) : value(std::forward<T>(t)) {}
```
davidchisnall, line 36: Is this supposed to be a forwarding constructor (in which case it should probably be a variadic template) or a move constructor that assumes a constructed `T`?  A comment would be good to see which.

src/rt/cpp/cown.h
```diff
@@ -0,0 +1,165 @@
...
+    return allocated_cown;
+  }
+
+  /**
+   * Construct a new cown ptr object, actually allocates a runtime cown.
+   *
+   * This is internal, and the `make_cown` below is the public interface,
+   * which has better behaviour for implicit template arguments.
+   */
+  cown_ptr(T&& t) : allocated_cown(new ActualCown(std::forward<T>(t))) {}
```
davidchisnall, line 61: Same question as above.  This looks as if it's move-constructing (or possible copy-constructing?) a cown from an existing `T`, but the use of `std::forward` means that this looks as if it should actually be emplacing a `T`.

I *think* that the `std::forward` here should be `std::move`.

src/rt/cpp/cown.h
```diff
@@ -0,0 +1,165 @@
...
+
+  ~cown_ptr()
+  {
+    // Condition to handle moved cown ptrs.
+    // And a defence against the destructor being called twice.
+    if (allocated_cown != nullptr)
+    {
+      auto& alloc = verona::rt::ThreadAlloc::get();
+      verona::rt::Cown::release(alloc, allocated_cown);
+      allocated_cown = nullptr;
```
davidchisnall, line 92: Note that this will almost certainly be optimised away above O0 because it is a dead store according to the abstract machine (it is UB to read from a field after the destructor has finished).

src/rt/cpp/cown.h
```diff
@@ -0,0 +1,165 @@
...
+    }
+  }
+
+  // Required as acquired_cown has to reach inside.
+  template<typename TT>
+  friend class acquired_cown;
+
+  template<typename TT>
+  friend cown_ptr<TT> make_cown(TT&& t);
```
davidchisnall, line 101: Should this class be friends with all variants of `acquired_cown` and `make_cown` or only ones with matching `T`?

src/rt/cpp/cown.h
```diff
@@ -0,0 +1,165 @@
...
+
+  template<typename...>
+  friend class When;
+};
+
+/**
+ * Used to construct a new cown_ptr to `t`.
+ *
+ * TODO: Need to improve the forwarding versus copy behaviour here.
```
davidchisnall, line 110: This should be a variadic template if it is intended to be forwarding the arguments.

src/rt/cpp/cown.h
```diff
@@ -0,0 +1,165 @@
...
+ * underlying access.
+ */
+template<typename T>
+class cown_ptr : cown_ptr_trait
+{
+private:
+  /**
+   * Internal Verona runtime cown for this type.
+   */
+  class ActualCown : public VCown<ActualCown>
```
davidchisnall, line 30: It would be good to be consistent about naming.  Here the type starts with a capital (which, I believe, is the style that we've used everywhere else), but `cown_ptr` and `acquired_cown` use lower case and underscore separators.

src/rt/cpp/cown.h
```diff
@@ -0,0 +1,165 @@
...
+  {
+    return origin_cown.allocated_cown->value;
+  }
+
+  operator T&()
+  {
+    return origin_cown.allocated_cown->value;
+  }
+
+  acquired_cown(acquired_cown&&) = delete;
```
davidchisnall, line 161: Please add a doc comment explaining why these are deleted (i.e. it's the closest that you can do to a no-capture requirement in C++: something can incorrectly capture a reference to this, but can't capture a copy of its internal state).

src/rt/cpp/when.h
```diff
@@ -0,0 +1,125 @@
...
+ *
+ *   when (cown1, ..., cownn) << closure;
+ *
+ * Allows the variadic number of cowns to occur before the closure.
+ */
+template<typename... Args>
+class When
+{
+  template<typename... Args2>
+  friend When<Args2...> when(Args2... args);
```
davidchisnall, line 27: Should this be friends with any `when` instantiation or only the one with `Args...`?

src/rt/test/perf/banking/verona_banks.cc
```diff
@@ -0,0 +1,113 @@
...
+
+  // We don't want to use the same account.
+  while (to_idx == from_idx)
+    to_idx = worker->rand.next() % NUM_ACCOUNTS;
+
+  // Schedule a transfer
+  int64_t amount = 100;
+  Accounts& accounts = *worker->accounts;
+  when(accounts[from_idx], accounts[to_idx])
+    << [=](acquired_cown<Account> from, acquired_cown<Account> to) {
```
davidchisnall, line 61: Can this be:

```c++
    << [=](auto from, auto to) {
```

Not a suggested change, but if this works then it would be nice for a comment to saying that type inference works here and you can use `auto` to name the `cown`s in the lambda, but the types are given here explicitly to make the example clearer.

src/rt/test/perf/dining_phil/verona_dining_phil.cc
```diff
@@ -0,0 +1,112 @@
...
+  // This code handles tables by splitting the index into the table and then
+  // the philosopher index.  It wraps the fork around for the last philosoher
+  // on each table.
+  size_t table_size = NUM_PHILOSOPHERS / NUM_TABLES;
+  size_t table = index / table_size;
+  size_t next_fork = (index + 1) % table_size;
+  return forks[(table * table_size) + next_fork];
+}
+
+struct Philosopher
```
davidchisnall, line 43: I guess that, in Verona, this would always be allocated in a new region?  The `unique_ptr` below is used as an approximation of `iso`?

src/rt/test/perf/dining_phil/verona_dining_phil.cc
```diff
@@ -0,0 +1,112 @@
...
+{
+  // This code handles tables by splitting the index into the table and then
+  // the philosopher index.  It wraps the fork around for the last philosoher
+  // on each table.
+  size_t table_size = NUM_PHILOSOPHERS / NUM_TABLES;
+  size_t table = index / table_size;
+  size_t next_fork = (index + 1) % table_size;
+  return forks[(table * table_size) + next_fork];
+}
+
```
davidchisnall, line 42: This is a lot more code than the pthread example.  For showing Verona in a good light, it would be nice if you could use the same code for accessing the left / right fork in a circular array of `T` and use `std::mutex` or `cown_ptr<Fork>` as `T` so that it's clearer what the differences are.

src/rt/test/perf/dining_phil/verona_dining_phil.cc
mjp41: Yeah, unique_ptr is the approximation of iso, and shared_ptr is the approximation of imm(utable).

src/rt/cpp/cown.h
mjp41: It should just be where TT=T, but C++ dislikes me and says things like:
```
.../verona/src/rt/./cpp/cown.h:105:16: error: explicit specialization of undeclared template class 'acquired_cown'
  friend class acquired_cown<T>;
```
If I try and do it differently.

src/rt/cpp/cown.h
rengolin: It's actually helpful when using LLVM containers/features that C++ didn't have in some previous standard but does have now. So, when we deprecate previous compilers (and bump the minimum standard), it's a lot easier to spot the LLVM ones from the standard ones.

I know it's not a very strong argument, but I'm also not sure we need to mimic the C++ standard naming here either.

src/rt/cpp/cown.h
```diff
@@ -0,0 +1,193 @@
...
+    other.allocated_cown = nullptr;
+  }
+
+  ~cown_ptr()
+  {
+    // Condition to handle moved cown ptrs.
+    if (allocated_cown != nullptr)
+    {
+      auto& alloc = verona::rt::ThreadAlloc::get();
+      verona::rt::Cown::release(alloc, allocated_cown);
```
rengolin, line 105: In the copy ctor above, does this release after the new copy acquire it? As is this the intended semantics?

src/rt/test/perf/banking/pthread_banks.cc
```diff
@@ -0,0 +1,66 @@
...
+{
+  size_t a, b;
+  xoroshiro::p128r64 rand;
+  rand.set_state(id + 1);
+
+  for (size_t i = 0; i < (TRANSACTIONS / NUM_WORKERS); i++)
+  {
+    a = rand.next() % NUM_ACCOUNTS;
+    b = a;
+    while (b == a)
```
rengolin, line 35: If `NUM_ACCOUNTS == 1`, this will loop forever. :)

src/rt/cpp/cown.h
```diff
@@ -0,0 +1,165 @@
...
+  template<typename...>
+  friend class When;
+
+private:
+  // Underlying cown that has been acquired.
+  // TODO: Look to optimise away the reference count here, as the
+  // runtime already holds one for the duration of the `when`.
+  cown_ptr<T> origin_cown;
+
+  acquired_cown(const cown_ptr<T>& origin) : origin_cown(origin) {}
```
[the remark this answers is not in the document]
mjp41: This needs further investigation, so left a TODO in the code.

src/rt/cpp/cown.h
mjp41: It needs to be like this.  If you want to optimise, the caller can use the move constructor.

src/rt/test/perf/banking/pthread_banks.cc
mjp41: This is checked here:
https://github.com/microsoft/verona/pull/572/files#diff-c36710122675c2f451aed2355723cde9536ee9f011ed0ffd7a9d138ac1d09b27R22-R26

PR #578, opened 2022-04-26
Fix #552

### LINKED ISSUE #552: Infrequent memory leak in func-con-external_event_source
We are occassionally hitting a bug on this test:
https://dev.azure.com/ProjectVeronaCI/Project%20Verona/_build/results?buildId=3717&view=logs&j=b0abc04d-bfd5-55cd-41dd-470d60a21bba&t=0a299330-b715-5f5b-5b0e-8fa56ea0992f

We should enable better systematic testing on this, now it can cover external threads, to see if it is deterministically reproducible.

mjp41:
Fixes #552.  Their was a sporadic memory leak only observed on Windows.  This PR adds some logging that helped diganose the issue, and fixes the issue. 

The poller cown is decrefed by the external source, however this decref can occur
after the runtime has executed the teardown, and thus is leads to a memory
leak.

This commit changes the order so the last thing the external event source does
is decrease the event source count, and thus the runtime cannot begin teardown
until the external thread has completed all its actions.


ihaller:
Great job finding the bug. The change looks good to me.

mjp41:
Fixes https://github.com/microsoft/verona/issues/552. Their was a sporadic memory leak only observed on Windows. This PR adds some logging that helped diganose the issue, and fixes the issue.

The poller cown is decrefed by the external source, however this decref can occur
after the runtime has executed the teardown, and thus is leads to a memory
leak.

This commit changes the order so the last thing the external event source does
is decrease the event source count, and thus the runtime cannot begin teardown
until the external thread has completed all its actions.

PR #582, opened 2022-04-27
Add Sean Allen to people

SeanTAllen:


rengolin:
Do you have a MSR page and a picture? Otherwise just shows up a weird question mark and no link.

PR #589, opened 2022-05-05
Add manual lock ordering to dining phils benchmark

lukecheeseman:


mjp41 (approved):
LGTM, the fail is transient and not your fault. I haven't managed to work out what is causing it.

mjp41 (approved):

Issue #588, opened 2022-05-03
check for failure in Library.alloc


saaramar:
The parent calls `lib.alloc` in order to allocate memory inside the sandbox, and does not check for failure. Such allocations could fail (the sandbox has a fixed size).

An example is in `handle_getaddrinfo` (libsandbox.cc):

```c++
          // Allocate space for everything.  No overflow checking here, but the
          // called code is trusted and we could only overflow if it returned a
          // nonsense `ai_addrlen`.
          char* buffer = lib.alloc<char>((sizeof(addrinfo) * count) + extra);
          // The allocated space will be an array of `addrinfo`s, followed by
          // (variable-sized) `sockaddr`s.
          addrinfo* ais = reinterpret_cast<addrinfo*>(buffer);
          char* sockaddrs = buffer + (sizeof(addrinfo) * count);
          size_t i = 0;
          // Copy each list element into the new buffer, updating its next
          // pointer and copying the payload into the end of the buffer.
          for (addrinfo* cur = res; cur != nullptr; cur = cur->ai_next, i++)
          {
            ais[i] = *cur;
            memcpy(sockaddrs, cur->ai_addr, cur->ai_addrlen);
            ais[i].ai_addr = reinterpret_cast<sockaddr*>(sockaddrs);
            sockaddrs += cur->ai_addrlen;
            ais[i].ai_next = &ais[i + 1];
          }
          // Add the null terminator in the list.
          ais[count - 1].ai_next = nullptr;
          *unsafeSandboxRes = ais;
          netpolicy.freeaddrinfo(res);
```

Issue #586, opened 2022-05-01
`handle` doesn't check the return value of send


saaramar:
The `handle` function (in libsandbox.cc) does not check the return value of `socket.send`. In this commit https://github.com/microsoft/verona/commit/cb3c14d62cd3d5340114cc94ccb73e6a47564cc7 , there is a new code that verifies the sandbox behaves correctly, and if not - terminates the sandbox.

`handle` should do the same. As a rule - if the parent cannot properly communicate with the sandbox, it's better to assume the sandbox does something wrong/malicious, and just terminate the communication.

These issues could be fixed by set all the send/receive functions to have [[nodiscard]].

@davidchisnall

Issue #584, opened 2022-04-29
Issue in `handle` - use of uninitialized stack due to lack of return value check


saaramar:
The function `handle` in libsandbox.cc looks as follows:

```c++
   /**
     * Handle a request.
     */
    void handle(Library& lib)
    {
      CallbackRequest req;
      platform::Handle in_fd;
      // FIXME: This should not block, but it can if the sandbox doesn't write
      // anything into the socket.
      socket.receive(&req, sizeof(req), in_fd);

      CallbackHandlerBase::Result ret;
      if (req.kind < handlers.size())
      {
        ret = handlers[req.kind]->invoke(lib, req, std::move(in_fd));
      }
      socket.send(&ret.integer, sizeof(ret.integer), ret.handle);
    }
```

Please note two important observations here:
1. The `req` field is uninitialized.
2. The function `socket.receive` returns a `bool`, and `handle` fails to check for that. Which means, that attackers can make the call to `socket.receive` to fail, and `req` has uninitialized stack values (in the parent).

You can see that `CallbackRequest` does not have default values in its definition:
```c++
  struct CallbackRequest
  {
    /**
     * The kind of message.
     */
    CallbackKind kind;
    /**
     * The size of the message payload (`data`).  All messages are fixed size,
     * and so it should be possible to deduce this from the `kind`, but this
     * acts as an extra sanity check.
     */
    size_t size;
    /**
     * The payload of this message.  This is a pointer into the sandbox's heap.
     * It is up to the receiver of this to validate this value.  In a CHERI
     * coprocess world, this would require rederivation and so introduce
     * confused deputy attacks that could bypass in-sandbox memory safety.
     * That should be fixed by replacing the use of UNIX domain sockets with a
     * cocall or similar abstraction.
     */
    uintptr_t data;
  };
```

Now, the sandbox could trigger dereference and use of uninitialized stack values in the parent, which is a security issue.

I have the following POC, which sends only 4 bytes (to set `req.kind`) and closes the socket, which creates the scenario of `req.size` and `req.data` uninitialized in the parent. The output is as follows:

```
root@da99df90fb82:/verona/experiments/process_sandbox/build-ninja# ./tests/test-sandbox-uninit-req
[parent] call socket.receive()
[receive_msg] ret == 24
[parent] req.kind == 0x0
[parent] req.size == 0x10
[parent] req.data == 0x7f0c8000c080
[receive_msg] ret == 8
[parent] call socket.receive()
[receive_msg] ret == 24
[...]

PR #592, opened 2022-05-19
Security fixes

### LINKED ISSUE #583: cross sandbox DOS in handle
The function `handle` in libsandbox.cc has a blocking `recvmsg` call, which means that if the sandbox won't write anything, it could make the parent block. This means other sandbox instances won't get responses for their requests.

A similar bug existed in `run` (https://github.com/microsoft/verona/issues/566) and it was fixed (https://github.com/microsoft/verona/pull/567). A similar fixed should be applied here.

There is already a comment on that in the code:

```c++
    /**
     * Handle a request.
     */
    void handle(Library& lib)
    {
      CallbackRequest req;
      platform::Handle in_fd;
      // FIXME: This should not block, but it can if the sandbox doesn't write
      // anything into the socket.
      socket.receive(&req, sizeof(req), in_fd);

      CallbackHandlerBase::Result ret;
      if (req.kind < handlers.size())
      {
        ret = handlers[req.kind]->invoke(lib, req, std::move(in_fd));
      }
      socket.send(&ret.integer, sizeof(ret.integer), ret.handle);
    }
```

The underlying functionality is in socketpair_posix.h in `receive_msg`:
```c++
          do
          {
            ret = recvmsg(fd, &header, MSG_NOSIGNAL);
          } while ((ret == -1) && (errno == EAGAIN));
```

@davidchisnall

davidchisnall:
Two commits that should fix #583, #584, #586, and #588.

Kept as two commits for (maybe) easier review.

@saaramar, please can you take a look?

### DIFF
--- a/.github/workflows/update-snmalloc.yaml
+++ b/.github/workflows/update-snmalloc.yaml
@@ -51,11 +51,11 @@ jobs:
         LOCAL_REV=$(git ls-tree HEAD $SUBMODULE_PATH | awk '{ print $3 }')
 
         # Export these so they can be used in the rest of the job.
-        echo ::set-env name=SUBMODULE_PATH::$SUBMODULE_PATH
-        echo ::set-env name=REMOTE_REV::$REMOTE_REV
-        echo ::set-env name=REMOTE_SHORT_REV::$(git rev-parse --short $REMOTE_REV)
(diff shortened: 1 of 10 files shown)


saaramar:
Thanks @davidchisnall, looks great to me!

PR #595, opened 2022-06-07
Read-only Cowns

lukecheeseman:
Cowns can be acquired for either reading or writing:
- a behaviour requiring to read a cown can be dispathced as long as there are no writers (but there may be writers)
- a behaviour requring to write a cown can be dispatched and as long as there no readers or writers
- The causallity changes for behaviours b and b':
  - Take the functions:
    - $writecownsof(b)$ gives only the writeable cowns required by $b$.
    - $readcownsof(b)$ gives only the read-only cowns required by $b$.
    - $allcownsof(b) = readcownsof(b) \cup writecownsof(b)$.
  - Then the 'overlap' part of $b$ happens before $b'$ is:
    -  $(allcownsof(b) \cap writecownsof(b')) \cup (allcownsof(b') \cap writecownsof(b)) \neq \emptyset$
    - note: prior to this PR, there was 'overlap' if $allcownsof(b) \cap allcownsof(b') \neq \emptyset$ as for all $b''$ $allcownsof(b'') = writecownsof(b'')$ .
  - in other words, $b$ happens before $b'$ iff $b$ writes a cown that $b'$ requires in anyway or vice versa.
  - this means there are happens before between write/write, read/write, write/read on the same cown
  - but not between read/read of the same cown
  - If a behaviour acquires a cown as both read and write, both are granted and the happens before order will not allow the behaviour to execute in parallel with other behaviours using the same cown (aside: I think acquiring the same cown twice currently deadlocks in the runtime).

In the modern cpp API:
  - requiring a `cown_ptr<T>` is a write request
  - requiring a `cown_ptr<const T>` is a read request
  - a `read` operator transforms a `cown_ptr<T>` to a `cown_ptr<const T>`
  - Thus we can read cowns as:
```
cown_ptr<Account> mycown = make_cown<Account(10)>;
when(read(mycown)) << [](acquired_cown<const Account> myreadonlycown) { ... };
```

Scheduling changes:
- Each cown has a read reference count.
  - The bottom bit is reserved for signalling a pending write (note: there can only be one pending write as this will be the head of the cowns queue). 
  - The top bits are the actual read ref count.
  - Thus read_ref modifications for reads are all of the form fetch_{sub/add}(2)
- Each request for a cown is accompanied with an access mode - read or write
- When a scheduler thread is working on a cown:
  - peek (and do not dequeue) the next message.
  - if the next message is for writing and there are reader's of the cown
     -  atomically fetch and set the bottom bit
     - if between the read and write the ref count hit 0 the clear the bottom bit and continue handling the message
     - otherwise, unschedule this cown
   - otherwise, dequeue and process the message:
     - if this is a read request
       - increment the read ref count of the cown
       - reschedule this cown
       - If this is the last cown to be acquired, we schedule the cown immediately-ish, but note that this thread should not schedule this cown after the behaviour.
  - after a behaviour executes:
    - decrement the read ref count of all read-only cowns
      - if any cown will hit 0 but the write bit is set (i.e. fetch_sub(2) == 3), either:
        - schedule that cown (if it is not this cown)
        - reinstate that this scheduler thread should reschedule this cown (if it is this cown).
    - schedule all writeable cowns that are not this cown.


src/rt/sched/cown.h
```diff
@@ -764,6 +877,33 @@ namespace verona::rt
...
+          {
+            // if in the time between reading and writing the ref count, it
+            // became zero, we can now process the write, so clear the flag
+            // and continue
+            if (read_ref_count.fetch_add(1) == 0)
+              read_ref_count.fetch_sub(1);
+            else
+              return false;
+          }
+        }
```
mjp41, line 905: Add a yield and check that `read_ref_count == 0`

src/rt/sched/cown.h
```diff
@@ -647,23 +701,64 @@ namespace verona::rt
...
+          else if (request.mode == AccessMode::READ)
+          {
+            // If this read is the last read of a cown, and that cown's next
+            // message requires writing, clear the flag and either:
+            //   - schedule the cown if it is not this cown, or
+            //   - continue processing this cown's message on this scheduler
+            //   thread
+            //     (overwriting the previous decision to stop processing this
+            //     cown).
+            size_t rc = cown->read_ref_count.fetch_sub(2);
```
mjp41, line 744: Just to check my understanding.  If `rc` is made `0` by this subtract, it is fine.  It may or not be scheduled, it is just up to if it has been marked sleeping.

mjp41 (changes requested):
I really like this solution.  It cleanly integrates with the existing code. 

The thing I found a little confusing a cown can have read behaviours executing on it, but be "sleep"ing.  This is actually really nice.  Sleep just means the queue is known to be empty, and thus anyone that sends it a message will have to schedule it.

I think the most complex case is understanding that if the refcount has the bottom bit set, then it is neither sleeping nor scheduled.  It is waiting for the last reader to reschedule it.  

CI is complaining about a few minor things that need fixing.  

I think the PR comment is a little confusing for what rocowns means.  If a cown is requested as both ro and wr, I am unsure what is happening.  Does `cownsof` contain cowns that are both ro and wr?

src/rt/sched/cown.h
lukecheeseman: Yes, if `cown->read_ref_count` hits `0` here it only signals that there are no more reads happening on this cown right now, there may be nothing left in its queue to process, or we may not have processed the next mesage yet.

src/rt/sched/cown.h
```diff
@@ -52,18 +52,41 @@ namespace verona::rt
...
    */
+
+  enum class AccessMode
+  {
+    WRITE,
+    READ
+  };
+
+  struct Request
+  {
```
mjp41, line 85: Could you make this more abstract with accessors, so that we can use the bottom bit of the pointer for the mode.  You don't need to do the bit packing, but make a class that could be changed without affecting anything else.

src/rt/sched/cown.h
```diff
@@ -137,6 +160,13 @@ namespace verona::rt
      **/
     std::atomic<size_t> weak_count{1};
 
+    /*
+     * Cown's read ref count.
```
mjp41, line 164: Could you make a class that encapsulates this logic?

In particular, there are just a few operations

* `fetch_add(2)` - add read
* try_write - true for no readers, false marks that a writer is waiting.
* release_read 

Then all this logic can be nicely encapsulated in one spot.  Currently it is spread out a bit more than might be easy to follow.

mjp41 (changes requested):
I really like this.  Just a few bits to improve the encapsulation.

mjp41 (approved):

mjp41:
Sorry I broke early release with the merge.

src/rt/cpp/when.h
```diff
@@ -55,24 +78,30 @@ namespace verona::cpp
...
      */
     template<typename C>
-    static acquired_cown<C> actual_cown_to_acquired(ActualCown<C>* c)
+    static acquired_cown<C> access_to_acquired(Access<C> c)
     {
-      assert(c != nullptr);
-      return acquired_cown<C>(*c);
+      assert(c.t != nullptr);
+      if constexpr (std::is_const<C>())
+        return acquired_cown<C>(*reinterpret_cast<ActualCown<const C>*>(c.t));
```
ihaller, line 102: Could the acquired_cown constructor be improved to not need the reinterpret_cast? Or a private conversion in ActualCown. Intuition says that adding const should not require reinterpret-cast.

src/rt/sched/behaviour.h
```diff
@@ -11,6 +11,36 @@ namespace verona::rt
...
+    Cown* _cown;
+
+    static const size_t READ_FLAG = 1;
+
+    Request() : _cown(nullptr) {}
+    Request(Cown* cown) : _cown(cown) {}
+
+    Cown* cown()
+    {
+      return (Cown*)((size_t)_cown & ~READ_FLAG);
```
ihaller, line 25: Could this size_t be uintptr_t instead?

src/rt/sched/behaviour.h
```diff
@@ -11,6 +11,36 @@ namespace verona::rt
...
+    }
+
+    static Request write(Cown* cown)
+    {
+      return Request(cown);
+    }
+
+    static Request read(Cown* cown)
+    {
+      return Request((Cown*)((size_t)cown | READ_FLAG));
```
ihaller, line 40: Could this size_t be uintptr_t instead?

src/rt/sched/behaviour.h
```diff
@@ -11,6 +11,36 @@ namespace verona::rt
...
+    Request(Cown* cown) : _cown(cown) {}
+
+    Cown* cown()
+    {
+      return (Cown*)((size_t)_cown & ~READ_FLAG);
+    }
+
+    bool is_read()
+    {
+      return ((size_t)_cown & READ_FLAG);
```
ihaller, line 30: Not as critical as the other two, since the pointer is never regenerated, but potentially uintptr_t is still better to be used here.

src/rt/cpp/lambdabehaviour.h
```diff
@@ -70,6 +70,13 @@ namespace verona::rt
       count, cowns, std::forward<T>(f));
   }
 
+  template<TransferOwnership transfer = NoTransfer, typename T>
+  static void schedule_lambda(size_t count, Request* requests, T f)
```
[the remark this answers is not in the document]
mjp41: agreed.

src/rt/cpp/when.h
mjp41: I think I have done what you are suggesting in:
[d3fd1a0](https://github.com/microsoft/verona/pull/595/commits/d3fd1a0c97b377e6475b94220097517320e06d5c)

src/rt/cpp/cown.h
```diff
@@ -186,6 +206,11 @@ namespace verona::cpp
   template<typename T, typename... Args>
   cown_ptr<T> make_cown(Args&&... ts)
   {
+    static_assert(
```
mjp41, line 209: @lukecheeseman  I've added an assert here to stop people conflicting `const` means.

mjp41 (approved):
I'm really happy with this.  @Lukechesseman could you sign off that you are okay with my modifications to your branch.

src/rt/cpp/when.h
ihaller: Excellent!

ihaller:
LGTM

src/rt/cpp/cown.h
```diff
@@ -229,32 +229,40 @@ namespace verona::cpp
...
+    ActualCown<std::remove_const_t<T>>& origin_cown;
 
     /// Constructor is private, as only `When` can construct one.
-    acquired_cown(ActualCown<T>& origin) : origin_cown(origin) {}
+    acquired_cown(ActualCown<std::remove_const_t<T>>& origin) : origin_cown(origin) {}
 
   public:
     /// Get a handle on the underlying cown.
-    cown_ptr<T> cown() const
+    cown_ptr<std::remove_const_t<T>> cown() const
```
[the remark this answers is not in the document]
lukecheeseman: That sounds good to me.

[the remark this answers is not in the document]
lukecheeseman: That sounds good to me.

src/rt/cpp/cown.h
lukecheeseman: That sounds good to me, there isn't really a use case for an always const cown anyway as you should probably use a shared immutable.

PR #596, opened 2022-06-07
Updating snmalloc

mjp41:
A minimal change to upgrade to the latest snmalloc.

I think the process sandbox could potentially be simplified, but I was trying to make the smallest (and fastest) change.

### DIFF
--- a/experiments/process_sandbox/include/process_sandbox/sandbox.h
+++ b/experiments/process_sandbox/include/process_sandbox/sandbox.h
@@ -68,6 +68,9 @@ namespace sandbox
     {
     private:
       /**
+       * Concrete instance of a pagemap.  This is updated only with
+       * `pagemap_lock` held.
+       *
        * This pagemap spans the entire address space (i.e. does not use the
        * fixed-range option) because it covers all sandboxes.  Each sandbox is
        * (currently) a fixed range within the global address space.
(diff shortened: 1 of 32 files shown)


experiments/process_sandbox/include/process_sandbox/sandbox.h
```diff
@@ -67,24 +66,15 @@ namespace sandbox
...
-       * Pagemap entry.
-       */
-      using Entry = SandboxMetaEntry;
-
     private:
       /**
-       * Concrete instance of a pagemap.  This is updated only with
-       * `pagemap_lock` held.
-       *
        * This pagemap spans the entire address space (i.e. does not use the
```
mjp41, line 71:
```suggestion
       * Concrete instance of a pagemap.  This is updated only with
       * `pagemap_lock` held.
       *
       * This pagemap spans the entire address space (i.e. does not use the
```
Dropped by mistake

experiments/process_sandbox/include/process_sandbox/sandbox.h
```diff
@@ -111,6 +101,8 @@ namespace sandbox
       inline static std::recursive_mutex pagemap_lock;
 
     public:
+      using Entry = PagemapEntry;
```
mjp41, line 104:
```suggestion
      /**
       * Pagemap entry.
       */
      using Entry = PagemapEntry;
```

davidchisnall (approved):
Process sandbox changes look good to me.

PR #602, opened 2022-07-11
Fix process sandbox for double faults.

davidchisnall:
The `snmalloc::message` function calls `fsync` on `stderr`.  With
Capsicum, this is an explicit permission and so tests that hit a code
path that did some logging (specifically, the curl test) in debug builds
during system call emulation would receive the second trap.

The first symptom of this is fixed by adding fsync permission to the
standard out / error file descriptors.  With that change, all of the
tests pass but the underlying problem remains: the signal is delivered
*after* return from the first system call, at which point it does not
have a sensible `ucontext` and things break in exciting ways
(sufficiently exciting that a debugger can't tell you anything useful).

The root cause is addressed by adding `SA_NODEFER` when registering the
signal handler.  This allows system calls that are used to emulate
other signal handlers to deliver their signals on top of the stack.  If
a system call is used in emulating itself, this still goes horribly
wrong, but in a way that's much easier to debug.

At the same time, use the SIGCAP support that is in process of being
upstreamed to FreeBSD.  This makes debugging Capsicum failures easier
because the debugger does not try to hide the signal from the
application whose correctness depends on handling it.


mjp41 (approved):
LGTM

PR #603, opened 2022-07-12
Fix bug in C++ cown api

mjp41:


src/rt/cpp/cown.h
```diff
@@ -120,19 +121,30 @@ namespace verona::cpp
...
+     */
+    void clear()
     {
       // Condition to handle moved cown ptrs.
       if (allocated_cown != nullptr)
       {
         auto& alloc = verona::rt::ThreadAlloc::get();
         verona::rt::Cown::release(alloc, allocated_cown);
       }
+      allocated_cown = nullptr;
```
ihaller, line 142: Nitpick, but this could just go in the conditional right? Although it is more readable this way and maybe I am unaware of some codegen benefits.

ihaller (approved):
LGTM

PR #605, opened 2022-08-18
Fix cown freeing policy

hlef:
Currently the scheduler thread collects cown stubs (frees them) whenever
the number of freed but uncollected cowns reaches the total number of
ever allocated cowns divided by two. As total_cowns is strictly
increasing, the system can reach a point where total_cowns is so big
that cowns don't get freed before the system falls out of memory.

This commit introduces a fixed threshold of freed but uncollected cowns,
`FREE_COWNS_THRESHOLD`, with a default of 100, after which these cowns
are systematically collected. The value of 100 is arbitrary and has been
validated in a performance setting and did not result in any negative
performance difference compared to the previous setting.

Signed-off-by: Hugo Lefeuvre <t-hlefeuvre@microsoft.com>

### DIFF
--- a/src/rt/sched/schedulerthread.h
+++ b/src/rt/sched/schedulerthread.h
@@ -41,6 +41,7 @@ namespace verona::rt
     template<typename Owner>
     friend class Noticeboard;
 
+    static constexpr uint64_t FREE_COWNS_THRESHOLD = 100;
     static constexpr uint64_t TSC_QUIESCENCE_TIMEOUT = 1'000'000;
 
     T* token_cown = nullptr;
@@ -162,7 +163,7 @@ namespace verona::rt
       while (true)
       {
         if (
-          (total_cowns < (free_cowns << 1))
+          (free_cowns > FREE_COWNS_THRESHOLD)
 #ifdef USE_SYSTEMATIC_TESTING
           || Systematic::coin()
 #endif


hlef:
FYI this test is failing on Windows RT (and only there):
```
 23/424 Test  #13: runtime/func-con-external_event_source ..........Exit code 0xc0000409
***Exception:  28.70 sec
Harness starting.
 D:/a/1/s/build/func-con-external_event_source.exe --seed 1486664878
Seed: 1486664878
debug_is_empty: found non-empty allocator: size=0x80 (0x7)
```

Not certain why, could it be that the number of freed cowns does not reach 100? Could that be an issue?

mjp41:
That test failure is not you. I have been trying to find the underlying cause for months...

mjp41:
I think the correct fix is to change:
https://github.com/microsoft/verona/blob/950cd20d08b11b73e4064c99fe11f85a0a972766/src/rt/sched/schedulerthread.h#L734
To be 
```
  free_cowns -= count;
  total_cowns -= count;
```
I think `total_cowns` should be the length of the list.

hlef:
That would also address the issue, although I am not 100% convinced that it is a perfect fix either. In situations where a large number of cowns are allocated, cowns will only be freed in very large batches (half of the number of allocated cowns), resulting in increased memory pressure and therefore potentially a performance drop.

I also think that having such a variable threshold to free cowns might make the performance behaviour of Verona harder to understand and predict for users, as opposed to a fixed threshold.

Of course the fixed threshold has its own drawbacks, because we then need to choose an appropriate one in the general case ("100 ought to be enough for everybody")

mjp41:
Because the cowns are stored in a singly linked list they are removed by a linear scan.  This has a performance cost of O(total_cowns), if we do that for a fixed limit, then we actually make the code considerably worse in terms of complexity class.  Doing as a ratio means we still have an amortized linear overhead rather than quadratic with what you are proposing.

mjp41:
Superseeded by #608

PR #607, opened 2022-08-30
Optimise the RC for cown_ptr

mjp41:
The previous implementation required a lot of copies of `cown_ptr` that resulted in potentially poor performance.  This implementation does not suffer from that problem as it uses a "univeral reference" for each cown to enable passing without causing decref and increfs.

### DIFF
--- a/src/rt/cpp/cown.h
+++ b/src/rt/cpp/cown.h
@@ -212,7 +212,7 @@ namespace verona::cpp
     ActualCown<T>& origin_cown;
 
     /// Constructor is private, as only `When` can construct one.
-    acquired_cown(ActualCown<T>* origin) : origin_cown(*origin) {}
+    acquired_cown(ActualCown<T>& origin) : origin_cown(origin) {}
 
   public:
     /// Get a handle on the underlying cown.
--- a/src/rt/cpp/cown.h
+++ b/src/rt/cpp/cown.h
@@ -34,6 +34,32 @@ namespace verona::cpp
   template<typename... Args>
   class When;
 
+  template<typename T>
+  class cown_ptr;
+
+  /**
+   * Internal Verona runtime cown for this type.
+   */
+  template<typename T>
+  class ActualCown : public VCown<ActualCown<T>>
+  {
+  private:
+    T value;
+
+    template<typename... Args>
+    ActualCown(Args&&... ts) : value(std::forward<Args>(ts)...)
+    {}
+
+    template<typename TT>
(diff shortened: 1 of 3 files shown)


src/rt/cpp/cown.h
```diff
@@ -195,33 +206,32 @@ namespace verona::cpp
   private:
     /// Underlying cown that has been acquired.
     /// Runtime is actually holding this reference count.
-    cown_ptr<T> origin_cown;
+    ActualCown<T>* origin_cown;
```
davidchisnall, line 209: Can this not be a reference?  We don't ever have a null pointer here do we?

src/rt/cpp/when.h
```diff
@@ -100,25 +94,44 @@ namespace verona::cpp
...
+              };
 
             std::apply(lift_f, cown_tuple);
           });
       }
     }
   };
 
+  /**
+   * Internal: required for template inference.
```
davidchisnall, line 111: Do we require C++17?  If so, you can probably do this with a [deduction guide](https://en.cppreference.com/w/cpp/language/class_template_argument_deduction) instead of a wrapper function.

src/rt/cpp/cown.h
```diff
@@ -31,6 +31,32 @@ namespace verona::cpp
     friend class cown_ptr;
   };
 
+  template<typename T>
+  class cown_ptr;
+
+  /**
+   * Internal Verona runtime cown for this type.
```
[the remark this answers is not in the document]
mjp41: I'll expand the comment

src/rt/cpp/cown.h
mjp41: Correct

ihaller:
LGTM

src/rt/cpp/cown.h
```diff
@@ -206,32 +209,32 @@ namespace verona::cpp
   private:
     /// Underlying cown that has been acquired.
     /// Runtime is actually holding this reference count.
-    ActualCown<T>* origin_cown;
+    ActualCown<T>& origin_cown;
 
     /// Constructor is private, as only `When` can construct one.
-    acquired_cown(ActualCown<T>* origin) : origin_cown(origin) {}
+    acquired_cown(ActualCown<T>* origin) : origin_cown(*origin) {}
```
davidchisnall, line 215:
```suggestion
    acquired_cown(ActualCown<T> &origin) : origin_cown(origin) {}
```

PR #608, opened 2022-08-30
Fix `total_cowns` invariant

mjp41:
The variable `total_cowns` should be the length of the per-thread cown list.
This is not the case, this commit checks that this is the case, and correctly
updates it.

### DIFF
--- a/src/rt/sched/schedulerthread.h
+++ b/src/rt/sched/schedulerthread.h
@@ -78,8 +78,12 @@ namespace verona::rt
 
     // The number of cowns in the per-thread list `list`.
     size_t total_cowns = 0;
+
     // The number of cowns that have been collected in the per-thread list
-    // `list`.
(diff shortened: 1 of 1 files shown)


mjp41:
@hlef this is my preferred solution to #605.

src/rt/sched/schedulerthread.h
```diff
@@ -75,7 +75,11 @@ namespace verona::rt
     SchedulerStats stats;
 
     T* list = nullptr;
+
+    // The number of cowns in the per-thread list `list`.
     size_t total_cowns = 0;
+    // The number of cowns that have been collected in the per-thread list
```
davidchisnall, line 81: Why is this atomic if it is per thread?

mjp41: Other threads can collect it, but not remove from the list.  I will expand the comment

PR #609, opened 2022-09-01
Consolidate allocation of message body

mjp41:
Prior to this commit the multimessage was composed of three allocations, the body, the cowns array and the behaviour/closure to execute.  This PR merges the three structures into a single more complex allocation. On @theodus's perf-con-ubench, we see an 18% improvement in message flow.  

It also includes two small code tidies.

### DIFF
--- a/src/rt/sched/multimessage.h
+++ b/src/rt/sched/multimessage.h
@@ -60,8 +60,10 @@ namespace verona::rt
         auto body = new (alloc.alloc(size)) Body(count);
         new ((Be*)&(body->get_behaviour())) Be(std::forward<Args>(args)...);
 
-        static_assert(alignof(Be) <= sizeof(void*), "Alignment not supported, yet!");
-        static_assert(sizeof(Body) % alignof(Be) == 0, "Alignment not supported, yet!");
+        static_assert(
+          alignof(Be) <= sizeof(void*), "Alignment not supported, yet!");
+        static_assert(
+          sizeof(Body) % alignof(Be) == 0, "Alignment not supported, yet!");
 
         return body;
       }
--- a/src/rt/sched/multimessage.h
+++ b/src/rt/sched/multimessage.h
@@ -15,7 +15,7 @@ namespace verona::rt
   class MultiMessage
   {
     /**
-     * This represents a message that is send to a behaviour.
+     * This represents a message that is sent to a behaviour.
      *
      * The layout requires that after the allocation there is space for
      * `count` cown pointers, and then the behaviour's body.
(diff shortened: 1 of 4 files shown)


src/rt/sched/multimessage.h
```diff
@@ -14,29 +14,61 @@ namespace verona::rt
...
-    MultiMessageBody* body;
+    // The body of the actual message.
+    Body* body;
     friend verona::rt::MPSCQ<MultiMessage>;
     friend class Cown;
 
     std::atomic<MultiMessage*> next{nullptr};
 
-    inline MultiMessageBody* get_body()
+    inline Body* get_body()
```
davidchisnall, line 66: This can return `nullptr`, but it probably should be an assertion failure if it would.  Can we make it return a `Body&` and add `assert((Body*)((uintptr_t)body & ~Object::MARK_MASK) != nullptr)`?

src/rt/sched/multimessage.h
```diff
@@ -14,29 +14,61 @@ namespace verona::rt
...
+      {
+        size_t size = sizeof(Body) + (sizeof(Cown*) * count) + behaviour_size;
+        return new (alloc.alloc(size)) Body(count);
+      }
     };
 
   private:
-    MultiMessageBody* body;
+    // The body of the actual message.
+    Body* body;
```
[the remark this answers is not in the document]
mjp41: The way we use snmalloc at the moment doesn't wouldn't work like that.  But I can leave a TODO.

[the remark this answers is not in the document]
mjp41: Actually, it can't be a unique_ptr for an actual multimessage, the body will be shared between all the other messages.

src/rt/sched/multimessage.h
mjp41: So I looked at this change, but we currently do
```
alloc.dealloc(m->get_body());
```
and I didn't really like that with the reference. It felt wrong.

src/rt/sched/multimessage.h
```diff
@@ -14,29 +14,61 @@ namespace verona::rt
 
   class MultiMessage
   {
-    struct MultiMessageBody
+    /**
+     * This represents a message that is send to a behaviour.
```
ihaller, line 18: send -> sent?

src/rt/sched/multimessage.h
davidchisnall: I think it feels wrong because it is wrong: it leaves body as a dangling pointer.  We should be deallocating and nulling out the pointer at the same time, possibly with a `clear_body` method.

ihaller: I agree with David on a more dedicated method. I am just considering if there is an argument to consider how ownership flows outside of this class. For example, should ownership flow to `Scheduler::local()->message_body`?

mjp41: I think this is a more complex refactor than I want in this PR. I will create a second PR address this nicely.

PR #611, opened 2022-09-01
Token source

mjp41:
This PR provides an API for giving a back pressure like behaviour, but purely at the C++ level without special runtime support (well it uses std::atomic internally, so would be considered part of the runtime).

This is based on my initial prototyping, and then @hlef usage of that API, and subsequent requests for change.

I am raising a PR to get feedback on the design, and see if this is the kind of thing we should add.

### DIFF
--- a/src/rt/cpp/cown.h
+++ b/src/rt/cpp/cown.h
@@ -46,6 +46,11 @@ namespace verona::cpp
   private:
     T value;
 
+      void schedule(bool fifo = true)
+      {
+        VCown<ActualCown>::schedule(fifo);
+      }
+
     template<typename... Args>
     ActualCown(Args&&... ts) : value(std::forward<Args>(ts)...)
     {}
@@ -58,6 +63,8 @@ namespace verona::cpp
 
     template<typename TT, typename... Args>
     friend cown_ptr<TT> make_cown(Args&&... ts);
+
+    friend class Token;
   };
 
   /**
@@ -83,7 +90,7 @@ namespace verona::cpp
     /**
      * Accesses the internal Verona runtime cown for this handle.
      */
-    Cown* underlying_cown()
+    ActualCown<T>* underlying_cown()
     {
       return allocated_cown;
     }
@@ -176,6 +183,8 @@ namespace verona::cpp
 
     template<typename...>
     friend class When;
+
+    friend class Token;
   };
 
   /**
--- a/src/rt/cpp/token.h
+++ b/src/rt/cpp/token.h
@@ -116,6 +116,11 @@ namespace verona::cpp
       }
 
     public:
+      size_t available_tokens()
+      {
+        return state->max_inflight - state->inflight.load();
+      }
+
       template<typename F>
       void get_token(F f) &&
       {
--- a/src/rt/cpp/token.h
+++ b/src/rt/cpp/token.h
@@ -95,6 +95,8 @@ namespace verona::cpp
       // Reference to the underlying state.
       State* state;
 
+      size_t local_tokens = 0;
+
       Source(State* state) : state(state) {}
 
       /**
@@ -104,7 +106,13 @@ namespace verona::cpp
       {
         if (state)
         {
+          // Return any locally held tokens.
+          state->inflight -= local_tokens;
+          local_tokens = 0;
+
+          // Return the sources count.
           auto old = state->inflight--;
+
           // If this is the last reference to the state,
           // then it is responsible for deallocating it.
           if (old == 1)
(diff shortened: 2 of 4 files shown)


src/rt/cpp/token.h
```diff
@@ -0,0 +1,305 @@
...
+   * its destructor runs, and that enables the source to provide another token.
+   */
+  class Token
+  {
+    /**
+     * Internal function that provides a cown that can be waited on.
+     * This is like a create promise, just the minimal thing required for this
+     * example.
+     */
+    static cown_ptr<size_t> make_wait()
```
marioskogias, line 31: How about using verona::rt::Promise for that? It might be easier to follow.

marioskogias:
The code makes sense and it's easy to understand and follow.

However, the usecase is not super clear to me. If I were to implement the mechanism in the example, I would make the slow_work to schedule generate_work and close the loop this way without the need for tokens. The level of parallelism can be controlled with the number of initial parallel calls to generate_work, which I guess is the equivalent with the initial number of tokens. In a way Verona already has a built-in token mechanism which is the messages themselves. As long as the number of messages stays bounded, and when consuming one message another one is generated, the system is stable. Are there any usecases that require the token solution and cannot be solved with the slow bit closing the loop?

src/rt/cpp/token.h
mjp41: Yeah, I would like to tidy up the Promise code a little to make it fit nicely here, but then I was going to make a larger change.  I'll move to promises before we integrate this, if we decide too.

mjp41:
> The code makes sense and it's easy to understand and follow.
> 
> However, the usecase is not super clear to me. If I were to implement the mechanism in the example, I would make the slow_work to schedule generate_work and close the loop this way without the need for tokens. The level of parallelism can be controlled with the number of initial parallel calls to generate_work, which I guess is the equivalent with the initial number of tokens. In a way Verona already has a built-in token mechanism which is the messages themselves. As long as the number of messages stays bounded, and when consuming one message another one is generated, the system is stable. Are there any usecases that require the token solution and cannot be solved with the slow bit closing the loop?

The PRs proposal would have up to `n` things running where there are `n` tokens.  I think if we close the loop in the way you are proposing, there would always be `n` things running, and then we wouldn't be able to stop polling and let other work come in. The `when` body would then have to do the scale up and down of the number of the behaviours.  

So, I did initially implement it just using `when`.  That required effectively building a count of how many things you don't want to run at the moment.

The main issue this has is you need to provide a few different pieces as lambdas to it.  With the proposed API that becomes a more sequential programming task, rather than a functional programming one.

I think this is the kind of example I was thinking about based on what you and @hlef have talked about.

```C++
main_loop(TokenSource ts)
{
  size_t i = ts.available_tokens();

  if (i == 0)
  {
     // We need to wait as the system is overloaded
     ...
  }

  for (; i > 0; i--)
  {
    auto result = get_packet();
    if (result == nullptr)
      break;  // Out of packets yield for a bit
    when () << [result, token = std::move(ts.get_token())]
    { ... }
  }

  // Yield for a bit
  when () << [ts = std::move(ts)](){ main_loop(std::move(ts)); }
}
```

marioskogias:
You can also have up to n things running my proposal. The number of outstanding messages, or the i in the above example, should be internal state of the cown in the context of which main_loop runs and there is another message/lambda that decrements/increments that value. I can write that in code to make it more clear.

My second concern about this solution, apart from redundancy, is that it is effectively a semaphore that also allows for busy waiting. If there are no available tokens, nothing prevents me from trying again, instead of calling wait_for_token, right?

mjp41:
@Marioskogias I am sure we can write it without, just not sure how clean it will be.

The semaphore point is very good and I had not considered it.

ihaller:
The Token mechanism allows you to detect that the event has been processed end-to-end without needing to add extra state to the event itself. So not just `n` outstanding messages, but `n` events that have not been fully processed yet being somewhere in the system.

My only comment would be that we might want to give the main_loop pattern as a function taking lambdas so that users have a clear method of using token without needing to implement it correctly. But we might want a couple more examples of usage first to generalize the pattern better.

src/rt/cpp/cown.h
```diff
@@ -83,7 +90,7 @@ namespace verona::cpp
     /**
      * Accesses the internal Verona runtime cown for this handle.
      */
-    Cown* underlying_cown()
+    ActualCown<T>* underlying_cown()
```
ihaller, line 93: Can we have a comment to potentially revert this if a better interface comes along for the inactive cown in the future?

mjp41:
Moved to runtime repo

PR #612, opened 2022-09-05
Decoupling SchedulerThreads & Cown queues

aghosn:
Introducing the `Core` layer in the Verona Runtime scheduler.
Cores hold the cown queues while SchedulerThreads are assigned to a core.
This refactoring enables to add new scheduler threads as needed during the execution of the runtime.

### DIFF
--- a/src/rt/ds/dllist.h
+++ b/src/rt/ds/dllist.h
@@ -1,0 +1,197 @@
+#pragma once
+
+#include <cassert>
+#include <type_traits>
+
+// Inspired by the snmalloc implementation of a double linked list.
+namespace verona::rt
+{
+  template<
+    class T,
+    class Terminator = std::nullptr_t,
+    bool delete_on_clear = false>
+  class DLList final
+  {
+  private:
+    static_assert(
+      std::is_same<decltype(T::prev), T*>::value, "T->prev must be a T*");
+    static_assert(
+      std::is_same<decltype(T::next), T*>::value, "T->next must be a T*");
+
+    T* head = Terminator();
+    T* tail = Terminator();
+
+  public:
+    ~DLList()
+    {
+      clear();
+    }
+
+    DLList() = default;
+
+    DLList(DLList&& o) noexcept
+    {
+      head = o.head;
+      tail = o.tail;
+
+      o.head = nullptr;
+      o.tail = nullptr;
+    }
+
+    DLList& operator=(DLList&& o) noexcept
+    {
+      head = o.head;
+      tail = o.tail;
+
+      o.head = nullptr;
+      o.tail = nullptr;
+      return *this;
+    }
+
+    bool is_empty()
+    {
+      return head == Terminator();
+    }
+
+    T* get_head()
+    {
+      return head;
+    }
+
+    T* get_tail()
+    {
+      return tail;
+    }
+
+    T* pop()
+    {
+      T* item = head;
+      if (item != Terminator())
+        remove(item);
+
+      return item;
+    }
+
+    T* pop_tail()
+    {
+      T* item = tail;
+      if (item != Terminator())
+        remove(item);
+      return item;
+    }
+
+    void insert(T* item)
+    {
+#ifndef NDEBUG
+      debug_check_not_contains(item);
+#endif
+      item->next = head;
+      item->prev = Terminator();
+      if (head != Terminator())
+        head->prev = item;
+      else
+        tail = item;
+      head = item;
+#ifndef NDEBUG
+      debug_check();
+#endif
+    }
+
+    void insert_back(T* item)
+    {
+#ifndef NDEBUG
+      debug_check_not_contains(item);
+#endif
+      item->prev = tail;
+      item->next = Terminator();
+
+      if (tail != Terminator())
+        tail->next = item;
+      else
+        head = item;
+      tail = item;
+#ifndef NDEBUG
+      debug_check();
+#endif
+    }
+
+    void remove(T* item)
+    {
+#ifndef NDEBUG
+      debug_check_contains(item);
+#endif
+      if (item->next != Terminator())
+        item->next->prev = item->prev;
+      else
+        tail = item->prev;
+
+      if (item->prev != Terminator())
+        item->prev->next = item->next;
+      else
+        head = item->next;
+#ifndef NDEBUG
+      debug_check();
+#endif
+    }
+    
+    void clear()
+    {
+      while (head != nullptr)
+      {
+        auto c = head;
+        remove(c);
+        if (delete_on_clear)
+        {
+          delete c;
+        }
+      }
+    }
+
+    void debug_check_contains(T* item)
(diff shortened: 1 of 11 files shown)


src/rt/sched/threadpool.h
```diff
@@ -21,7 +24,7 @@ namespace verona::rt
   inline void nop() {}
 
   using namespace snmalloc;
-  template<class T>
+  template<class T, class E>
```
mjp41, line 27: Is `E` going to be `Cown` here?  I think a comment, and possibly a different letter would help.

src/rt/sched/threadpool.h
```diff
@@ -82,13 +83,24 @@ namespace verona::rt
 
     ThreadState state;
 
+    /// Pool of cores shared by the scheduler threads.
+    CorePool<ThreadPool<T, E>, E>* core_pool = nullptr;
+
+    /// Systematic ids.
+    size_t systematic_ids = 0;
```
mjp41, line 90: Does this need to be `atomic`?  The ids might appear in logs even if systematic testing is not running.

src/rt/sched/threadpool.h
```diff
@@ -330,67 +336,77 @@ namespace verona::rt
...
+      while (t != nullptr)
       {
-        T* next = t->next;
-        delete t;
-        t = next;
-      } while (t != first_thread);
+        T* prev = t;
+        t = threads->popFree();
+        delete prev;
+      }
```
mjp41, line 368: Perhaps add this as a method on `SchedulerList`? It is easier to understand the correctness in there.

src/rt/sched/threadpool.h
```diff
@@ -61,17 +64,15 @@ namespace verona::rt
     ThreadSync<T> sync;
 #endif
 
-    std::atomic_uint64_t barrier_count = 0;
     uint64_t barrier_incarnation = 0;
 
-    T* first_thread = nullptr;
+    /// List of instantiated scheduler threads.
+    /// Contains both free and active threads; protects accesses with a lock.
+    SchedulerList<T>* threads = nullptr;
```
mjp41, line 71: Could this be a `SchedulerList<T>`, then we don't have to worry about its lifetime? Rather than a separate allocation.

src/rt/sched/threadpool.h
```diff
@@ -82,13 +83,24 @@ namespace verona::rt
 
     ThreadState state;
 
+    /// Pool of cores shared by the scheduler threads.
+    CorePool<ThreadPool<T, E>, E>* core_pool = nullptr;
```
mjp41, line 87: Does this need to be a separate allocation?

mjp41 (changes requested):
It looks good.  A few minor bits here that should be addressed.

src/rt/sched/threadpool.h
aghosn: I get an error from snmalloc:
```
error: variable does not have a constant initializer
 SNMALLOC_REQUIRE_CONSTINIT static ThreadPool<T, C> global_thread_pool;
```

src/rt/sched/threadpool.h
aghosn: Same as above apparently.

src/rt/sched/schedulerthread.h
```diff
@@ -72,36 +74,36 @@ namespace verona::rt
     EpochMark prev_epoch = EpochMark::EPOCH_B;
 
     ThreadState::State state = ThreadState::State::NotInLD;
-    SchedulerStats stats;
 
     T* list = nullptr;
```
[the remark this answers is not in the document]
aghosn: If I remember correctly, this list is for cowns that need to be collected, right?
Having it per core would require extra synchronization on the collect path.

src/rt/sched/threadpool.h
aghosn: Switch to `C`

src/rt/sched/schedulerthread.h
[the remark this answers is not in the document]
mjp41: Well the statistic about this list are now per-core.  `free_cowns` and `total_cowns` refer to the list.

It also means that all thread would need to be scheduled to clear cowns.  An "extra" thread could be owning cowns, but then never restarted, and we will leak the base of those cowns.

src/rt/sched/schedulerlist.h
```diff
@@ -0,0 +1,131 @@
...
+    snmalloc::FlagWord m;
+
+    DLList<T> active;
+    DLList<T> free;
+
+  public:
+    SchedulerList() {}
+    ~SchedulerList()
+    {
+      snmalloc::FlagLock lock(m);
```
mjp41, line 30: I like having debug information here, but it might prevent the inlining of the structure into the thread pool.  As this is not perf sensitive, then it might be fine to leave it as a separate allocation.

src/rt/sched/threadpool.h
aghosn: Done.

src/rt/sched/core.h
```diff
@@ -0,0 +1,134 @@
...
+    /// Progress and synchronization between the threads.
+    //  These counters represent progress on a CPU core, not necessarily on
+    //  the core's queue. This is necessary to take into account core-stealing
+    //  to avoid spawning many threads on a core hogged by a long running
+    //  behaviour but with an empty cown queue.
+    std::atomic<std::size_t> progress_counter = 0;
+    std::atomic<std::size_t> servicing_threads = 0;
+    std::atomic<std::size_t> last_worker = 0;
+
+    // The number of cowns in the per-thread list `list`.
```
mjp41, line 31:
```suggestion
    // The number of cowns in the per-core list `list`.
```

src/rt/sched/core.h
```diff
@@ -0,0 +1,134 @@
...
+    std::atomic<size_t> total_cowns = 0;
+
+    // The number of cowns that have been collected in the per-thread list
```
mjp41, line 34:
```suggestion
    // The number of cowns that have been collected in the per-core list
```

src/rt/ds/dllist.h
```diff
@@ -0,0 +1,199 @@
...
+    T* head = Terminator();
+    T* tail = Terminator();
+
+  public:
+    ~DLList()
+    {
+      clear();
+    }
+
+    DLList() = default;
```
mjp41, line 32:
```suggestion
    constexpr DLList() = default;
```

src/rt/sched/schedulerthread.h
```diff
@@ -698,7 +701,13 @@ namespace verona::rt
         default:;
       }
 
-      T** p = &list;
+      assert(core != nullptr);
+      // TODO not sure about the logic here.
```
mjp41, line 705: Let's discuss in person.

mjp41 (approved):

PR #617, opened 2022-09-27
Remove leak detector

mjp41:
The leak detector's current design is far from the ideal design we want going forward.  It is designed to run frequently, but since changing the language semantics to use strong and weak reference counts on cowns, this is not the expected use.

The PR removes the code to make future changes easier.


src/rt/test/func/cowngc1/cowngc1.cc
```diff
@@ -304,20 +300,18 @@ struct Ping : public VBehaviour<Ping>
         Cown::schedule<Pong>(c2, c2);
       }
 
-      // Randomly introduce a few leaks. We don't want to do this for every
-      // Ping, only about a quarter.
+      // Randomly introduce a pointer nulling operatins. We don't want to do
```
ihaller, line 303: `operations`

ihaller:
LGTM

ihaller (approved):

PR #618, opened 2022-09-28
adding == and != comparison operators for cown_ptr and nullptr

lukecheeseman:


mjp41:
LGTM.  CI is complaining about `nullptr_t` should be `std::nullptr_t`.

davidchisnall:
Do we still need to support C++17?  If not, can we replace this with a spaceship and `std::partial_ordering` (or possibly `strong_ordering`, since I think `cown`s need to have a total ordering)?

PR #623, opened 2022-10-10
Some refactoring of reference count manipulation

mjp41:
Overall, reduces the amount of reference count required on a cown from one per message to one per scheduler queue.


mjp41:
@lukecheeseman you might like to see if this improves performance for you.

PR #624, opened 2022-10-10
Fix bug in weak reference counting

mjp41:


mjp41:
@septract, @mattwindsor91 this is the bug I was discussing with you.

The PR is in draft as I need to improve the writing, but I think the proof and code are pretty reasonable.

PR #626, opened 2022-10-13
Small optimisations and refactors around message send and receive

mjp41:


src/rt/sched/multimessage.h
```diff
@@ -69,39 +83,66 @@ namespace verona::rt
       }
     };
 
+    struct Delivered
+    {
+      Body& body;
```
ihaller, line 88: Can we have a comment here to document that this is an owning reference. Should the members be potentially const (the closest we can get to public read private write)?

src/rt/sched/multimessage.h
```diff
@@ -69,39 +83,66 @@ namespace verona::rt
...
     {
-      MultiMessage* m = make(alloc, body);
-      Logging::cout() << "MultiMessage " << m << " payload " << body
-                      << Logging::endl;
-      return m;
+      return (body_and_mode & 1) == 1;
     }
 
-    inline size_t size()
+    Delivered deliver(Alloc& alloc)
```
ihaller, line 134: Can we have a comment here about ownership transfer for the body object?

mjp41:
Moved to new repo https://github.com/microsoft/verona-rt/pull/1

PR #645, opened 2023-10-04
Reinstate old test suite

mjp41:
This is a simple output comparison based testsuite.

The previous version of this is on the `old-version` branch.


src/parse.cc
```diff
@@ -75,21 +75,21 @@ namespace verona
     p("start",
       {
         // Blank lines terminate.
-        "\n(?:[[:blank:]]*\n)+([[:blank:]]*)" >>
+        "\r?\n(?:[[:blank:]]*(\r\n|\n))+([[:blank:]]*)" >>
```
mjp41, line 78:
```suggestion
        "\r?\n(?:[[:blank:]]*\r?\n)+([[:blank:]]*)" >>
```

PR #649, opened 2023-12-06
Adding language motivation document

lnsolar:
Okay, so I added a first draft of an attempt at summarizing the verona language. At the moment it mostly contains parts which I have already talked about with some of you in the weekly type system meetings. It would be great if you have a read through and comment here.

### DIFF
--- a/docs/motivation.md
+++ b/docs/motivation.md
@@ -56,7 +56,10 @@ intuitive syntax. For example, if we want to describe the type signature of a fu
 together, this would be
 ```verona
 f(n1: Number & ToString, n2: Number & ToString) : Number
-// TODO add ToStringNumber version
+```
+or equivalently
+```verona
+f(n1: ToStringNumber, n2: ToStringNumber) : Number
 ```
 
 In contrast, in a nominal system like Java, we would have to define a new
@@ -244,9 +247,9 @@ class D : C {
 If we want to describe a relation between `C` and `D`, we can only do so by
 subtyping them under a trait.
 
-* Question: There is a case to be made for not allowing inheritance from a
+There is a case to be made for not allowing inheritance from a
   class, since this allows unanticipated code reuse. This might however also be
-  a positive. The question if we want to allow it remains open.
+  a positive.
 
 ##### Trait inheritance
 Traits can be used for inheritance with their default methods.


viewedlobster:
Some things i need to do

* Make sure syntax is correct in code examples
* Add comparisons with other languages

davidchisnall:
It would be nice to start such a document talking about the two key differentiating features of the language: region-based ownership and behaviour-oriented concurrency.  I worry that someone coming to the language and seeing this would not see what makes Verona interesting.

viewedlobster:
@microsoft-github-policy-service agree

viewedlobster:
> It would be nice to start such a document talking about the two key differentiating features of the language: region-based ownership and behaviour-oriented concurrency. I worry that someone coming to the language and seeing this would not see what makes Verona interesting.

I agree. I think we aimed to explain the main features of the type system, with less focus on capabilities, regions and BoC. I could try to add parts about these as well, or perhaps change the introduction to more clearly state what the document is supposed to be.

docs/motivation.md
```diff
@@ -0,0 +1,535 @@
...
+type ToString = {
+    toString(s: Self) : String
+}
+
+type Printable = {
+    print(s: Self) : Unit // this will have to be defined if Self </: ToString
+    print(s: Self) : Unit where (Self <: ToString) {
+        sys.io.out(s.toString())
+    }
+}
```
EliasC, line 520: Is there always a most specific constraint? There is a parallel to pattern matching where you can have a matching order that does not depend on the order that cases are written, but on their specificity, and then you need to look out for patterns that cannot be ordered, e.g., the patterns for tuples of lists `(nil, l2)` and `(l1, nil)`. I wonder if a similar problem can appear here.

EliasC (commented):
This is a really useful document to have! I have added some comments and questions, and also opened a pull request with spelling fixes.

mjp41 (changes requested):

mjp41:
@viewedlobster do you think you will be able to make the requested changes?

docs/motivation.md
```diff
@@ -0,0 +1,535 @@
...
+In Verona, a trait with multiple methods can be further simplified by breaking
+up each method into what we call an "atomic" trait and combining them with a
+conjunction. E.g. the definition of `Number` above is simply syntactic sugar for
+```verona
+type Number = {
+    add(self: Self, other: Number): Number
+} & {
+    sub(self: Self, other: Number): Number
+}
+```
```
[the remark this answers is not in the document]
viewedlobster: It probably doesn't make much difference to most people. On the other hand, I don't think it hurts to mention it?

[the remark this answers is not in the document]
EliasC: I think it might even help with the intuition for `&`, reading it again now. But maybe it should be more of a note than a proposed simplification (I would argue that it is less simple to compose atomic traits than to write a composite trait).

docs/motivation.md
```diff
@@ -0,0 +1,535 @@
...
+
+* TODO: compare with
+    - objective c: categories
+    - ruby: open classes
+    - C#: extension methods
+
+### Disjunction types
+Verona also includes disjunction types. Assuming two classes `C` and `D` we can
```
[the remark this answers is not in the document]
viewedlobster: I think we use them to describe the pattern match which I think requires them to be classes? We could perhaps generalize the first part and then adding the assumption of classes when we describe pattern matching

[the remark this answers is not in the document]
EliasC: Sounds good!

docs/motivation.md
```diff
@@ -0,0 +1,535 @@
...
+define the new type
+```verona
+type COrD = C | D
+```
+
+Since in Verona there is no inheritance between classes, this will allow us to
```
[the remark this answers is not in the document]
viewedlobster: 👍

docs/motivation.md
```diff
@@ -0,0 +1,535 @@
...
+  ```verona
+    type A[T] = {
+        type Self
+        type Anon = {
+            type Self
+            g(self: A[T].Anon.Self) : A[T].Anon.Self
+        }
+        f(self: A[T].Self) : A[T].Anon
+    }
+  ```
```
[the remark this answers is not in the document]
viewedlobster: I think this desugaring is wrong, since self is basically an implicit type parameter

docs/motivation.md
```diff
@@ -0,0 +1,535 @@
...
+let z : ToString = if (...) { C::create() } else { D::create() }
+let u = z.toString() // dynamic dispatch, since ToString is an open type
+```
+
+### Type predicates
+Type predicates allows us to describe assumptions about subtyping. These can be
+written on both type and method level, and both has their distinct uses.
+Specifically, predicates on type level allow us to express F-bounded
+polymorphism, while method level predicates are useful in reducing code
+duplication, somewhat akin to type classes.
```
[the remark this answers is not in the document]
viewedlobster: true

docs/motivation.md
viewedlobster: don't know, we should discuss later

docs/motivation.md
```diff
@@ -0,0 +1,535 @@
...
+// class C definition
+class C {
+    toString(s: C) : String
+    ...
+}
+
+// file 2
+// declaration that C implements Printable together with its default method
+class C : Printable
+```
```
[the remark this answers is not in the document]
viewedlobster: probably, we should discuss where we should put this

docs/motivation.md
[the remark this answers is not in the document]
viewedlobster: yeah, i agree, i'll see if i have time to write something before the meeting

docs/motivation.md
```diff
@@ -0,0 +1,550 @@
...
+This document aims to cover the parts of the type system that are similar to (though in some places diverge in interesting ways from) other languages.
+
+Verona is an object oriented reference capability language with structural
+subtyping and bounded polymorphism. This document aims to describe the aspects
+of the type system that are independent from both the concurrency model and
+reference capabilities.
+
+## Structural type system
+
+A type system is all about managing a complexity budget. Make it too complicated
```
TobiasWrigstad, line 14: Nitpick: I'm not sure it is *all* about this! But I morally agree :)

docs/motivation.md
```diff
@@ -0,0 +1,550 @@
...
+This gives us a flexible way to define combinations of types on the fly, with an
+intuitive syntax. For example, if we want to describe the type signature of a function
+`f` printing two numbers and then returning the result of adding them
+together, this would be
+```verona
+f(n1: Number & ToString, n2: Number & ToString) : Number
+// TODO add ToStringNumber version
+```
+
+In contrast, in a nominal system like Java, we would have to define a new
```
davidchisnall, line 62: The Java equivalent of this is `Object<Number, ToString>`.

docs/motivation.md
sylvanc: But this is important. We need to avoid such phrases. Perhaps: "one aspect of a type system is".

docs/motivation.md
```diff
@@ -0,0 +1,550 @@
...
+Verona is an object oriented reference capability language with structural
+subtyping and bounded polymorphism. This document aims to describe the aspects
+of the type system that are independent from both the concurrency model and
+reference capabilities.
+
+## Structural type system
+
+A type system is all about managing a complexity budget. Make it too complicated
+and programmers will find it difficult to use and be less productive. Make it too
+simple and it will give you less guarantees to lean on as a programmer.
```
sylvanc, line 16: Similarly, we're provoking argument with an implicit productivity claim. For this to be effective, we need to back up the claim. Also s/less/fewer/.

docs/motivation.md
```diff
@@ -0,0 +1,550 @@
...
+
+We opt to use a structural type system since we beleive it simplifies some parts
+of reasoning, e.g. we don't need co/contra-variance notation on type parameters.
```
sylvanc, line 19: Let's avoid the word "believe". The more concrete claim is that we are eliminating type hierarchies and improving extension (classes may be subtypes of structural types that are added later).

docs/motivation.md
```diff
@@ -0,0 +1,550 @@
...
+This allows us to spend this complexity elsewhere, e.g. introducing reference
+capabilities to enforce isolation and region topology.
+
+We do however treat classes nominally and have opted to not have inheritance
+subtyping. This means that there is never any subtyping between two distinct
+classes.
```
sylvanc, line 25: I would prefer to avoid the word "opted", and state our concete motivations. In this case, we are distinguishing classes from types - a class is a closed binary layout.

We also need to highlight early on here that the type system is both structural and algebraic, allowing both "open world" and "closed world" subtyping, in both cases without hierarchies.

docs/motivation.md
```diff
@@ -0,0 +1,553 @@
...
+```
+both `C` and `D` are subtypes of `Eq`, but neither `C` or `D` are subtypes of
+the other.
+
+#### Inheritance
+There are two forms of inheritance in
+Verona. One between classes and one between class and trait.
+
+##### Class inheritance
+Inheritance between classes works as you would expect in a language like
```
matajoh, line 230: It is not clear from the document (at this stage) whether you allow multiple inheritance. Is this the purpose of "language like Java" here?

docs/motivation.md
```diff
@@ -0,0 +1,553 @@
...
+```verona
+class C {
+    equal(self: C, other: C): Bool {...}
+    f(...) : ... {...}
+}
+
+// here we reuse functionality of C in D, specifically the method f, but
+// we do not get a subtyping relation D <: C
+class D : C {
+    equal(self: D, other: D): Bool {...}
```
nwf-msr, line 242: What would have happened if `D` hadn't redefined `equal` like this?  Would it have picked up an `equal : (C,C) -> Bool` that then could be called with `::` as in `D::equal(c1,c2)` but not via `.`, since `D </: C`?

docs/motivation.md
```diff
@@ -0,0 +1,550 @@
...
+This allows us to spend this complexity elsewhere, e.g. introducing reference
+capabilities to enforce isolation and region topology.
+
+We do however treat classes nominally and have opted to not have inheritance
+subtyping. This means that there is never any subtyping between two distinct
+classes.
+
+### Traits
+Verona is a structural type system based on traits, i.e. types can be described
+using combination of methods on an object. E.g. we can describe a type
```
sylvanc, line 29: This is so minor, but: "e.g." is "for example" in Latin, requires a comma after it, and is bad form to use at the beginning of a sentence. Here it can be "object, e.g., we". "i.e." is "inter alia" and also requires a comma. Using both in the same sentence is an indicator that you probably had an entirely different structure in mind and you might want to rewrite :)

docs/motivation.md
```diff
@@ -0,0 +1,550 @@
...
+```verona
+type ToString = {
+    toString(self: Self): String
+}
+```
+A type describing numbers implementing addition and subtraction can be described with
+```verona
+type Number = {
+    add(self: Self, other: Number): Number
+    sub(self: Self, other: Number): Number
```
sylvanc, line 40: This is an opportunity to use symbolic function names:

type Number =
{
  +(self: Self, other: Number): Self
  -(self: Self, other: Number): Self
}

Also, a "self types" section is needed.

docs/motivation.md
```diff
@@ -0,0 +1,553 @@
...
+class C {
+    m1(self: Self) : Self
+    m2(self: C) : C
+    // in verona, these two method signatures are equivalent up to method name.
+}
+```
+
+For trait types, Self acts as as a reference to the underlying concrete type of the instantiation. This allows us to write e.g.
+```verona
+type Collection[T] = {
```
matajoh, line 308: Have you explained the `[T]` notation so far? This seemed to come out of nowhere.

docs/motivation.md
```diff
@@ -0,0 +1,550 @@
...
+type Number = {
+    add(self: Self, other: Number): Number
+    sub(self: Self, other: Number): Number
+}
+```
+Traits also support default implementation as described below in section [Trait Inheritance](#trait-inheritance).
+
+
+### Conjunction types
+We can easily combine traits using conjunctions, so if we, e.g., want to describe
```
sylvanc, line 47: Just remove the "e.g.".

docs/motivation.md
```diff
@@ -0,0 +1,550 @@
...
+```java
+interface ToStringNumber extends Number, ToString { ... }
+
+class C {
+    static Number f(ToStringNumber n1, ToStringNumber n2) { ... }
+}
+```
+Furthermore we would have to change the definition of every class that we want
+to implement this new trait.
+
```
sylvanc, line 76: This belongs in the discussion of structural types, not algebraic types.

docs/motivation.md
viewedlobster: haha, yeah, wrote that in a heated moment 😓

docs/motivation.md
```diff
@@ -0,0 +1,550 @@
...
+In Verona, a trait with multiple methods can be further simplified by breaking
```
sylvanc, line 77: This isn't a simplification. It's true mathematically, but to a human, assigning names to complex concepts is a simplification, and this is the opposite.

docs/motivation.md
```diff
@@ -0,0 +1,550 @@
...
+up each method into what we call an "atomic" trait and combining them with a
```
sylvanc, line 78: I would not call this an "atomic" trait. Avoid using any concurrency words in text that isn't about concurrency. Also avoid adding terminology when explaining terminology.

docs/motivation.md
```diff
@@ -0,0 +1,550 @@
...
+conjunction. For example, the definition of `Number` above is simply syntactic sugar for
```
sylvanc, line 79: While this is true, I don't think this clarifies anything, and can be removed.

docs/motivation.md
```diff
@@ -0,0 +1,550 @@
...
+```verona
+type Number = {
+    add(self: Self, other: Number): Number
+} & {
+    sub(self: Self, other: Number): Number
+}
+```
+
+#### Capabilities and conjunction
+In Verona, each concrete type is tagged with a *reference capability*. The reference
```
sylvanc, line 89: This is false. Types don't have reference capabilities. References have reference capabilities. And such references may not have a concrete type.

docs/motivation.md
mjp41: I didn't know that, which version of Java added that?

docs/motivation.md
```diff
@@ -0,0 +1,550 @@
...
+capability decides which operations are permitted through and with a reference
+depending on the viewpoint. For example, a reference with capability `mut` allows us to
+read and mutate the object, while a capability `iso` is unique and opaque to all
+read/write operations.
```
sylvanc, line 93: Partially introducing reference capabilities will lead to confusion. The full set and their meaning needs to be discussed at once.

docs/motivation.md
```diff
@@ -0,0 +1,553 @@
...
+    C => y.toString() // allows method call inlining since call is now static
+    D => y.toString()
+}
+
+let z : ToString = if (...) { C::create() } else { D::create() }
+let u = z.toString() // dynamic dispatch, since ToString is an open type
+```
+
+### Type predicates
+Type predicates allow us to describe assumptions about subtyping. These can be
```
matajoh, line 342: I think for most readers the distinction between subtyping here and what we say above (i.e. " inheritance
subtyping. This means that there is never any subtyping between two distinct classes.") may be lost, i.e. they would react with "but I thought there was no subtyping in this language". Maybe it is just me, but it seems a bit confusing.

docs/motivation.md
```diff
@@ -0,0 +1,550 @@
...
+```
+
+#### Capabilities and conjunction
+In Verona, each concrete type is tagged with a *reference capability*. The reference
+capability decides which operations are permitted through and with a reference
+depending on the viewpoint. For example, a reference with capability `mut` allows us to
+read and mutate the object, while a capability `iso` is unique and opaque to all
+read/write operations.
+
+The conjunction type operator plays well together with capabilities, allowing us
```
sylvanc, line 95: Avoid phrases like "plays well". There is no explanation or motivation here.

docs/motivation.md
```diff
@@ -0,0 +1,553 @@
+# The Verona language: Core type system
+
+Verona is a language with a behaviour-oriented concurrency model and a region abstraction for ownership.
```
matajoh, line 3: General question: who is the audience for this? And what do we want the effect of reading this document to be on them?

docs/motivation.md
```diff
@@ -0,0 +1,550 @@
...
+
+The conjunction type operator plays well together with capabilities, allowing us
+to express capability tags as part of the regular type system. For example, given a class
+`C` we can express the type of a reference to an object of class type `C` with
+reference capability `mut` as
+```verona
+type CMut = C & mut
+```
+
+By extension this means that polymorphism in Verona doesn't need special
```
sylvanc, line 103: Why the negative? There is no reason to syuppose "special handling" in the first place. Avoid negative constructions.

docs/motivation.md
```diff
@@ -0,0 +1,550 @@
...
+handling of capabilities in type parameters. A type `A[T]` parameterized on `T`
+can simply be instantiated with a type like `C & mut`. The method types of `A`
```
sylvanc, line 105: Avoid "simple" and "simply".

docs/motivation.md
```diff
@@ -0,0 +1,550 @@
...
+are then able to express further restrictions on `T` that allow us to define
+operations for each instantiation.
+```verona
+type A[T] = {
```
sylvanc, line 109: This needs a more careful explanation of ref caps where the distinction between A[T & imm] and A[T] & imm can be made clear.

docs/motivation.md
```diff
@@ -0,0 +1,532 @@
+# The Verona language: Features and Motivation 
+
+Verona is an object oriented reference capability language with structural
```
sophiaIC, line 7: I think here we should also say which parts of the type system are novel, and which are more standard. 

The capabilities part seems to me not to be fully explained here. Is that so? Is that on purpose? Should there be a reference to OOPSLA?

docs/motivation.md
```diff
@@ -0,0 +1,550 @@
...
+}
+```
+
+* TODO: compare with
+    - objective c: categories
+    - ruby: open classes
+    - C#: extension methods
+
+### Disjunction types
+Verona also includes disjunction types. Assuming two types `C` and `D` we can
```
sylvanc, line 121: These sections need introduction as "algebraic types". Having top level sections of Structural, followed by Predicated, followed by Algebraic, for example (order to be discussed!). In Algebraic, a discussion of "what can be composed", allowing a distinction between tuples, classes, traits, and reference capabilities to be made.

docs/motivation.md
```diff
@@ -0,0 +1,550 @@
...
+define the new type
+```verona
+type COrD = C | D
+```
+This describes the union of the two types.
+
+If we assume `C` and `D` above are classes, we can furthermore write pattern
+matches to distinguish between the two types dynamically.
+
+Since in Verona there is no subtyping between classes, we are able to check that
```
sylvanc, line 131: Pattern matching can be completely elided here. If it's to be discussed, it needs a separate section, after everything else. And exhaustive matching is a "late concept" within pattern matching.

docs/motivation.md
```diff
@@ -0,0 +1,550 @@
...
+
+f (x : t) (g : t -> s) =
+    match x with
+        C y => ...
+        _   => g x
+```
+Note that `g` cannot have a more refined type than `t -> s`.
+
+#### Type discrimination
+In ML we might want to construct a type such as
```
sylvanc, line 166: Don't introduce ML when trying to explain Verona. Instead, discuss open and closed world types, without reference to other languages.

docs/motivation.md
davidchisnall: 1.0

davidchisnall: Well, technically, Objective-C 4 (I think, might have been 3), but Java inherited it from there.

docs/motivation.md
```diff
@@ -0,0 +1,550 @@
...
+type s = A of int
+         B of int
+```
+A corresponding Verona type might look like
+```verona
+type S = I32 | I32
+```
+but it is now impossible to discriminate between the two cases. This is however
+easily fixed by defining `S` instead as
+```verona
```
sylvanc, line 177: This is really about defining data types. There's no need to address discriminated vs. anonymous unions here - that's for a language comparison doc.

docs/motivation.md
```diff
@@ -0,0 +1,550 @@
...
+class A { var i : I32 }
+class B { var i : I32 }
+
+type S = A | B
+```
+
+#### Explicit nullability
+We can easily encode nullability with a class `None`
```
sylvanc, line 185: This is not "Null". A null is a value that can inhabit many types. Instead, this can be addressed as "verona uses algebraic types to express options and absence".

docs/motivation.md
```diff
@@ -0,0 +1,550 @@
...
+        }
+        _ => { // x : None
+            ...
+        }
+    }
+}
+```
+
+### Subtyping and inheritance
+Verona is an object-oriented system with classes and traits. When it comes to
```
sylvanc, line 203: This isn't quite right. We can remove the use of "nominal" everywhere. Classes are concrete, traits are structural (there's no nominal type relationship).

docs/motivation.md
```diff
@@ -0,0 +1,550 @@
...
+subtyping, classes are treated nominally, and traits are treated structurally.
+Furthermore there is no subtyping between classes. This means a class
+can be a subtype of a trait, but never a subtype of a class other than itself.
```
sylvanc, line 206: This can be covered early in a Structural section. Introduce `Eq` separately in a section that shows how to build useful structural primitives, and covers F-bound polymorphism and Self types.

sylvanc (changes requested):

docs/motivation.md
mjp41: Add a section on the audience, expectation of the understanding the reader will get, and purpose of the document.

mjp41: Newspaper style... important first, and less important as you go.

mjp41: Tree of documents versus newspaper style?  To be decided

docs/motivation.md
mjp41: Novelty should be part of the top level description.

docs/motivation.md
```diff
@@ -0,0 +1,550 @@
+# The Verona language: Core type system
+
+Verona is a language with a behaviour-oriented concurrency model and a region abstraction for ownership.
+These are fundamentally new abstractions in language design and are covered separately.
+This document aims to cover the parts of the type system that are similar to (though in some places diverge in interesting ways from) other languages.
+
+Verona is an object oriented reference capability language with structural
+subtyping and bounded polymorphism. This document aims to describe the aspects
```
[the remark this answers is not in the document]
mjp41: Predicated polymorphism.

docs/motivation.md
mjp41: Evidence on productivity?  Jan Vitek paper about productivity citation.  We approximate dynamic programming through the really cool typesystem.

docs/motivation.md
mjp41: Changing structure to address this.

docs/motivation.md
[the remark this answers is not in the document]
sophiaIC: "syntactic sugar" ==> "can also be understood as" 

I think our readers do not care about syntactic sugar.

docs/motivation.md
mjp41: Not all humans will find this simplifying

docs/motivation.md
mjp41: All our types are types of references...

docs/motivation.md
EliasC: Is the comma issue a British vs. American English thing? I agree on not starting a sentence with either though.

mjp41: I think I would put comma's in like Sylvan says, e.g., like in this sentence.
But mostly, I use
> "for example" instead of "e.g."

and

> "that is" instead of "i.e."

docs/motivation.md
sylvanc: Yes. Caveat: mentioning your own type could also be treated as `Self`, so that D got `equal: (Self, Self)->Bool`, that needs discussion (it may be that such a trick leads to too many type errors in the body of the function).

More generally: I think we should favor the term "code reuse" instead of "inheritance".

docs/motivation.md
```diff
@@ -0,0 +1,553 @@
...
+    f(...) : ... {...}
+}
+
+// here we reuse functionality of C in D, specifically the method f, but
+// we do not get a subtyping relation D <: C
+class D : C {
+    equal(self: D, other: D): Bool {...}
+    g(...): ... { ... f(...) ... }
+}
+// TODO: find a realistic example
```
sylvanc, line 245: Meta comment: we don't need realistic examples. In fact, they can be actively harmful. Examples of languages to emulate for language docs and language specs include: Go, Rust, C#, and to some degree TypeScript (that one is problematic as it spends a lot of time referencing JavaScript).

docs/motivation.md
```diff
@@ -0,0 +1,553 @@
...
+```
+If we want to describe a relation between `C` and `D`, we can only do so by
+subtyping them under a trait.
+
+There is a case to be made for not allowing inheritance from a
+  class, since this allows unanticipated code reuse. This might however also be
+  a positive.
```
sylvanc, line 252: This can be removed. Language docs aren't the right place to have design discussions.

docs/motivation.md
```diff
@@ -0,0 +1,553 @@
...
+// here we reuse functionality of C in D, specifically the method f, but
+// we do not get a subtyping relation D <: C
+class D : C {
+    equal(self: D, other: D): Bool {...}
+    g(...): ... { ... f(...) ... }
+}
+// TODO: find a realistic example
+```
+If we want to describe a relation between `C` and `D`, we can only do so by
+subtyping them under a trait.
```
sylvanc, line 248: This isn't quite right. There may exist many relationships between `C` and `D`, based purely on structural commonality. There's no need to write a trait for such commonality to exist - this is a key point in structural subtyping.

docs/motivation.md
```diff
@@ -0,0 +1,553 @@
...
+```
+Here, the definition `ToString` contains a default implementation for the print
+method.
+
+* Question: Should traits be allowed to define default fields?
+
+###### External trait inheritance
+A possibility is to allow externally declared trait inheritance, i.e.
+declaration of trait inheritance can be separated from class definition.
+```verona
```
sylvanc, line 279: I don't understand what this means, or what the example is illustrating. Questions and design discussions are great! Let's move them to a different doc.

docs/motivation.md
```diff
@@ -0,0 +1,553 @@
...
+type Collection[T] = {
+    add(self: Self, e: T) : Self
+}
+```
+where the `add` method returns something of the concrete type, e.g.
+`list.add(x)` returns the type `List[T]` rather than `Collection[T]`.
+
+
+## Dynamic and static dispatch
+Method calls can be dynamically or statically dispatched based on the type of
```
sylvanc, line 317: I think it's too early to try to give optimization tips. Perhaps the interesting thing to cover here is call syntax?

docs/motivation.md
sylvanc: Yes, I think this is important. We can avoid the word "inheritance" entirely. For describing predicates, I think we'll need to try several times :) We're looking for the "simplest possible" language - with no metric for what that is!

docs/motivation.md
```diff
@@ -0,0 +1,553 @@
...
+let z : ToString = if (...) { C::create() } else { D::create() }
+let u = z.toString() // dynamic dispatch, since ToString is an open type
+```
+
+### Type predicates
+Type predicates allow us to describe assumptions about subtyping. These can be
+written on both type and method level, and both has their distinct uses.
+Specifically, predicates on both type and method level allow us to express
+F-bounded polymorphism, while method level predicates are further useful in
+reducing code duplication, somewhat akin to type classes.
```
sylvanc, line 346: Calling out "what something isn't" leads to confusion. Either the audience doesn't have that context, or worse yet, has it - and makes unfounded additional assumptions! Go/Rust/C# docs are very good at explaining without reference to other languages or out-of-language terminology.

docs/motivation.md
mjp41: Motivation, refer to spec when we have it.

docs/motivation.md
mjp41: Motivation, refer to spec when we have it.

docs/motivation.md
mjp41: Stylistic say positive not negative

docs/motivation.md
```diff
@@ -0,0 +1,553 @@
...
+The conjunction type operator plays well together with capabilities, allowing us
+to express capability tags as part of the regular type system. For example, given a class
+`C` we can express the type of a reference to an object of class type `C` with
+reference capability `mut` as
+```verona
+type CMut = C & mut
+```
+
+By extension this means that polymorphism in Verona doesn't need special
+handling of capabilities in type parameters. A type `A[T]` parameterized on `T`
```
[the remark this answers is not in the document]
mjp41: Square bracket notation - language spec to cover.  Probably should say a bit more here.

docs/motivation.md
mjp41: Need language spec to cover in detail

docs/motivation.md
mjp41: Table of contents for language spec really needs to think about the order of these.  Use this discussion to guide choices.

docs/motivation.md
mjp41: Define first, and compare second.

Perhaps want language comparison docs for each similar language. 

X-lang for Y programmers docs - often negative signposting.. e.g., common misunderstandings.

docs/motivation.md
mjp41: https://github.com/microsoft/verona/pull/649/files#r1530627703

docs/motivation.md
mjp41: Precision of Null versus option types.

docs/motivation.md
mjp41: Nominal comes with baggage from nominal subtyping, so avoid the baggage.

mjp41: Open and close world types discussion too

docs/motivation.md
mjp41: Inheritance, perhaps use "reuse"?

Make clear that you can have multiple traits.

docs/motivation.md
mjp41: Realism versus teaching.  Need to teach

docs/motivation.md
mjp41: Third type of doc... Design decisions.  Neither Spec nor motivation.

docs/motivation.md
```diff
@@ -0,0 +1,553 @@
...
+    toString(s: C) : String {
+        ...
+    }
+}
+// here C will inherit the default implementation of print
+```
+Here, the definition `ToString` contains a default implementation for the print
+method.
+
+* Question: Should traits be allowed to define default fields?
```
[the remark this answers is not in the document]
mjp41: Should be addressed by the language spec.

PR #653, opened 2024-03-22
Initial TOC draft for the language specification.

sylvanc:


spec/spec.md
```diff
@@ -0,0 +1,174 @@
...
+- `_` as "don't care".
+
+### Strings
+
+### Constants
+
+- Boolean
+- Various integers and floats
+
+> How to talk about integer literals not having a type yet?
```
sophiaIC, line 18: no strings?

spec/spec.md
```diff
@@ -0,0 +1,174 @@
...
+
+- `_` as "don't care".
+
+### Strings
+
+### Constants
+
+- Boolean
+- Various integers and floats
+
```
sophiaIC, line 17: no string constants?

sophiaIC, line 17: are strings not constant?

spec/spec.md
```diff
@@ -0,0 +1,174 @@
...
+> How to talk about integer literals not having a type yet?
+
+### Lambdas
+
+### Object Literals
+
+- Open issues: code reuse, possible conflict with "real" `new`.
+
+### Expression Termination
+
```
sophiaIC, line 27: syntactic termination?

spec/spec.md
```diff
@@ -0,0 +1,174 @@
...
+### Dots
+
+- Selectors
+- Reverse application
+
+### Universal Call Syntax
+
+- Object-like vs. operator-like local names
+  - Type names are `create` sugar
```
sophiaIC, line 69: do no issues arise about order of intiailization of objects?

spec/spec.md
```diff
@@ -0,0 +1,174 @@
...
+
+### Generics and Predicates
+
+### Type Lists
+
+- Generics for tuple arity.
+
+### Capabilities
+
+> Chicken and egg with regions.
```
mjp41, line 129: I think we should discuss in the Teams call.

spec/spec.md
```diff
@@ -0,0 +1,174 @@
...
+- Boolean
+- Various integers and floats
+
+> How to talk about integer literals not having a type yet?
+
+### Lambdas
+
+### Object Literals
+
+- Open issues: code reuse, possible conflict with "real" `new`.
```
TobiasWrigstad, line 24: ?

spec/spec.md
mjp41, line 18: By this we mean `0` is the zero that is in U64, float, etc.

spec/spec.md
```diff
@@ -0,0 +1,174 @@
...
+### Try
+
+- "Give me the value instead of returning"
+
+### Pattern Matching
+
+- Do we need to talk about type lists here?
+
+## Types
+
```
TobiasWrigstad, line 104: Where under here (or elsewhere) do the Store[] types go?

spec/spec.md
nwf-msr, line 18: Someone somewhere surely has said something useful about Haskell's handling of (initially numeric, and now also string and list) literals.  https://www.haskell.org/tutorial/numbers.html is surely someone and somewhere, but as to something useful, I'll leave judgement to the reader.

spec/spec.md
```diff
@@ -0,0 +1,174 @@
...
+- Fields
+- Functions
+- "Left-hand" functions
+- Nesting
+- Default values
+- Visibility and encapsulation
+- Generics: what needs to be discussed here as opposed to in `Types`?
+  - Treat it as template-like, discuss type checking later
+- Do we need to talk about type lists here?
+
```
TobiasWrigstad, line 45: What is the exception story?

spec/spec.md
```diff
@@ -0,0 +1,174 @@
...
+### Comments
+
+### Identifiers and Symbols
+
+- `_` as "don't care".
+
+### Strings
+
+### Constants
```
mjp41, line 13: @davidchisnall Rename to "Primitive literal", rather than constants

spec/spec.md
mjp41: Perhaps call the section "String literals" and move after "object literals"

mjp41: Do we have "array literals"?

spec/spec.md
mjp41: Currently the parser uses `new` for string literals, but that might not be the final choice, and we should decide.

spec/spec.md
mjp41: Comments from discussion
* Static initialisers in other languages, and when things happen. 
* Immutable static data initialisation, probably not a problem here.
* Initialisation order is complex in some languages.
* `create` as a factory method, and allocates after it has found all the bits. So create can fail, make sure we allow failure returns from `create`.

spec/spec.md
```diff
@@ -0,0 +1,174 @@
+# Verona Language Spec
+
```
mjp41, line 2: High-level comment.

Rationale subsections, or a separate rationale document.

spec/spec.md
mjp41: In the assignment section above?

spec/spec.md
mjp41: Comments:
* Describe the region model before the types to capture and enforce it.  
* Object model section before types - pretty pictures
* Types turn pretty pictures into maths.

PR #667, opened 2024-04-18
Fixes #651

### LINKED ISSUE #651: Shadowing of Self type name leads to looping in structure pass
```
type Copyable[Self] = { // This loops forever due to parameter name self
  copy(self: Self) : Self
} & (Self < Copyable[Self])
```

Running the compiler on the above program leads to looping in structure pass.

mjp41:
This bug was caused by error checking code.  Basic issue was matching using `!T(NS...)` and replacing with error, when `Error` is not in `NS...`.  This results in an infinite loop.


mjp41:
Fixed in Trieste.  Semantics of Trieste has been changed to not match terms with Error contained in them.

PR #682, opened 2025-07-08
updating documentation to specify clang-format-15

SophiaSR:
Updating the docs to match new requirement for clang-format-15 instead of 9, and fixing links to point to 15.


mjp41 (approved):
LGTM

